PR #244, opened 2020-07-24
Steve/errors

saolsen:
Stack traces as part of error message.
Bad attribute lookup errors.
- [x] Python
- [x] Ruby
- [x] Java

### DIFF
--- a/polar/src/lib.rs
+++ b/polar/src/lib.rs
@@ -298,6 +298,25 @@ pub extern "C" fn polar_question_result(query_ptr: *mut Query, call_id: u64, res
 }
 
 #[no_mangle]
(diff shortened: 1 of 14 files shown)


polar/src/lib.rs
```diff
@@ -313,6 +313,20 @@ pub extern "C" fn polar_question_result(query_ptr: *mut Query, call_id: u64, res
     })
 }
 
+#[no_mangle]
+pub extern "C" fn polar_error(query_ptr: *mut Query, message: *mut c_char) -> i32 {
```
dhatch, line 317: Maybe `polar_application_error` or `polar_set_error`? To make this clearly separate from the other get error function?

PR #248, opened 2020-07-24
allow() -> is_allowed() CODE change

leina05:
PR checklist:

- [x] Added changelog entry.


languages/python/oso/oso.py
```diff
@@ -37,7 +37,7 @@ def __init__(self):
         self.register_class(PathMapper)
 
     # TODO (dhatch): should we name this 'is_allowed'?
```
dhatch, line 39:
```suggestion
```

samscott89: Finally! That TODO has been there for so long.

PR #253, opened 2020-07-25
Java installation instructions

saolsen:
Instructions on where to download the jar and how to put it on the classpath.

Is there any way we can dynamically inject the latest version? I didn't add maven instructions because they're pretty tedious. Will add those in when we're on maven central and it's easy to add to your project.


saolsen:
Added the release thing we talked about in slack and made sure it's in sync with all other versions.

docs/getting-started/download/java.rst
```diff
@@ -2,7 +2,13 @@
 Java
 ====
 
-.. todo:: Java installation
+The Java version of oso is available on Github. Go to the `Maven Repository <https://github.com/osohq/oso/packages/321403>`_ and download the latest jar.
+
+To use it, add it to the classpath for your project.
+
+    $ javac -classpath "oso-|release|.jar:." MyProject.java
```
samscott89, line 9: Does this actually work? I'm surprised sphinx lets you do substitutions mid-text

saolsen: Yep, 
Screen Shot 2020-07-25 at 6 14 18 PM

docs/spelling_allowed_words.txt
```diff
@@ -58,5 +58,5 @@ iterables
 sqlite
 lookups
 classpath
-Github
+GitHub
```
samscott89, line 61:
```suggestion
```

(its on there already) - line 43

PR #255, opened 2020-07-25
Host language debugger

plotnick:
Debugger works in all three host languages. Displays application instances using their native string representations:
```
>>> from oso import Oso, Http
>>> o = Oso()
>>> o.load_str("d(x) if debug(x);")
>>> o.query_predicate("d", Http(hostname="foo")).success
Welcome to the debugger!
debug(Http(hostname="foo"))
debug> bindings
Http = <class 'oso.extras.Http'>
Datetime = <class 'datetime.datetime'>
Timedelta = <class 'datetime.timedelta'>
PathMapper = <class 'oso.extras.PathMapper'>
_x_1 = Http(hostname="foo")
_x_2 = Http(hostname="foo")
debug> continue
True
```

PR checklist:
- [x] Docs

### DIFF
--- a/polar/src/formatting.rs
+++ b/polar/src/formatting.rs
@@ -299,7 +299,11 @@ pub mod to_polar {
 
     impl ToPolarString for ExternalInstance {
         fn to_polar(&self) -> String {
-            format!("^{{id: {}}}", self.instance_id)
+            if let Some(ref repr) = self.repr {
+                repr.clone()
+            } else {
+                format!("^{{id: {}}}", self.instance_id)
+            }
         }
     }
 
--- a/docs/more/dev-tools/debugger.rst
+++ b/docs/more/dev-tools/debugger.rst
@@ -5,39 +5,40 @@ Debugger
 The Polar debugger allows debugging of policy rules. It can be helpful to see
 why a rule is behaving differently than expected.
 
-.. highlight:: polar
-
 ********************
 Running the debugger
 ********************
 
-The debugger can be entered through two mechanisms:
+The debugger is entered through the ``debug()`` predicate. It acts like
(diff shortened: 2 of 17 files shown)


polar/src/formatting.rs
```diff
@@ -299,7 +299,11 @@ pub mod to_polar {
 
     impl ToPolarString for ExternalInstance {
         fn to_polar(&self) -> String {
-            format!("^{{id: {}}}", self.instance_id)
+            if let Some(ref repr) = self.repr {
+                repr.clone()
+            } else {
+                format!("^{{id: {}}}", self.instance_id)
```
samscott89, line 305: Hmm, actually not sure this should go in ToPolarString... It's not quite correct. ToPolarString should be parseable back into a term.

samscott89:
Reminded by @gj we need docs + tests for these. I think the docs are actually stable enough now that these don't _need_ to be done on the `docs/` branch. But I'd be okay with doing tests here + docs there if easier.

polar/src/formatting.rs
plotnick: `ExternalInstance` was and is the exception to that. There is no parseable notation for an external instance.

samscott89: Oh you're totally right, I thought we had parser code still for `^{...}`. I think this is still unfortunate though. Other than implementing a _new_ format trait which uses ToPolarString for most things, I can't see a good way to solve this. We should revisit this though.

plotnick: I don't think it's a problem. E.g., in Python, the default instance repr is not parsable; Common Lisp has `print-unreadable-object` for the same reason. Most instances don't have a parsable representation in most languages.

PR #256, opened 2020-07-25
Add tests for not + lookup precedence ordering.

samscott89:
Currently failing.


polar/tests/integration_tests.rs
```diff
@@ -460,7 +460,7 @@ fn test_not() {
     assert!(qeval(
         &mut polar,
         "
-        d = {x: 1} and not d.x = 1
+        d = {x: 1} and not d.x = 2
```
samscott89, line 463: Oh woops. The test was actually correctly failing for the wrong reason.

PR #266, opened 2020-07-27
Warn about unknown specifiers

plotnick:


polar/tests/integration_tests.rs
```diff
@@ -693,7 +693,8 @@ fn test_load_with_query() {
 
 #[test]
 fn test_externals_instantiated() {
-    let polar = Polar::new(None);
+    let mut polar = Polar::new(None);
+    polar.register_constant(sym!("Foo"), term!(true));
```
gj, line 697: caught by the warning?

polar/tests/integration_tests.rs
```diff
@@ -968,11 +971,14 @@ fn test_singleton_vars() {
...
+               001: f(x:X,y:Y,z:Z) if z = z;
                       ^
                Singleton variable y
-               001: f(x,y,z) if z = z;
-                        ^
+               001: f(x:X,y:Y,z:Z) if z = z;
+                          ^
+               Unknown specializer Z
+               001: f(x:X,y:Y,z:Z) if z = z;
+                                ^
```
gj, line 981: excellent

languages/python/polar/polar.py
```diff
@@ -22,13 +22,17 @@ class Polar:
...
+        self.host = Host(self.ffi_polar)
         self.load_queue = []
 
         # Register built-in classes.
         self.register_class(datetime, name="Datetime")
         self.register_class(timedelta, name="Timedelta")
 
+        # Pre-registered classes.
+        for name, cls in classes.items():
+            self.register_class(cls, name=name, from_polar=constructors[name])
```
gj, line 34: Should the equivalent change be made for Ruby and Java?

polar/src/polar.rs
```diff
@@ -248,7 +252,11 @@ impl Polar {
         for (sym, singleton) in singletons {
             if let Some(term) = singleton {
                 let mut writer = self.output.write().unwrap();
-                let _ = writeln!(&mut writer, "Singleton variable {}", sym);
+                let _ = if let Value::Pattern(..) = term.value() {
+                    writeln!(&mut writer, "Unknown specializer {}", sym)
+                } else {
+                    writeln!(&mut writer, "Singleton variable {}", sym)
```
leina05, line 258:
```suggestion
                    writeln!(&mut writer, "Singleton variable {}. This variable may be unused or undefined, see https://docs-preview.oso.dev/using/polar-syntax.html#variables ", sym)
```
We would need to update the link when the docs are live

polar/tests/integration_tests.rs
plotnick: Yup.

languages/python/polar/polar.py
plotnick: No, because they don't have the decorator that frobs the globals that this handles. They always explicitly use `register_class` after the `Polar` instance is instantiated.

polar/src/polar.rs
plotnick: I strongly object to a warning that long. We could leave out the "Singleton variable" part, and make it `The variable {} is unused or undefined, see <...>`, but that's still pretty long. I really think it's best as-is, with an FAQ entry that points to the doc.

plotnick: OK, settled on a variant, we'll use a link shortener later.

PR #267, opened 2020-07-27
Update python query API

leina05:
`query()` and `query_rule()` are now generators in Python, to match the other language libraries.

**Implications**
- These methods used to return `QueryResult`, which had the method `success()`. Success must now be determined by converting the result to a list. E.g., `assert list(polar.query("foo(x)"))
- `QueryResult` has the `results` attribute, which gives a list. Now to get a list of results, convert to a list directly as above
- We are not doing anything with `trace()`... the infrastructure is still there but we simply never create a `QueryResult` anymore

In Java: `query()` now accepts either a String or a Predicate, and `queryRule()` accepts variable args to match the other libraries.

PR checklist:

- [ ] Added changelog entry.

### DIFF
--- a/languages/java/oso/src/main/java/com/osohq/oso/Polar.java
+++ b/languages/java/oso/src/main/java/com/osohq/oso/Polar.java
@@ -85,23 +85,23 @@ public class Polar {
      * @param query String string
      * @return Query object (Enumeration of resulting variable bindings).
      */
-    public Query query(String query) throws Exceptions.OsoException {
+    public Query queryStr(String query) throws Exceptions.OsoException {
         loadQueuedFiles();
         return new Query(ffiPolar.newQueryFromStr(query), host.clone());
     }
 
     /**
-     * Query for a predicate.
+     * Query for a rule.
      *
-     * @param predicate Predicate name, e.g. "f" for predicate "f(x)".
-     * @param args List of predicate arguments.
+     * @param rule Rule name, e.g. "f" for rule "f(x)".
+     * @param args List of rule arguments.
      * @return Query object (Enumeration of resulting variable bindings).
      * @throws Exceptions.OsoException
      */
-    public Query query(String predicate, List<Object> args) throws Exceptions.OsoException {
+    public Query queryRule(String rule, List<Object> args) throws Exceptions.OsoException {
         loadQueuedFiles();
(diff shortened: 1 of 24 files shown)


languages/python/polar/polar.py
```diff
@@ -76,28 +87,18 @@ def query(self, query, single=False):
...
-            results.append(result)
-            if single:
-                break
-        return QueryResult(results)
+        for res in Query(query, host=host).run():
+            yield res["bindings"]
 
-    def query_predicate(self, name, *args, **kwargs):
-        """Query for predicate with name ``name`` and args ``args``.
+    def query_rule(self, name, *args, **kwargs):
```
dhatch, line 93: You can drop **kwargs, it's unused.

languages/python/polar/query.py
```diff
@@ -60,17 +56,17 @@ def run(self, ffi_query=None):
             if kind == "MakeExternal":
                 self.handle_make_external(data)
             if kind == "ExternalCall":
-                self.handle_external_call(ffi_query, data)
+                self.handle_external_call(self.ffi_query, data)
```
dhatch, line 59: This doesn't have to happen on this PR, but we can just access ``self.ffi_query`` from within the handler methods.

languages/java/oso/src/main/java/com/osohq/oso/Polar.java
```diff
@@ -85,23 +85,23 @@ public void loadStr(String str) throws Exceptions.OsoException {
      * @param query String string
      * @return Query object (Enumeration of resulting variable bindings).
      */
-    public Query query(String query) throws Exceptions.OsoException {
+    public Query queryStr(String query) throws Exceptions.OsoException {
```
dhatch, line 88: This is still just called `query` in other languages.  Ok to leave for now.

leina05: Made this fix and updated the Java query API to be the same as the others

languages/python/polar/query.py
leina05: fixed.

languages/python/polar/polar.py
leina05: fixed

PR #269, opened 2020-07-27
Documentation proof edits

dhatch:
Contains edits for getting started guides.


docs/getting-started/policies/index.rst
```diff
@@ -162,12 +167,6 @@ So instead, we can refactor the role check into its own rule:
 
     role(user, role_name) if user.role = role_name;
 
-
-The ``role(user, "accountant")`` is yet another example of matching happening
```
samscott89, line 166: Hmm, I personally think that removing this  - and the previous change to de-emphasize matching, though I was less bothered about that by itself -  is a loss for helping people understand what is going on when writing policies. I think this is a good chance to get people to build up some intuition about it without becoming full prolog experts.

dhatch: Ok, I can change the wording and keep these.  I removed it because we no longer have the intro section that introduces matching.

samscott89: Okay, that would be great.

dhatch: Reverted this, we actually do mention matching in the beginning of this doc.

PR #270, opened 2020-07-27
Fix python decorator bugs.

dhatch:


dhatch:
duplicated a fix in #266 .

PR #272, opened 2020-07-27
Handle None/nil/null returns from host methods

leina05:
How we handle None returns:
- Stored as a host instance

What this means:
- unification with something that returns `None` will simply fail, but not error
- lookups on `None` will throw Application Errors

PR checklist:

- [ ] Added changelog entry.

### DIFF
--- a/languages/java/oso/src/main/java/com/osohq/oso/Host.java
+++ b/languages/java/oso/src/main/java/com/osohq/oso/Host.java
@@ -48,7 +48,8 @@ public class Host implements Cloneable {
      * Store a Java class in the cache by name.
      *
      * @param name
-     * @throws Exceptions.DuplicateClassAliasError If the class is already registered.
+     * @throws Exceptions.DuplicateClassAliasError If the class is already
+     *                                             registered.
      */
     public String cacheClass(Class cls, Function<Map, Object> constructor, String name)
             throws Exceptions.DuplicateClassAliasError {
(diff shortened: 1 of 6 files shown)


languages/java/oso/src/main/java/com/osohq/oso/Host.java
```diff
@@ -170,37 +172,38 @@ public boolean isa(long instanceId, String classTag)
...
      * @param value Java Object to be converted to Polar.
-     * @return JSONObject Polar term of form: {@code {"id": _, "offset": _, "value": _}}.
+     * @return JSONObject Polar term of form: {@code {"id": _, "offset": _, "value":
+     *         _}}.
      */
     public JSONObject toPolarTerm(Object value) throws Exceptions.OsoException {
         // Build Polar value
         JSONObject jVal = new JSONObject();
-        if (value.getClass() == Boolean.class) {
+        if (value != null && value.getClass() == Boolean.class) {
```
samscott89, line 181: How about handling the null check first?

leina05: yeah, I just wanted to use the same code as the external instance, so I'd have to duplicate that at the top. I thought this was slightly clearer but can change.

leina05: we can do this as a follow-up if needed, merging now

PR #273, opened 2020-07-27
Don't loop on a variable bound to itself

plotnick:
Before:
```console
$ python -m oso
query> x = x
[1]    6778 segmentation fault  python -m oso
```

After:
```console
$ python -m oso
query> x = x
{'x': 'x'}
```

### DIFF
--- a/polar/tests/integration_tests.rs
+++ b/polar/tests/integration_tests.rs
@@ -621,6 +621,9 @@ fn test_non_instance_specializers() {
 #[test]
 fn test_bindings() {
     let mut polar = Polar::new(None);
+    assert_eq!(qvar(&mut polar, "x=1", "x"), vec![value!(1)]);
+    assert_eq!(qvar(&mut polar, "x=x", "x"), vec![value!(sym!("x"))]);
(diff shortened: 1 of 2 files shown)


polar/tests/integration_tests.rs
```diff
@@ -621,6 +621,9 @@ fn test_non_instance_specializers() {
 #[test]
 fn test_bindings() {
     let mut polar = Polar::new(None);
+    assert_eq!(qvar(&mut polar, "x=1", "x"), vec![value!(1)]);
+    assert_eq!(qvar(&mut polar, "x=x", "x"), vec![value!(sym!("x"))]);
```
samscott89, line 625:
```suggestion
    assert_eq!(qvar(&mut polar, "x=x", "x"), vec![value!(sym!("x"))]);
    assert_eq!(qvar(&mut polar, "x=y,y=x", "x"), vec![value!(sym!("x"))]);
```
Would this work?

plotnick: More or less.

PR #288, opened 2020-07-28
Assignment Operator

leina05:
Introduced `:=` as assignment operator. 
Currently, assignment is only valid when the LHS is an unbound variable. Open to input on that.

PR checklist:

- [x] Added changelog entry.

### DIFF
--- a/docs/changelogs/vNEXT.rst
+++ b/docs/changelogs/vNEXT.rst
@@ -24,14 +24,12 @@ Link to migration guide
 New features
 ==============
 
-Feature 1
-=========
-
-- summary
-- of
-- user facing changes
-
-Link to relevant documentation section
+Assignment Operator
+===================
+- The operator ``:=`` may now be used to assign values to unbound variables. Unlike the unify operator (``=``),
+the assignment operator will NOT evaluate equality.
+- Attempting to assign to a non-variable will result in a parse error.
+- Attempting to assign to a bound variable will result in a runtime error.
 
 
 Other bugs & improvements
--- a/docs/using/polar-syntax.rst
+++ b/docs/using/polar-syntax.rst
@@ -224,7 +224,7 @@ said to *unify* if they are equal or if there is a consistent set of
 variable bindings that makes them equal. Unification is defined
 recursively over compound types (e.g., lists and dictionaries):
 two compound values unify if all of their corresponding elements
-unify.
+unify.:warn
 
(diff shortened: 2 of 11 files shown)


languages/python/tests/test_polar.py
```diff
@@ -378,7 +378,7 @@ def test_parser_errors(polar):
 
 def test_runtime_errors(polar, query):
     rules = """
-    foo(a,b) := a in b;
+    foo(a,b) if a in b;
```
srenatus, line 381: I don't believe it's of much consequence, but I've stumbled upon the `polar_str` in traces -- does this need to be updated, too? https://github.com/osohq/oso/blob/ff8fc2d524d6c3b7eccb3be2935a32f4dae45237/polar/src/formatting.rs#L256

samscott89: Yep! Good catch.

docs/changelogs/vNEXT.rst
```diff
@@ -24,14 +24,12 @@ Link to migration guide
...
-- of
-- user facing changes
-
-Link to relevant documentation section
+Assignment Operator
+===================
+- The operator ``:=`` may now be used to assign values to unbound variables. Unlike the unify operator (``=``),
+the assignment operator will NOT evaluate equality.
+- Attempting to assign to a non-variable will result in a parse error.
+- Attempting to assign to a bound variable will result in a runtime error.
```
gj, line 32: Is blowing up at runtime preferable to allowing the variable reassignment?

docs/using/polar-syntax.rst
```diff
@@ -224,7 +224,7 @@ said to *unify* if they are equal or if there is a consistent set of
 variable bindings that makes them equal. Unification is defined
 recursively over compound types (e.g., lists and dictionaries):
 two compound values unify if all of their corresponding elements
-unify.
+unify.:warn
```
gj, line 227: Is this a valid rst directive?

polar-core/src/lexer.rs
```diff
@@ -48,31 +48,31 @@ pub enum Token {
...
+    Add,   // +
+    Sub,   // -
+    Eq,    // ==
+    Neq,   // !=
+    Leq,   // <=
+    Geq,   // >=
+    Lt,    // <
+    Gt,    // >
+    Unify, // =
+    Assign,
```
gj, line 73: nit: alignment purposes

PR #297, opened 2020-07-30
Improve integer/floating-point comparisons

plotnick:
Should be able to compare any two integer & floating-point numbers with full machine precision.

Dropped implicit support for Ɛ comparisons. We can add a separate `~` operator later if we want that, but I think `==` should be as strict an equality relation as we can make it.

Also move numeric implementations to separate `numerics` module.

And follow up to #290 to support, e.g., `1e2`.

### DIFF
--- a/polar/src/lexer.rs
+++ b/polar/src/lexer.rs
@@ -340,9 +340,17 @@ impl<'input> Lexer<'input> {
 
         last = self.match_digits(last);
 
+        if let Some((i, '.')) = self.c {
+            self.push_char('.');
+            last = i;
+            parse_as_float = true;
+
+            last = self.match_digits(last);
+        }
+
         if let Some((i, char)) = self.c {
             match char {
-                '.' | 'e' | 'E' => {
+                'e' | 'E' => {
                     self.push_char(char);
                     last = i;
                     parse_as_float = true;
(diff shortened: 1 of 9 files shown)


polar/src/lexer.rs
```diff
@@ -340,19 +340,23 @@ impl<'input> Lexer<'input> {
 
         last = self.match_digits(last);
 
-        if let Some((i, '.')) = self.c {
-            self.push_char('.');
-            last = i;
-            parse_as_float = true;
-
-            last = self.match_digits(last);
+        if let Some((i, char)) = self.c {
```
saolsen, line 343: This isn't right anymore. Is this because you wanted to allow 1e4 without a `.`? Currently as this is it'd allow `1ee2` which isn't valid.

saolsen: added a fix

polar/src/numerics.rs
```diff
@@ -0,0 +1,135 @@
...
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        // Compare the integer `i` with the float `f`.
+        // Adapted from MongoDB's `compareLongToDouble`.
+        let partial_cmp = |i: i64, f: f64| {
+            if f.is_nan() {
+                None
+            } else if -MOST_POSITIVE_EXACT_FLOAT < i && i < MOST_POSITIVE_EXACT_FLOAT {
+                // The integer is exactly representable as a float.
+                (i as f64).partial_cmp(&f)
+            } else if f > MOST_POSITIVE_INTEGER as f64 {
```
plotnick, line 106: For edge case testing: is this `>` or `>=`?

PR #299, opened 2020-07-31
Add clippy target to Makefile

plotnick:
Arguments should match those in the GitHub workflow.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: test rust-test rust-build python-build python-test
+.PHONY: test rust-test rust-build python-build python-test ruby-test java-test docs-test fmt clippy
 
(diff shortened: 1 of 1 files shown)


saolsen:
Even this doesn't show errors for me that actions shows. Not sure why.

Makefile
```diff
@@ -32,3 +32,6 @@ docs-test: python-build
 fmt:
 	cargo fmt
 	$(MAKE) -C languages/python fmt
+
+clippy:
+	cargo clippy --all-features --all-targets -- -D warnings
```
gj, line 37: if you really want this to catch _everything_, I'd recommend prefixing it with `touch polar/src/lib.rs && cargo clippy...`

PR #301, opened 2020-07-31
Updates to parser and traces

samscott89:
- Main parser refactor work: https://github.com/osohq/oso/commit/2ac5d7002c7302c3aab74f86d23a71482c9e72cf
  adds a macro for constructing `Terms` from `Value`s as a spanned value. Changes many of the parser terms to be just Values.
 The exception is expressions which are mostly built up out of terms.

- Rework how errors are constructed, so errors reliably have some error context on them, and this is used to produce nicer error messages. Cleans up a bunch of the error code. (https://github.com/osohq/oso/commit/6847b365c7999420b255ca815b8a3d7670f1271e)

- Used this work to try out some nicer trace output format. (https://github.com/osohq/oso/commit/29d3e0ca08fcedfede6e6fcf90033fe447e53c70)

Before:

```
  f(x) if x = 1 and x = 1; [
    _x_3 = 1 and _x_3 = 1 [
      _x_3 = 1 [
      ]
      _x_3 = 1 [
      ]
    ]
```

After:

```
f(1) [
  f(x) if
    x = 1 and x = 1 [
      x = 1 []
      x = 1 []
  ]
]
```

Pros/cons:
- Former shows the intermediate values more clearly, instead of many many `x`s
- Latter is a bit easier to read

### DIFF
--- a/polar-core/src/parser.rs
+++ b/polar-core/src/parser.rs
@@ -263,4 +263,19 @@ mod tests {
             }
         ));
     }
+
+    #[test]
+    fn test_primitive_methods() {
+        let q = r#""abc".startswith("a")"#;
+        assert_eq!(
+            parse_query(q),
+            term!(op!(Dot, term!("abc"), term!(pred!("startswith", ["a"])))),
+        );
+
+        let q = r#"x.("invalid-key")"#;
+        assert_eq!(
+            parse_query(q),
+            term!(op!(Dot, term!(sym!("x")), term!("invalid-key"))),
+        );
+    }
 }
--- a/polar-core/src/parser.rs
+++ b/polar-core/src/parser.rs
@@ -73,9 +73,9 @@ pub fn parse_query(src_id: u64, src: &str) -> PolarResult<Term> {
 }
 
 #[cfg(test)]
-pub fn parse_rules(src: &str) -> PolarResult<Vec<Rule>> {
+pub fn parse_rules(src_id: u64, src: &str) -> PolarResult<Vec<Rule>> {
     RULES_PARSER
-        .parse(0, Lexer::new(src))
+        .parse(src_id, Lexer::new(src))
         .map_err(|e| to_parse_error(e).into())
 }
 
@@ -85,8 +85,16 @@ mod tests {
     use crate::formatting::ToPolarString;
     use pretty_assertions::assert_eq;
 
+    fn parse_term(src: &str) -> Term {
+        super::parse_term(src).unwrap()
+    }
+
     fn parse_query(src: &str) -> Term {
         super::parse_query(0, src).unwrap()
+    }
+
+    fn parse_rule(src: &str) -> Rule {
+        super::parse_rules(0, src).unwrap().pop().unwrap()
     }
 
     fn parse_lines(src: &str) -> Vec<Line> {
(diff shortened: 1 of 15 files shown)


polar/src/polar.lalrpop
```diff
@@ -120,79 +130,42 @@ Object<T>: BTreeMap<Symbol, Term> = {
     }
 };
 
-Dictionary<T>: Term = <start:@L> <fields:Object<T>> => {
+// ****** Strange Values ******* //
```
dhatch, line 135: What makes these "strange"?

polar/src/polar.lalrpop
```diff
@@ -120,79 +130,42 @@ Object<T>: BTreeMap<Symbol, Term> = {
...
 
 BuiltinOperator: Operator = {
     "debug" => Operator::Debug,
     "print" => Operator::Print,
 };
 
-pub BuiltinOperation: Operation = {
-    <op:BuiltinOperator> "(" <mut args:(<Exp2<"Term">> ",")*> <arg:Exp2<"Term">?> ")" => {
+BuiltinOperation: Value = {
+    <op:BuiltinOperator> "(" <mut args:(<TermExp> ",")*> <arg:TermExp?> ")" => {
```
dhatch, line 163: I'm not sure this should be a full `TermExp` as the argument.  Is `print(x in 1)` valid? Or `print(1 = x)`? Or even `print(y = 1 and x = 1)`

Those would all parse with this change.

polar/src/polar.lalrpop
```diff
@@ -202,207 +175,228 @@ RewritableOperator: Operator = {
...
-    "(" <exp:Exp<T>> ")" => exp // seems sorta impossible to support parrens for expressions and list literals.
-}
+New: Value = "new" <literal:Spanned<InstanceLiteralTerm>> => {
+    let args = vec![literal];
+    let op = Operation{operator: Operator::New, args};
+    Value::Expression(op)
+};
 
-Exp10<T>: Term = {
+SpecialExp<T>: Value = {
```
dhatch, line 195: Why is this a special exp? I think the naming based on precedence (Exp10) is clearer.

polar/src/polar.lalrpop
```diff
@@ -202,207 +175,228 @@ RewritableOperator: Operator = {
...
-    <start:@L> "cut" => {
+    <New> if T == "Term",
+    "cut" => {
         let args = vec![];
         let op = Operation{operator: Operator::Cut, args};
-        Term::new_from_parser(src_id, start, Value::Expression(op))
+        Value::Expression(op)
     },
-    <start:@L> "forall" "(" <arg1:Exp2<"Term">> "," <arg2:Exp2<"Term">> ")" => {
+    "forall" "(" <arg1:TermExp> "," <arg2:TermExp> ")" => {
```
dhatch, line 203: Same question here, what's the implication of changing this from `Exp2`?

polar/src/vm.rs
```diff
@@ -1986,21 +1989,26 @@ impl PolarVirtualMachine {
         }
     }
 
+    fn set_error_context(
+        &self,
+        term: &Term,
+        error: impl Into<error::PolarError>,
+    ) -> error::PolarError {
+        let source = self.source(term);
+        if source.is_none() {
```
dhatch: This could just be an expect, right?

polar/src/polar.lalrpop
samscott89: They already do:

```
query> print(x in 1)
x in 1
True
```

Exp2 is already matching everything but `x and y`. It's not really achieving much though, you're right

polar/src/vm.rs
samscott89: This was a placeholder :) Caught it in a follow up commit.

polar/src/polar.lalrpop
samscott89: I can update this to something more meaningful. I found them strange because they aren't "pure" values - they contain expressions, and they are different whether its a term or a pattern.

polar/src/polar.lalrpop
samscott89: Only allowing a few more possible combinations. Again, `x and y` now.

```
query> forall(x in [1, 2, 3], x > 0 and x < 4)
Parse error:  did not expect to find the token 'and' at line 1, column 30
query> forall(x in [1, 2, 3], (x > 0 and x < 4))
True
```

samscott89: With this change the former would be valid too.

polar/src/polar.lalrpop
samscott89: This one is not ideal because there isn't really a common thread for why these are grouped.

What I like about this pattern: the name of the parser rule matches the thing it is parsing. So we have AddExp, MatchExp, MulExp - mostly all descriptive names. The naming of levels is still there, e.g.

```
Exp10<T>: Term = {
    <Spanned<SpecialExp<T>>>,
    <Exp11<T>>,
}
```

Part of this is a side effect from wanting to refactor out all the term creations. You either have expressions are values, and the term wrapping is done in the rule, e.g. `AddOp: Value = <Spanned<Exp1>> "+" <Spanned<Exp2>>` or whatever,
Or all Exps are terms (so you don't need the Spanned) and you need to somehow wrap `AddOp` in `Spanned` (which is what we have here).

dhatch: Keyword Operation? `new` & `cut` are keywords.  `forall` could be moved to be with debug & print possibly.

samscott89: Moved them into BuiltinOperators and made them part of `Value` - they don't actually need to be parsed as expressions since they work as standalone values.

polar-core/src/parser.rs
```diff
@@ -263,4 +263,19 @@ mod tests {
             }
         ));
     }
+
+    #[test]
+    fn test_primitive_methods() {
```
samscott89, line 268: @dhatch noticed this while I was addressing your comments. 

`"abc".startswith("a")` did not previously work.

The fix was to let the LHS of dot be any Exp9 expression (so basically any value, term, or parenthesized expression): https://github.com/osohq/oso/pull/301/commits/6f40f08df25402f22b412c0dcad45ff9a00a9a48#diff-bf4d21bbb0fffe459f1185ea62ce26a4R242

PR #304, opened 2020-08-03
Built-in types

plotnick:
Can now specialize on `Boolean`, `Integer`, `Float`, `List`, `Dictionary`, and `String`.

PR checklist:
- [x] Python
- [x] Ruby
- [x] Java
- [x] Add changelog entry.


languages/python/tests/test_polar.py
```diff
@@ -556,8 +557,44 @@ def __eq__(self, other):
...
+    assert query("foo(true)")
+    assert not query("foo(false)")
+
+    # Integer
+    assert query("foo(3)")
+    assert not query("foo(2)")
+    assert not query("foo(-2)")
+
+    # Float
+    assert query("foo(4)")
```
dhatch, line 579: Why does this match the float rule instead of the integer one?

polar/src/vm.rs
```diff
@@ -1371,8 +1376,25 @@ impl PolarVirtualMachine {
...
+
+        // Coerce booleans to integers.
+        fn to_int(x: bool) -> i64 {
+            if x {
+                1
+            } else {
+                0
+            }
+        }
+        if let Value::Boolean(x) = left_term.value() {
```
dhatch, line 1390: Why did you add this?

languages/python/tests/test_polar.py
plotnick: It doesn't. Thanks.

polar/src/vm.rs
plotnick: In Python, `bool` is a subclass of `int`, and so rules specialized on `Integer` are also applicable to instances of `Boolean` (i.e., `true` and `false`). Coercing booleans to ints here (and we should consider it for arithmetic operations, too) lets everything work in a natural way instead of throwing horrible errors.

dhatch: Does this behavior make sense in other languages besides Python?

plotnick: I like languages that do this particular coercion. Encoding `true` & `false` as `1` and `0` is very common across many domains, so it comes in handy. I can't think of a case off-hand where it would fail to make sense, but I guess it depends on how strict you like your typing.

languages/ruby/lib/oso/polar/polar.rb
```diff
@@ -4,6 +4,24 @@
...
+    [TrueClass, Boolean].include? cls
+  end
+end
+
+# Monkey-patch Ruby false type.
+class FalseClass
+  def is_a?(cls)
+    [FalseClass, Boolean].include? cls
+  end
+end
```
gj, line 23:
```suggestion
# Missing Ruby type.
module PolarBoolean; end
# Monkey-patch Ruby true type.
class TrueClass; include PolarBoolean; end
# Monkey-patch Ruby false type.
class FalseClass; include PolarBoolean; end
```

languages/ruby/lib/oso/polar/polar.rb
```diff
@@ -15,6 +33,14 @@ def initialize
         @ffi_polar = FFI::Polar.create
         @host = Host.new(ffi_polar)
         @load_queue = Set.new
+
+        # Register built-in classes.
+        register_class(Boolean, name: 'Boolean')
```
gj, line 38:
```suggestion
        register_class(PolarBoolean, name: 'Boolean')
```

plotnick: To avoid clashes in the global namespace.

languages/java/oso/src/main/java/com/osohq/oso/Polar.java
```diff
@@ -20,6 +20,15 @@ public Polar() throws Exceptions.OsoException {
         ffiPolar = Ffi.get().polarNew();
         host = new Host(ffiPolar);
         loadQueue = new HashMap<String, String>();
+
+        // Register built-in classes.
+        // FIXME: These constructors are garbage.
+        registerClass(Boolean.class, m -> (boolean) m.get("value") ? false : true, "Boolean");
```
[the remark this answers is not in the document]
plotnick: Yes, probably. But they're going away today, which is why I didn't bother to really make them work. They're just stubs until boa-constructors land.

languages/ruby/spec/oso/language_parity_spec.rb
```diff
@@ -67,16 +67,26 @@ def foo
...
-
-
+oso.register_class Math
+oso.load_str '?= Math.acos(1.0) = 0.0;'
+
+# Test built-in type specializers.
+# rubocop:disable Layout/EmptyLineAfterGuardClause
+oso.query('builtinSpecializers(true)').next
+raise unless oso.query('builtinSpecializers(false)').to_a.empty?
+oso.query('builtinSpecializers(2)').next
```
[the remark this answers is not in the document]
plotnick: Fixed, thanks.

PR #310, opened 2020-08-05
Python windows support

saolsen:
Only works for 64bit python. Can't build one for 32bit on CI but hopefully nobody really wants to be using that anyway.

If you want to try out the built wheels you can download them here
https://github.com/osohq/oso/suites/1012234333/artifacts/13211355


languages/python/polar/build.py
```diff
@@ -15,7 +15,14 @@
     "CI": "native",
 }
 env = os.environ.get("ENV", "DEVELOPMENT")
-lib = lib_dirs[env] + "/libpolar.a"  # @TODO: Windows
+libs = []
+if sys.platform.startswith("win"):
+    libs.append(lib_dirs[env] + "/polar.lib")
+    libs.append("Ws2_32.lib")
```
samscott89, line 21: How did you find out this is what's needed?

saolsen: googling unknown symbol link errors one at a time to see what lib they're in

samscott89: Nice

PR #311, opened 2020-08-05
Make python API the same as the other ones

leina05:
Updates to the FFI API. Split FFI functions among appropriate classes (`Polar`, `Query`, `QueryEvent`, `Error`). Upshot is that now the FFI `lib` functions only get called from `ffi.py`

### DIFF
--- a/languages/python/polar/ffi.py
+++ b/languages/python/polar/ffi.py
@@ -4,7 +4,7 @@ import json
 
 from _polar_lib import ffi, lib
 
-from .errors import raise_error
+from .errors import get_python_error
 from .exceptions import PolarRuntimeException
 
 
@@ -75,7 +75,7 @@ class Query:
         check_result(lib.polar_application_error(self.ptr, message))
 
     def next_event(self):
-        return ffi_deserialize(lib.polar_next_query_event(self.ptr))
+        return QueryEvent(check_result(lib.polar_next_query_event(self.ptr)))
 
     def debug_command(self, command):
         check_result(lib.polar_debug_command(self.ptr, ffi_serialize(command)))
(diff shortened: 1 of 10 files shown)


languages/python/polar/ffi.py
```diff
@@ -3,91 +3,119 @@
...
-        return json.loads(ffi.string(string).decode())
-    finally:
-        if not is_null(string):
-            lib.string_free(string)
+class QueryEvent:
+    def __init__(self, ptr):
+        self.ptr = ptr
 
+    def get(self):
+        return ffi.string(self.ptr).decode()
```
saolsen, line 90: do we ever actually call del on the QueryEvent to make sure it gets freed?

leina05: Oh good call. We don't but I thought the GC will do it?

saolsen: It wont always call __del__ when it goes out of scope like you’d want it to. Python does some weird object caching stuff. I think this came up before and we found some stackoverflow stuff about how __del__ shouldn’t be used for memory management like this. If you want an automatic thing you have to use a with statement but I think in this case this is fine we should just call del in the main event loop after handling an event.

leina05: Added

languages/python/polar/query.py
```diff
@@ -36,9 +36,10 @@ def run(self):
         """Run the event loop and yield results."""
         assert self.ffi_query, "no query to run"
         while True:
-            event_str = self.ffi_query.next_event().get()
-            event = json.loads(event_str)
+            ffi_event = self.ffi_query.next_event()
+            event = json.loads(ffi_event.get())
             if event == "Done":
```
saolsen, line 41: I think this should happen on line 41 after the event is parsed. if it’s here it’ll only be called for done events

languages/python/polar/query.py
```diff
@@ -40,15 +30,16 @@ def __init__(self, ffi_query, *, host=None):
...
     def run(self):
         """Run the event loop and yield results."""
         assert self.ffi_query, "no query to run"
         while True:
-            event_s = lib.polar_next_query_event(self.ffi_query)
-            event = ffi_deserialize(event_s)
+            ffi_event = self.ffi_query.next_event()
+            event = json.loads(ffi_event.get())
             if event == "Done":
+                del ffi_event
```
saolsen, line 42:
```suggestion
            ffi_event = self.ffi_query.next_event()
            event = json.loads(ffi_event.get())
            del ffi_event
            if event == "Done":
```

leina05: oh yeah you're totally right, I was thinking we still need it while the event loop is running but we don't

PR #312, opened 2020-08-05
remove lazy loading

leina05:
Removed lazy loading; all policy files loaded when the method is called. A bit of logic for detecting duplicates, but nothing to explicitly manage issues that could arise with unregistered classes as a side-effect of decorators. The tests passed without this, so though more edge case testing could be added, it seems that this is a fine way for loading to work. 

PR checklist:

- [ ] Added changelog entry.


srenatus:
Cool. This lazy loading had been a bit of a surprise for me, too.

languages/python/polar/polar.py
```diff
@@ -176,3 +188,19 @@ def wrap(cls):
         return wrap
 
     return wrap(_cls)
+
+
+def hash_file(fname):
+    # BUF_SIZE is totally arbitrary, change for your app!
```
gj, line 194: Is this comment accurate?

languages/ruby/lib/oso/polar/polar.rb
```diff
@@ -28,11 +31,25 @@ def clear
...
-      def load_file(name)
+      def load_file(name) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
         raise PolarFileExtensionError unless ['.pol', '.polar'].include? File.extname(name)
         raise PolarFileNotFoundError, name unless File.file?(name)
 
-        load_queue << name
+        hash = hash_file(name)
+
+        if @loaded_names.key?(name)
+          raise PolarRuntimeError, "File #{name} has already been loaded." unless @loaded_names[name] != hash
```
gj, line 41: bit easier to understand than the double negative, imo

```suggestion
          raise PolarRuntimeError, "File #{name} has already been loaded." if @loaded_names[name] == hash
```

languages/ruby/lib/oso/polar/polar.rb
```diff
@@ -28,11 +31,25 @@ def clear
...
       # @raise [PolarFileNotFoundError] if provided filename does not exist.
-      def load_file(name)
+      def load_file(name) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
         raise PolarFileExtensionError unless ['.pol', '.polar'].include? File.extname(name)
         raise PolarFileNotFoundError, name unless File.file?(name)
 
-        load_queue << name
+        hash = hash_file(name)
+
+        if @loaded_names.key?(name)
```
gj, line 40: the point of having the `private` `attr_reader :loaded_names` below is so that we don't need to directly refer to the instance variable :slightly_smiling_face:

```suggestion
        if loaded_names.key?(name)
```

languages/ruby/lib/oso/polar/polar.rb
```diff
@@ -28,11 +31,25 @@ def clear
...
+          raise PolarRuntimeError, "File #{name} has already been loaded." unless @loaded_names[name] != hash
+
+          raise PolarRuntimeError, "A file with the name #{name}, but different contents,
+            has already been loaded."
```
gj, line 44: Actually on second thought I kinda like the readability a bit more if we reverse the order of these two `raise`s so that the conditional check mirrors the error text:

```suggestion
          raise PolarRuntimeError, "A file with the name #{name} but different contents
          has already been loaded." if loaded_names[name] != hash

          raise PolarRuntimeError, "File #{name} has already been loaded."
```

languages/ruby/lib/oso/polar/polar.rb
```diff
@@ -28,11 +31,25 @@ def clear
...
+        elsif @loaded_contents.key?(hash)
+          raise PolarRuntimeError, "A file with the same contents as #{name} named #{@loaded_contents[hash]}
+            has already been loaded."
+        else
+          File.open(name) { |file| load_str(file.read, filename: name) }
```
gj, line 49: Could we read the file once, then hash it, and then reuse the existing contents here instead of reading it twice?

languages/ruby/lib/oso/polar/polar.rb
gj: nit: would prefer these were new error types instead of `PolarRuntimeError`s with inline messages

languages/ruby/lib/oso/polar/polar.rb
```diff
@@ -28,11 +31,25 @@ def clear
       # @param name [String]
       # @raise [PolarFileExtensionError] if provided filename has invalid extension.
       # @raise [PolarFileNotFoundError] if provided filename does not exist.
-      def load_file(name)
+      def load_file(name) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
         raise PolarFileExtensionError unless ['.pol', '.polar'].include? File.extname(name)
         raise PolarFileNotFoundError, name unless File.file?(name)
```
gj, line 36: as another nit, since we aren't lazy loading we should now just try reading the file here and handle errors instead of checking for existence and then reading the file if it exists.

languages/ruby/lib/oso/polar/polar.rb
leina05: yeah, was like this but got a rubocop error

leina05: I can put the guard clause on the other error

languages/python/polar/polar.py
leina05: nope lol some stack overflow things

languages/python/polar/polar.py
```diff
@@ -54,16 +58,34 @@ def load_file(self, policy_file):
...
             raise PolarApiException(f"Could not find file: {policy_file}")
-        if policy_file not in self.load_queue:
-            self.load_queue.append(policy_file)
 
-    def load_str(self, string):
+        fname = str(policy_file)
+        fhash = hash_file(policy_file)
+        if fname in self.loaded_names.keys():
+            if self.loaded_names.get(fname) == fhash:
+                raise PolarRuntimeException(f"File {fname} has already been loaded.")
```
gj, line 70: We should create the same new error types for Python

languages/python/polar/polar.py
```diff
@@ -54,16 +58,34 @@ def load_file(self, policy_file):
         until a query is made."""
         policy_file = Path(policy_file)
         extension = policy_file.suffix
-        if extension not in (".pol", ".polar"):
-            raise PolarApiException(f"Polar files must have .pol or .polar extension.")
+        if extension not in (".polar"):
```
gj, line 61: nit: no need to keep this as a tuple

languages/python/polar/polar.py
```diff
@@ -54,16 +58,34 @@ def load_file(self, policy_file):
...
+            raise PolarApiException(f"Polar files must have a .polar extension.")
         if not policy_file.exists():
             raise PolarApiException(f"Could not find file: {policy_file}")
```
gj, line 64: I think we should make the same changes as in the Ruby library: try reading the file, translating ENOENT into the appropriate PolarError, and also avoid reading it twice

languages/ruby/lib/oso/polar/polar.rb
```diff
@@ -43,11 +46,31 @@ def clear
       # @param name [String]
       # @raise [PolarFileExtensionError] if provided filename has invalid extension.
       # @raise [PolarFileNotFoundError] if provided filename does not exist.
-      def load_file(name)
+      def load_file(name) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
         raise PolarFileExtensionError unless ['.pol', '.polar'].include? File.extname(name)
```
gj, line 50: You removed `.pol` on the Python version but not here. I haven't checked Java yet, but we should ensure it's the same across all three.

languages/ruby/lib/oso/polar/polar.rb
```diff
@@ -43,11 +46,31 @@ def clear
...
 
-        load_queue << name
+        begin
+          file = File.open(name)
+          file_data = file.read
+          hash = Digest::MD5.hexdigest(file_data)
+          file.close
+        rescue Errno::ENOENT
+          raise PolarFileNotFoundError, name unless File.file?(name)
+        end
```
gj, line 59: nit: I would structure this as
```suggestion
        content = File.open(name) { |f| f.read }
        hash = Digest::MD5.hexdigest(content)
```
and then down at the end of the method you can `rescue` like this:
```rb
def my_method
  ...
rescue Errno::ENOENT
  raise PolarFileNotFoundError, name
end
```

languages/ruby/lib/oso/polar/polar.rb
```diff
@@ -43,11 +46,31 @@ def clear
...
+
+        if loaded_names.key?(name)
+          raise RepeatLoadError, "File #{name} has already been loaded." if loaded_names[name] == hash
```
gj, line 62: nit: I think these two errors should be different types. E.g., in the new JS lib I called 'em:
```ts
export class PolarFileAlreadyLoadedError extends PolarError {
  constructor(file: string) {
    super(`File '${file}' already loaded.`);
    Object.setPrototypeOf(this, PolarFileAlreadyLoadedError.prototype);
  }
}

export class PolarFileContentsChangedError extends PolarError {
  constructor(file: string) {
    super(`File '${file}' already loaded, but contents have changed.`);
    Object.setPrototypeOf(this, PolarFileContentsChangedError.prototype);
  }
}
```
fine w/ whatever names you decide on, but I think they should be different in case end users want to handle them differently w/o string matching on the message.

PR #313, opened 2020-08-05
Java on windows.

saolsen:
You can try out the built jar by downloading it from here.
https://github.com/osohq/oso/suites/1013151849/artifacts/13231221


.github/workflows/release.yml
```diff
@@ -291,10 +291,20 @@ jobs:
...
+        run: |
           javac -cp "../jar/oso-${{ steps.version.outputs.oso_version }}.jar:." Test.java
           java -cp "../jar/oso-${{ steps.version.outputs.oso_version }}.jar:." Test
         working-directory: test
+      - name: Run tests
+        if: runner.os == 'Windows'
+        env:
+          github_ref: ${{ github.ref }}
+        run: |
+          javac -cp "..\jar\oso-$($env:github_ref.Substring(11)).jar;." Test.java
```
samscott89, line 305: Wait what is this doing?

saolsen: Pulling the version number off the `github_ref` environment variable which is something like `refs/tags/v0.0.0`

samscott89: Ahhh, okay from index 11 onwards. Makes sense.

PR #314, opened 2020-08-06
Enable RuboCop + Solargraph checks in CI

gj:
Came across this stuff while using the Ruby lib as the template for the forthcoming JS lib.


saolsen:
Can we add rubocop to the tests like we do for rustfmt, clippy and black?

gj:
> Can we add rubocop to the tests like we do for rustfmt, clippy and black?

That's a good idea. I'll look into it.

gj:
Well, yes it's trivially possible to add it to the tests, but we currently have 57 offenses haha. Will fix; shouldn't take long.

plotnick:
Could you please add a `Makefile` target, like `clippy`? I like to run these things locally before I push.

gj:
> Could you please add a `Makefile` target, like `clippy`? I like to run these things locally before I push.

#317

PR #318, opened 2020-08-06
Create polar-wasm-api crate

gj:


polar-wasm-api/Makefile
```diff
@@ -0,0 +1,10 @@
+.PHONY: build test
+
+build:
+	wasm-pack build --dev --target nodejs
+	rm pkg/.gitignore pkg/package.json
+	cp --recursive pkg/. ../languages/js/lib
+	rm -rf pkg
+
+test:
+	wasm-pack test --node
```
saolsen, line 10: If they also pass on a headless browser might be worth just running those tests now too to make doing the browser javascript api easier later. If they don't it'll take work anyway and we can just do that later.

gj: They pass in Chrome on my machine -- I'll try running 'em in CI and see what happens. Firefox tests work but not headless, and I don't have Safari on this machine.

gj: On second thought, after a bit of googling I can't figure out how to run the same set of tests on Node + a browser since (AFAICT) there's a special macro to alert wasm-pack that they should be run in a browser:

```rs
wasm_bindgen_test::wasm_bindgen_test_configure!(run_in_browser);
```

and I don't really feel like having two identical copies of the same test files, one with that macro and one without, which is all my tired brain can come up with right now.

We can circle back when browser support becomes necessary.

polar-wasm-api/Makefile
```diff
@@ -0,0 +1,10 @@
+.PHONY: build test
+
+build:
+	wasm-pack build --dev --target nodejs
```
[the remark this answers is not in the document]
gj: Yeah, we can do that when we actually have something to release :slightly_smiling_face:

gj:
> It looks real good. Do we wanna add any tests of loading and calling these methods from javascript as part of this or just wait till the work on the javascript library?

Gonna wait on those until the JS lib. The tests in this PR assert against the JS types from the `js_sys` crate, so they're pretty solid tests of the JS API already.

PR #319, opened 2020-08-06
Boa constructors

plotnick:
Java does not support keywords, so cross-language classes now use boa ("By Order of Arguments") constructors exclusively; e.g., `new User(name)`. Also made certain "constructors" static methods, and renamed, e.g., `by_id/byId` → `id` to be consistent across languages.

PR checklist:
- [ ] Java `@PolarClass` annotation
- [x] Docs
- [x] Changelog entry

### DIFF
--- a/docs/examples/abac/java/TestAbac.java
+++ b/docs/examples/abac/java/TestAbac.java
@@ -47,17 +47,17 @@ public class TestAbac {
     public static void testRbac02() throws Exception {
         Oso oso = setupOso();
         oso.loadFile("02-rbac.polar");
-        oso.loadStr("role(_: User { name: \"sam\" }, \"admin\", __: Project { id: 2 });");
+        oso.loadStr("role(_: User { name: \"sam\" }, \"admin\", _: Project { id: 2 });");
 
         Expense expense = new Expense(50, "steve", "NYC", 0);
         if (oso.isAllowed(new User("sam"), "view", expense)) {
             throw new Exception("ABAC docs test failed!");
         }
 
-        /*expense = new Expense(50, "steve", "NYC", 2);
+        expense = new Expense(50, "steve", "NYC", 2);
         if (!oso.isAllowed(new User("sam"), "view", expense)) {
             throw new Exception("ABAC docs test failed!");
-        }*/
+        }
     }
 
     public static void testHierarchy03() throws Exception {
--- a/docs/examples/abac/java/TestAbac.java
+++ b/docs/examples/abac/java/TestAbac.java
@@ -11,11 +11,11 @@ public class TestAbac {
 
     public static Oso setupOso() throws Exception {
         Oso oso = new Oso();
-        oso.registerClass(Expense.class, (args) -> Expense.byId((Integer) args.get("id")));
-        oso.registerClass(User.class, (args) -> new User((String) args.get("name")));
-        oso.registerClass(Project.class, (args) -> Project.byId((Integer) args.get("id")));
-        oso.registerClass(Team.class, (args) -> Team.byId((Integer) args.get("id")));
-        oso.registerClass(Organization.class, (args) -> Organization.byId((Integer) args.get("id")));
+        oso.registerClass(Expense.class);
+        oso.registerClass(User.class);
+        oso.registerClass(Project.class);
+        oso.registerClass(Team.class);
+        oso.registerClass(Organization.class);
 
         return oso;
 
@@ -53,11 +53,11 @@ public class TestAbac {
         if (oso.isAllowed(new User("sam"), "view", expense)) {
             throw new Exception("ABAC docs test failed!");
         }
-        expense = new Expense(50, "steve", "NYC", 2);
 
+        /*expense = new Expense(50, "steve", "NYC", 2);
         if (!oso.isAllowed(new User("sam"), "view", expense)) {
             throw new Exception("ABAC docs test failed!");
-        }
+        }*/
     }
 
     public static void testHierarchy03() throws Exception {
--- a/languages/java/oso/src/main/java/com/osohq/oso/Host.java
+++ b/languages/java/oso/src/main/java/com/osohq/oso/Host.java
@@ -33,9 +33,6 @@ public class Host implements Cloneable {
 
     /**
      * Get a registered Java class.
-     *
-     * @param name
-     * @throws Exceptions.UnregisteredClassError
      */
     public Class<?> getClass(String name) throws Exceptions.UnregisteredClassError {
         if (classes.containsKey(name)) {
@@ -48,9 +45,8 @@ public class Host implements Cloneable {
     /**
      * Store a Java class in the cache by name.
      *
-     * @param name
-     * @throws Exceptions.DuplicateClassAliasError If the class is already
-     *                                             registered.
+     * @param name The name of the class from within Polar.
+     * @throws Exceptions.DuplicateClassAliasError If the name is already registered.
      */
     public String cacheClass(Class<?> cls, Constructor<?> constructor, String name)
             throws Exceptions.DuplicateClassAliasError {
@@ -64,9 +60,6 @@ public class Host implements Cloneable {
 
     /**
      * Get a cached Java instance.
-     *
-     * @param instanceId
-     * @throws Exceptions.UnregisteredInstanceError
      */
     public Object getInstance(long instanceId) throws Exceptions.UnregisteredInstanceError {
         if (hasInstance(instanceId)) {
@@ -78,8 +71,6 @@ public class Host implements Cloneable {
 
     /**
      * Determine if a Java instance has been cached.
-     *
-     * @param instanceId
      */
     public boolean hasInstance(long instanceId) {
         return instances.containsKey(instanceId);
@@ -87,10 +78,6 @@ public class Host implements Cloneable {
 
     /**
      * Cache an instance of a Java class.
-     *
-     * @param instance
-     * @param id
-     * @throws Exceptions.OsoException
      */
     public Long cacheInstance(Object instance, Long id) throws Exceptions.OsoException {
         if (id == null) {
@@ -104,10 +91,6 @@ public class Host implements Cloneable {
     /**
      * Make an instance of a Java class from a {@code Map<String, Object>} of
      * fields.
-     *
-     * @param className
-     * @param fields
-     * @param id
      */
     public Object makeInstance(String className, List<Object> initargs, long id)
             throws Exceptions.OsoException
@@ -147,12 +130,6 @@ public class Host implements Cloneable {
 
     /**
      * Check if a class specializer is more specific than another class specializer.
-     *
-     * @param instanceId
-     * @param leftTag
-     * @param rightTag
-     * @return
-     * @throws Exceptions.UnregisteredClassError
      */
     public boolean subspecializer(long instanceId, String leftTag, String rightTag)
             throws Exceptions.UnregisteredClassError {
@@ -178,13 +155,6 @@ public class Host implements Cloneable {
 
     /**
      * Check if a Java instance is an instance of a class.
-     *
-     * @param instance
-     * @param classTag
-     * @return
-     * @throws Exceptions.UnregisteredClassError
-     * @throws Exceptions.UnregisteredInstanceError
-     * @throws Exceptions.UnexpectedPolarTypeError
      */
     public boolean isa(JSONObject instance, String classTag)
             throws Exceptions.UnregisteredClassError,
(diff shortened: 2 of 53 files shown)


languages/java/oso/src/main/java/com/osohq/oso/Host.java
```diff
@@ -103,37 +91,50 @@ public Long cacheInstance(Object instance, Long id) throws Exceptions.OsoExcepti
...
+                if (argTypes.length == paramTypes.length) {
+                    for (int i = 0; i < paramTypes.length; i++) {
+                        if (!paramTypes[i].isAssignableFrom(argTypes[i])) {
+                            continue search;
+                        }
+                    }
+                    constructor = c;
+                    break search;
+                }
+            }
```
leina05, line 116: Is there a reason not to use [`getConstructor()`](https://docs.oracle.com/javase/7/docs/api/java/lang/Class.html#getConstructor(java.lang.Class...))? If it's because of `isAssignableFrom`, why is this necessary?

languages/java/oso/src/main/java/com/osohq/oso/Query.java
```diff
@@ -113,22 +106,32 @@ private void handleCall(String attrName, JSONArray jArgs, JSONObject polarInstan
...
                     }
-                    String clsName = data.getJSONObject("instance").getString("tag");
-                    JSONObject jFields = data.getJSONObject("instance").getJSONObject("fields").getJSONObject("fields");
-                    host.makeInstance(clsName, host.polarDictToJava(jFields), id);
+                    JSONObject constructor = data.getJSONObject("constructor").getJSONObject("value");
+                    JSONArray initargs;
+                    if (constructor.has("InstanceLiteral")) {
+                        // Keyword initargs are not supported in Java.
+                        className = constructor.getJSONObject("InstanceLiteral").getString("tag");
+                        throw new Exceptions.InstantiationError(className);
```
leina05, line 114: This is what I was envisioning to allow support of both kwargs and positional args, but now I'm confused. Are kwarg constructors still supported for other languages?

languages/python/polar/query.py
```diff
@@ -67,10 +67,17 @@ def run(self):
...
+        if "InstanceLiteral" in constructor:
+            cls_name = constructor["InstanceLiteral"]["tag"]
+            fields = constructor["InstanceLiteral"]["fields"]["fields"]
+            initargs = {k: self.host.to_python(v) for k, v in fields.items()}
+        elif "Call" in constructor:
+            cls_name = constructor["Call"]["name"]
+            initargs = [self.host.to_python(arg) for arg in constructor["Call"]["args"]]
+        else:
+            raise PolarApiException("Bad constructor")
+        self.host.make_instance(cls_name, initargs, id)
```
leina05, line 80: Ah okay I see, so construction can still be done by instance literal.

languages/python/tests/parity/policies/test_api.polar
```diff
@@ -6,19 +6,19 @@ actorInRole(actor, role, resource: Widget) if
     role = resource.company.role(actor);
 
 allow(actor, "get", _: Http{path: path}) if
-    new PathMapper{template: "/widget/{id}"}.map(path) = {id: id} and
+    new PathMapper("/widget/{id}").map(path) = {id: id} and
```
leina05, line 9: If it's still possible to construct with instance literals in python, why have all of these been changed?

languages/ruby/lib/oso/polar/query.rb
```diff
@@ -118,9 +118,18 @@ def start # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metri
...
+                cls_name = constructor['InstanceLiteral']['tag']
+                fields = constructor['InstanceLiteral']['fields']['fields']
+                initargs = Hash[fields.map { |k, v| [k.to_sym, host.to_ruby(v)] }]
+              elsif constructor.key? 'Call'
+                cls_name = constructor['Call']['name']
+                initargs = constructor['Call']['args'].map { |arg| host.to_ruby(arg) }
+              else
+                raise InvalidConstructorError
+              end
+              host.make_instance(cls_name, initargs: initargs, id: id)
```
leina05, line 132: Could this be split into a separate function?

languages/ruby/spec/oso/oso_spec.rb
```diff
@@ -80,14 +80,14 @@ def initialize(id:)
         subject.register_class(Widget)
         subject.load_str <<~POLAR
           allow(actor, "get", _: Http{path: path}) if
-              new PathMapper{template: "/widget/{id}"}.map(path) = {id: id} and
+              new PathMapper("/widget/{id}").map(path) = {id: id} and
               allow(actor, "get", new Widget{id: id});
           allow(_actor, "get", widget: Widget{}) if widget.id = "12";
         POLAR
-        widget12 = Oso::Http.new(path: '/widget/12')
+        widget12 = Oso::Http.new('host', '/widget/12', {})
```
leina05, line 87: What is this doing?

polar-core/src/parser.rs
```diff
@@ -228,13 +228,18 @@ mod tests {
...
         assert_eq!(results[0].to_polar(), r#"a(x) if x = new Foo{a: 1};"#);
     }
 
+    #[test]
+    fn test_parse_new_boa_constructor() {
+        let f = r#"a(x) if x = new Foo(1, 2);"#;
+        let results = parse_rules(0, f).unwrap();
+        assert_eq!(results[0].to_polar(), r#"a(x) if x = new Foo(1, 2);"#);
+    }
+
```
leina05, line 242: Could we include a test of the error that results from attempting to call with a combination of positional and kwargs? E.g., `Foo(1, bar: 2)`. Not sure how common it would be but if the syntax is similar enough to Ruby people might try it and it should have a decent error.

docs/examples/abac/java/TestAbac.java
```diff
@@ -53,11 +53,11 @@ public static void testRbac02() throws Exception {
         if (oso.isAllowed(new User("sam"), "view", expense)) {
             throw new Exception("ABAC docs test failed!");
         }
-        expense = new Expense(50, "steve", "NYC", 2);
 
+        /*expense = new Expense(50, "steve", "NYC", 2);
         if (!oso.isAllowed(new User("sam"), "view", expense)) {
             throw new Exception("ABAC docs test failed!");
-        }
+        }*/
```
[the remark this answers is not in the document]
plotnick: Apparently not, but I don't really understand why or why not.

languages/java/oso/src/main/java/com/osohq/oso/Host.java
plotnick: Consider this constructor: `public Http(String hostname, String path, Map<String, String> query)`. This is the right signature for it to have, but `Map` is an interface; no actual argument will ever have `arg.class == Map`. But a `HashMap`, say, `isAssignableFrom` a `Map`, as is any subclass.

The `Class.getMethod` method does all this, but `getConstructor` doesn't. I don't think there's a good reason for this, other than "because Java".

languages/java/oso/src/main/java/com/osohq/oso/Query.java
plotnick: Yes.

languages/python/tests/parity/policies/test_api.polar
plotnick: Because we have shared policy code that uses `PathMapper` and `Http`, they can only use the common denominator of boa constructors.

languages/ruby/lib/oso/polar/query.rb
plotnick: I don't think that would buy anything. This logic is inherent to the task of instantiating an instance in one of several ways.

languages/ruby/spec/oso/oso_spec.rb
plotnick: See above; because Java has to use this interface, we have to use positional args, so a host and query must be provided.

polar-core/src/parser.rs
plotnick: Ugh. These errors are not handled well at all, but it's not specific to constructors. Can we please make this a follow-up task? E.g.,
```
python -m oso
query> f(x:x)
  File "/home/alex/oso/oso/languages/python/polar/polar.py", line 124, in repl
    ffi_query = self.ffi_polar.new_query_from_str(query)
  File "/home/alex/oso/oso/languages/python/polar/ffi.py", line 30, in new_query_from_str
    check_result(lib.polar_new_query(self.ptr, to_c_str(query_str), 0))
  File "/home/alex/oso/oso/languages/python/polar/ffi.py", line 108, in check_result
    raise Error().get()
polar.exceptions.UnrecognizedToken: did not expect to find the token ':' at line 1, column 4

During handling of the above exception, another exception occurred:

  File "/home/alex/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/home/alex/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/alex/oso/oso/languages/python/oso/__main__.py", line 3, in <module>
    Oso().repl(load=True)
  File "/home/alex/oso/oso/languages/python/polar/polar.py", line 126, in repl
    print("Parse error: ", str(e.value()))
AttributeError: 'UnrecognizedToken' object has no attribute 'value'
```

That said, I think this is the right syntax for us to use for mixed positional/keyword args. I think it's more consistent with the rest of the language than a Python-style `f(a=b)`.

languages/python/polar/query.py
plotnick: Right.

languages/java/oso/src/main/java/com/osohq/oso/Host.java
leina05: ugh wow

languages/ruby/lib/oso/polar/query.rb
leina05: Right, but then maybe it should go in `#make_instance` instead of in `#start`, and `#make_instance` gets called with the `constructor` JSON object?

polar-core/src/parser.rs
leina05: Sure, can take this as a follow-up. That error is bad, but that's not actually the syntax I meant to check. I meant something like `query> x = new Foo(id: 1)`.

languages/ruby/lib/oso/polar/query.rb
plotnick: I'm sorry, my comment was poorly phrased. What I meant was that this is the decoding logic for the two cases we have now (boa and kwargs), and I like the strict separation we have now between decoding the messages (in `Query`) and implementing the host-specific bits (in `Host`). So I don't think it belongs in `Host.make_instance`.

polar-core/src/parser.rs
leina05: This is what I get with that, which actually isn't too bad: 

```
polar = <polar.polar.Polar object at 0x10b5da610>, qeval = <function qeval.<locals>._qeval at 0x10b6c0d40>

    def test_boa_constructor(polar, qeval):
        class Foo:
            def __init__(self, id=0):
                self.id = 1
    
        polar.register_class(Foo)
>       polar.load_str("f(x) if x = new Foo(id: 1);")

tests/test_polar.py:663: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
polar/polar.py:66: in load_str
    self.ffi_polar.load_str(string, None)
polar/ffi.py:26: in load_str
    check_result(lib.polar_load(self.ptr, string, filename))
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

result = 0

    def check_result(result):
        if result == 0 or is_null(result):
>           raise Error().get()
E           polar.exceptions.UnrecognizedToken: did not expect to find the token ':' at line 1, column 23

polar/ffi.py:108: UnrecognizedToken
```

leina05:
LGTM as once docs + changelog are updated

plotnick:
If anybody has a few minutes to proof read the doc changes, that'd be great.

PR #320, opened 2020-08-07
Initial Node.js library

gj:
PR checklist:

- [ ] Added changelog entry.


gj:
Follow-ups:

- [ ] Test REPL and create `npm run oso repl` / `yarn oso repl` CLI entrypoint.
- [ ] More tests on JS data types, especially the various generator constructions in `Query#registerCall`.
- [ ] Documentation, especially of divergence from other language libs like the lack of custom constructors, no support for kwargs (but the JS convention of using an object to simulate kwargs works fine), weirdness arising from using `==` for `Host#unify` (not in this PR, but will be changed to that in a follow-up), etc.
- [ ] Changelog entry (when we're ready to announce Node.js support).

PR #321, opened 2020-08-07
musl support

saolsen:
Builds musl wheels that can be used on alpine.
Doesn't use the musl-static build target for the regular linux builds.


scripts/test_musl.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env sh
+apk add build-base libffi-dev
+cd oso
+pip install oso==$OSO_VERSION -f musl-wheel
+cd test
+python test.py
```
dhatch, line 6: Should we run the full test suite against this?

PR #322, opened 2020-08-07
Fix bug with checking for element in an empty list.

samscott89:
PR checklist:

- [x] Added changelog entry.


polar-core/tests/integration_tests.rs
```diff
@@ -1120,6 +1120,10 @@ fn test_in() {
     assert!(qnull(&mut polar, "not (1 in [1,2,3])"));
     assert!(qnull(&mut polar, "not (2 in [1,2,3])"));
     assert!(qnull(&mut polar, "not (3 in [1,2,3])"));
+
+    // empty lists
+    assert!(qnull(&mut polar, "x in []"));
+    assert!(qeval(&mut polar, "not x in []"));
```
dhatch, line 1126: Let's add a few tests for non variable cases (`2 in []`, etc.).

PR #326, opened 2020-08-09
μ-optimizations

plotnick:
Benchmark | Time | Change
----------|------|-------
`many_rules` | 254.53 us | -45.642%
`fib` |  407.14 us | -16.342%

All other benchmarks within noise threshold.


dhatch:
It was super easy to change Term to use `Arc` because we put it inside the Term type.  What do you think about doing the same here? Or, we could have a type alias at least `ArcRule` or `BoxedRule` (so it's generic to the pointer type used internally a bit more)?

plotnick:
>Or, we could have a type alias at least `ArcRule` or `BoxedRule` (so it's generic to the pointer type used internally a bit more)?

That's a good suggestion, thanks. I think we should also have a `BoxedGoal = Rc<Goal>` for consistency.

samscott89:
> > Or, we could have a type alias at least `ArcRule` or `BoxedRule` ...
> 
> That's a good suggestion, thanks. I think we should also have a `BoxedGoal = Rc<Goal>` for consistency.

Are we doing this in this PR?

plotnick:
> > > Or, we could have a type alias at least `ArcRule` or `BoxedRule` ...
> > 
> > That's a good suggestion, thanks. I think we should also have a `BoxedGoal = Rc<Goal>` for consistency.
>
> Are we doing this in this PR?

I haven't gotten to it yet. I could do it as part of indexing, though. I don't think it needs to go in with this.

plotnick:
Do we need a changelog entry for this? "Improved performance on our own micro-benchmarks" doesn't really sound that great.

samscott89:
> Do we need a changelog entry for this? "Improved performance on our own micro-benchmarks" doesn't really sound that great.

No, I don't think we do :)

PR #328, opened 2020-08-10
Index rules

plotnick:
Make `Value` implement `Hash` so that we can index on non-variable, unspecialized rule parameters.

Benchmark | Time | Change
----------|------|-------
`many_rules` | 55.444 us | -78.514%

Change is relative to #326.

PR checklist:
- [x] Add changelog entry.
- [x] Fix fib bench


polar-core/src/numerics.rs
```diff
@@ -70,11 +70,8 @@ impl Hash for Numeric {
         H: Hasher,
     {
         match self {
-            Numeric::Integer(i) => i,
-            Numeric::Float(f) => unsafe {
-                #[allow(clippy::transmute_ptr_to_ptr)]
-                std::mem::transmute(f)
-            },
+            Numeric::Integer(i) => *i as u64,
+            Numeric::Float(f) => f.to_bits(),
```
samscott89, line 74: You might want to hash the [discriminant](https://doc.rust-lang.org/std/mem/fn.discriminant.html) first so that ints and floats dont hash to the same thing in certain circumstances? Unless that's intentional?

plotnick: That is a good idea, thanks.

polar-core/src/numerics.rs
```diff
@@ -61,9 +66,67 @@ impl PartialEq for Numeric {
...
+                *i as u64
+            }
+            Numeric::Float(f) => match f.classify() {
+                FpCategory::Zero => {
+                    // Canonicalize zero representations.
+                    discriminant(&Numeric::Integer(0)).hash(state);
+                    0u64
+                }
+                FpCategory::Nan => {
+                    // Randomize NaN hashes so they always miss.
```
samscott89, line 97: This is raising a bunch of red flags.

polar-core/src/numerics.rs
```diff
@@ -61,9 +66,67 @@ impl PartialEq for Numeric {
...
+                    discriminant(&FpCategory::Nan).hash(state);
+                    f64::from_bits(f.to_bits() | random::<u64>()).to_bits()
```
samscott89, line 99: Not sure what ORing with f.bits() is achieving (`f64::NAN.to_bits()` returns `111111111111000000000000000000000000000000000000000000000000000`), so it's just setting the top 12 bits to 1?)

polar-core/src/numerics.rs
plotnick: Here's my rationale: `NaN != NaN`, for any `NaN`, and therefore `hash(NaN) != hash(NaN)` for any `NaN`. The only alternative I've thought of is to treat `NaN` as non-ground, but this seems simpler and cleaner, if a little tricky.

polar-core/src/numerics.rs
plotnick: No, in a `NaN` the exponent bits (11 of the top 12) are always 1. So this keeps it a `NaN`, but randomizes its payload.

plotnick: Shoot. Realized just now that if the `NaN` is already all 1s, this will not do anything. Need to `xor` it.

polar-core/src/types.rs
```diff
@@ -308,6 +325,7 @@ impl SourceInfo {
 
 /// Represents a concrete instance of a Polar value
 #[derive(Debug, Clone, Serialize, Deserialize)]
+#[allow(clippy::derive_hash_xor_eq)]
```
samscott89, line 328:
```suggestion
```

polar-core/src/numerics.rs
samscott89: (Repeating what I said in person): I think it's okay for NaNs to hash equals to each other, and for us to treat rule filtering as not being a strict process. I.e. it can let more things through than it should. We will still unify args/params when selecting a rule so it doesn't invalidate that.

Alternatively/additionally, marking NaN as not ground I think would be a good idea.

Personally I disagree its simpler/cleaner: introduces a new dependency which relies on OS features == potentially more trouble. And it makes hashing non-deterministic which is sort of worrying.

polar-core/tests/integration_tests.rs
```diff
@@ -1465,4 +1473,6 @@ fn test_numeric_applicability() {
     assert!(qeval(&mut polar, "f(9223372036854775807)"));
     assert!(qeval(&mut polar, "f(-9223372036854775807)"));
     assert!(qeval(&mut polar, "f(9223372036854776000.0)"));
+    assert!(qnull(&mut polar, "f(nan1)"));
+    assert!(qnull(&mut polar, "f(nan2)"));
```
samscott89, line 1477: 👍

polar-core/src/numerics.rs
plotnick: Abandoned.

polar-core/src/formatting.rs
```diff
@@ -106,6 +104,16 @@ pub fn format_params(args: &[Parameter], sep: &str) -> String {
         .join(sep)
 }
 
+/// Formats a vector of rules as a string-separated list.
+#[allow(clippy::ptr_arg)]
+pub fn format_rules(rules: &Rules, sep: &str) -> String {
```
gj, line 109: We could probably make this + `format_params` (above) generic over `&[T]` where `T: ToPolarString`, but it doesn't have to block this PR.

polar-core/src/numerics.rs
```diff
@@ -195,5 +255,77 @@ mod tests {
     }
 
     #[test]
-    fn test_sanity() {}
+    fn test_numeric_hash() {
+        let nan1 = f64::NAN;
+        let nan2 = f64::from_bits(f64::NAN.to_bits() | 0xDEADBEEF); // frob the payload
```
gj, line 260: :cow2:

PR #332, opened 2020-08-10
Wrap IDs after 52 bits of precision

gj:
This limits us to 9,007,199,254,740,991 unique IDs for calls, instances, etc. per KB and then wraps back around to 1.

### DIFF
--- a/polar-core/src/types.rs
+++ b/polar-core/src/types.rs
@@ -574,6 +574,8 @@ pub struct KnowledgeBase {
     pub inline_queries: Vec<Term>,
 }
 
+const MAX_ID: u64 = (1 << 53) - 1;
+
 impl KnowledgeBase {
     pub fn new() -> Self {
         Self {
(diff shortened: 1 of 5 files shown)


polar-core/src/types.rs
```diff
@@ -588,8 +590,16 @@ impl KnowledgeBase {
     }
 
     /// Return a monotonically increasing integer ID.
+    ///
+    /// Wraps around at 52 bits of precision so that it can be safely coerced to an IEEE-754
+    /// double-float (f64).
     pub fn new_id(&self) -> u64 {
-        self.id_counter.fetch_add(1, Ordering::SeqCst)
+        if self.id_counter.load(Ordering::SeqCst) == MAX_ID {
```
samscott89, line 597: If you do the check after the fetch, you save an operation on most calls.

samscott89: Actually, might need to do https://doc.rust-lang.org/std/sync/atomic/struct.AtomicUsize.html#method.compare_exchange instead. Otherwise, there's a weird race condition where two threads might both set the counter to 1.
(load -> load -> store -> store).

gj: Yeah I think you're right. Changing now.

PR #337, opened 2020-08-11
Start 0.4.1-rc0 version.

samscott89:


docs/changelogs/0.4.0.rst
```diff
@@ -0,0 +1,98 @@
...
+Breaking changes
+================
+
+.. warning:: This release contains breaking changes. Be sure
+   to follow migration steps before upgrading.
+
+Breaking change 1
+-----------------
+
+- summary of breaking change
```
samscott89, line 16: Still need to update for the lazy loading breaking change.

languages/ruby/lib/oso/version.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module Oso
-  VERSION = '0.4.0'
+  VERSION = '0.4.1-rc0'
```
saolsen, line 4: You have to update `Gemfile.lock` too.

languages/java/oso/pom.xml
```diff
@@ -97,7 +97,7 @@
       <plugin>
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-javadoc-plugin</artifactId>
-          <version>2.9.1</version>
+          <version>3.0.1</version>
```
saolsen, line 100: I don't think we can upgrade this without testing that it's compatible with the maven central stuff.

PR #339, opened 2020-08-11
Add CI & release documentation for flask_oso

dhatch:
I also made a few changes to the main oso release workflow to:

- quiet down maven output
- use the existing tag with a release (`v0.5.0`) instead of making a new duplicate `0.5.0` tag.


gj:
This PR would probably be a good candidate to squash into fewer commits so that the ultimate squash commit to main has a cleaner commit message.

PR #349, opened 2020-08-14
Build JS Package in CI

saolsen:
Also includes fixes for the release job. Pre-releases and python builds are both broken right now.


gj:
Will address requested changes on #320

PR #356, opened 2020-08-18
Separate attribute and method lookups

samscott89:
Attribute lookups and method calls (with no args) are no longer elided into the same thing.
`x.foo` will only look for an attribute on `x` called `foo`, and `x.foo()` will only look for a method on `x` called `foo`.

Todo:
- [x] Rust changes
- [x] Python
- [x] Ruby
- [x] Java
- [x] Javascript
- [x] Test for returning method as attribute
- [x] Documentation
- [x] Examples/doctests

PR checklist:

- [x] Added changelog entry.

### DIFF
--- a/languages/java/oso/src/main/java/com/osohq/oso/Exceptions.java
+++ b/languages/java/oso/src/main/java/com/osohq/oso/Exceptions.java
@@ -194,6 +194,16 @@ public class Exceptions {
         }
     }
 
+    public static class InvalidAttributeError extends PolarRuntimeException {
+        public InvalidAttributeError(String className, String attrName) {
+            super("Invalid attribute `" + attrName + "` on class " + className);
+        }
+
+        public InvalidAttributeError(String msg) {
+            super(msg);
+        }
+    }
+
     public static class InvalidConstructorError extends PolarRuntimeException {
         public InvalidConstructorError(String msg) {
             super(msg);
--- a/languages/java/oso/src/main/java/com/osohq/oso/Query.java
+++ b/languages/java/oso/src/main/java/com/osohq/oso/Query.java
@@ -59,9 +59,12 @@ public class Query implements Enumeration<HashMap<String, Object>> {
     /**
      * Helper for `ExternalCall` query events
      */
-    private void handleCall(String attrName, JSONArray jArgs, JSONObject polarInstance, long callId)
+    private void handleCall(String attrName, Optional<JSONArray> jArgs, JSONObject polarInstance, long callId)
             throws Exceptions.OsoException {
-        List<Object> args = host.polarListToJava(jArgs);
+        Optional<List<Object>> args = Optional.empty();
+        if (jArgs.isPresent()) {
+            args =  Optional.of(host.polarListToJava(jArgs.get()));
+        }
         try {
             registerCall(attrName, args, callId, polarInstance);
         } catch (Exceptions.InvalidCallError e) {
@@ -126,7 +129,11 @@ public class Query implements Enumeration<HashMap<String, Object>> {
                     instance = data.getJSONObject("instance");
                     callId = data.getLong("call_id");
                     String attrName = data.getString("attribute");
-                    JSONArray jArgs = data.getJSONArray("args");
+
+                    Optional<JSONArray> jArgs = Optional.empty();
+                    if (!data.get("args").equals(null)) {
+                        jArgs = Optional.of(data.getJSONArray("args"));
+                    }
                     handleCall(attrName, jArgs, instance, callId);
                     break;
                 case "ExternalIsa":
@@ -177,7 +184,7 @@ public class Query implements Enumeration<HashMap<String, Object>> {
      * @param polarInstance JSONObject containing either an instance_id or an
      *                      instance of a built-in type.
      */
-    public void registerCall(String attrName, List<Object> args, long callId, JSONObject polarInstance)
+    public void registerCall(String attrName, Optional<List<Object>> args, long callId, JSONObject polarInstance)
             throws Exceptions.InvalidCallError,
                    Exceptions.UnregisteredInstanceError,
                    Exceptions.UnexpectedPolarTypeError {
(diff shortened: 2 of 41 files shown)


languages/java/oso/src/main/java/com/osohq/oso/Query.java
```diff
@@ -192,24 +199,27 @@ public void registerCall(String attrName, List<Object> args, long callId, JSONOb
...
+                    throw new Exceptions.InvalidCallError(cls.getName(), attrName, argTypes);
+                }
+            } else {
                 // Look for a field with the given name.
                 try {
                     Field field = cls.getField(attrName);
                     result = field.get(instance);
                 } catch (NoSuchFieldException f) {
-                    throw new Exceptions.InvalidCallError(cls.getName(), attrName, argTypes);
+                    throw new Exceptions.InvalidCallError(cls.getName(), attrName);
```
gj, line 222: Feels like this should be a distinct type of error, something like `Exceptions.InvalidFieldAccessError`

polar-core/src/vm.rs
```diff
@@ -988,12 +995,18 @@ impl PolarVirtualMachine {
...
-        let (field_name, args) = match &field.value() {
+        let (field_name, args) = match self.deref(field).value() {
             Value::Call(Predicate { name, args }) => (
                 name.clone(),
-                args.iter().map(|arg| self.deref(arg)).collect(),
+                Some(args.iter().map(|arg| self.deref(arg)).collect()),
             ),
-            _ => unreachable!("call must be a predicate"),
+            Value::String(field) => (Symbol(field.clone()), None),
+            v => {
```
gj, line 1004: Is this path actually reachable?

polar-core/src/vm.rs
```diff
@@ -1374,15 +1387,15 @@ impl PolarVirtualMachine {
...
         assert_eq!(args.len(), 3);
         let object = self.deref(&args[0]);
-        let field = &args[1];
+        let field = self.deref(&args[1]);
         let value = &args[2];
 
         match object.value() {
             // Push a `Lookup` goal for simple field lookups on dictionaries.
-            Value::Dictionary(dict) if !matches!(field.value(), Value::Call(predicate) if !predicate.args.is_empty()) => {
+            Value::Dictionary(dict) if matches!(field.value(), Value::String(_) | Value::Variable(_)) => {
```
gj, line 1395: We allow the RHS of a lookup to be a variable at this point?

languages/java/oso/src/main/java/com/osohq/oso/Query.java
samscott89: Agreed, just did the same in TS.

polar-core/src/vm.rs
samscott89: Yep, I changed the parser so a `x.y` parses as `.(x, "y")`

gj: To be clear, I'm asking about the catch-all path here, `v => {`

polar-core/src/vm.rs
gj: That's really cool

polar-core/src/vm.rs
samscott89: Ohh right. I guess you could if you do `x.(1)` or something, or manually construct `.(d, 1, value)`.

languages/python/oso/polar/query.py
```diff
@@ -99,9 +94,15 @@ def handle_external_call(self, data):
...
-            if callable(attr):  # If it's a function, call it with the args.
+            if (
+                callable(attr) and not data["args"] is None
+            ):  # If it's a function, call it with the args.
+                args = [self.host.to_python(arg) for arg in data["args"]]
                 result = attr(*args)
+            elif not data["args"] is None:
+                raise RuntimeError(
+                    f"tried to call '{attribute}' but it is not callable"
+                )
```
gj, line 105: nit: would prefer this as a new exception type that can be caught by users of the library

languages/java/oso/src/main/java/com/osohq/oso/Exceptions.java
```diff
@@ -194,6 +194,16 @@ public InvalidCallError(String msg) {
...
     }
 
+    public static class InvalidAttributeError extends PolarRuntimeException {
+        public InvalidAttributeError(String className, String attrName) {
+            super("Invalid attribute `" + attrName + "` on class " + className);
+        }
+
+        public InvalidAttributeError(String msg) {
+            super(msg);
+        }
```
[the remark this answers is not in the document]
samscott89: We don't you're right. I just copied it from the InvalidCallError

languages/js/src/Polar.test.ts
```diff
@@ -200,6 +209,7 @@ describe('#registerClass', () => {
         'animal',
       ]);
       expect(qvar(p, `what_is(${canine}, r)`, 'r')).toStrictEqual([
+        undefined, // Canine has no species, so looking up the 'species' prop returns undefined.
```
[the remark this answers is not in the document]
samscott89: I guess this is where we're inconsistent on handling None/nil/undefined?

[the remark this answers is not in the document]
gj: Not sure what the broader implications are, but I just added this comment in the code (and highlighted it here) to raise awareness of that behavior.

languages/java/oso/src/main/java/com/osohq/oso/Exceptions.java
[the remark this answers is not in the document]
samscott89: It's not doing much harm. Leaving it for now.

languages/js/src/Polar.test.ts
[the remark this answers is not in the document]
samscott89: Thanks. I linked it to the Notion task we have for handling None better

languages/python/oso/polar/query.py
samscott89: This is consistent with how other exceptions are done in the Python library. Don't want to solve here.

PR #357, opened 2020-08-18
Publish Node.js package to NPM

gj:


.github/workflows/release.yml
```diff
@@ -497,8 +496,8 @@ jobs:
         uses: actions/download-artifact@v1
         with:
           name: oso_js_node
-      - name: Zip file
-        run: zip -r oso-js-node.zip oso_js_node
+      - name: Create JS release artifact
+        run: npm pack
```
saolsen, line 500: I think this has to be run inside `oso_js_node`

.github/workflows/release.yml
```diff
@@ -568,9 +567,9 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           upload_url: ${{ steps.create_release.outputs.upload_url }}
-          asset_path: ./oso-js-node.zip
-          asset_name: oso-js-node-${{ steps.version.outputs.oso_version }}.zip
-          asset_content_type: application/zip
+          asset_path: ./oso-${{ steps.version.outputs.oso_version }}.tgz
```
saolsen, line 570: And I think this would be `./oso_js_node/oso-0.0.0.tgz`

PR #358, opened 2020-08-18
Accept a custom equality fn to override == equality for the JS lib

gj:


languages/js/src/Oso.test.ts
```diff
@@ -1,8 +1,29 @@
...
   expect(o.isAllowed(1, 2, 3)).toBe(true);
   expect(o.isAllowed(3, 2, 1)).toBe(false);
 });
+
+describe('Equality function used for unification', () => {
+  test('defaults to loose equality (==)', () => {
+    const o = new Oso();
+    o.registerConstant('undefined', undefined);
+    o.registerConstant('null', null);
+    expect(() => o.loadStr('?= undefined = null;')).not.toThrow(
```
plotnick, line 16: Nit: can we please use `query` instead of inline queries? I find it much clearer to separate the two, and would prefer that we only use inline queries in policy files.

languages/js/src/Polar.test.ts
```diff
@@ -70,12 +70,9 @@ describe('#registerClass', () => {
...
     const p = new Polar();
-    p.registerClass(ConstructorNoArgs, 'A');
-    p.registerClass(ConstructorNoArgs, 'B');
-    p.registerClass(ConstructorArgs, 'C');
-    expect(() =>
-      p.loadStr('?= new A() = new B() and not new A() = new C();')
-    ).not.toThrow();
+    p.registerClass(A, 'A');
+    p.registerClass(A, 'B');
+    expect(() => p.loadStr('?= new A().a() = new B().a();')).not.toThrow();
```
plotnick, line 75: Soooooo much nicer.

languages/js/src/types.ts
```diff
@@ -189,3 +189,9 @@ export interface QueryEvent {
 export type QueryResult = Generator<Map<string, any>, void, never>;
 
 export type obj = { [key: string]: any };
+
+export type EqualityFn = (x: any, y: any) => boolean;
+
+export interface OsoOptions {
```
plotnick, line 195: Can we please call this `Options`, or `ConstructorOptions`? Including `oso` in the name seems inconsistent. `PolarOptions` would be more accurate, but I understand wanting to avoid that.

languages/js/src/Oso.test.ts
```diff
@@ -13,17 +13,15 @@ describe('Equality function used for unification', () => {
     const o = new Oso();
     o.registerConstant('undefined', undefined);
     o.registerConstant('null', null);
-    expect(() => o.loadStr('?= undefined = null;')).not.toThrow(
-      InlineQueryFailedError
-    );
+    o.loadStr('f() if undefined = null;');
+    expect(Array.from(o.query('f()'))).toStrictEqual([map()]);
```
plotnick, line 17: OK, but why not just `o.query("undefined = null")`?

languages/js/src/types.ts
```diff
@@ -192,6 +192,6 @@ export type obj = { [key: string]: any };
 
 export type EqualityFn = (x: any, y: any) => boolean;
 
-export interface OsoOptions {
+export interface Options {
```
plotnick, line 195: Thanks.

languages/js/src/Oso.test.ts
gj: just wasn't the first thing I thought of. updated

PR #359, opened 2020-08-19
Fix list bug.

saolsen:
Also fixed one thing by matching Rest variables in isa and deref lists.

Main bug that's not fixed is this.
```
query> delete([1,2,3,2,1],2,x)
{'x': [1, [3, [1, []]]]}
```


saolsen:
I tried some hacky list deref logic to expand the final element if it comes from a rest variable and it did actually fix my example.
```
query> delete([1,2,3,2,1],2,x)
{'x': [1, 3, 1]}
```

It broke another test though so I'm not sure it's really correct.

saolsen:
Fixed

polar-core/src/vm.rs
```diff
@@ -584,6 +584,28 @@ impl PolarVirtualMachine {
...
+                        if let Value::List(terms) = last_term.value() {
+                            derefed.append(&mut terms.clone());
+                        } else {
+                            derefed.push(last_term);
+                        }
+                    }
+                }
+                term.clone_with_value(Value::List(derefed))
+            }
             Value::Variable(symbol) | Value::RestVariable(symbol) => {
```
samscott89, line 609: Does it ever make sense to deref a rest variable outside of the above case with lists? If not, then we could move this arm up into the list case and do something like:

```
if let Some(Value::RestVariable(symbol)) = derefed.last() {
    if let Value::List(terms) = self.value(symbol).value() {
        derefed.pop();
       derefed.append(&mut terms.clone())
   }
```

And you wouldn't need the `rest` logic, nor the case where you pop/push the same term back on.

saolsen: You can definitely deref a rest variable outside derefing a list.
```
query> [1,2,3] = [1,*bs] and x = bs
x => [2, 3]
bs => [2, 3]
```

The push pop logic is needed because the thing the rest variable derefs to could be a list or it could just be the deref variable still (if there's no binding for it yet).
Doing it with a pop/push is simpler than trying to check if it's a list and then doing the pop append because of rusty borrowing reasons.

samscott89: Ah okay. So in my mind that `bs` would be a regular symbol bound to a list? How do we know the `bs` on the RHS is a rest variable?

polar-core/tests/integration_tests.rs
```diff
@@ -1484,3 +1486,20 @@ fn test_external_unify() {
     let (results, _externals) = query_results_with_externals(query);
     assert!(results.is_empty());
 }
+
+#[test]
+fn test_list_results() {
```
samscott89, line 1491: This is an _awesome_ test. But could we have a few simpler examples? This logic seems really hard to get right. How about:

- `[1, 2] = [1, *ys]` outputs `xs= [2]`
- `[1, 2, *xs] = [1, *ys] and [1, 2, 3] = [1, *ys]` outputs `xs = [3], ys = [2, 3]`
- `xs = [2], [1, 2] = [1, *xs]` succeeds
And then failure version of each of those.

samscott89: That second one in particular seems tough

polar-core/src/vm.rs
saolsen: Hmm you might be right, I don't understand your first suggestion though. Once things are derefed it's no longer a list so I don't see how you can avoid doing the check for it being a rest variable before and doing the append after.

saolsen: Oh n/m I get how it works if you don't have the case for RestVariable. I don't know how to test if we need that at the top level thoughl

polar-core/tests/integration_tests.rs
saolsen: Those all pass (after I removed the typos). I don't know what you mean by "the failure version of each of those".

polar-core/src/vm.rs
saolsen: Ok, I tried it like that and it breaks some of these tests, you end up getting non derefed restvariables back as things so I think it has to stay how it is. We could play with simplifying it more later but I'd really like to get this merged since it's broken now.

saolsen: > Ah okay. So in my mind that `bs` would be a regular symbol bound to a list? How do we know the `bs` on the RHS is a rest variable?

I believe that `bs` the normal variable derefs to `*bs` the rest variable which derefs to `[2, 3]` so you can deref a rest variable directly and that's why that version doesn't work.

polar-core/tests/integration_tests.rs
samscott89: So maybe something like `[1, 2] = [2, *ys]` should fail, `[1, 2, *xs] = [1, *ys] and [1, 2, *ys]` should fail.

saolsen: `[1, 2] = [2, *ys]` fails correctly.
`[1, 2, *xs] = [1, *ys] and [1, 2, *ys]` doesn't make sense.

samscott89: My bad, `[1, 2, *xs] = [1, *ys] and [1, 2, 3] = [1, 2, *ys]`

polar-core/src/vm.rs
samscott89: Oh interesting. I find that a little weird. Agree we can revisit later.

PR #362, opened 2020-08-20
Additional FFI tests

samscott89:
Added test cases for:
- unifying (in a head) for a class. This is potentially useful for doing "model-level" authZ. E.g. `allow(user, "read", ModelClass)`.
- That terms are sufficiently derefed before crossing boundary. (`x = 1 and [x].index(1)` is a nice one). This one hit a few issues:
   - Java `getMethod` looks for _exact_ class matches. Used `common-beanutils` to address this.
   - We weren't actually doing a full deref for most calls (only make_external did the correct thing).

PR checklist:

- [ ] Added changelog entry.


languages/java/oso/src/main/java/com/osohq/oso/Query.java
```diff
@@ -208,12 +209,11 @@ public void registerCall(String attrName, Optional<List<Object>> args, long call
...
                     .toArray(new Class[0]);
-                try {
-                    Method method = cls.getMethod(attrName, argTypes);
-                    result = method.invoke(instance, args.get().toArray());
-                } catch (NoSuchMethodException e) {
+                Method method = MethodUtils.getMatchingAccessibleMethod(cls, attrName, argTypes);
+                if (method == null) {
                     throw new Exceptions.InvalidCallError(cls.getName(), attrName, argTypes);
                 }
+                result = method.invoke(instance, args.get().toArray());
```
leina05, line 216: Can't this still throw an exception?

samscott89: Nope, it catches them and returns `null` for some reason 😑

samscott89: Oh wait, not sure about this one

samscott89: Thinking about it some more - it won't throw a `NoSuchMethodException`. That is covered by the null check above. But invoking the method can presumably raise any other kind of error, as it could before. Which we don't handle.

polar-core/src/vm.rs
```diff
@@ -1721,7 +1722,7 @@ impl PolarVirtualMachine {
                     instance_id: right_instance,
                     ..
                 }),
-            ) if left_instance != right_instance => {
+            ) => {
```
gj, line 1725: highlighting this change for visibility

samscott89: Oh thanks, I meant to ping @plotnick on this one. The reason you wouldn't want to outright skip the unify if the instance IDs are equals are in the case where the two classes are not self-equals ,e.g. `nan != nan`.

languages/js/test/parity.ts
```diff
@@ -71,6 +81,8 @@ if (
     oso.queryRule('testMethodCalls', new A('hello'), new B.C('hello')).next()
       .done,
     oso.queryRule('testOr').next().done,
+    oso.queryRule('testUnifyClass', A).next().done,
+    oso.queryRule('testDerefJava').next().done,
```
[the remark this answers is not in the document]
samscott89: That was a happy accident.
How about `testDerefJavas`?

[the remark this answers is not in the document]
samscott89: These are a little unfortunate in that it's supposed to be a parity test, but the languages don't agree on a common method for this.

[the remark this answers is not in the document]
gj: For the other parity tests that required language specific stuff, I just loaded each rule string separately in the language-specific file instead of adding it to the shared polar file

polar-core/src/vm.rs
gj: maybe a counterpoint (and definitely splitting hairs): isn't this more of an "is this the same thing" check instead of an "are these two things equal" check?

samscott89: Hmm, that's true. And would be more consistent with our rationale of letting `==` be overloadable, but not `=`.

polar-core/src/vm.rs
```diff
@@ -1721,11 +1723,15 @@ impl PolarVirtualMachine {
...
                     ..
                 }),
-            ) if left_instance != right_instance => {
-                self.push_goal(Goal::UnifyExternal {
-                    left_instance_id: *left_instance,
-                    right_instance_id: *right_instance,
-                })?;
+            ) => {
+                // If the two have identical IDs, the two _are_ the same
+                // value so unify.
```
gj, line 1728:
```suggestion
                // If IDs match, they're the same _instance_ (not just the same _value_), so unify.
```

polar-core/src/vm.rs
```diff
@@ -1291,8 +1297,7 @@ impl PolarVirtualMachine {
                     matches!(result.value(), Value::Variable(_)),
                     "Must have result variable as second arg."
                 );
-                let mut constructor = args.pop().unwrap();
-                constructor.map_replace(&mut |t| self.deref(t));
```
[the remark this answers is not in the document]
samscott89: Indeed. And for the record, the idea was to colocate all the `deep_deref`s with `QueryEvent`s. So that it (should) be obvious when a type is crossing the FFI boundary and needs to be derefed.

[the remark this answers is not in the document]
samscott89: We could actually encode this pretty cleanly with a newtype.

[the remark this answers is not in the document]
gj: SGTM

PR #363, opened 2020-08-20
Update REPLs

gj:
- Working Node.js REPL
- Update Ruby + Python `Polar#repl` methods to accept a list of files instead of a boolean parameter.
- Hook Ruby + Python REPLs up to readline

### DIFF
--- a/languages/js/Makefile
+++ b/languages/js/Makefile
@@ -34,3 +34,6 @@ parity: wasm
 
 test: wasm
 	yarn test
+
+repl: build
+	./bin/repl.js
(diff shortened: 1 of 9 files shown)


gj:
> This looks great, and all work for me. I'm wondering if the `RESET` code for JS is correct because I get output in yellow, not white like the others.
> Screen Shot 2020-08-24 at 1 30 44 PM
> 
> I am also not a JS person, so can't really review for convention/style.

The RESET code is an ANSI escape sequence so it's standard across all of the languages -- it's for the terminal; not for the language itself.

The JS booleans being printed out in yellow sounds like [this](https://nodejs.org/docs/latest/api/util.html#util_customizing_util_inspect_colors), but since `colors: false` is the default for `util.inspect`, I'm not sure why that would be happening. `console.log` says it shells out to `util.format`, which also apparently defaults to `colors: false`, so... I'm not sure. We can explicitly set `colors: false` if you think it's worth it.

PR #365, opened 2020-08-24
Move file checking into core.

samscott89:
Reduces the duplication needed, and unlocks handling conflicts
in the core (e.g. remove all rules whose source matches the
new filename).

Made the usage of `load` and `load_str` a little more consistent
across core + host languages.

### DIFF
--- a/languages/java/oso/src/main/java/com/osohq/oso/Host.java
+++ b/languages/java/oso/src/main/java/com/osohq/oso/Host.java
@@ -2,7 +2,6 @@ package com.osohq.oso;
 
 import java.lang.reflect.Constructor;
 import java.util.*;
-import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.DoubleStream;
 import java.util.stream.IntStream;
@@ -46,7 +45,8 @@ public class Host implements Cloneable {
      * Store a Java class in the cache by name.
      *
      * @param name The name used to reference the class from within Polar.
-     * @throws Exceptions.DuplicateClassAliasError If the name is already registered.
+     * @throws Exceptions.DuplicateClassAliasError If the name is already
+     *                                             registered.
      */
     public String cacheClass(Class<?> cls, Constructor<?> constructor, String name)
             throws Exceptions.DuplicateClassAliasError {
@@ -92,16 +92,13 @@ public class Host implements Cloneable {
      * Make an instance of a Java class from a {@code Map<String, Object>} of
      * fields.
      */
-    public Object makeInstance(String className, List<Object> initargs, long id)
-            throws Exceptions.OsoException
-    {
+    public Object makeInstance(String className, List<Object> initargs, long id) throws Exceptions.OsoException {
         Constructor<?> constructor = constructors.get(className);
         if (constructor == null) {
             // Try to find a constructor applicable to the supplied arguments.
             Class<?> cls = classes.get(className);
-            Class<?>[] argTypes = initargs.stream().map(arg -> arg.getClass())
-                .collect(Collectors.toUnmodifiableList())
-                .toArray(new Class[0]);
+            Class<?>[] argTypes = initargs.stream().map(arg -> arg.getClass()).collect(Collectors.toUnmodifiableList())
+                    .toArray(new Class[0]);
             search: for (Constructor<?> c : cls.getConstructors()) {
                 Class<?>[] paramTypes = c.getParameterTypes();
                 if (argTypes.length == paramTypes.length) {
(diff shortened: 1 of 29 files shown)


languages/java/oso/src/main/java/com/osohq/oso/Polar.java
```diff
@@ -2,15 +2,11 @@
```
samscott89, line 2: Bunch of autoformat changes here, sorry. Can back out if necessary

languages/js/src/Polar.test.ts
```diff
@@ -358,26 +355,22 @@ describe('#loadFile', () => {
   test('throws if two files with the same contents are loaded', async () => {
     const p = new Polar();
     expect(p.loadFile(await tempFile('', 'a.polar'))).resolves;
-    expect(p.loadFile(await tempFile('', 'b.polar'))).rejects.toThrow(
-      PolarFileDuplicateContentError
-    );
+    expect(p.loadFile(await tempFile('', 'b.polar'))).rejects.toThrow();
```
samscott89, line 358: @gj couldn't work out the right way to check that the exception thrown has kind `RuntimeError::FileLoading` as specifier in the wasm errors.rs (without manually catching the error and checking it).

polar-core/src/polar.rs
```diff
@@ -70,10 +75,58 @@ impl Polar {
         Self {
             kb: Arc::new(RwLock::new(KnowledgeBase::new())),
             messages: MessageQueue::new(),
+            loaded_content: Arc::new(RwLock::new(HashMap::new())),
+            loaded_files: Arc::new(RwLock::new(HashSet::new())),
         }
     }
 
-    pub fn load_file(&self, src: &str, filename: Option<String>) -> PolarResult<()> {
+    fn check_file(&self, src: &str, filename: &str) -> PolarResult<()> {
```
samscott89, line 83: Simply relying on the hashing algorithm used in `HashSet` and `HashMap` to check for collisions. This isn't guaranteed to be stable across different versions of rustc. Do we think this is a problem?

languages/js/src/Polar.test.ts
gj: If you pass a string to `toThrow()` it will do a regex match on the error output

samscott89: Ohh, regex. Nice.

languages/java/oso/src/main/java/com/osohq/oso/Host.java
```diff
@@ -92,16 +92,13 @@ public Long cacheInstance(Object instance, Long id) throws Exceptions.OsoExcepti
      * Make an instance of a Java class from a {@code Map<String, Object>} of
      * fields.
      */
-    public Object makeInstance(String className, List<Object> initargs, long id)
-            throws Exceptions.OsoException
-    {
+    public Object makeInstance(String className, List<Object> initargs, long id) throws Exceptions.OsoException {
```
leina05, line 95: Are these formatting changes due to version differences in rustfmt?

languages/js/src/Polar.test.ts
samscott89: Added a regex test check in the latest.

languages/java/oso/src/main/java/com/osohq/oso/Polar.java
```diff
@@ -100,7 +72,7 @@ public void loadFile(String filename) throws IOException, OsoException {
      * @param filename Name of the source file.
      */
     public void loadStr(String str, String filename) throws Exceptions.OsoException {
```
leina05, line 74: I wonder if this method could be private

polar-core/src/polar.rs
```diff
@@ -70,10 +75,58 @@ impl Polar {
         Self {
             kb: Arc::new(RwLock::new(KnowledgeBase::new())),
             messages: MessageQueue::new(),
+            loaded_content: Arc::new(RwLock::new(HashMap::new())),
+            loaded_files: Arc::new(RwLock::new(HashSet::new())),
```
leina05, line 79:
```suggestion
            loaded_content: Arc::new(RwLock::new(HashMap::new())),    // file content -> file name
            loaded_files: Arc::new(RwLock::new(HashSet::new())),            // set of file names
```

polar-core/src/polar.rs
```diff
@@ -70,10 +75,58 @@ impl Polar {
...
+            .unwrap()
+            .insert(filename.to_string());
+
+        Ok(())
+    }
+
+    pub fn load(&self, src: &str, filename: Option<String>) -> PolarResult<()> {
+        if let Some(ref filename) = filename {
+            self.check_file(src, filename)?;
+        }
```
leina05, line 129: Way cleaner, this is great

polar-core/src/polar.rs
leina05: What's the implication of this? These checks won't always work?

samscott89: I'm pretty sure its a non-issue. The checks wouldn't match if we somehow made it possible to do hot-swapped oso updates. But even then it's pretty nebulous.

languages/java/oso/src/main/java/com/osohq/oso/Host.java
samscott89: We don't have a standardised java formatter, but apparently my IDE does.

polar-core/src/polar.rs
samscott89: :)

languages/java/oso/src/main/java/com/osohq/oso/Host.java
gj: It would be good to pick one and standardize on it (not as part of this PR)

PR #366, opened 2020-08-24
Trace Logging

saolsen:
Trace output during execution of a query that shows what's happening. Should make debugging a bit easier.

Notable features include.
* Showing all the applicable rules when querying a call.
   This shows which rules and the order they will be tried after filtering and sorting has been done.
* Showing Queries, Unifys, Matches, Comparisons and Math operations.
* ShowingFFIi lookups and the results that come back.
* Showing relevant bindings (bindings for variables used in the expressions being traced) alongside so it's easier to see what's happening.

Looks something like this.
```
query> f(1);
[trace]   QUERY: 'f(1)', BINDINGS: {}
[trace]     FILTERING RULES
[trace]       UNIFY: '1' = '_x_4', BINDINGS: {}
[trace]       UNIFY: '1' = '_x_6', BINDINGS: {}
[trace]       UNIFY: '1' = '_x_7', BINDINGS: {}
[trace]     SORTING RULES
[trace]     APPLICABLE_RULES: [
[trace]       f(x) if new (Foo{x: x}, _instance_2) and .(_instance_2, foo(), _value_1) and _value_1 = x;
[trace]       f(x) if x = 1;
[trace]       f(x) if x + 1 and _op_3 == 2;
[trace]     ]
[trace]     RULE:
[trace]     f(x) if
[trace]       new Foo{x: x}.foo() = x
[trace]     UNIFY: '1' = '_x_10', BINDINGS: {}
[trace]       QUERY: 'new (Foo{x: _x_10}, _instance_2_11) and .(_instance_2_11, foo(), _value_1_12) and _value_1_12 = _x_10', BINDINGS: {"_x_10": "1"}
[trace]         QUERY: 'new (Foo{x: _x_10}, _instance_2_11)', BINDINGS: {"_x_10": "1"}
[trace]           UNIFY: '_instance_2_11' = 'Foo{x: 1}', BINDINGS: {}
[trace]         QUERY: '.(_instance_2_11, foo(), _value_1_12) and _value_1_12 = _x_10', BINDINGS: {"_instance_2_11": "Foo{x: 1}", "_x_10": "1"}
[trace]           QUERY: '.(_instance_2_11, foo(), _value_1_12)', BINDINGS: {"_instance_2_11": "Foo{x: 1}"}
[trace]             LOOKUP: Foo{x: 1}.foo()
[trace]             => 2
[trace]           QUERY: '_value_1_12 = _x_10', BINDINGS: {"_x_10": "1", "_value_1_12": "2"}
[trace]             UNIFY: '_value_1_12' = '_x_10', BINDINGS: {"_value_1_12": "2", "_x_10": "1"}
[trace]             UNIFY: '2' = '1', BINDINGS: {}
[trace]             LOOKUP: Foo{x: 1}.foo()
[trace]             => No more results.
[trace]     RULE:
[trace]     f(x) if
[trace]       x = 1
[trace]     UNIFY: '1' = '_x_13', BINDINGS: {}
[trace]       QUERY: '_x_13 = 1', BINDINGS: {"_x_13": "1"}
[trace]         UNIFY: '_x_13' = '1', BINDINGS: {"_x_13": "1"}
[trace]         UNIFY: '1' = '1', BINDINGS: {}
True
[trace]     RULE:
[trace]     f(x) if
[trace]       x + 1 == 2
[trace]     UNIFY: '1' = '_x_14', BINDINGS: {}
[trace]       QUERY: '_x_14 + 1 and _op_3_15 == 2', BINDINGS: {"_x_14": "1"}
[trace]         QUERY: '_x_14 + 1', BINDINGS: {"_x_14": "1"}
[trace]           MATH: '1' + '1' = '_op_3_15', BINDINGS: {}
[trace]           UNIFY: '2' = '_op_3_15', BINDINGS: {}
[trace]         QUERY: '_op_3_15 == 2', BINDINGS: {"_op_3_15": "2"}
[trace]           CMP: '2' == '2', BINDINGS: {}
True
query>
```

### DIFF
--- a/docs/changelogs/vNEXT.rst
+++ b/docs/changelogs/vNEXT.rst
@@ -24,14 +24,11 @@ performs a lookup, and ``x.foo()`` invokes a zero-arity method.
 New features
 ==============
 
-Feature 1
+Debug Mode
 =========
 
-- summary
-- of
-- user facing changes
-
-Link to relevant documentation section
+Oso can be run in debug mode by setting an environment variable `POLAR_DEBUG=1`
+This prints debug output when polar queries are evaluated to show what's happening.
 
 Other bugs & improvements
 =========================
--- a/docs/changelogs/vNEXT.rst
+++ b/docs/changelogs/vNEXT.rst
@@ -27,7 +27,7 @@ New features
 Debug Mode
 =========
 
-Oso can be run in debug mode by setting an environment variable `POLAR_DEBUG=1`
+Oso can be run in debug mode by setting an environment variable ``POLAR_DEBUG=1``
 This prints debug output when polar queries are evaluated to show what's happening.
 
 Other bugs & improvements
--- a/docs/changelogs/vNEXT.rst
+++ b/docs/changelogs/vNEXT.rst
@@ -39,3 +39,4 @@ Other bugs & improvements
 - Improved performance of policies with many rules having ground (constant) parameters.
 - Improved performance of ``in`` operator (list membership) with many ground elements.
 - Stack traces return the original policy source instead of the internal version.
+- New FFI methods for passing print and warning messages from the core runtime to the language libraries.
--- a/docs/more/dev-tools/tracing.rst
+++ b/docs/more/dev-tools/tracing.rst
@@ -1,0 +1,62 @@
+########
+Tracing
+########
+
+Polar tracing shows logs of how a query is evaluated.
+
+********************
+Enabling Tracing
+********************
+
+Debug mode is enabled by setting an environment variable ``POLAR_LOG=1``
+
+It will print out ``[debug]`` messages during query evaluation that shows how the query is executed.
+Notable things include
+
+* Queries and subquery evaluations.
+* Showing values that are bound to variables for each expression.
+* Showing which rules a polar predicate call will evaluate.
+* Showing values returned from application classes.
+
+Example
+-------
+
+.. code-block:: oso
+
+  query> f(12);
+  [debug]   QUERY: f(12), BINDINGS: {}
+  [debug]     APPLICABLE_RULES: [
+  [debug]       f(x) if new (Foo{x: x}, _instance_2) and .(_instance_2, foo(), _value_1) and _value_1 = x;
+  [debug]       f(x) if x = 1;
+  [debug]       f(x) if x + 1 and _op_3 == 2;
+  [debug]     ]
+  [debug]     RULE:
+  [debug]     f(x) if
+  [debug]       new Foo{x: x}.foo() = x
+  [debug]       QUERY: new (Foo{x: _x_10}, _instance_2_11) and .(_instance_2_11, foo(), _value_1_12) and _value_1_12 = _x_10, BINDINGS: {"_x_10": "12"}
+  [debug]         QUERY: new (Foo{x: _x_10}, _instance_2_11), BINDINGS: {"_x_10": "12"}
+  [debug]         QUERY: .(_instance_2_11, foo(), _value_1_12) and _value_1_12 = _x_10, BINDINGS: {"_instance_2_11": "Foo{x: 12}", "_x_10": "12"}
+  [debug]           QUERY: .(_instance_2_11, foo(), _value_1_12), BINDINGS: {"_instance_2_11": "Foo{x: 12}"}
+  [debug]             LOOKUP: Foo{x: 12}.foo()
+  [debug]             => 13
+  [debug]           QUERY: _value_1_12 = _x_10, BINDINGS: {"_x_10": "12", "_value_1_12": "13"}
+  [debug]             BACKTRACK
+  [debug]             LOOKUP: Foo{x: 12}.foo()
+  [debug]             => No more results.
+  [debug]             BACKTRACK
+  [debug]     RULE:
+  [debug]     f(x) if
+  [debug]       x = 1
+  [debug]       QUERY: _x_13 = 1, BINDINGS: {"_x_13": "12"}
+  [debug]         BACKTRACK
+  [debug]     RULE:
+  [debug]     f(x) if
+  [debug]       x + 1 == 2
+  [debug]       QUERY: _x_14 + 1 and _op_3_15 == 2, BINDINGS: {"_x_14": "12"}
+  [debug]         QUERY: _x_14 + 1, BINDINGS: {"_x_14": "12"}
+  [debug]           MATH: 12 + 1 = _op_3_15, BINDINGS: {}
+  [debug]         QUERY: _op_3_15 == 2, BINDINGS: {"_op_3_15": "13"}
+  [debug]           CMP: 13 == 2, BINDINGS: {}
+  [debug]           BACKTRACK
+  [debug]           HALT
+  False
--- a/docs/more/dev-tools/tracing.rst
+++ b/docs/more/dev-tools/tracing.rst
@@ -8,15 +8,15 @@ Polar tracing shows logs of how a query is evaluated.
 Enabling Tracing
 ********************
 
-Debug mode is enabled by setting an environment variable ``POLAR_LOG=1``
+Debug mode is enabled by setting an environment variable: ``POLAR_LOG=1``.
 
-It will print out ``[debug]`` messages during query evaluation that shows how the query is executed.
-Notable things include
+It will print out ``[debug]`` messages during query evaluation that show how the query is executed.
+Notable things include:
 
 * Queries and subquery evaluations.
 * Showing values that are bound to variables for each expression.
-* Showing which rules a polar predicate call will evaluate.
-* Showing values returned from application classes.
+* Showing which rules a Polar predicate call will evaluate.
+* Showing values returned from calls into the application.
 
 Example
 -------
--- a/docs/more/dev-tools/tracing.rst
+++ b/docs/more/dev-tools/tracing.rst
@@ -13,7 +13,7 @@ Debug mode is enabled by setting an environment variable: ``POLAR_LOG=1``.
 It will print out ``[debug]`` messages during query evaluation that show how the query is executed.
 Notable things include:
 
-* Queries and subquery evaluations.
+* Queries and sub-query evaluations.
 * Showing values that are bound to variables for each expression.
 * Showing which rules a Polar predicate call will evaluate.
 * Showing values returned from calls into the application.
--- a/polar-core/src/vm.rs
+++ b/polar-core/src/vm.rs
@@ -529,7 +529,7 @@ impl PolarVirtualMachine {
             if !include_temps && self.is_temporary_var(&var) {
                 continue;
             }
-            bindings.insert(var.clone(), self.deref(value));
+            bindings.insert(var.clone(), self.deep_deref(value));
         }
         bindings
     }
@@ -575,6 +575,10 @@ impl PolarVirtualMachine {
             .rev()
             .find(|binding| binding.0 == *variable)
             .map(|binding| &binding.1)
+    }
+
+    pub fn deep_deref(&self, term: &Term) -> Term {
+        term.cloned_map_replace(&mut |t| self.deref(t))
     }
 
     /// Recursively dereference a variable.
@@ -1065,7 +1069,7 @@ impl PolarVirtualMachine {
         let (field_name, args): (Symbol, Option<Vec<Term>>) = match self.deref(field).value() {
             Value::Call(Predicate { name, args }) => (
                 name.clone(),
-                Some(args.iter().map(|arg| self.deref(arg)).collect()),
+                Some(args.iter().map(|arg| self.deep_deref(arg)).collect()),
             ),
             Value::String(field) => (Symbol(field.clone()), None),
             v => {
@@ -1103,7 +1107,7 @@ impl PolarVirtualMachine {
 
         Ok(QueryEvent::ExternalCall {
             call_id,
-            instance: Some(instance.clone()),
+            instance: Some(self.deep_deref(instance)),
             attribute: field_name,
             args,
         })
@@ -1117,6 +1121,7 @@ impl PolarVirtualMachine {
         let result = self.kb.read().unwrap().gensym("isa");
         let call_id = self.new_call_id(&result);
 
+        self.bind(&result, Term::new_temporary(Value::Boolean(false)));
         self.push_goal(Goal::Unify {
             left: Term::new_temporary(Value::Variable(result)),
             right: Term::new_temporary(Value::Boolean(true)),
@@ -1124,7 +1129,7 @@ impl PolarVirtualMachine {
 
         Ok(QueryEvent::ExternalIsa {
             call_id,
-            instance: instance.clone(),
+            instance: self.deep_deref(instance),
             class_tag: literal.tag.clone(),
         })
     }
@@ -1137,6 +1142,7 @@ impl PolarVirtualMachine {
         let result = self.kb.read().unwrap().gensym("unify");
         let call_id = self.new_call_id(&result);
 
+        self.bind(&result, Term::new_temporary(Value::Boolean(false)));
         self.push_goal(Goal::Unify {
             left: Term::new_temporary(Value::Variable(result)),
             right: Term::new_temporary(Value::Boolean(true)),
@@ -1152,7 +1158,7 @@ impl PolarVirtualMachine {
     pub fn make_external(&self, constructor: &Term, instance_id: u64) -> QueryEvent {
         QueryEvent::MakeExternal {
             instance_id,
-            constructor: constructor.clone(),
+            constructor: self.deep_deref(&constructor),
         }
     }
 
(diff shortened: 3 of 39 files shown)


samscott89:
Could we also do a `DEBUG` version that outputs similar information but not as verbose?

saolsen:
> Could we also do a `DEBUG` version that outputs similar information but not as verbose?

I specifically made this a lot less verbose than the old logging. I'm not sure what subset of this would still be useful.

samscott89:
> I specifically made this a lot less verbose than the old logging. I'm not sure what subset of this would still be useful.

Okay. Will try this locally with a few examples and see how it goes. There are just many unifies and queries that are mostly noise.

polar-core/src/vm.rs
```diff
@@ -825,6 +882,16 @@ impl PolarVirtualMachine {
             "Called isa with bare dictionary!"
         );
 
+        if self.polar_log {
+            let relevant_bindings = self.relevant_bindings(&[left, right]);
+            self.log(&format!(
+                "ISA:'{}' matches '{}', BINDINGS: {:?}",
+                left.to_string(),
+                right.to_string(),
+                relevant_bindings
```
plotnick, line 891: Could we do something more like:
```suggestion
                "ISA: {} matches {}, BINDINGS: {}",
                left.to_polar(),
                right.to_polar(),
                format_bindings(relevant_bindings)
```
I find the single quotes around Polar terms distracting, and the bindings should be non-debug, too, right?

saolsen:
> > I specifically made this a lot less verbose than the old logging. I'm not sure what subset of this would still be useful.
> 
> Okay. Will try this locally with a few examples and see how it goes. There are just many unifies and queries that are mostly noise.

The queries I think are important but maybe the unify's aren't really needed since you have the bindings.

polar-core/src/vm.rs
```diff
@@ -741,6 +794,7 @@ impl PolarVirtualMachine {
         if self.log {
             self.print("⇒ backtrack");
```
plotnick, line 795: Should this be checking `self.polar_log`?

saolsen: This is the old log, which I haven't removed yet. I'm not logging backtracks for this new one because I didn't find them useful, rather I have the log indent based on subqueries which is a lot easier to follow.

polar-core/src/vm.rs
saolsen: :+1:

samscott89:
How about only outputting failed unifies?

polar-core/src/vm.rs
```diff
@@ -825,6 +882,16 @@ impl PolarVirtualMachine {
             "Called isa with bare dictionary!"
         );
 
+        if self.polar_log {
+            let relevant_bindings = self.relevant_bindings(&[left, right]);
+            self.log(&format!(
+                "ISA:{} matches {}, BINDINGS: {:?}",
```
samscott89, line 888: This should probably be `MATCHES` since that's the user-facing term.

saolsen:
> How about only outputting failed unifies?

I like it a lot without any unifys. You still see the queries and the bindings so when it's something like a=b failed it's still obvious.

```
query> f(2);
[trace]   QUERY: f(2), BINDINGS: {}
[trace]     FILTERING RULES
[trace]     SORTING RULES
[trace]     APPLICABLE_RULES: [
[trace]       f(x) if new (Foo{x: x}, _instance_2) and .(_instance_2, foo(), _value_1) and _value_1 = x;
[trace]       f(x) if x = 1;
[trace]       f(x) if x + 1 and _op_3 == 2;
[trace]     ]
[trace]     RULE:
[trace]     f(x) if
[trace]       new Foo{x: x}.foo() = x
[trace]       QUERY: new (Foo{x: _x_10}, _instance_2_11) and .(_instance_2_11, foo(), _value_1_12) and _value_1_12 = _x_10, BINDINGS: {"_x_10": "2"}
[trace]         QUERY: new (Foo{x: _x_10}, _instance_2_11), BINDINGS: {"_x_10": "2"}
[trace]         QUERY: .(_instance_2_11, foo(), _value_1_12) and _value_1_12 = _x_10, BINDINGS: {"_instance_2_11": "Foo{x: 2}", "_x_10": "2"}
[trace]           QUERY: .(_instance_2_11, foo(), _value_1_12), BINDINGS: {"_instance_2_11": "Foo{x: 2}"}
[trace]             LOOKUP: Foo{x: 2}.foo()
[trace]             => 3
[trace]           QUERY: _value_1_12 = _x_10, BINDINGS: {"_x_10": "2", "_value_1_12": "3"}
[trace]             LOOKUP: Foo{x: 2}.foo()
[trace]             => No more results.
[trace]     RULE:
[trace]     f(x) if
[trace]       x = 1
[trace]       QUERY: _x_13 = 1, BINDINGS: {"_x_13": "2"}
[trace]     RULE:
[trace]     f(x) if
[trace]       x + 1 == 2
[trace]       QUERY: _x_14 + 1 and _op_3_15 == 2, BINDINGS: {"_x_14": "2"}
[trace]         QUERY: _x_14 + 1, BINDINGS: {"_x_14": "2"}
[trace]           MATH: 2 + 1 = _op_3_15, BINDINGS: {}
[trace]         QUERY: _op_3_15 == 2, BINDINGS: {"_op_3_15": "3"}
[trace]           CMP: 3 == 2, BINDINGS: {}
[trace]           HALT
False
query>
```

samscott89:
> 
> 
> > How about only outputting failed unifies?
> 
> I like it a lot without any unifys. You still see the queries and the bindings so when it's something like a=b failed it's still obvious.

Nice, yeah that's good. Would it be too noisy to put in something to indicate a backtrack?

saolsen:
> > > How about only outputting failed unifies?
> > 
> > 
> > I like it a lot without any unifys. You still see the queries and the bindings so when it's something like a=b failed it's still obvious.
> 
> Nice, yeah that's good. Would it be too noisy to put in something to indicate a backtrack?

I played around with that a bit. It was a lot easier to follow when I did the subquery indentation than actually logging the backtracks.

saolsen:
Ok, dropped unify, really wanna just keep this to stuff that's useful for people and not noise. Also refactored a bit to contain the logging stuff more.

saolsen:
Added back backtrack, it is useful after all. The only problem is that there are backtracks for stuff that doesn't exactly make sense with the limited other stuff we output. 

```
query> f(3);
[trace]   QUERY: f(3), BINDINGS: {}
[trace]     BACKTRACK
[trace]     BACKTRACK
[trace]     APPLICABLE_RULES: [
[trace]       f(x) if new (Foo{x: x}, _instance_2) and .(_instance_2, foo(), _value_1) and _value_1 = x;
[trace]       f(x) if x = 1;
[trace]       f(x) if x + 1 and _op_3 == 2;
[trace]     ]
[trace]     RULE:
[trace]     f(x) if
[trace]       new Foo{x: x}.foo() = x
[trace]       QUERY: new (Foo{x: _x_10}, _instance_2_11) and .(_instance_2_11, foo(), _value_1_12) and _value_1_12 = _x_10, BINDINGS: {"_x_10": "3"}
[trace]         QUERY: new (Foo{x: _x_10}, _instance_2_11), BINDINGS: {"_x_10": "3"}
[trace]         QUERY: .(_instance_2_11, foo(), _value_1_12) and _value_1_12 = _x_10, BINDINGS: {"_x_10": "3", "_instance_2_11": "Foo{x: 3}"}
[trace]           QUERY: .(_instance_2_11, foo(), _value_1_12), BINDINGS: {"_instance_2_11": "Foo{x: 3}"}
[trace]             LOOKUP: Foo{x: 3}.foo()
[trace]             => 4
[trace]           QUERY: _value_1_12 = _x_10, BINDINGS: {"_x_10": "3", "_value_1_12": "4"}
[trace]             BACKTRACK
[trace]             LOOKUP: Foo{x: 3}.foo()
[trace]             => No more results.
[trace]             BACKTRACK
[trace]     RULE:
[trace]     f(x) if
[trace]       x = 1
[trace]       QUERY: _x_13 = 1, BINDINGS: {"_x_13": "3"}
[trace]         BACKTRACK
[trace]     RULE:
[trace]     f(x) if
[trace]       x + 1 == 2
[trace]       QUERY: _x_14 + 1 and _op_3_15 == 2, BINDINGS: {"_x_14": "3"}
[trace]         QUERY: _x_14 + 1, BINDINGS: {"_x_14": "3"}
[trace]           MATH: 3 + 1 = _op_3_15, BINDINGS: {}
[trace]         QUERY: _op_3_15 == 2, BINDINGS: {"_op_3_15": "4"}
[trace]           CMP: 4 == 2, BINDINGS: {}
[trace]           BACKTRACK
[trace]           HALT
False
query>
```

saolsen:
Ok, this is really nice now. You have backtracks so you can see when a query fails so you really don't need to log unifys anymore. I also muted trace output during sorting and filtering of rules so we don't get a bunch of extra backtrack traces.

```
query> f(3);
[trace]   QUERY: f(3), BINDINGS: {}
[trace]     APPLICABLE_RULES: [
[trace]       f(x) if new (Foo{x: x}, _instance_2) and .(_instance_2, foo(), _value_1) and _value_1 = x;
[trace]       f(x) if x = 1;
[trace]       f(x) if x + 1 and _op_3 == 2;
[trace]     ]
[trace]     RULE:
[trace]     f(x) if
[trace]       new Foo{x: x}.foo() = x
[trace]       QUERY: new (Foo{x: _x_10}, _instance_2_11) and .(_instance_2_11, foo(), _value_1_12) and _value_1_12 = _x_10, BINDINGS: {"_x_10": "3"}
[trace]         QUERY: new (Foo{x: _x_10}, _instance_2_11), BINDINGS: {"_x_10": "3"}
[trace]         QUERY: .(_instance_2_11, foo(), _value_1_12) and _value_1_12 = _x_10, BINDINGS: {"_instance_2_11": "Foo{x: 3}", "_x_10": "3"}
[trace]           QUERY: .(_instance_2_11, foo(), _value_1_12), BINDINGS: {"_instance_2_11": "Foo{x: 3}"}
[trace]             LOOKUP: Foo{x: 3}.foo()
[trace]             => 4
[trace]           QUERY: _value_1_12 = _x_10, BINDINGS: {"_value_1_12": "4", "_x_10": "3"}
[trace]             BACKTRACK
[trace]             LOOKUP: Foo{x: 3}.foo()
[trace]             => No more results.
[trace]             BACKTRACK
[trace]     RULE:
[trace]     f(x) if
[trace]       x = 1
[trace]       QUERY: _x_13 = 1, BINDINGS: {"_x_13": "3"}
[trace]         BACKTRACK
[trace]     RULE:
[trace]     f(x) if
[trace]       x + 1 == 2
[trace]       QUERY: _x_14 + 1 and _op_3_15 == 2, BINDINGS: {"_x_14": "3"}
[trace]         QUERY: _x_14 + 1, BINDINGS: {"_x_14": "3"}
[trace]           MATH: 3 + 1 = _op_3_15, BINDINGS: {}
[trace]         QUERY: _op_3_15 == 2, BINDINGS: {"_op_3_15": "4"}
[trace]           CMP: 4 == 2, BINDINGS: {}
[trace]           BACKTRACK
[trace]           HALT
False
query> 
```

saolsen:
> I think this is definitely an improvement from the unreadable mess of debugging output we had before.
> 
> My main concern with this in its current state is the embedding directly into the VM. I would much rather have a few functions in `formatting` that the VM calls out to. I'm also confused by the relation between (on the VM) `self.log` and `self.polar_log` and the conditions you have to check before calling `self.log`; can we have those calls be unconditional and have it check the necessary fields in just one place?
> 
> I also have a few concerns about the formatting itself. Why do some terms use `to_string()` but others `to_polar()`? I'd also prefer to drop the single quotes, and not use debug formatting anywhere.

Refactored this a bit to keep just about all the logic in `self.log()` I hope that helps a bit. Actually pulling it out of the VM is tricky because we need to pull context stuff and write to the message queue, this is I hope an acceptable comprimise.

samscott89:
> Ok, this is really nice now. You have backtracks so you can see when a query fails so you really don't need to log unifys anymore. I also muted trace output during sorting and filtering of rules so we don't get a bunch of extra backtrack traces.
> 

Super pedantic then: should this therefore be the DEBUG mode? And we could expose _moar_ goals etc. in trace mode that would only really be useful to debug internals as opposed to user policies?

saolsen:
> > Ok, this is really nice now. You have backtracks so you can see when a query fails so you really don't need to log unifys anymore. I also muted trace output during sorting and filtering of rules so we don't get a bunch of extra backtrack traces.
> 
> Super pedantic then: should this therefore be the DEBUG mode? And we could expose _more_ goals etc. in trace mode that would only really be useful to debug internals as opposed to user policies?

Yeah, I would call this debug mode. I can make the prefix of the printing `[debug]` if you want. The env var is already `POLAR_DEBUG`.

samscott89:
> Refactored this a bit to keep just about all the logic in `self.log()` I hope that helps a bit. Actually pulling it out of the VM is tricky because we need to pull context stuff and write to the message queue, this is I hope an acceptable compromise.
> 

It looks like most of the debugging is done on the information contained in the `Goal` anyway? Could we have a debug method that takes in a VM and a Goal and outputs the same kind of information? Even if a little more duplicative it would keep the vm code from getting messy.

> Yeah, I would call this debug mode. I can make the prefix of the printing [debug] if you want. The env var is already POLAR_DEBUG.

I think that would be good.

This is looking super useful btw.

saolsen:
> It looks like most of the debugging is done on the information contained in the `Goal` anyway? Could we have a debug method that takes in a VM and a Goal and outputs the same kind of information? Even if a little more duplicative it would keep the vm code from getting messy.

No it's not really goal based anymore. Stuff like applicable methods, special math and comparison logging, external results all have to be done in different places than just before a goal is run.

saolsen:
Gonna leave it like this for now. When we want the debugger to access the same information it'll probably be clearer what the best way to organize it all is.

dhatch:
Could we add this to the developer tools documentation section or somewhere else so that people can find it? It might also be useful to add a top level page for what to do if a policy doesn't work as you expect -- something like Troubleshooting Policies?

saolsen:
> Could we add this to the developer tools documentation section or somewhere else so that people can find it? It might also be useful to add a top level page for what to do if a policy doesn't work as you expect -- something like Troubleshooting Policies?

Added a docs page about how to enable it and some stuff it shows. I agree with having a `Troubleshooting Policies` guide but that's a bigger task and should probably talk about how to use all the tools together and have some example bugs.

docs/changelogs/vNEXT.rst
```diff
@@ -24,14 +24,11 @@ performs a lookup, and ``x.foo()`` invokes a zero-arity method.
...
-Feature 1
+Debug Mode
 =========
 
-- summary
-- of
-- user facing changes
-
-Link to relevant documentation section
+Oso can be run in debug mode by setting an environment variable `POLAR_DEBUG=1`
```
gj, line 30: Doesn't this have to be double-backticked for rst?
```suggestion
Oso can be run in debug mode by setting an environment variable ``POLAR_DEBUG=1``
```

docs/more/dev-tools/tracing.rst
```diff
@@ -0,0 +1,62 @@
...
+Tracing
+########
+
+Polar tracing shows logs of how a query is evaluated.
+
+********************
+Enabling Tracing
+********************
+
+Debug mode is enabled by setting an environment variable ``POLAR_LOG=1``
```
gj, line 11:
```suggestion
Debug mode is enabled by setting an environment variable: ``POLAR_LOG=1``.
```

docs/more/dev-tools/tracing.rst
```diff
@@ -0,0 +1,62 @@
...
+
+It will print out ``[debug]`` messages during query evaluation that shows how the query is executed.
```
gj, line 13:
```suggestion
It will print out ``[debug]`` messages during query evaluation that show how the query is executed.
```

docs/more/dev-tools/tracing.rst
```diff
@@ -0,0 +1,62 @@
...
+Notable things include
```
gj, line 14: I'm not sure if this is the right way to get rst to show a literal colon
```suggestion
Notable things include:
```

docs/more/dev-tools/tracing.rst
```diff
@@ -0,0 +1,62 @@
...
+
+* Queries and subquery evaluations.
+* Showing values that are bound to variables for each expression.
+* Showing which rules a polar predicate call will evaluate.
```
gj, line 18:
```suggestion
* Showing which rules a Polar predicate call will evaluate.
```

docs/more/dev-tools/tracing.rst
```diff
@@ -0,0 +1,62 @@
...
+* Showing values returned from application classes.
```
gj, line 19:
```suggestion
* Showing values returned from calls into the application.
```

docs/more/dev-tools/tracing.rst
saolsen: looks right rendered :+1:

PR #342, opened 2020-08-12
Spring boot example app + docs

leina05:
PR checklist:

- [ ] Create new repository for Spring app + link to it
- [x] Ideally the HTTP response would be the same from both apps (e.g., 401 vs. 403 for unauthorized)
- [x] Tests don't do anything--should either make them real or remove

PR #367, opened 2020-08-24
Tiny tweak to phrasing in quickstart

callahad:
The quickstart includes Python and Ruby policies which check:

    expense.submitted_by = actor;

However, the example Java policy uses a camelCase attribute name:

    expense.submittedBy = actor;

The docs then say that "oso looks up the `submitted_by` field," but this
is only true for Ruby and Python. In Java, the field is `submittedBy`.
The discrepancy could mislead Java developers into thinking that oso
somehow converts identifier formats when evaluating Polar policies.

Long term, it might be better to completely avoid referencing variable
names that differ between languages. For instance, the examples could be
reworded to use `submitter` instead of `submittedBy` / `submitted_by`.

### DIFF
--- a/docs/getting-started/quickstart.rst
+++ b/docs/getting-started/quickstart.rst
@@ -573,7 +573,7 @@ To accomplish that, we can **replace** our existing rule with:
       :caption: :fa:`oso` expenses.polar
       :class: copybutton
 
-Behind the scenes, oso looks up the ``submitted_by`` field on the provided
+Behind the scenes, oso looks up the associated field on the provided
 ``Expense`` instance and compares that value against the provided **actor**.
 And just like that, an actor can only see an expense if they submitted the expense.
 


callahad:
This is so minor as to almost escape notice, but I was speaking with Graham and... well... :shrug:

samscott89:
> 
> This is so minor as to almost escape notice, but I was speaking with Graham and... well... 🤷

Thank you! I like the suggested change. It's still an open question how we solve this more broadly (see [this comment](https://github.com/osohq/oso/pull/342/files?file-filters%5B%5D=.cmd&file-filters%5B%5D=.jar&file-filters%5B%5D=.java&file-filters%5B%5D=.polar&file-filters%5B%5D=.properties&file-filters%5B%5D=.rst&file-filters%5B%5D=.sql&file-filters%5B%5D=.xml&file-filters%5B%5D=No+extension&w=1#r475748582) for example).

samscott89:
Hey @callahad ! Thank again for the suggestion. We made the change as part of #368 and credited you [here](https://docs.osohq.com/changelogs/0.5.0.html).

PR #368, opened 2020-08-24
Node.js documentation

gj:
PR checklist:

- [x] TypeScript API docs
- [x] Node.js version of Quickstart
- [x] Node.js version of Application Types
- [x] Working with Node.js Types
- [x] Node.js version of RBAC
- [x] Node.js version of ABAC
- [x] Node.js version of Context
- [ ] Node.js version of Multiple Actor Types
- [ ] Node.js version of Resources with Inheritance
- [x] Node.js version of REPL docs
- [x] Grep docs to ensure "Node.js" is added anywhere we currently enumerate supported languages.
- [x] Automate build of TypeScript API docs
- [x] Link built TypeScript API docs on Libraries page
- [x] Added changelog entry.
- [x] Install instructions
- [x] JavaScript Library Page


docs/examples/abac/nodejs/01-simple.js
```diff
@@ -0,0 +1,92 @@
+const { Oso } = require('oso');
+
+const oso = new Oso();
+
+const EXPENSES = [
+  { amount: 500, submitted_by: 'alice', location: 'NYC', project_id: 2 },
```
leina05, line 6: Is it weird to use camel case in JavaScript?

gj: camelCase is idiomatic; snake_case is not

docs/using/libraries/node/index.rst
```diff
@@ -0,0 +1,196 @@
...
+your policy (see :ref:`application-types`).
+
+Class Instances
+^^^^^^^^^^^^^^^
+
+Any ``new``-able JavaScript object (including ES6-style classes) can be
+constructed from inside an oso policy using the :ref:`operator-new` operator if
+the constructor (a ``class`` or ``function`` that responds to `the new operator
+<https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/new>`_)
+has been **registered** using the ``#registerClass`` method. An example of
```
leina05, line 42: Is the `#` a JS convention also, or is this left over from ruby?

gj: it's not a JS convention; we could drop it

docs/examples/abac/nodejs/01-simple.js
leina05: Okay, there are a few other places where snake case is used in the JS examples, but I assume that's just for consistency with the rest of the docs for now. We need to go through all the docs for this issue anyway, can be a follow-up.

docs/using/libraries/node/index.rst
leina05: It's not used in other places to refer to JS methods (that I saw), so maybe best to drop it.

docs/examples/abac/nodejs/01-simple.js
gj: > Okay, there are a few other places where snake case is used in the JS examples, but I assume that's just for consistency with the rest of the docs for now. We need to go through all the docs for this issue anyway, can be a follow-up.

assumption is correct; agree on the follow-up

docs/using/libraries/node/index.rst
gj: agree; follow-up

PR #376, opened 2020-08-25
Specify which Inline query failed

saolsen:
PR checklist:

- [x] Added changelog entry.


languages/README.md
```diff
@@ -83,6 +83,7 @@ Execute a Polar query through the FFI event interface.
 | `results`           |        |         | x            |         |
 | `getCall`           |        |         | x            |         |
 | `run`               | `run`  | `start` | `nextResult` | `start` |
+| `source`            | x      | x       | x            |         |
```
saolsen, line 86:
```suggestion
| `source`            | x      | x       | x            | x       |
```

polar-core/src/polar.rs
```diff
@@ -36,6 +37,10 @@ impl Query {
     pub fn next_message(&self) -> Option<Message> {
         self.vm.messages.next()
     }
+
+    pub fn source(&self) -> String {
+        self.vm.term_source(&self.source)
```
samscott89, line 42: Maybe we should include the context too? From `source.source`? Will have line number for example.

saolsen:
Messages look like this now. (example in python)
`polar.exceptions.PolarRuntimeException: Inline query in file failed. 1 = 2 at line 1, column 3 in file foo.polar`

languages/js/src/Query.ts
```diff
@@ -38,6 +38,10 @@ export class Query {
     this.results = this.start();
   }
 
+  public source(): string {
```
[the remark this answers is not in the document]
gj: Also please add API docs -- see the rest of the methods in this file for examples. This one should probably be marked `@internal`

languages/java/oso/src/main/java/com/osohq/oso/Exceptions.java
```diff
@@ -225,8 +225,8 @@ public InstantiationError(String className, Exception e) {
     }
 
     public static class InlineQueryFailedError extends PolarRuntimeException {
-        public InlineQueryFailedError() {
-            super(null, null);
+        public InlineQueryFailedError(String source) {
+            super("Inline Query Failed " + source);
```
gj, line 229: I think it would be worth standardizing this error messages across the libs. I know we need a general standardization pass that goes beyond this single error type, but no reason not to go for incremental improvement.

polar-core/src/vm.rs
```diff
@@ -694,7 +694,7 @@ impl PolarVirtualMachine {
         }
     }
 
-    fn source(&self, term: &Term) -> Option<Source> {
+    pub fn source(&self, term: &Term) -> Option<Source> {
```
gj, line 697: Why does this have to be `pub`?

languages/js/src/Polar.ts
```diff
@@ -151,10 +151,13 @@ export class Polar {
...
       this.processMessages();
       if (query === undefined) break;
+      const source = query.source();
       const { results } = new Query(query, this.#host);
       const { done } = await results.next();
       results.return();
-      if (done) throw new InlineQueryFailedError(name);
+      if (done) {
+        throw new InlineQueryFailedError(source, name);
+      }
```
gj, line 160: This is the style we use across the rest of the lib
```suggestion
      if (done) throw new InlineQueryFailedError(source, name);
```

languages/js/src/Query.ts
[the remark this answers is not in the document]
gj: Now that I look at it more closely, we aren't actually using this method at all. You're calling `.source()` directly on the FFI query instead of on the TS `Query` instance.

languages/js/src/errors.ts
```diff
@@ -35,11 +35,11 @@ export class DuplicateInstanceRegistrationError extends PolarError {
...
 
 export class InlineQueryFailedError extends PolarError {
-  constructor(file?: string) {
+  constructor(source: string, file?: string) {
     if (file !== undefined) {
-      super(`Inline query failed while loading '${file}'.`);
+      super(`Inline query failed while loading '${file}', ${source}.`);
     } else {
-      super('Inline query failed.');
+      super(`Inline query failed, ${source}.`);
```
gj, line 42: I think these commas would be better as colons, maybe with the source indented on the following line. Something like:
```
Inline query failed while loading file policies/production.polar:
  ?= 1 = 2;
```

languages/ruby/lib/oso/polar/ffi/query.rb
```diff
@@ -74,6 +75,15 @@ def process_messages
             message.process
           end
         end
+
+        # @return [String]
```
gj, line 79: This doesn't look like it returns a string -- did you mean to return `res.to_s`?

languages/ruby/lib/oso/polar/ffi/query.rb
```diff
@@ -74,6 +75,15 @@ def process_messages
...
+        # @raise [FFI::Error] if the FFI call returns an error.
+        def source
+          res = Rust.source(self)
+          raise FFI::Error.get if res.null?
```
gj, line 83: Judging by your CFFI code in polar-c-api/src/lib.rs, it doesn't look like it's possible for this FFI call to return null

polar-core/src/polar.rs
```diff
@@ -36,6 +37,19 @@ impl Query {
     pub fn next_message(&self) -> Option<Message> {
         self.vm.messages.next()
     }
+
+    pub fn source(&self) -> String {
+        let mut source_str = self.vm.term_source(&self.source);
+        if let Some(source) = self.vm.source(&self.source) {
```
gj, line 43: Could you explain what's going on here? It's a bit hard to follow - the `Query#source` method gets the `term_source` for its own `source` field which calls `VM#source` to get the term source info and then we call `VM#source` again (???), passing in the query's `source` field again... and I've already lost the plot of what this is actually supposed to result in 😄

polar-core/src/vm.rs
saolsen: So we can call it from the method on Query.

languages/ruby/lib/oso/polar/ffi/query.rb
saolsen: If there's a panic it would still return null.

polar-core/src/polar.rs
saolsen: Renamed a bunch of stuff.

languages/README.md
```diff
@@ -83,6 +83,7 @@ Execute a Polar query through the FFI event interface.
 | `results`           |        |         | x            |         |
 | `getCall`           |        |         | x            |         |
 | `run`               | `run`  | `start` | `nextResult` | `start` |
+| `source`            | x      | x       | x            | x       |
```
gj, line 86: This isn't a method on the Query object -- we're making a direct FFI call.

languages/js/src/errors.ts
```diff
@@ -35,12 +35,8 @@ export class DuplicateInstanceRegistrationError extends PolarError {
...
 
 export class InlineQueryFailedError extends PolarError {
-  constructor(file?: string) {
-    if (file !== undefined) {
-      super(`Inline query failed while loading '${file}'.`);
-    } else {
-      super('Inline query failed.');
-    }
+  constructor(source: string) {
+    super(`Inline query failed: ${source}`);
```
gj, line 39: Did we remove the optional filename handling because `source` will contain the same info?

languages/python/oso/polar/polar.py
```diff
@@ -140,7 +140,8 @@ def load_str(self, string, filename=None):
                 try:
                     next(Query(query, host=self.host.copy()).run())
                 except StopIteration:
-                    raise PolarRuntimeException("Inline query in file failed.")
+                    source = query.source()
+                    raise PolarRuntimeException(f"Inline query failed: {source.get()}")
```
gj, line 144: Nit: it would be good to create an error type for this so that people can rescue it specifically without matching on the message, but that can be part of a (necessary) broader Python error handling cleanup.

languages/ruby/lib/oso/polar/polar.rb
```diff
@@ -105,7 +105,7 @@ def load_str(str, filename: nil) # rubocop:disable Metrics/MethodLength
           begin
             Query.new(next_query, host: host).results.next
           rescue StopIteration
-            raise InlineQueryFailedError
+            raise InlineQueryFailedError, format('Inline query failed: %<source>s', source: next_query.source)
```
gj, line 108: This formatting should be handled in the `InlineQueryFailedError#initialize` method:
```rb
    class InlineQueryFailedError < PolarRuntimeError # rubocop:disable Style/Documentation
      # @param file [String]
      def initialize(source)
        super(format('Inline query failed: %<source>s', source: source))
      end
    end
```

and then this becomes:
```suggestion
            raise InlineQueryFailedError, next_query.source
```

polar-core/src/polar.rs
```diff
@@ -36,6 +37,19 @@ impl Query {
     pub fn next_message(&self) -> Option<Message> {
         self.vm.messages.next()
     }
+
+    pub fn source_info(&self) -> String {
+        let mut source_info = self.vm.term_source_string(&self.term);
+        if let Some(source) = self.vm.source(&self.term) {
```
gj, line 43: It still seems suboptimal that we're calling `self.vm.term_source_string(&self.term)`, which calls `self.vm.source(&self.term)`, and then we're calling `self.vm.source(&self.term)` _again_

languages/js/src/errors.ts
saolsen: Yeah. If it's from a file the location and filename are part of the message.

languages/python/oso/polar/exceptions.py
```diff
@@ -45,6 +45,10 @@ class PolarFileNameChangedError(PolarRuntimeException):
     pass
 
 
+class InlineQueryFailedError(PolarRuntimeException):
```
gj, line 48: 😍

polar-core/src/vm.rs
```diff
@@ -2258,36 +2258,58 @@ impl PolarVirtualMachine {
...
+        source_string
+    }
+
+    pub fn term_source_info(&self, term: &Term) -> String {
+        self.term_source(term, true)
+    }
+
+    pub fn term_source_string(&self, term: &Term) -> String {
+        self.term_source(term, false)
     }
```
gj, line 2292: nit: I would personally do away with these two helpers and just call `vm.term_source(term, [true|false])` at the callsites b/c I don't love the naming of `term_source_info` and `term_source_string` but can't come up with anything better 😄

PR #377, opened 2020-08-25
Update docs for release.

samscott89:


docs/_static/css/custom.css
```diff
@@ -64,8 +64,22 @@
 }
 
 
+@media only screen and (max-width:44.9375em) {
```
gj, line 67: 44.9375em is a pretty specific size

PR #379, opened 2020-08-26
Fix parser bug on lists in patterns.

samscott89:
PR checklist:

- [x] Added changelog entry.

### DIFF
--- a/polar-core/src/polar.lalrpop
+++ b/polar-core/src/polar.lalrpop
@@ -446,10 +446,13 @@ pub Pattern: Value = {
     <Variable>,
     <DictionaryPattern>,
     <InstanceLiteralPattern>,
-    "[" <terms:(<Spanned<Pattern>> ",")*> <term:Spanned<RestVar<"Pattern">>?> "]" => {
-        match term {
-            Some(term) => Value::List(vec![term]),
-            None => Value::List(terms)
+    "[" <mut head:(<Spanned<Pattern>> ",")*> <last:Spanned<RestVar<"Pattern">>?> "]" => {
+        match last {
+            Some(last) =>{ 
+                head.push(last);
+                Value::List(head)
+            },
+            None => Value::List(head)
         }
     },
 };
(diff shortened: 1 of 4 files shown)


polar-core/tests/integration_tests.rs
```diff
@@ -1137,6 +1137,10 @@ fn test_matches() {
     qeval(&mut polar, "x = {foo: 1} and x matches {foo: 1}");
     qnull(&mut polar, "x = {foo: 1} and x matches {foo: 1, bar: 2}");
     qnull(&mut polar, "x = {foo: 1} and x matches {foo: 2}");
+
+    qeval(&mut polar, "[1, 2, 3] matches [1, 2, 3]");
+    qeval(&mut polar, "[1, 2, 3] matches [1, 2, *rest]");
+    qnull(&mut polar, "[1, 2, 3] matches [1, 2]");
```
gj, line 1143: What's the thinking behind not wanting this to match? I'm not saying it should; just curious why you think it shouldn't.

Would also be nice to add a few more test cases:
- RHS is a superset of LHS
- both lists have the same elements but ordered differently
- both lists are empty
- RHS is the empty list and LHS is non-empty
- RHS is non-empty and LHS is the empty list
- both lists have the same number of elements but no overlapping elements
- `*rest` on the LHS (is this legal?)
- `*rest` on both sides (is this legal?)

samscott89: I agree with you, but really didn't want to spend too much time on this before getting this fix in. This was fixing a small parser bug not about the logic. Will add a follow up task for revisiting matches on lists.

I personally don't think lists should have subset matching logic. My main argument is that ordering is important for lists. `[1, 2]` shouldn't match `[2, 1]`.

We should perhaps add sets to address this. (That would be nice actually, `list = [3, 1, 2] and set(list1) matches { 1, 2 }`

gj: > I agree with you, but really didn't want to spend too much time on this before getting this fix in. This was fixing a small parser bug not about the logic. Will add a follow up task for revisiting matches on lists.

works for me

> I personally don't think lists should have subset matching logic. My main argument is that ordering is important for lists. `[1, 2]` shouldn't match `[2, 1]`.

I agree with that

polar-core/src/polar.lalrpop
```diff
@@ -463,13 +466,13 @@ pub Value: Value = {
     <BuiltinOperation>,
     <RewrittenOperation>,
     <Predicate>,
-    "[" <mut terms:(<Term> ",")*> <term:Spanned<RestVar<"Term">>?> "]" => {
-        match term {
-            Some(term) => {
-                terms.push(term);
-                Value::List(terms)
+    "[" <mut head:(<Term> ",")*> <last:Spanned<RestVar<"Term">>?> "]" => {
```
gj, line 469: nit: I'd prefer this as `head` + `tail`

samscott89: Yes, but punt and it gets changed in #390 anyway

PR #381, opened 2020-08-26
Node.js docs follow-ups and general docs cleanup

gj:
- [x] Update Node.js REPL docs


docs/using/examples/context.rst
```diff
@@ -54,9 +54,13 @@ variables.
            :caption: :fab:`java` Env.java
```
gj, line 54: as a meta comment, I don't think `Context` is a very descriptive name for this docs page

docs/more/dev-tools/repl.rst
```diff
@@ -186,15 +219,15 @@ plus ``oso``, and then use the ``Oso.repl()`` API method to start the REPL:
...
-            const { Expense, User } = require("./models");
-            const { Oso } = require("oso");
+            const { Expense, User } = require('./models');
+            const { Oso } = require('oso');
 
             const oso = new Oso();
             oso.registerClass(Expense);
             oso.registerClass(User);
-            await oso.repl();
+            oso.repl();
```
dhatch, line 233: Why did you remove `await` here?  This is still an async method.

gj: Because if someone copies this code into a script and runs `node path/to/that/script.js` it'll work either way, but if someone opens up the Node REPL and copies and pastes these lines in it'll blow up if they try to use a top-level `await` (unless they started Node with the `--experimental-repl-await` feature) but it'll work fine if they skip the `await`.

PR #384, opened 2020-08-27
Missing semicolon in node example.

dhatch:


gj:
This may turn out to be the most important PR we ever ship

PR #387, opened 2020-08-28
Fix Node.js REPL

gj:
PR checklist:

- [x] Added changelog entry.

### DIFF
--- a/languages/js/src/Polar.ts
+++ b/languages/js/src/Polar.ts
@@ -23,12 +23,7 @@ import { readFile, repr } from './helpers';
 let RESET = '';
 let FG_BLUE = '';
 let FG_RED = '';
-if (
-  typeof stdout.hasColors === 'function' &&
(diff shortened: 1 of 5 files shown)


languages/js/src/Polar.ts
```diff
@@ -183,77 +167,89 @@ export class Polar {
...
     let loadError;
     try {
       if (files?.length) await Promise.all(files.map(f => this.loadFile(f)));
     } catch (e) {
       loadError = e;
     }
-    if (loadError !== undefined) {
-      console.error(FG_RED + 'One or more files failed to load.' + RESET);
-      console.error(loadError.message);
+    if (loadError !== undefined) printError(loadError);
```
dhatch, line 193: Why not just print these error at 191?

gj: I think this is a holdover from... something. But I can't remember why I did it this way. Removed in 6207cb6

PR #388, opened 2020-08-29
Refactor types

plotnick:
The `types` module is gone. The new modules that replace it are:
- `events`
- `kb`
- `messages`
- `rules`
- `sources`
- `terms`
- `traces`

Only one substantive code change was made: we avoided a cycle between `terms` and `sources` by splitting `Sources.get_source(&self, term: &Term)` into `Sources.get_source(&self, src_id: u64)` and `Term.get_source_id(&self)`. Usage is `term.get_source_id().and_then(|id| source.get_source(id))...`

[New dependency graph for `polar-core`:](https://github.com/osohq/oso/files/5146019/dep.pdf)
dep

I think the only remaining cycle is between `debugger` and `vm`, but that was there before and apparently not bothering anything.


polar-core/src/events.rs
```diff
@@ -0,0 +1,66 @@
...
+use super::terms::*;
+use super::traces::*;
+use serde::{Deserialize, Serialize};
+
+#[allow(clippy::large_enum_variant)]
+#[must_use]
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub enum QueryEvent {
+    None,
+    Debug {
```
gj, line 11: the newline between every variant except the first two made my eye twitch

polar-core/src/messages.rs
```diff
@@ -0,0 +1,51 @@
+use serde::{Deserialize, Serialize};
+use std::collections::VecDeque;
+use std::sync::{Arc, Mutex};
+#[derive(Debug, Clone, Serialize, Deserialize)]
```
gj, line 4: nit
```suggestion
use std::sync::{Arc, Mutex};

#[derive(Debug, Clone, Serialize, Deserialize)]
```

polar-core/src/numerics.rs
```diff
@@ -1,10 +1,16 @@
+use serde::{Deserialize, Serialize};
 use std::cmp::Ordering;
 use std::hash::{Hash, Hasher};
 use std::mem::discriminant;
 use std::num::FpCategory;
 use std::ops::{Add, Div, Mul, Sub};
 
-use super::types::*;
+/// A number. See the [`numerics`] module for implementations.
```
gj, line 8: the call is coming from inside the house

polar-core/src/rewrites.rs
```diff
@@ -10,6 +12,17 @@ fn and_wrap(a: &mut Term, b: Term) {
     a.replace_value(new_value);
 }
 
+#[cfg(test)]
```
gj, line 15: looks like this was added -- nice catch

polar-core/src/sources.rs
```diff
@@ -0,0 +1,66 @@
...
+    },
+
+    /// Created as a temporary variable
+    TemporaryVariable,
+
+    /// From an FFI call
+    Ffi,
+
+    /// Created for a test
+    Test,
```
gj, line 23: should we `#[cfg(test)]` this variant?

polar-core/src/sources.rs
```diff
@@ -0,0 +1,66 @@
...
+}
+
+#[derive(Clone, Debug, Default, Serialize, Deserialize)]
+pub struct Source {
+    pub filename: Option<String>,
+    pub src: String,
+}
+
+pub struct Sources {
+    // Pair of maps to go from Term ID -> Source ID -> Source.
```
gj, line 39: comment needs to be updated, and is this struct still necessary?

polar-core/src/terms.rs
```diff
@@ -0,0 +1,476 @@
...
+#[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq, Hash)]
+pub struct ExternalInstance {
+    pub instance_id: u64,
+    pub constructor: Option<Term>,
+    pub repr: Option<String>,
+}
+
+// Context stored somewhere by id.
+
+// parser outputs
```
gj, line 81: are these comments still up-to-date/needed?

polar-core/src/terms.rs
```diff
@@ -0,0 +1,476 @@
...
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Context {
+    pub file: String,
+    // TODO: more things
+
+    // maybe for ffi, you say the method on what python class you called or whatever.
```
gj, line 88: Can we update these comments?

polar-core/src/terms.rs
```diff
@@ -0,0 +1,476 @@
...
+}
+
+impl Value {
+    pub fn symbol(self) -> Result<Symbol, error::RuntimeError> {
+        match self {
+            Value::Variable(name) => Ok(name),
+            Value::RestVariable(name) => Ok(name),
+            _ => Err(error::RuntimeError::TypeError {
+                msg: format!("Expected symbol, got: {}", self.to_polar()),
+                stack_trace: None, // @TODO
```
gj, line 223: What is this TODO for? (and all of the identical ones below this in the `impl Value` block)

polar-core/src/terms.rs
```diff
@@ -0,0 +1,476 @@
...
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash)]
+pub enum Value {
+    Number(Numeric),
+    String(String),
+    Boolean(bool),
+    ExternalInstance(ExternalInstance),
+    // TODO (dhatch) Remove this type so that it is no longer possible to even make an
+    // instance literal value!
```
gj, line 205: @dhatch highlighting this TODO in case we can update it.

polar-core/src/terms.rs
```diff
@@ -0,0 +1,476 @@
...
+    InstanceLiteral(InstanceLiteral),
+    Dictionary(Dictionary),
+    Pattern(Pattern),
+    Call(Predicate), // @TODO: Do we just want a type for this instead?
```
gj, line 209: @saolsen this appears to be your TODO from a while ago -- can we update it?

polar-core/src/vm.rs
```diff
@@ -753,7 +761,11 @@ impl PolarVirtualMachine {
                         continue;
                     }
                     let _ = write!(st, "\n  ");
-                    let source = { self.kb.read().unwrap().sources.get_source(&t) };
+
+                    let source = {
+                        t.get_source_id()
+                            .and_then(|src_id| self.kb.read().unwrap().sources.get_source(src_id))
+                    };
                     if let Some(source) = source {
```
gj, line 769:
```suggestion
                    if let Some(source) = self.source(t) {
```

polar-core/src/sources.rs
plotnick: That's an interesting question. The layout of the `enum` would then be different in test & prod, which I suppose shouldn't worry me but does. Does anyone else have opinions?

polar-core/src/sources.rs
plotnick: Comment updated. The struct itself is admittedly simple, but I think offers a worthwhile abstraction.

polar-core/src/terms.rs
plotnick: I think it's the "Ensure all errors contain stack traces" task, right?

polar-core/src/terms.rs
saolsen, line 88:
```suggestion
```

polar-core/src/terms.rs
saolsen, line 205:
```suggestion
```

polar-core/src/terms.rs
saolsen, line 209:
```suggestion
    Call(Predicate),
```

polar-core/src/sources.rs
gj: #TrustInRustC

More seriously, I think Rust intentionally makes no guarantees about [layout for enums using the default representation](https://doc.rust-lang.org/reference/type-layout.html#representations), so I'd be shocked if different layouts led to different runtime characteristics.

polar-core/src/sources.rs
gj: just wary of abstraction leading to unnecessary layers of duplication, like the KB having a `sources` property that is a `Sources` that has a `sources` property 😄

polar-core/src/terms.rs
gj: If you think it would be worth the reminder, could you update the comments to something like
```suggestion
                stack_trace: None, // @TODO: Ensure all errors contain stack traces.
```

polar-core/src/terms.rs
gj: @saolsen couple more potentially obsolete comments to look at

gj (approved):
Few outstanding comments, but nothing major. LGTM

Issue #389, opened 2020-08-29
Parsing error when trying to assign list variable


tappleby:
When trying to assign a variable in a rule where one list item references an object property, I am running into a syntax parsing error "did not expect to find the token ']' at line X"

```polar
resource_scope(actor: Person, "read", "Person", filters) if
    filters = ["id", "=", actor.id];
```

The same rule works fine if assigning a dictionary instead of list:

```polar
resource_scope(actor: Person, "read", "Person", filters) if
    filters = { id: actor.id };
```

As a workaround, the following will parse correctly:

```polar
resource_scope(actor: Person, "read", "Person", filters) if
    field = actor.id and
    filters = ["id", "=", field];
```

samscott89:
Thank you for finding and opening an issue!

The issue looks like a parsing error: currently expressions aren't valid inside lists. So similarly `x = [1+1]` also fails. We will have a fix up soon!

samscott89:
Thanks again for opening this issue. The fix was including in v0.5.1

PR #390, opened 2020-08-31
Allow parsing expressions in lists

samscott89:
Addresses #389.

PR checklist:

- [x] Added changelog entry.

### DIFF
--- a/polar-core/tests/integration_tests.rs
+++ b/polar-core/tests/integration_tests.rs
@@ -1178,6 +1178,10 @@ fn test_matches() {
     // This doesn't fail because `y` is parsed as an unknown specializer
     // assert!(qnull(&mut polar, "x = 1 and y = 2 and x matches y"));
     assert!(qeval(&mut polar, "x = {foo: 1} and x matches {foo: 1}"));
+    assert!(qeval(
+        &mut polar,
+        "x = {foo: 1, bar: 2} and x matches {foo: 1}"
+    ));
     assert!(qnull(
         &mut polar,
         "x = {foo: 1} and x matches {foo: 1, bar: 2}"
@@ -1638,11 +1642,23 @@ fn test_list_matches() {
     );
     assert!(qnull(&mut polar, "[1, 2, 3] matches [1, x]"));
 
-    assert!(qeval(&mut polar, "[] matches [*ys]"));
-    assert!(qeval(&mut polar, "[*xs] matches []"));
-    assert!(qeval(&mut polar, "[*xs] matches [1]"));
-    assert!(qeval(&mut polar, "[1] matches [*ys]"));
+    assert_eq!(qvar(&mut polar, "[] matches [*ys]", "ys"), vec![value!([])]);
+    assert_eq!(qvar(&mut polar, "[*xs] matches []", "xs"), vec![value!([])]);
+    assert_eq!(
+        qvar(&mut polar, "[*xs] matches [1]", "xs"),
+        vec![value!([1])]
+    );
+    assert_eq!(
+        qvar(&mut polar, "[1] matches [*ys]", "ys"),
+        vec![value!([1])]
+    );
     assert!(qeval(&mut polar, "[*xs] matches [*ys]"));
-    assert!(qeval(&mut polar, "[1, 2, 3] matches [1, 2, *rest]"));
-    assert!(qeval(&mut polar, "[1, 2, *xs] matches [1, 2, 3, *ys]"));
+    assert_eq!(
+        qvar(&mut polar, "[1, 2, 3] matches [1, 2, *rest]", "rest"),
+        vec![value!([3])]
+    );
+    assert_eq!(
+        qvar(&mut polar, "[1, 2, *xs] matches [1, 2, 3, *ys]", "xs"),
+        vec![value!([3, Value::RestVariable(Symbol::new("ys"))])]
+    );
 }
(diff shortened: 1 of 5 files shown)


polar-core/tests/integration_tests.rs
```diff
@@ -1535,3 +1575,65 @@ fn test_list_results() {
...
+    let mut polar = Polar::new();
+
+    assert!(qeval(&mut polar, "[] matches []"));
+    assert!(qnull(&mut polar, "[1] matches []"));
+    assert!(qnull(&mut polar, "[] matches [1]"));
+    assert!(qnull(&mut polar, "[1, 2, 3] matches [1, 2]"));
+    assert!(qnull(&mut polar, "[1, 2] matches [1, 2, 3]"));
+    assert!(qnull(&mut polar, "[2, 1] matches [1, 2]"));
+    assert!(qeval(&mut polar, "[1, 2, 3] matches [1, 2, 3]"));
+    assert!(qnull(&mut polar, "[1, 2, 3] matches [1, 2]"));
```
gj, line 1621: You have the same case four lines above
```suggestion
```

polar-core/tests/integration_tests.rs
```diff
@@ -1535,3 +1575,65 @@ fn test_list_results() {
...
+
+    assert!(qeval(&mut polar, "[] matches []"));
```
gj, line 1623: duplicate case
```suggestion
```

polar-core/tests/integration_tests.rs
```diff
@@ -1535,3 +1575,65 @@ fn test_list_results() {
...
+    assert!(qnull(&mut polar, "[x] matches []"));
+    assert!(qnull(&mut polar, "[] matches [x]"));
+    assert!(qnull(&mut polar, "[1, 2, x] matches [1, 2]"));
+    assert!(qnull(&mut polar, "[1, x] matches [1, 2, 3]"));
+    assert!(qnull(&mut polar, "[2, x] matches [1, 2]"));
+    assert!(qeval(&mut polar, "[1, 2, x] matches [1, 2, 3]"));
```
gj, line 1629: Should we use `qvar` here to test that `x` gets bound to 3?

polar-core/tests/integration_tests.rs
```diff
@@ -1535,3 +1575,65 @@ fn test_list_results() {
...
+    assert!(qnull(&mut polar, "[1, 2, 3] matches [1, x]"));
+
+    assert!(qeval(&mut polar, "[] matches [*ys]"));
+    assert!(qeval(&mut polar, "[*xs] matches []"));
+    assert!(qeval(&mut polar, "[*xs] matches [1]"));
+    assert!(qeval(&mut polar, "[1] matches [*ys]"));
+    assert!(qeval(&mut polar, "[*xs] matches [*ys]"));
+    assert!(qeval(&mut polar, "[1, 2, 3] matches [1, 2, *rest]"));
+    assert!(qeval(&mut polar, "[1, 2, *xs] matches [1, 2, 3, *ys]"));
```
gj, line 1638: Should we use `qvar` here to test the bindings?

Also what happens if you have a *rest variable that's not in the terminal position of a list? Parse error?

polar-core/tests/integration_tests.rs
```diff
@@ -1133,12 +1169,16 @@ fn test_in_op() {
...
-    qnull(&mut polar, "x = 1 and y = 2 and x matches y");
-    qeval(&mut polar, "x = 1 and y = 1 and x matches y");
-
-    qeval(&mut polar, "x = {foo: 1} and x matches {foo: 1}");
-    qnull(&mut polar, "x = {foo: 1} and x matches {foo: 1, bar: 2}");
-    qnull(&mut polar, "x = {foo: 1} and x matches {foo: 2}");
+    assert!(qnull(&mut polar, "1 matches 2"));
+    assert!(qeval(&mut polar, "1 matches 1"));
+    // This doesn't fail because `y` is parsed as an unknown specializer
+    // assert!(qnull(&mut polar, "x = 1 and y = 2 and x matches y"));
```
gj, line 1175: What's the goal with this comment? That the behavior will be fixed to allow this in the future?

polar-core/tests/integration_tests.rs
```diff
@@ -1133,12 +1169,16 @@ fn test_in_op() {
...
+    assert!(qeval(&mut polar, "x = {foo: 1} and x matches {foo: 1}"));
+    assert!(qnull(
+        &mut polar,
+        "x = {foo: 1} and x matches {foo: 1, bar: 2}"
+    ));
```
gj, line 1180: can we add the inverse of this, a superset test?

gj:
We should add a changelog entry as well

polar-core/tests/integration_tests.rs
samscott89: Yes and yes. Added a test for latter.

polar-core/tests/integration_tests.rs
samscott89: This was never functioning, but there was a non-checked test for it. I didn't want to outright remove it, because it does seem like an incorrect case.

polar-core/tests/integration_tests.rs
samscott89: Yup, good catch

polar-core/tests/integration_tests.rs
samscott89: Yes

Issue #291, opened 2020-07-29
Support for Rust


lfdominguez:
I known that is written in Rust... but the support for Rust is missing? Reading the documentation I can see any Rust type.

gj:
Rust support is on our roadmap :slightly_smiling_face:. As you noted, the oso core is written in Rust, and we're huge fans of the language. We're definitely excited to work on a Rust version of the library!

I can't give an exact estimate at this point, but the next few languages we have mapped out are JavaScript/TypeScript (via WebAssembly), Golang, and Rust. Community interest is a large factor in determining the selection and ordering of new language libraries, so thank you very much for opening the issue!

lfdominguez:
thanks for response and waiting for this amazing tool

samscott89:
> 
> 
> I was surprised that, being written in rust, doesn't support rust. :(

We agree! It's coming soon!

richerarc:
@samscott89 Looking forward to it!

saolsen:
Some initial work has been merged in #391  Doing cleanup and docs now and will hopefully have something people can play with in a week or two.

gneray:
We shipped [initial support for Rust](https://docs.osohq.com/using/libraries/rust/index.html) today. We will continue to add features and improvements in the coming weeks and months, but please feel free to open another issue or [join our Slack](https://join-slack.osohq.com/) if you have feedback!

lfdominguez:
Great work... and thanks

PR #391, opened 2020-08-31
Initial Rust library

saolsen:
# Rust library!

This is a big PR and there are many different things that need to be reviewed before we merge it. I'm going to break down some categories here and will ask people to review the PR for specific things. (Obviously you can comment on anything but hopefully having a theme in your head helps you filter out some of the noise.

## Reviews
Some categories of review (please suggest more) and some thoughts on things to think about while reviewing.

### Public Facing API
Do the publicly exposed functions match those of the other language integrations. Can you do all the same things you can do in other languages. Do names of functions / error cases / error messages all match up?

### Ops
Can we package up the rust library and run the tests on each platform.
Does the release job include these tests.
Can that package be published to crates.io.

### Library Testing
Are the tests run?
Are all the polar tests we have in other integrations tested in the rust tests?
Are the parity tests equivalent to 
Do we have enough tests of the user class system.

### Tooling
Do the repl / debugger / printing / warnings / debug_mode, etc all work like in the other languages?

### Correctness
Are all operations that happen in the library correct?
* Unify
* Isa
* Lookups
Do they handle errors in the same way that the other language libraries do?

### Classes
The rust class system is complicated so having some reviewers focusing specifically on that will be very helpful.
Does it work the way you would expect it to (though verbosely)
Procedural macros to automate registration are out of scope for this pr and will be added in as a followup.

## Outstanding Work
Here's outstanding work either already known or revealed in review. Things above the line need to be accomplished before merging this PR and things below will be followups.

### TODO as followups

- [ ] Package crate for crates.io
        This currently doesn't work because the crate depends on rust-core which is an unpublished crate. We either need a way to embed it or rust-core itself has to be a published crate (which also causes some problems for packaging and testing as you can't package a crate that depends on a non published crate).
- [ ] Procedural macros to automate registering classes and methods via attributes.
- [ ] Docs / Quickstart and examples.
- [ ] Error handling.
- [ ] Debugger.
- [ ] Handle Messages.
  - [ ] Make sure prints always print and debug logs are turned on when debug mode is enabled.
- [ ] Memory model (Arc/mutexes/etc)
- [ ] Eq check

## Related issues

- #291


plotnick (commented):
I pushed a commit containing my REPL suggestions, and PR #400 is a minor follow-up for the other languages.

There is no debugger to review.

Errors from queries are reported correctly, but `POLAR_LOG` doesn't work at all.

saolsen:
> I pushed a commit containing my REPL suggestions, and PR #400 is a minor follow-up for the other languages.

Thanks!

> There is no debugger to review.

Added to todo list.

> Errors from queries are reported correctly, but `POLAR_LOG` doesn't work at all.

Added "Handle Messages" to the todo list.

languages/rust/src/polar.rs
```diff
@@ -0,0 +1,127 @@
...
+
+    pub fn register_constant<V: crate::host::ToPolar>(
+        &mut self,
+        name: &str,
+        value: &V,
+    ) -> anyhow::Result<()> {
+        let mut host = self.host.lock().unwrap();
+        self.inner.register_constant(
+            polar_core::types::Symbol(name.to_string()),
+            host.value_to_polar(value),
```
dhatch, line 123: Why do we use this here versus `value.to_polar(host)`?

languages/rust/src/query.rs
```diff
@@ -0,0 +1,282 @@
...
+                ),
+                QueryEvent::Debug { message } => self.handle_debug(message),
+            };
+            if let Err(e) = result {
+                self.application_error(e);
+            }
+        }
+    }
+
+    fn check_messages(&mut self) {
```
dhatch, line 92: This duplicates the implementation on `Polar`.  Can we just have one?

languages/rust/src/query.rs
```diff
@@ -0,0 +1,282 @@
...
+impl Query {
+    pub fn new(inner: polar_core::polar::Query, host: Arc<Mutex<crate::host::Host>>) -> Self {
+        Self {
+            calls: HashMap::new(),
+            inner,
+            host,
+        }
+    }
+
+    pub fn next_result(&mut self) -> Option<anyhow::Result<ResultSet>> {
```
dhatch, line 30: What are we getting with this `anyhow` crate? We don't use it for error handling in the core. (This can be tabled since I know error handling is a big open item).

languages/rust/src/query.rs
```diff
@@ -0,0 +1,282 @@
...
+                } => self.handle_external_is_subspecializer(
+                    call_id,
+                    instance_id,
+                    left_class_tag,
+                    right_class_tag,
+                ),
+                QueryEvent::Debug { message } => self.handle_debug(message),
+            };
+            if let Err(e) = result {
+                self.application_error(e);
```
dhatch, line 87: Should we return on an error?

languages/rust/src/query.rs
```diff
@@ -0,0 +1,282 @@
...
+        eprintln!("TODO: {}", message);
+        self.check_messages();
+        Ok(())
+    }
+}
+
+#[derive(Clone)]
+pub struct ResultSet {
+    pub bindings: polar_core::types::Bindings,
+    pub host: Arc<Mutex<crate::host::Host>>,
```
dhatch, line 255: TODO why is host needed here?

languages/rust/src/query.rs
```diff
@@ -0,0 +1,282 @@
...
+                return Some(Err(e.into()));
+            }
+            let event = event.unwrap();
+            tracing::debug!(event=?event);
+            let result = match event {
+                QueryEvent::None => Ok(()),
+                QueryEvent::Done => return None,
+                QueryEvent::Result { bindings, .. } => {
+                    return Some(Ok(ResultSet {
+                        bindings,
```
dhatch, line 44: In other APIs, type conversion happens immediately here.

languages/rust/src/polar.rs
```diff
@@ -0,0 +1,127 @@
...
+use std::io::Read;
+use std::rc::Rc;
+use std::sync::{Arc, Mutex};
+
+use crate::host::Host;
+
+#[derive(Clone)]
+pub struct Polar {
+    inner: Rc<crate::PolarCore>,
+    host: Arc<Mutex<Host>>,
```
dhatch, line 12: I'm concerned all the locking on host will mean queries across threads (since they share a host) will have significant contention.  Can we either move locking of data structures internal to Host so that callers do not need to manage calling `lock()` and we can have more granular locking only on portions of operations that need it?

Alternatively, should we consider having a single host class per query?

Is this thread-safety issue something we may have missed when implementing this in other languages?

languages/rust/src/host/mod.rs
```diff
@@ -0,0 +1,476 @@
...
+        debug_assert!(self.instances.get(&id).is_none());
+        let fields = fields; // TODO: use
+        let instance = class.init(fields, self);
+        self.cache_instance(instance, Some(id));
+    }
+
+    pub fn unify(&self, left: u64, right: u64) -> bool {
+        let _left = self.get_instance(left).unwrap();
+        let _right = self.get_instance(right).unwrap();
+        todo!("left == right")
```
dhatch, line 110: Note: not implemented

languages/rust/src/query.rs
```diff
@@ -0,0 +1,282 @@
...
+pub struct ResultSet {
+    pub bindings: polar_core::types::Bindings,
+    pub host: Arc<Mutex<crate::host::Host>>,
+}
+
+impl ResultSet {
+    pub fn get<T: crate::host::FromPolar>(&self, name: &str) -> Option<T> {
+        self.bindings
+            .get(&Name(name.to_string()))
+            .and_then(|term| T::from_polar(term, &mut self.host.lock().unwrap()))
```
dhatch, line 262: What happens in this case:

```
# policy.polar

allow(1);
allow("1");
allow(1.1);
allow([1, 2]);
```

```
// test.rs

fn test_policy() {
    let oso = Oso::new();
    // I don't think we have a Variable type right now, anyway...
    let query = oso.query_rule("allow", Variable("x"));

    for result in query {
        let x = result.get("x");
        println("{}", x);
    }
}
```

The above will fail to compile because rust won't know the type of `T`.

```
// test_typed.rs

fn test_policy() {
    let oso = Oso::new();
    // I don't think we have a Variable type right now, anyway...
    let query = oso.query_rule("allow", Variable("x"));

    for result in query {
        let x: u64 = result.get("x");
        println("{}", x);
    }
}
```

This will succeed for the first rule, but return `None` for the rest of them.

Since we are going from dynamic to static types here, I think we need a way to have `result.get()` return an enum so that the user can have rules that potentially return multiple types, then match and extract the value as they want.

I also think we need to return an Error type from `get()` so that we can distinguish between the absence of a value (i.e no binding named "x") and an incorrect type (there is a binding named "x", but it cannot be converted to the result type `T`).

languages/rust/src/host/mod.rs
dhatch: Do we have a plan for how to implement this? It doesn't seem like we actually are able to downcast into the real underlying type in order to call the `Eq` trait.

languages/rust/src/host/mod.rs
```diff
@@ -0,0 +1,476 @@
...
+            Value::ExternalInstance(ExternalInstance { instance_id, .. }) => {
+                let instance = host.get_instance(*instance_id)?;
+                instance.instance.downcast_ref::<C>().cloned()
+            }
+            _ => None,
+        }
+    }
+}
+
+impl<C: 'static + Clone + HostClass> ToPolar for C {
```
dhatch, line 463: Are there any common use cases that would be eliminated by requiring this to be `'static`?

languages/rust/src/query.rs
```diff
@@ -0,0 +1,282 @@
...
+            Value::InstanceLiteral(InstanceLiteral { .. }) => todo!("instantiate from literal"),
+            Value::Call(Predicate { name, args }) => {
+                let _instance = host.make_instance(name, args.clone(), instance_id);
+            }
+            _ => panic!("not valid"),
+        }
+        Ok(())
+    }
+
+    fn register_call(
```
leina05, line 131: Nit: argument order is different from other libs.

languages/rust/src/query.rs
```diff
@@ -0,0 +1,282 @@
+use std::collections::HashMap;
```
leina05, line 1: Obviously some still unimplemented, but the organization looks great.

languages/rust/tests/parity.rs
```diff
@@ -0,0 +1,139 @@
...
+fn test() {
+    let mut oso = Oso::new();
+    tracing_subscriber::fmt::init();
+
+    Class::with_constructor(A::new)
+        .name("A")
+        .add_attribute_getter("x", |a_self: &A| a_self.x.clone())
+        .add_method("foo", A::foo)
+        .register(&mut oso)
+        .unwrap();
```
leina05, line 75: I really don't understand what's going on so this might be way off base, but why can't a lot of this configuration be done through the `HostClass` trait? 

Also, are we anticipating that classes would actually be registered this way? I find it more confusing to call `register()` on the class with `oso` as the arg, than the other way around.

leina05 (commented):
Left some comments, but started to get kind of lost in everything so I might revisit again later but at least wanted to get this review in. 

This is very impressive, but also quite different from the other libraries so it's pretty hard to compare in terms of the API. I think that it's roughly consistent, but obviously there's still some implementation missing. I'm really interested in talking about the classes/types stuff in person, because it's really hard to wrap my brain around it from reading the code. 

Another general comment would be that more inline documentation/comments would definitely help with readability, especially at the tops of files.

samscott89:
> > Errors from queries are reported correctly, but `POLAR_LOG` doesn't work at all.
> 
> Added "Handle Messages" to the todo list.

Ah, yeah I used tracing to handle all messages. So you can get them through `RUST_LOG`. Should configure tracing to use `POLAR_LOG` for that instead though.

languages/rust/src/lib.rs
```diff
@@ -0,0 +1,46 @@
...
+pub mod query;
+
+pub use polar_core::polar::Polar as PolarCore;
+
+use host::ToPolar;
+
+#[derive(Clone, Default)]
+pub struct Oso(polar::Polar);
+
+impl std::ops::Deref for Oso {
```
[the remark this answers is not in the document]
samscott89: Agreed. Did this to be consistent with the other languages, but we should expose the methods manually.

languages/rust/src/polar.rs
```diff
@@ -0,0 +1,127 @@
...
+        self.inner.load(&policy, Some(file.to_string()))?;
+        self.check_inline_queries()
+    }
+
+    pub fn load_str(&mut self, s: &str) -> anyhow::Result<()> {
+        self.inner.load(s, None)?;
+        self.check_inline_queries()
+    }
+
+    pub fn query(&mut self, s: &str) -> anyhow::Result<crate::query::Query> {
```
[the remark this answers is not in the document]
samscott89: I'm pretty sure this `&mut` is not needed. This is already carrying around an `Rc` (which probably should be an `Arc` and a `Arc<Mutex>`).

Let's add a follow-up though to decide what the appropriate locking/etc. mechanisms are and where they should be.

languages/rust/src/lib.rs
```diff
@@ -0,0 +1,46 @@
+pub mod host;
+pub mod polar;
+pub mod query;
+
+pub use polar_core::polar::Polar as PolarCore;
```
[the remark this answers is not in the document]
samscott89: Probably not

languages/rust/src/polar.rs
samscott89: Same comment here: https://github.com/osohq/oso/pull/391/files/33fe4fb3bedac9a9ff6b302ddba03a23e94e2cd9#r482588557

Let's revisit. I'm sure there are plenty of improvements on this, but this was the "good enough" version without worrying about borrows + references.

languages/rust/src/polar.rs
```diff
@@ -0,0 +1,127 @@
...
+}
+
+impl Polar {
+    pub fn new() -> Self {
+        let inner = Rc::new(crate::PolarCore::new());
+        let mut host = Host::new(Rc::downgrade(&inner));
+
+        // register all builtin constants
+        for (name, class) in crate::host::builtins::constants() {
+            let value = host.cache_class(class, name.clone());
```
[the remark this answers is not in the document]
samscott89: I think it probably can now. I think this is vestigial from the first way I implemented the builtin logic.

languages/rust/src/polar.rs
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    pub fn clear(&mut self) {
+        *self = Self::new();
+    }
+
+    fn check_messages(&mut self) {
+        while let Some(msg) = self.inner.next_message() {
+            match msg.kind {
+                polar_core::types::MessageKind::Print => tracing::trace!("{}", msg.msg),
```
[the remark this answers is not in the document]
samscott89: >  Isn't there some general logging module for libraries that abstracts the API of whatever the consumer uses and emits messages to that directly?

This is what both `tracing` and `log` do, we can enable the `log` feature in tracing and support both: https://docs.rs/tracing/0.1.19/tracing/#emitting-log-records

The printing itself for this module is enabled in tests + the repl and uses a simple env logger approach: https://docs.rs/tracing-subscriber/0.2.11/tracing_subscriber/fmt/index.html

languages/rust/src/polar.rs
```diff
@@ -0,0 +1,127 @@
...
+                polar_core::types::MessageKind::Warning => tracing::warn!("{}", msg.msg),
+            }
+        }
+    }
+
+    fn check_inline_queries(&mut self) -> anyhow::Result<()> {
```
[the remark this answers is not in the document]
samscott89: We should be using https://crates.io/crates/thiserror for library errors, and anyhow for the application (tests, repl + debugger).

We're using it here because it's super convenient and easy to get something initially working.

languages/rust/src/polar.rs
```diff
@@ -0,0 +1,127 @@
...
+        self.check_messages();
+        Ok(())
+    }
+
+    pub fn load_file(&mut self, file: &str) -> anyhow::Result<()> {
+        let mut f = File::open(&file)?;
+        let mut policy = String::new();
+        f.read_to_string(&mut policy)?;
+        self.inner.load(&policy, Some(file.to_string()))?;
+        self.check_inline_queries()
```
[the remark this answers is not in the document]
samscott89: But `load_str` doesn't take a filename?

languages/rust/src/polar.rs
```diff
@@ -0,0 +1,127 @@
...
+                    anyhow::bail!("error in inline query: {}", e);
+                }
+            }
+        }
+        self.check_messages();
+        Ok(())
+    }
+
+    pub fn load_file(&mut self, file: &str) -> anyhow::Result<()> {
+        let mut f = File::open(&file)?;
```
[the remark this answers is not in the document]
samscott89: Yes, good point

languages/rust/src/host/class/builtins.rs
```diff
@@ -0,0 +1,43 @@
...
+    Class::<i64>::with_default().name("Integer")
+}
+
+fn float() -> Class<f64> {
+    Class::<f64>::with_default().name("Float")
+}
+
+fn number() -> Class<Numeric> {
+    Class::<Numeric>::with_constructor(|| Numeric::Integer(0)).name("Number")
+}
```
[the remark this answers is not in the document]
samscott89: So that you can specialize on anything numeric? Do we not do this in the other languages?

languages/rust/src/host/class/builtins.rs
```diff
@@ -0,0 +1,43 @@
...
+    Class::<HashMap<Name, Term>>::with_default().name("Dictionary")
+}
+
+fn string() -> Class<String> {
+    Class::<String>::with_default()
+        .add_method("ends_with", |s: &String, pat: String| s.ends_with(&pat))
+}
+
+/// Returns the builtin types, the name, class, and instance
+pub fn constants() -> Vec<(Name, Class)> {
```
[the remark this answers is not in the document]
samscott89: No, classes is a better name

languages/rust/src/host/mod.rs
samscott89: We can follow the same approach as some of the similar methods, have an `eq_check: Arc<Fn(...)>` stored on the class which downcasts the args and checks equality. 
If we wanted to do this in a more generic way, we'll need to look into downcasting to trait objects using something like this: https://crates.io/crates/query_interface

languages/rust/src/host/mod.rs
```diff
@@ -0,0 +1,476 @@
...
+        let _right = self.get_class(right_tag).unwrap();
+
+        todo!("????")
+    }
+
+    pub fn operator(&self, _op: Operator, _args: [Instance; 2]) -> bool {
+        todo!()
+    }
+
+    pub fn value_to_polar(&mut self, value: &dyn ToPolar) -> Term {
```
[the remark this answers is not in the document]
samscott89: Yeah, I was trying to stay consistent with the other languages which have a `to_polar` and `to_host` method. But didn't really work out here.

languages/rust/src/host/mod.rs
```diff
@@ -0,0 +1,476 @@
...
+        value.to_polar(self)
+    }
+}
+
+pub trait ToPolar {
+    fn to_polar_value(&self, host: &mut Host) -> Value;
+
+    fn to_polar(&self, host: &mut Host) -> Term {
+        Term::new_from_ffi(self.to_polar_value(host))
```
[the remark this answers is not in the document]
samscott89: By us or by anyone? Not really, it was just easier if implementations only need to return a value and have a default to_polar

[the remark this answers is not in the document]
samscott89: Incidentally, I'm not sure _any_ of the APIs should be sending terms across the FFI boundary. They should all just send Values and then Polar can construct the Term.

languages/rust/src/host/mod.rs
samscott89: Types with references?

We can try and relax this and add lifetimes to `Polar` instances so that we know that any cached value doesn't exceed the lifetime of the polar instance.

languages/rust/src/polar.rs
```diff
@@ -0,0 +1,127 @@
...
+            args,
+        });
+        let query_term = polar_core::types::Term::new_from_ffi(query_value);
+        let query = self.inner.new_query_from_term(query_term, false);
+        self.check_messages();
+        let query = crate::query::Query::new(query, self.host.clone());
+        Ok(query)
+    }
+
+    pub fn register_class(&mut self, class: crate::host::Class) -> anyhow::Result<()> {
```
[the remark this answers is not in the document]
samscott89: I originally had this with an optional parameter that gets threaded though. However, given that the `Class` currently needs to be constructed through the long builder-style method, it seemed really unnatural to then specify an optional parameter in the register method, as opposed to `.name("Foo")`.

It looks like there are other questions about why the class registration is written in the builder pattern, so maybe we can come back to it after discussing that part.

languages/rust/src/polar.rs
```diff
@@ -0,0 +1,127 @@
...
+        let mut host = self.host.lock().unwrap();
+        let name = class.name.clone();
+        let name = polar_core::types::Symbol(name);
+        let class_term = host.cache_class(class, name.clone());
+        self.inner.register_constant(name, class_term);
```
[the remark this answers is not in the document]
samscott89: Sounds worth a try. I think this works if we do your other suggestion of implementing `ToPolar` for `Class`. The difference with other languages though is that we cache a `Class` but register an `Instance` of a `Class<Type>`. I think it will work as you describe though

languages/rust/src/polar.rs
samscott89: Trying to be consistent with the other languages in the use of `host.to_polar`. But it's a bit unfortunate because of the trait clash. Probably not worth having both.

languages/rust/src/query.rs
samscott89: Temporary convenience. 

https://github.com/osohq/oso/pull/391/files/33fe4fb3bedac9a9ff6b302ddba03a23e94e2cd9#r482598147

languages/rust/src/query.rs
samscott89: Isn't the idea we return errors back to polar-core and let it decide what to do?

languages/rust/src/query.rs
samscott89: Probably, yes.

languages/rust/src/query.rs
```diff
@@ -0,0 +1,282 @@
...
+
+    fn call_result_none(&mut self, call_id: u64) -> anyhow::Result<()> {
+        Ok(self.inner.call_result(call_id, None)?)
+    }
+
+    fn application_error(&mut self, error: anyhow::Error) {
+        self.inner.application_error(error.to_string())
+    }
+
+    fn handle_make_external(&mut self, instance_id: u64, constructor: Term) -> anyhow::Result<()> {
```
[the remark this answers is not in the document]
samscott89: Isn't this handled here: https://github.com/osohq/oso/pull/391/files/33fe4fb3bedac9a9ff6b302ddba03a23e94e2cd9#diff-90409a94612032cbd35e08d4256dfd20R101

Not sure whether debug_assert is correct for this though

languages/rust/src/query.rs
samscott89: It's a good nit. My logic was:

`call_id: instance.name(args)` is roughly how you would read this

languages/rust/src/query.rs
samscott89: Will respond to this one: https://github.com/osohq/oso/pull/391/files/33fe4fb3bedac9a9ff6b302ddba03a23e94e2cd9#r482499569

languages/rust/src/query.rs
samscott89: Easy one first:

> I also think we need to return an Error type from get() so that we can distinguish between the absence of a value

💯 

> Since we are going from dynamic to static types here, I think we need a way to have result.get() return an enum so that the user can have rules that potentially return multiple types, then match and extract the value as they want.

Yeah @saolsen raised the same issue. Our plan was to have an option (maybe even the default?) be to just return the `Value` enum. You can do what you want with it. But also provide a helper method, like what we have here, that also does the conversion for you. What do you think?

languages/rust/tests/parity.rs
samscott89: We can cover the implementation specifics after going through it. A couple of points on the API though:

- This is the long-form registration. Most would be handled by a proc macro along the lines of:
```
#[register_class(name="Foo")]
struct Foo { ... }
```
 - Maybe adding register as a method on `Class` was overkill. There are a couple of subtle things it's doing that makes it useful though

languages/rust/src/host/mod.rs
gj: w/r/t @samscott89's comment, it would also be cool to reduce the amount of cloning we're currently doing in this implementation.

languages/rust/src/host/class/builtins.rs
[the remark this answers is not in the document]
gj: No, in the other languages you can specialize on Float or Integer.

languages/rust/src/polar.rs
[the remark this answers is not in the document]
samscott89: This worked nicely I think

samscott89 (approved):

PR #395, opened 2020-09-01
Django integration

dhatch:
PR checklist:

- [ ] Added changelog entry.


languages/python/django-oso/django_oso/tests.py
```diff
@@ -0,0 +1,3 @@
+from django.test import TestCase
```
leina05, line 1: Is this file still necessary?

leina05 (commented):
This looks very cool! I don't have experience with django but it seems like this would be quite easy to use. Happy to review again once more is in. I think having tests for how the model registration works as you outlined will be good.

languages/python/django-oso/django_oso/oso.py
```diff
@@ -0,0 +1,62 @@
...
+
+def init_oso():
+    # Register all models.
+    for app in apps.get_app_configs():
+        for model in app.get_models():
+            Oso.register_class(model)
+
+    # Custom registration for auth (AnonymousUser)
+    if apps.is_installed("django.contrib.auth"):
+        from django.contrib.auth.models import AnonymousUser
```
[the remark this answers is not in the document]
dhatch: It's not actually a model, so the bit above doesn't get it.  Registering it allows it to be used as a specializer.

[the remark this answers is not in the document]
leina05: ah ok

docs/using/frameworks/django.rst
```diff
@@ -0,0 +1,166 @@
...
+:py:data:`django_oso.oso.Oso`.
+
+Registering classes & models
+----------------------------
+
+Often, authorization rules will be expressed over django models.  Therefore,
+``django_oso`` will register every model for each installed app upon startup as
+a class with oso. The :py:class:`django.http.HttpRequest` is also registered
+under ``HttpRequest``.  Django models are referenced in a Polar file using the
+syntax ``app_name::ModelName``. If an app name contains `.`, for example
```
leina05, line 50:
```suggestion
syntax ``app_name::ModelName``. If an app name contains ``.``, for example
```

docs/using/frameworks/django.rst
```diff
@@ -0,0 +1,166 @@
...
+
+Additional classes can be registered as needed using
+:py:meth:`oso.Oso.register_class` on :py:data:`django_oso.oso.Oso`.
+
+Performing authorization
+------------------------
+
+To authorize a request, use the :py:func:`django_oso.auth.authorize` function.
+It accepts the same arguments as
+:py:meth:`~oso.Oso.is_allowed`, but provides sensible defaults for working with
```
leina05, line 62: It doesn't accept the same arguments (`request`)

docs/using/frameworks/django.rst
```diff
@@ -0,0 +1,166 @@
...
+:py:func:`~django_oso.decorators.authorize_request` decorator does this::
+
+    from django_oso.decorators import authorize_request
+
+    @authorize_request
+    def auth_route(request):
+        pass
+
+A policy can then be written controlling authorization based on request
+attributes, like the path:
```
leina05, line 114:
```suggestion
Rules can then be written using request
attributes, like the path:
```

docs/using/frameworks/django.rst
```diff
@@ -0,0 +1,166 @@
...
+        try:
+            expense = Expense.objects.get(pk=id)
+        except Expense.DoesNotExist:
+            return HttpResponseNotFound()
+
+        authorize(request, expense, action="read")
+        return HttpResponse(expense.json())
+
+Requiring authorization on every request
+----------------------------------------
```
leina05, line 92: Something that we shouldn't address on this PR, but it might be worth talking about how using this middleware (and the one in the Flask integration) will only effectively apply authorization if there is a corresponding rule in your policy for the actor, action, and resource.

leina05 (approved):
LGTM! Nice job with the namespacing.

PR #397, opened 2020-09-02
Fix Python readline prompt on Linux

samscott89:
Previously you would get this happening:

```
query> long_variable_name = 1
long_variable_name => 1
#  here I go to the beginning of the line and type x = 2, but readline isn't sure where that is
query> long_varix = 2 and able_name = 1
x => 2
long_variable_name => 1
# I pressed up and enter to the command from history, it actually entered correctly
query> x = 2 and long_variable_name = 1
x => 2
long_variable_name => 1
query>
```

### DIFF
--- a/docs/changelogs/0.5.2.rst
+++ b/docs/changelogs/0.5.2.rst
@@ -12,22 +12,33 @@ None
 New features
 ============
 
-Support for constructors with both positional and keyword arguments
--------------------------------------------------------------------
+Support for constructors with both positional and keyword arguments (`#405 <https://github.com/osohq/oso/pull/405>`_)
+---------------------------------------------------------------------------------------------------------------------
 
-- In languages that support both positional and keyword arguments (currently Python and Ruby), oso now supports registering classes with constructors that use both.
-- The syntax for calling a constructor with mixed arguments from an oso policy is::
+- In languages that support both positional and keyword arguments (currently
+  Python and Ruby), oso now supports registering classes with constructors that
+  use both.
+- The syntax for calling a constructor with mixed arguments from an oso policy is:
 
-    instance = new MyClass(1, 2, foo: 3)
+  .. code-block:: polar
 
-- As is the convention in Python and Ruby, positional arguments must come before keyword arguments.
-- See :ref:`operator-new`
+    new MyClass(1, 2, foo: 3)
+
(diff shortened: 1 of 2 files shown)


gj (commented):
This seems fine, but I'm now worried about the other prompts. Will check now.

gj:
> This seems fine, but I'm now worried about the other prompts. Will check now.

I can't reproduce this behavior with any of the language REPLs that use readline.

samscott89:
> 
> 
> > This seems fine, but I'm now worried about the other prompts. Will check now.
> 
> I can't reproduce this behavior with any of the language REPLs that use readline.

So ship it?

gj:
> So ship it?

😅 I'd rather figure out exactly what causes it. Do you see it on Linux and Windows?

samscott89:
> 😅 I'd rather figure out exactly what causes it. Do you see it on Linux and Windows?

Did you read the linked SO?

https://stackoverflow.com/a/9468954/390293

Not sure why it would be uniquely a Python issue if they're all just using the C readline library under the covers. Probably just a small difference in how they call readline under the covers.

gj:
> Did you read the linked SO?
> 
> [stackoverflow.com/a/9468954/390293](https://stackoverflow.com/a/9468954/390293)
> 
> Not sure why it would be uniquely a Python issue if they're all just using the C readline library under the covers. Probably just a small difference in how they call readline under the covers.

I did. My concern isn't that it's uniquely a Python issue. I couldn't reproduce it in _any_ of the language REPLs that use readline, including Python. That's why I think it might be a platform issue.

Are you able to reproduce this issue in any non-Python REPLs?

samscott89:
> I did. My concern isn't that it's uniquely a Python issue. I couldn't reproduce it in _any_ of the language REPLs that use readline, including Python. That's why I think it might be a platform issue.
> 
> Are you able to reproduce this issue in any non-Python REPLs?

Oh okay, I understand now.

The only other languages using readline are ruby and Node (right?) and those both seem to work fine. On windows it doesn't use readline.

samscott89:
Tried on:

- Python (linux/Windows)
- Ruby (linux)
- Node (Linux)

Problem only found before on Linux and confirmed fixed on Python/Linux after.

gj (approved):

PR #398, opened 2020-09-02
use caching for bench and test

saolsen:


samscott89:
Hmm, is this going to do that thing where clippy doesn't recheck because it's using a cached build?

saolsen:
I'll push a clippy break and see if it catches it.

saolsen:
What workflow is the clippy job on?

saolsen:
Clippy failed when I pushed an error so I think we're good.

PR #400, opened 2020-09-02
More consistent printing in REPL

plotnick:
The goal is to print syntactically valid Polar in the REPL, and that bindings should be indicated with `=` (unify).

### DIFF
--- a/languages/java/oso/src/main/java/com/osohq/oso/Polar.java
+++ b/languages/java/oso/src/main/java/com/osohq/oso/Polar.java
@@ -162,7 +162,11 @@ public class Polar {
             } else {
                 do {
                     HashMap<String, Object> result = query.nextElement();
-                    System.out.println(result.size() > 0 ? result.toString() : "true");
+                    if (result.size() == 0) {
+                        System.out.println("true");
+                    } else {
+                        result.forEach((variable, value) -> System.out.println(variable + " = " + value.toString()));
+                    }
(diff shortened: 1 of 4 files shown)


gj (commented):
I'm fine with the change, with a couple caveats:

- I don't think the REPL output will be "syntactically valid Polar", even after this change. The bound values of variables will be represented in the host language's syntax, which might not be valid Polar. E.g., `True`/`False` in Python, `Instance { ... }` repr of a class instance in JS, etc.
- I'm fine keeping stam/rustoso-1 as the base branch for this PR until that merges, but I don't want to merge this PR into that one. I would rather wait until that one ships and then merge this into main b/c I don't want the change to get rolled up into that giant, unrelated PR.

plotnick:
>   The bound values of variables will be represented in the host language's syntax...

Righty-o. Closer to the goal, anyway.

>  I'm fine keeping stam/rustoso-1 as the base branch for this PR until that merges, but I don't want to merge this PR into that one.

OK. How about we push this without the Rust change to `main`, and push the Rust change to the `stam/rustoso-1` branch?

gj:
> OK. How about we push this without the Rust change to `main`, and push the Rust change to the `stam/rustoso-1` branch?
> 
> Rebased locally, can force push if that's ok.

WFM

plotnick:
> WFM

Cool, apparently need another official :+1:

gj (approved):

PR #401, opened 2020-09-03
Increase test coverage of docs examples

saolsen:
Tests for a few more examples. Still doesn't cover all of them.
There's quickstart tests, but only for python but that at least makes sure the quickstart polar code is valid.
Probably want someone to spend more time doing these but wanna at least get these in.


docs/examples/quickstart/python/server.py
```diff
@@ -4,7 +4,10 @@
 from expense import Expense, EXPENSES
 
 oso = Oso()
-oso.load_file("expenses.polar")
+
+
+def setup_oso():
```
samscott89, line 9: Why do we need this change?

saolsen: I did this so I can load the file in the test (to make sure it's valid python) without re-initing oso.

gj: I'm a -1 on this change since users can download this file via a link on the quickstart page. I think I would rather test the quickstart examples by actually running the server and programmatically making requests against it.

docs/examples/user_types/nodejs/02-userClasses.js
```diff
@@ -29,23 +29,25 @@ oso.registerClass(InternalUser);
...
+  if (userType === "internal") {
     const actor = new InternalUser(id);
-    if (actor.role() === 'account_manager') {
+    if (actor.role() === "account_manager") {
       return new AccountManager(id);
     } else {
       return actor;
     }
-  } else if (userType === 'customer') {
+  } else if (userType === "customer") {
```
gj, line 48: Can you make it so your autoformatter doesn't change all of these quotes to double quotes? It would make it easier to see what the actual changes are in the JS files that were touched.

docs/examples/user_types/python/test_example.py
```diff
@@ -0,0 +1,42 @@
+from pathlib import Path
+import pytest
+
+from oso import Oso
+
+# This example is sudocode mostly, queries a non existant database.
```
gj, line 6:
```suggestion
# This example is pseudo-code mostly, queries a nonexistent database.
```

docs/using/examples/user_types.rst
```diff
@@ -61,15 +61,15 @@ We can now write a simple policy over these Actor types:
 .. tabs::
   .. group-tab:: Python
 
-    .. literalinclude:: /examples/user_types/python/user_policy.polar
+    .. literalinclude:: /examples/user_types/user_policy.polar
```
gj, line 64: If these are all the same file now, we don't need tabs

docs/using/examples/user_types.rst
```diff
@@ -176,15 +176,15 @@ Then add the following rule to our policy:
...
         :caption: :fa:`oso` user_types.polar
         :language: polar
         :start-after: rbac-start
         :end-before: rbac-end
 
   .. group-tab:: Ruby
 
-    .. literalinclude:: /examples/user_types/ruby/user_policy.polar
+    .. literalinclude:: /examples/user_types/user_policy.polar
```
gj, line 187: If these are all the same file, we don't need tabs

gj (changes requested):

docs/examples/user_types/python/test_example.py
saolsen: `sudo code`

docs/examples/quickstart/python/server.py
saolsen: If the users download and run the file it'll work the same as it did before. I agree that actually testing the web servers is better but this still validates it more than we did before.

docs/examples/user_types/ruby/user_policy.polar
```diff
@@ -1,26 +0,0 @@
...
-
-# Customers only have access to the customer dashboard
-allow(actor: Customer, "view", "customer_dashboard");
-# simple-end
-
-# rbac-start
-# Internal users can access the accounts dashboard if
-# they are an account manager
-allow(actor: InternalUser, "view", "accounts_dashboard") if
-    actor.role = "account_manager";
```
gj, line 15: The Ruby version of this policy used idiomatic Ruby method calls (no parens for a no-arg call)

docs/examples/user_types/nodejs/user_policy.polar
```diff
@@ -1,26 +0,0 @@
...
-# rbac-end
-
-# manager-start
-# Account managers can access the accounts dashboard
-allow(actor: AccountManager, "view", "accounts_dashboard");
-
-# Account managers can access account data for the accounts
-# that they manage
-allow(actor: AccountManager, "view", resource: AccountData) if
-    resource.accountId = actor.customerAccounts();
```
gj, line 25: the Node.js and Java versions of this policy camelCased attributes and methods, which is idiomatic for those languages

gj (changes requested):
I think we should revert the consolidation of the user types examples -- see other comments on idiomatic touches lost in the consolidation

dhatch:
What's up with this?

docs/examples/user_types/nodejs/user_policy.polar
```diff
@@ -1,26 +1,26 @@
...
+allow(actor: InternalUser, "view", "accounts_dashboard") if	
+    actor.role() = "account_manager";	
+# rbac-end	
 
-# manager-start
-# Account managers can access the accounts dashboard
-allow(actor: AccountManager, "view", "accounts_dashboard");
+# manager-start	
+# Account managers can access the accounts dashboard	
+allow(actor: AccountManager, "view", "accounts_dashboard");
```
gj, line 20: your editor does the strangest things 😆

saolsen: wtf

docs/getting-started/quickstart.rst
```diff
@@ -181,6 +181,7 @@ start a REPL session and follow along:
 
     .. code-block:: pycon
 
+      >>> oso.register_class(Expense)
```
gj, line 184: Looks like the Ruby + Java versions could use this same change

gj (approved):

PR #402, opened 2020-09-03
cache by build instead of workflow

saolsen:


samscott89:
What's the advantage of this over per-workflow? The two wont have the same caches. `cargo build --release` wont build the bench targets for example?

saolsen:
https://github.com/osohq/oso/pull/399/files/db20c9eb230d0d75b83640b9555af1f87fc06198#diff-2e7822c9495b4cd7a05389935ddc44ed

I sorta might agree though, if bench has different dependencies than just the release build it might make sense to continue cacheing them separately.

gj:
> What's the advantage of this over per-workflow? The two wont have the same caches. `cargo build --release` wont build the bench targets for example?

Just tested this locally to confirm that `cargo bench` and `cargo build --release` overlap, significantly decreasing compile time over compiling both independently. `cargo bench` doesn't build the `polar-wasm-api` crate, and yes `cargo build --release` won't build the bench targets, but I think this is still worth doing.

gj (approved):
Wait for @samscott89's approval as well, but this is fine by me

samscott89 (approved):

PR #403, opened 2020-09-03
Add some common aliases to unknown specializer message

saolsen:
Not trying to be clever right now. Though in the future some algorithm that actually checks for "misspellings" and compares to registered types too would be cool.

For now just a hardcoded list of common aliases that other languages use for the builtin types we use in polar.


polar-core/src/warnings.rs
```diff
@@ -5,6 +5,42 @@ use super::terms::*;
...
+        "int" => Some("Number"),
+        "Integer" => Some("Number"),
+        "i32" => Some("Number"),
+        "i64" => Some("Number"),
+        "u32" => Some("Number"),
+        "u64" => Some("Number"),
+        "size_t" => Some("Number"),
+        "usize" => Some("Number"),
+        "float" => Some("Number"),
+        "Float" => Some("Number"),
```
samscott89, line 19: Wait, we do register `Float` and `Integer` as separate types right?

polar-core/src/warnings.rs
```diff
@@ -5,6 +5,42 @@ use super::terms::*;
 
 use std::collections::{hash_map::Entry, HashMap};
 
+pub fn common_misspellings(t: &str) -> Option<String> {
+    let misspelled_type = match t {
```
samscott89, line 9: This would be a little less painful if it was:

```
let misspelled_type = match t { 
   "int" => "Number",
   ...
  _ => return None
};
Some(misspelled_type)

polar-core/src/warnings.rs
gj: yeah these should be `Float` or `Integer`; not `Number`

samscott89 (approved):

samscott89 (approved):
A test would be nice (see `test_print`).

PR #405, opened 2020-09-04
Mixed positional + kwargs NON-BREAKING

leina05:
PR checklist:

- [x] Added changelog entry.
- [ ] Mixed method calls


samscott89:
Is this (intentionally or otherwise) only mixed pos+kw args for constructors? Or should it also work for methods? I couldn't see any changes in the register_call side of things for python for example.

polar-core/src/polar.lalrpop
```diff
@@ -99,15 +99,21 @@ Variable: Value  = <n:Name> => {
     Value::Variable(n)
 };
 
-Predicate: Value = {
+Call: Value = {
     <name:Name> "(" <mut args:(<TermExp> ",")*> <arg:TermExp?> ")" => {
```
gj, line 103: off-topic: does this technically mean that `f(1,)` is valid Polar?

polar-core/src/vm.rs
```diff
@@ -1117,7 +1117,11 @@ impl PolarVirtualMachine {
         check_errors: bool,
     ) -> PolarResult<QueryEvent> {
         let (field_name, args): (Symbol, Option<Vec<Term>>) = match self.deref(field).value() {
-            Value::Call(Predicate { name, args }) => (
+            Value::Call(Call {
+                name,
+                args,
+                kwargs: None,
```
gj, line 1123: Why are we only matching `None` for kwargs?

polar-core/src/types.rs
```diff
@@ -900,13 +908,14 @@ mod tests {
 
     #[test]
     fn serialize_test() {
-        let pred = Predicate {
+        let pred = Call {
             name: Symbol("foo".to_owned()),
             args: vec![Term::new_from_test(value!(0))],
+            kwargs: None,
```
gj, line 914: would be good to add an additional test case for non-null kwargs

polar-core/src/types.rs
```diff
@@ -114,9 +114,10 @@ impl Symbol {
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq, Hash)]
-pub struct Predicate {
+pub struct Call {
     pub name: Symbol,
     pub args: TermList,
+    pub kwargs: Option<BTreeMap<Symbol, Term>>,
```
gj, line 120: Is this a BTreeMap b/c of the derivation of Hash, or is there another reason? Just curious b/c I wouldn't think the order of the kwargs would matter (though maybe it does for printing purposes?

polar-core/src/parser.rs
```diff
@@ -240,6 +240,22 @@ mod tests {
...
+            results[0].to_polar(),
+            r#"a(x) if x = new Foo(1, 2, bar: 3);"#
+        );
+        let f = r#"a(x) if x = new Foo(bar: 3, baz: 4);"#;
+        let results = parse_rules(0, f).unwrap();
+        assert_eq!(
+            results[0].to_polar(),
+            r#"a(x) if x = new Foo(bar: 3, baz: 4);"#
+        );
+    }
```
gj, line 257: nit: It would be great to add a case that expects an error if a positional arg trails a kwarg

polar-core/src/parser.rs
```diff
@@ -240,6 +240,22 @@ mod tests {
         assert_eq!(results[0].to_polar(), r#"a(x) if x = new Foo(1, 2);"#);
     }
 
+    #[test]
+    fn test_parse_new_mixed_args() {
+        let f = r#"a(x) if x = new Foo(1, 2, bar: 3);"#;
```
gj, line 245: Could we add a second kwarg here? I'd just feel better about having multiple positional + multiple kwargs in the test.

languages/ruby/spec/oso/polar/polar_spec.rb
```diff
@@ -216,7 +216,7 @@ def initialize(bar:, baz:)
       it 'handles no args' do
         stub_const('Foo', Class.new)
         subject.register_class(Foo, from_polar: -> { Foo.new })
-        id = subject.host.make_instance('Foo', initargs: {}, id: 1)
+        id = subject.host.make_instance('Foo', args: [], kwargs: {}, id: 1)
```
gj, line 219: Can we add two new test cases here, one that tests args w/o kwargs and one that tests both args and kwargs at the same time?

languages/ruby/lib/oso/polar/query.rb
```diff
@@ -104,12 +104,37 @@ def handle_call(attribute, call_id:, instance:, args:)
         call_result(nil, call_id: call_id)
       end
 
+      def handle_make_external(data) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength, Metrics/PerceivedComplexity, Metrics/CyclomaticComplexity
+        id = data['instance_id']
+        raise DuplicateInstanceRegistrationError, id if host.instance? id
+
+        constructor = data['constructor']['value']
+        if constructor.key? 'InstanceLiteral'
```
gj, line 112: It looks like this PR removes the ability to parse this, no? Or is this still used for patterns or something?

polar-core/src/vm.rs
gj: I think this is related to @samscott89's comment: https://github.com/osohq/oso/pull/405#issuecomment-687264166

leina05:
> Is this (intentionally or otherwise) only mixed pos+kw args for constructors? Or should it also work for methods? I couldn't see any changes in the register_call side of things for python for example.

Yeah, currently it's only implemented for constructors. It would be easy to implement on methods as well, but I didn't think of it earlier. I think we should probably implement it for methods.

languages/java/oso/src/test/java/com/osohq/oso/PolarTest.java
```diff
@@ -209,8 +209,7 @@ public void testRegisterAndMakeClass() throws Exception {
 
     @Test
     public void testDuplicateRegistration() throws Exception {
-        assertThrows(Exceptions.DuplicateClassAliasError.class,
-                     () -> p.registerClass(MyClass.class, "MyClass"));
+        assertThrows(Exceptions.DuplicateClassAliasError.class, () -> p.registerClass(MyClass.class, "MyClass"));
```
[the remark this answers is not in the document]
leina05: VSCode has a default formatter but I don't actually know how to run it from the command line, so I can just turn my formatter off. I think others probably have a formatter also though.

polar-core/src/polar.lalrpop
leina05:
```
query> f(1,)
False
```
yerp

leina05: Fixed

polar-core/src/vm.rs
leina05: Correct, it's because there's no support for mixed args in method calls. But there could be.

polar-core/src/types.rs
leina05: Yeah, just did it because that's what `Fields` is (for instance literals).

languages/java/oso/src/test/java/com/osohq/oso/PolarTest.java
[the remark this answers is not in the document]
gj: Okay, picking one and adding it into CI would be a good follow-up task. I'll create one.

gj:
> Yeah, currently it's only implemented for constructors. It would be easy to implement on methods as well, but I didn't think of it earlier. I think we should probably implement it for methods.

Agree

languages/ruby/lib/oso/polar/query.rb
leina05: Ah yeah I forgot about the parser.

languages/ruby/spec/oso/polar/polar_spec.rb
leina05: Added here as well as to the python tests

leina05:
Okay, still no support for mixed method calls, but other comments have been addressed.

languages/ruby/lib/oso/polar/host.rb
```diff
@@ -118,24 +118,28 @@ def cache_instance(instance, id: nil)
...
-                     constructor.call(**initargs)
+                   elsif args.empty? && kwargs.is_a?(Hash)
+                     constructor.call(**kwargs)
+                   elsif kwargs.empty? && args.is_a?(Array)
+                     constructor.call(*args)
+                   elsif args.is_a?(Array) && kwargs.is_a?(Hash)
+                     constructor.call(*args, **kwargs)
                    else
                      raise PolarRuntimeError, "Bad initargs: #{initargs}"
                    end
```
gj, line 145: This whole thing looks like it can simplify down to

```suggestion
      def make_instance(cls_name, args:, kwargs:, id:)
        constructor = get_constructor(cls_name)
        instance = if constructor == :new
                     get_class(cls_name).__send__(:new, *args, **kwargs)
                   else
                     constructor.call(*args, **kwargs)
                   end
```

gj (changes requested):

gj:
Somewhat on-topic: do you remember why we added support for custom constructors to the Python lib? I copied that over to the Ruby lib and justified it before we added positional args support since it allowed users to pass a custom constructor that converted kwargs to positional args, but I'm not sure it makes sense to keep it now.

samscott89:
> 
> 
> Somewhat on-topic: do you remember why we added support for custom constructors to the Python lib? I copied that over to the Ruby lib and justified it before we added positional args support since it allowed users to pass a custom constructor that converted kwargs to positional args, but I'm not sure it makes sense to keep it now.

I think it was probably from before we had class methods. I kinda agree custom constructors don't make a whole lot of sense now.

languages/java/oso/src/main/java/com/osohq/oso/Query.java
```diff
@@ -7,6 +7,7 @@
 import java.io.IOException;
 import java.io.InputStreamReader;
 import java.util.*;
+import java.util.concurrent.ForkJoinWorkerThread;
```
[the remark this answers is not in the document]
leina05: nothing

languages/python/oso/tests/test_polar.py
```diff
@@ -461,44 +461,44 @@ def test_query(load_file, polar, query):
...
-            one=True,
-        )
-        == 2
+    assert instance.a == 1
+    assert instance.b == 2
+    assert instance.bar == 3
+    assert instance.baz == 4
+
+    instance = qvar(
+        "instance = new Foo(a: 1, b: 2, bar: 3, baz: 4)",
```
gj, line 494: nit: it would be cool to mix up the order here to ensure they're getting translated into true kwargs

gj (approved):

docs/changelogs/0.5.2.rst
```diff
@@ -0,0 +1,44 @@
...
+
+Support for constructors with both positional and keyword arguments
+===================================================================
+
+- In languages that support both positional and keyword arguments (currently Python and Ruby), oso now supports registering classes with constructors that use both.
+- The syntax for calling a constructor with mixed arguments from an oso policy is::
+
+    instance = new MyClass(1, 2, foo: 3)
+
+- As is the convention in Python and Ruby, positional arguments must come before keyword arguments.
```
gj, line 35:
```suggestion
- As is mandated in Python and Ruby, positional arguments must come before keyword arguments.
```

PR #406, opened 2020-09-04
Python exceptions overhaul

leina05:
PR checklist:

- [x] Added changelog entry.


languages/java/oso/.project
```diff
@@ -20,4 +20,15 @@
 		<nature>org.eclipse.jdt.core.javanature</nature>
 		<nature>org.eclipse.m2e.core.maven2Nature</nature>
 	</natures>
+	<filteredResources>
```
dhatch, line 23: What's this for?

languages/python/oso/polar/exceptions.py
```diff
@@ -3,71 +3,163 @@
...
+        self.message = message
+        self.details = details
+        self.stack_trace = details.get("stack_trace") if details else None
+        super().__init__(self.message)
 
-    def __init__(self, message, error=None):
-        super(PolarException, self).__init__(message)
-        self._inner = error
 
+class FFIErrorNotFound(OsoError):
```
dhatch, line 14: Could we add documentation strings to this module & exception types indicating when they would occur?

dhatch (commented):
Do we need to update the python integrations as well with the new exception names?

languages/java/oso/.project
leina05: oh this is from something else, will remove

languages/python/flask-oso/tests/test_flask_oso.py
```diff
@@ -55,6 +55,8 @@ def test_initialization(flask_app, oso, simple_policy, app_ctx, user):
     flask_oso.set_oso(oso)
     flask_oso.authorize(action="read", resource="resource")
 
+
+def test_initialization_with_init(flask_app, oso, simple_policy, app_ctx, user):
```
samscott89, line 59: @dhatch I had to split these tests, since initializing FlaskOso twice with the same Oso instance attempts to register Request twice, which now (correctly) raises an exception.

samscott89 (approved):

PR #409, opened 2020-09-07
Add benchmark for simple indexing + some simple optimisations 

samscott89:
Added a benchmark to check to that "trivial" indexing is approximately linear in complexity.
(This is a nice resource to see why we would expect that to be the case: https://github.com/Rufflewind/bench-maps) 

There were some performance optimisations to be made across the board, but also a specific one for indexing which has significantly increased that benchmark.

There was also a huge change from using [`iter_batched_ref`](http://bheisler.github.io/criterion.rs/criterion/struct.Bencher.html#method.iter_batched_ref) which, I think, basically removes the time to drop the entire KB after running the query. The before/after below is cherry-picking that change too

The change was moving the borrow of the kb rules so we can avoid cloning. Overall on the newly added benchmarks, this is the improvement:

```
indexed/1               time:   [4.9974 us 5.1745 us 5.3621 us]
                        change: [-39.293% -36.543% -33.355%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild

indexed/10              time:   [5.3793 us 5.5503 us 5.7522 us]
                        change: [-36.308% -33.545% -30.729%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild

indexed/100             time:   [7.0676 us 7.6192 us 8.2091 us]
                        change: [-69.387% -67.421% -65.173%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 6 outliers among 100 measurements (6.00%)
  6 (6.00%) high mild
```

### DIFF
--- a/polar-core/src/vm.rs
+++ b/polar-core/src/vm.rs
@@ -691,11 +691,6 @@ impl PolarVirtualMachine {
         rule
     }
 
-    /// Get a generic rule by name.
-    fn get_generic_rule(&self, name: &Symbol) -> Option<GenericRule> {
-        self.kb.read().unwrap().rules.get(name).cloned()
-    }
-
     /// Print a message to the output stream.
     fn print(&self, message: &str) {
         self.messages.push(MessageKind::Print, message.to_owned());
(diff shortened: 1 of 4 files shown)


polar-core/src/vm.rs
```diff
@@ -1286,8 +1300,8 @@ impl PolarVirtualMachine {
     /// Sort applicable rules by specificity.
     /// Create a choice over the applicable rules.
     fn query_for_predicate(&mut self, predicate: Predicate) -> PolarResult<()> {
-        match self.get_generic_rule(&predicate.name) {
-            None => self.push_goal(Goal::Backtrack)?,
+        let goals = match self.kb.read().unwrap().rules.get(&predicate.name) {
```
samscott89, line 1303: This is slightly worse to read than the previous version. But I think the only alternative would be to create a new wrapper type that holds the RwLock and derefs to predicate.

plotnick (approved):
Looks good.

PR #410, opened 2020-09-08
Handle ±∞ and NaN across FFI

plotnick:
JSON does not directly support ±∞ or NaN, so we encode them as the strings `Infinity`, `-Infinity`, and `NaN`.

### DIFF
--- a/languages/java/oso/src/main/java/com/osohq/oso/Host.java
+++ b/languages/java/oso/src/main/java/com/osohq/oso/Host.java
@@ -287,7 +287,7 @@ public class Host implements Cloneable {
                                     return Double.NaN;
                                 default:
                                     throw new Exceptions.OsoException(
-                                        "Expected a Float; received the string \"" + f + "\"");
+                                        "Expected a floating point number, got \"" + f + "\"");
                             }
                         }
                         return (Double) f;
--- a/languages/java/oso/src/main/java/com/osohq/oso/Host.java
+++ b/languages/java/oso/src/main/java/com/osohq/oso/Host.java
@@ -154,7 +154,7 @@ public class Host implements Cloneable {
      * Check if a Java instance is an instance of a class.
      */
     public boolean isa(JSONObject instance, String classTag) throws Exceptions.UnregisteredClassError,
-            Exceptions.UnregisteredInstanceError, Exceptions.UnexpectedPolarTypeError {
+            Exceptions.UnregisteredInstanceError, Exceptions.UnexpectedPolarTypeError, Exceptions.OsoException {
         Class<?> cls = getClass(classTag);
         return cls.isInstance(toJava(instance));
     }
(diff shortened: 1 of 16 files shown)


languages/js/src/Host.ts
```diff
@@ -201,7 +202,14 @@ export class Host {
         return { value: { Boolean: v } };
       case Number.isInteger(v):
         return { value: { Number: { Integer: v } } };
-      case typeof v === 'number' && Number.isFinite(v):
+      case typeof v === 'number':
+        if (v == Infinity) {
+          v = 'Infinity';
+        } else if (v == -Infinity) {
```
gj, line 208: Inconsequential here, but best practice generally.

```suggestion
        if (v === Infinity) {
          v = 'Infinity';
        } else if (v === -Infinity) {
```

languages/js/src/Host.ts
```diff
@@ -237,7 +245,19 @@ export class Host {
...
+        switch (f) {
+          case 'Infinity':
+            return Infinity;
+          case '-Infinity':
+            return -Infinity;
+          case 'NaN':
+            return NaN;
+          default:
+            if (typeof f !== 'number')
+              throw new PolarError('Expected a floating point number.');
```
gj, line 258: This message + the corresponding Java message should match.

languages/python/oso/polar/host.py
```diff
@@ -1,5 +1,7 @@
 """Translate between Polar and the host language (Python)."""
 
+from math import inf, isinf, isnan, nan
```
gj, line 3:
```suggestion
from math import inf, isnan, nan
```

languages/python/oso/tests/test_polar.py
```diff
@@ -1,5 +1,5 @@
 from datetime import datetime, timedelta
-from math import pi
+from math import inf, isnan, nan, pi
```
gj, line 2:
```suggestion
from math import inf, isnan, nan
```

languages/java/oso/src/main/java/com/osohq/oso/Host.java
```diff
@@ -268,7 +276,21 @@ public Object toJava(JSONObject term)
...
+                            switch ((String) f) {
+                                case "Infinity":
+                                    return Double.POSITIVE_INFINITY;
+                                case "-Infinity":
+                                    return Double.NEGATIVE_INFINITY;
+                                case "NaN":
+                                    return Double.NaN;
+                                default:
+                                    throw new Exceptions.OsoException(
+                                        "Expected a Float; received the string \"" + f + "\"");
```
gj, line 290: on second thought do we actually need this check since we control both ends of this serialization pipe and know this will always be one of these three strings?

plotnick: These checks are basically assertions. We "know" it will be one of the three special strings, but if it's *not*, I want to know about it.

gj: Word. In the spirit of keeping errors aligned across the lang libs, should we add the same assertion to Ruby and Python?

polar-core/src/numerics.rs
```diff
@@ -1,16 +1,78 @@
...
 }
 
+/// Since JSON does not support ±∞ or NaN (RFC 8259 §6),
+/// we encode them as magic strings.
+fn serialize_float<S>(f: &f64, s: S) -> Result<S::Ok, S::Error>
+where
+    S: Serializer,
+{
+    match f.classify() {
+        FpCategory::Zero => s.serialize_f64(*f),
```
samscott89, line 28: Merge with the last variant?

languages/java/oso/src/main/java/com/osohq/oso/Host.java
plotnick: Yes, I had that thought, and also thought, "Gabe's going to ask about the others." Thank you for not letting me ignore it.

gj (approved):

samscott89 (approved):
LGTM! Good stuff

PR #412, opened 2020-09-08
bug fix, parse kwargs only in constructor

saolsen:


polar-core/src/polar.lalrpop
```diff
@@ -117,12 +120,15 @@ Call: Value = {
         let kwargs = None;
         Value::Call(Call{name, args, kwargs})
     },
+};
+
+Call: Value = {
```
gj, line 125: nit: can we call this `Construction` and the other one `Call`?

samscott89 (approved):

polar-core/src/parser.rs
```diff
@@ -266,6 +266,12 @@ mod tests {
 
         let f = r#"a(x) if x = new Foo(bar: 3, baz: 4, 1, 2);"#;
         parse_rules(0, f).expect_err("parse error");
+
+        // Don't allow varargs in calls or dot ops.
```
gj, line 270: Matches the language we use elsewhere

```suggestion
        // Don't allow kwargs in calls or dot ops.
```

PR #417, opened 2020-09-09
Any specializer is more specific than no specializer

leina05:
PR checklist:

- [ ] Added changelog entry.

This needs more tests, but wondering if anyone has ideas about pushing the tests into Rust.

### DIFF
--- a/polar-core/src/vm.rs
+++ b/polar-core/src/vm.rs
@@ -2304,7 +2304,7 @@ impl PolarVirtualMachine {
                     }
                 }
                 // If the left rule has no specializer and the right does, it is NOT more specific,
-                // so we fall through to Backtrack (fail)
+                // so we Backtrack (fail)
                 (None, Some(_)) => return self.push_goal(Goal::Backtrack),
                 // If the left rule has a specializer and the right does not, the left IS more specific,
                 // so we return
--- a/polar-core/src/vm.rs
+++ b/polar-core/src/vm.rs
@@ -2309,7 +2309,7 @@ impl PolarVirtualMachine {
                 // If the left rule has a specializer and the right does not, the left IS more specific,
                 // so we return
                 (Some(_), None) => return Ok(()),
-                // If neither has a specializer, neither is more specific, so we fail
+                // If neither has a specializer, neither is more specific, so we continue to the next argument.
                 (None, None) => (),
             }
         }
--- a/polar-core/src/vm.rs
+++ b/polar-core/src/vm.rs
@@ -2274,41 +2274,46 @@ impl PolarVirtualMachine {
     fn is_more_specific(&mut self, left: &Rule, right: &Rule, args: &TermList) -> PolarResult<()> {
         let zipped = left.params.iter().zip(right.params.iter()).zip(args.iter());
         for ((left_param, right_param), arg) in zipped {
-            // TODO: Handle the case where one of the params has a specializer and the other does
-            // not. The original logic in the python code was that a param with a specializer is
-            // always more specific than a param without.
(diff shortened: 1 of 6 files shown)


polar-core/src/vm.rs
```diff
@@ -2289,41 +2289,46 @@ impl PolarVirtualMachine {
...
+                            Goal::Unify {
+                                left: Term::new_temporary(Value::Variable(answer)),
+                                right: Term::new_temporary(Value::Boolean(true)),
+                            },
+                        ]);
+                    }
                 }
+                // If the left rule has no specializer and the right does, it is NOT more specific,
+                // so we fall through to Backtrack (fail)
+                (None, Some(_)) => (),
```
dhatch, line 2323: This needs to either break from the loop or return backtrack immediately (probably preferable).  Otherwise, a multiple argument rule will not work properly (and we should add a test for that).

polar-core/src/vm.rs
```diff
@@ -2289,41 +2289,46 @@ impl PolarVirtualMachine {
...
+                // If the left rule has a specializer and the right does not, the left IS more specific,
+                // so we return
+                (Some(_), None) => return Ok(()),
+                // If neither has a specializer, neither is more specific, so we fail
+                (None, None) => (),
```
dhatch, line 2328: Here, I don't think we want to fail, but just want to continue.  Right now specificity is equal, so we must go to the next argument.

dhatch (changes requested):

polar-core/src/vm.rs
leina05: Isn't that what this does?

polar-core/src/vm.rs
leina05: Right. So the behavior is that the first specializer determines ordering.

dhatch: This is in a for loop so this will continue to the next specializer, right, not fail? I'm thinking of a call like:

```
left: f(a, b: B), right: f(a: A, b))
```

This would succeed because the first iteration would continue, and the second return Ok.  Instead, we want the first iteration to immediately backtrack.

polar-core/src/vm.rs
dhatch: Yes, in contrast to the first case though & the comment is incorrect.  We aren't failing.  We are just continuing.

polar-core/src/vm.rs
```diff
@@ -2289,43 +2289,47 @@ impl PolarVirtualMachine {
...
+                            },
+                            Goal::Unify {
+                                left: Term::new_temporary(Value::Variable(answer)),
+                                right: Term::new_temporary(Value::Boolean(true)),
+                            },
+                        ]);
+                    }
                 }
+                // If the left rule has no specializer and the right does, it is NOT more specific,
+                // so we fall through to Backtrack (fail)
```
dhatch, line 2322:
```suggestion
                // so we Backtrack (fail)
```

polar-core/src/vm.rs
```diff
@@ -2289,43 +2289,47 @@ impl PolarVirtualMachine {
...
+                (None, Some(_)) => return self.push_goal(Goal::Backtrack),
+                // If the left rule has a specializer and the right does not, the left IS more specific,
+                // so we return
+                (Some(_), None) => return Ok(()),
+                // If neither has a specializer, neither is more specific, so we fail
```
dhatch, line 2327:
```suggestion
                // If neither has a specializer, neither is more specific, so we continue to the next argument.
```

dhatch (approved):
lgtm, sorry this took so long!

PR #418, opened 2020-09-10
re-org to set up proc-macro crate

saolsen:


dhatch:
This can be closed now, right?

PR #419, opened 2020-09-11
Proc macros

saolsen:
Works. I need to clean up some of the attribute finding code and decide if bad attributes are errors or just ignored. There's some extra match branches in there just because I sort of expect those paths to be useful when we add more attributes and I don't want to forget about them.

Right now it lets you specify the class name and which fields you want getters for.
It derives a few functions on an interface, one that returns a full class definition and one that returns a partially filled out builder that you can then add more methods to.

I think it's almost good enough for v1. I have to make constructors be their own builder method since we can't get them up front from an attribute (yet). Right now it defaults to `new` but we still want people to be able to specify a custom one.

Over all this went pretty well.


languages/rust/oso/tests/test_polar.rs
```diff
@@ -244,3 +245,43 @@ fn test_external() {
...
+        .name("Baz")
+        .add_method("world", |receiver: &Foo| format!("{} world", receiver.a))
+        .build();
+    test.oso.register_class(class).unwrap();
+
+    test.qvar_one(
+        r#"new Baz("hello").world() = x"#,
+        "x",
+        "hello world".to_string(),
+    );
```
saolsen, line 286: Here's what the usage looks like.

languages/rust/oso/src/host/class.rs
```diff
@@ -160,13 +178,17 @@ impl<T> Class<T> {
...
+        if let Some(constructor) = &self.constructor {
+            let instance = constructor.invoke(fields, host)?;
+            Ok(Instance {
+                name: self.name.clone(),
+                instance,
+                attributes: Arc::new(self.attributes.clone()),
+                methods: Arc::new(self.instance_methods.clone()),
+            })
+        } else {
+            panic!("Invalid class, should already have a constructor")
```
samscott89, line 190: Classes don't _have_ to be constructable, right? No constructor seems okay, should be a regular error.

saolsen: Oh true, cool.

languages/rust/oso-derive/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+                    get_nested_attr(nested, oso_attrs);
+                }
+            }
+            _ => (),
+        }
+    }
+}
+
+#[proc_macro_derive(PolarClass, attributes(polar))]
+pub fn derive_testing_fn(ts: TokenStream) -> TokenStream {
```
dhatch, line 79: Why this function name?

dhatch (approved):
lgtm, todo clean up as follow up

languages/rust/oso/tests/test_polar.rs
```diff
@@ -244,3 +245,46 @@ fn test_external() {
...
+#[test]
+fn test_macros() {
+    // stub
+
+    let _ = tracing_subscriber::fmt::try_init();
+
+    #[derive(PolarClass)]
+    #[polar(class_name = "Bar")]
+    struct Foo {
+        #[polar(attribute)]
```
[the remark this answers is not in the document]
saolsen: Added

languages/rust/oso-derive/src/lib.rs
saolsen: old name, fixed

PR #420, opened 2020-09-11
Fix rule filtering bug + update POLAR_LOG formatting

gj:
PR checklist:
- [x] Add additional test coverage of rule filtering.


polar-core/src/vm.rs
```diff
@@ -2137,11 +2113,15 @@ impl PolarVirtualMachine {
...
             }
 
-            // Try to unify the arguments with renamed parameters.
-            // TODO: Think about using backtrack so that we don't
-            // leave temporary bindings around.
-            let args = self.rename_vars(args.clone());
+            // We need to check each argument for applicability,
+            // but unwind any bindings made during the check. So
+            // we will backtrack to this choice point on success.
+            self.push_choice(vec![vec![applicable]]);
```
samscott89, line 2119: I'm struggling to get enough context to reason about this without reading through the full method again. What does the choice stack look like at the end of this method?
Something like:

[
- check_applicable goals (unify goals followed by backtrack, so backtracks *if successful*)
- inapplicable rules - has a cut so applicable rules doesn't get executed afterwards?

], 
[
 - applicable rules <- this is pushed onto the "current" choice stack so is an alternative after the head (here represented above) is completed

]

samscott89: Feels like a double negative somehow? Why isn't the choice point: [check_applicable + backtrack + cut + applicable, inapplicable] for example?

samscott89: Wait, backtrack would clear the other goals wouldn't it. Hmm, I guess that makes sense. The use of `push_choice` feels a little too like manual stack manipulation to me.

samscott89: Also, if the applicable check succeeds, doesn't this mean we still would try the inapplicable branch too?

plotnick: No, it's cut. But I will work on a better comment or diagram to explain the admittedly confusing flow control.

samscott89: Oh I see, I missed https://github.com/osohq/oso/pull/420/files/433da70ca8ee5acac745276a73dbccfce5e72995#diff-99e04680f3976c570014561774238bdeR2138

polar-core/src/vm.rs
```diff
@@ -2158,8 +2138,19 @@ impl PolarVirtualMachine {
             check_applicability.push(Goal::Cut {
                 choice_index: self.choices.len(),
             });
-            check_applicability.push(applicable);
-            self.choose(vec![check_applicability, vec![inapplicable]])?;
+            check_applicability.push(Goal::Backtrack);
+
+            // If a check fails, we will backtrack to the inapplicable branch.
+            self.choose(vec![
```
samscott89, line 2144: Maybe replace this with a `choose_exactly_one` or something? Is that the intention? Or something like a choose_xor? Do applicable XOR inapplicable?

polar-core/src/vm.rs
```diff
@@ -2158,8 +2138,19 @@ impl PolarVirtualMachine {
             check_applicability.push(Goal::Cut {
```
dhatch, line 2138: Can we add a comment about what this is cutting (the inapplicable branch below)

polar-core/src/vm.rs
```diff
@@ -1272,7 +1248,7 @@ impl PolarVirtualMachine {
             Value::Expression(Operation {
                 operator: Operator::And,
                 args,
-            }) if args.len() == 1 => (),
+            }) if args.len() < 2 => (),
```
gj, line 1251: This change makes it so that we no longer print empty rule bodies.

dhatch (commented):
Discussed offline:

The implementation seems correct to me and makes sense. Main suggestion is to refactor the goal & choice stack manipulation as another function.  This is just an if / else construct.  We use the same in sort rules.  It makes sense to have a flow control helper that allows us to achieve that.

polar-core/src/vm.rs
gj: I could see it as something like `push_conditional(conditions: Vec<Goal>, consequent: Goal, alternative: Goal)`, so in this case:

```rust
push_conditional(check_applicability, applicable, inapplicable)
```

gj: Looks like @dhatch had similar thoughts: https://github.com/osohq/oso/pull/420#pullrequestreview-488075393

samscott89: Looks _so_ much better with this

samscott89 (approved):
With `choose_conditional` this looks great + will be nice to use for more things in the future.

PR #421, opened 2020-09-12
Don't translate errors that shouldn't ever occur into query failures

gj:
Don't translate errors that _shouldn't_ ever occur into query failures; let them propagate.


leina05 (approved):
LGTM

PR #423, opened 2020-09-14
Proof of concept: scope helpers for django.

samscott89:
Making filters easier to use with django: create an `OsoManager` that runs approved filters on the object set.

This works fairly well with the Django `Q` class. But an alternative would be to return strings, and have the user provided a dictionary from filter name to the filter operation (a lambda on the recordset?)

PR checklist:

- [ ] Added changelog entry.

### DIFF
--- a/languages/python/django-oso/django_oso/manager.py
+++ b/languages/python/django-oso/django_oso/manager.py
@@ -1,0 +1,41 @@
+import threading
+from django.db import models
+from django.db.models import Q
+
+from .oso import Oso
+from oso import Variable
+
+
+class OsoManager(models.Manager):
+    _requests = {}
+
+    @classmethod
+    def get_request(cls, default=None):
+        """
+        Retrieve the request object for the current thread, or the optionally
+        provided default if there is no current request.
+        """
+        return cls._requests.get(threading.current_thread(), default)
+
+    @classmethod
+    def set_request(cls, request):
+        """
+        Save the given request into storage for the current thread.
+        """
+        cls._requests[threading.current_thread()] = request
+
+    @classmethod
+    def del_request(cls):
+        """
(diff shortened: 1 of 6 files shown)


languages/python/django-oso/django_oso/manager.py
```diff
@@ -0,0 +1,41 @@
...
+
+from .oso import Oso
+from oso import Variable
+
+
+class OsoManager(models.Manager):
+    _requests = {}
+
+    @classmethod
+    def get_request(cls, default=None):
```
samscott89, line 13: This doesn't make sense. Maybe we should wrap this up in django_oso somewhere?

samscott89: (Doesn't make sense to be _here_)

languages/python/django-oso/tests/test_app/models.py
```diff
@@ -9,3 +10,14 @@ class Meta:
...
     class Meta:
         app_label = "test_app"
+
+
+class TestScope(models.Model):
+    public = models.BooleanField()
+    user = models.CharField(max_length=256)
+
+    objects = models.Manager()
+    authorized_objects = OsoManager()
```
samscott89, line 20: Could presumably provide this via a mixin?

languages/python/django-oso/django_oso/oso.py
```diff
@@ -19,8 +20,8 @@ def reset_oso():
     Useful as a test helper to clean state between tests, but generally should
     not be used otherwise.
     """
-    Oso.clear()
-    init_oso()
+    # Oso.clear()
```
samscott89, line 23: This was resulting in none of the classes being registered.

PR #424, opened 2020-09-14
Rust tuple struct test

saolsen:
Test to make sure tuple structs work. There's some stuff we can't do now but would be cool in the future like auto generating accessors.


dhatch (approved):
lgtm, can you add a task with this todo in notion? no engineers column of this cycle or next

PR #425, opened 2020-09-14
Rust result and option method calls

saolsen:
Errors if the result is Err, returns the value if it's Ok

I did it in calling the method rather than just implementing `ToPolar` for `Result<x,y>`. Doing it by implementing `ToPolar` would work fine in this case (returning the value on success and erring on error) but it wouldn't work for the `Option` case. In the `Option` case I think we want to return the value if it's Some and not return anything if it's None. By the time we get to serialization it's too late for that.

The option case works like the result case right now and throws an error if it's `None`. I don't think that's correct though, it should probably return nothing if it's a None. I think the easiest way to do that would just be return an empty iterator but we don't have those yet.

### DIFF
--- a/languages/rust/oso/src/host/class.rs
+++ b/languages/rust/oso/src/host/class.rs
@@ -5,10 +5,9 @@ use polar_core::terms::{Symbol, Term};
 use std::any::{Any, TypeId};
 use std::collections::HashMap;
 use std::fmt;
-use std::fmt::Debug;
 use std::sync::Arc;
 
-use crate::{FromPolar, ToPolar};
+use crate::FromPolar;
 
 use super::class_method::{ClassMethod, Constructor, InstanceMethod};
 use super::method::{Function, Method};
--- a/languages/rust/oso/src/host/class.rs
+++ b/languages/rust/oso/src/host/class.rs
@@ -5,6 +5,7 @@ use polar_core::terms::{Symbol, Term};
 use std::any::{Any, TypeId};
 use std::collections::HashMap;
 use std::fmt;
+use std::fmt::Debug;
 use std::sync::Arc;
 
 use crate::{FromPolar, ToPolar};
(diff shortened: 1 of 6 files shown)


languages/rust/oso/src/host/class.rs
```diff
@@ -130,6 +131,31 @@ impl<T> Class<T> {
         self
     }
 
+    pub fn add_result_method<F, Args, R, E>(mut self, name: &str, f: F) -> Self
+    where
+        Args: FromPolar,
+        F: Method<T, Args, Result = Result<R, E>> + 'static,
+        R: ToPolar + 'static,
+        E: Debug + 'static,
```
samscott89, line 139: This would be better if it just relied on `E: std::error::Error`, and then used `err.to_string()` later.

languages/rust/oso/src/host/to_polar.rs
```diff
@@ -143,3 +135,40 @@ impl<C: 'static + Clone + super::HostClass> ToPolar for C {
...
+
+use std::iter;
+
+pub type PolarIter = Box<dyn Iterator<Item = Result<Box<dyn ToPolar>, crate::OsoError>>>;
+
+// Trait for the return value of class methods.
+// This allows us to return polar values, as well as options and results of polar values.
+// Iterators and futures too once we get to them.
+pub trait ToPolarIter {
+    fn to_polar_iter(&self) -> PolarIter;
```
dhatch, line 147: Should this function consume self by value? That would remove the need for clone in some implementations.

dhatch (commented):
Small question.

languages/rust/oso/src/host/to_polar.rs
saolsen: This doesn't work because when I call it on a `dyn ToPolarIter` it cant move out of that because it doesn't know the size.

languages/rust/oso/tests/test_polar.rs
```diff
@@ -386,3 +386,52 @@ fn test_tuple_structs() {
...
+            None
+        }
+    }
+
+    let mut test = OsoTest::new();
+    test.oso
+        .register_class(
+            Foo::get_polar_class_builder()
+                .set_constructor(Foo::new)
+                .add_method("ok", |recv: &Foo| recv.ok())
```
samscott89, line 424: These can all be `Foo::ok` etc.

samscott89 (approved):
One _minor_ suggestion. LGTM.

PR #426, opened 2020-09-15
Add ExternalUnify to Java

gj:
PR checklist:

- [x] Added changelog entry.
- [x] Ensure Java docs are up-to-date.


languages/java/oso/src/test/java/com/osohq/oso/PolarTest.java
```diff
@@ -47,6 +47,18 @@ public static String myStaticMethod() {
     public String myReturnNull() {
       return null;
     }
+
+    @Override
+    public boolean equals(Object obj) {
```
samscott89, line 52: Would this work if it was `equals(MyClass other)` instead?

gj: That would be overloading instead of overriding -- might work here since we only pass other instances of `MyClass` into this `equals()` method, but I'm not sure there's a reason to prefer that over what's currently here.

samscott89: Oh right. I was thinking it would be more normal for someone to implement equals for specific instance instead of the generic equals for any object.

languages/java/oso/src/test/java/com/osohq/oso/PolarTest.java
```diff
@@ -325,6 +376,13 @@ public void testExternalIsSubSpecializer() throws Exception {
         result.equals(List.of(Map.of("x", 2))), "Failed to order rules based on specializers.");
   }
 
+  @Test
+  public void testExternalUnify() throws Exception {
+    assertFalse(p.query("new MyClass(\"foo\", 1) = new MyClass(\"foo\", 1)").results().isEmpty());
+    assertTrue(p.query("new MyClass(\"foo\", 1) = new MyClass(\"foo\", 2)").results().isEmpty());
+    assertTrue(p.query("new MyClass(\"foo\", 1) = new MyClass(\"bar\", 1)").results().isEmpty());
```
leina05, line 383: Could we do a test unifying with a different type?

leina05 (approved):
Asked for one additional test but approving now so you can ship whenever that's in. Nice work :)

PR #427, opened 2020-09-15
[Rust] Implement external unify.

dhatch:
Some error checking and fixes going on here as well.

PR checklist:

- [x] fix todos
- [x] remove random prints and do proper logging
- [x] remove debug constraint on `set_equality_check`. Would be nice if we were able to include Debug somehow for logging if the type implements debug.

### DIFF
--- a/languages/rust/oso/src/host/class.rs
+++ b/languages/rust/oso/src/host/class.rs
@@ -126,15 +126,12 @@ where T: 'static
     pub fn set_equality_check<F>(mut self, f: F) -> Self
     where
         F: Fn(&T, &T) -> bool + 'static,
-        T: std::fmt::Debug
     {
         self.equality_check = Arc::new(move |a, b| {
-            println!("equality check");
+            tracing::trace!("equality check");
 
             let a = downcast(a).map_err(|e| e.user())?;
             let b = downcast(b).map_err(|e| e.user())?;
-
-            println!("{:?} == {:?}", a, b);
 
             Ok((f)(a, b))
         });
@@ -143,7 +140,7 @@ where T: 'static
     }
 
     pub fn with_equality_check(self) -> Self
-    where T: PartialEq<T> + std::fmt::Debug {
+    where T: PartialEq<T> {
         self.set_equality_check(|a, b| PartialEq::eq(a, b))
     }
 
@@ -282,7 +279,7 @@ impl std::fmt::Debug for Instance {
 impl Instance {
     /// Return `true` if the `instance` of self equals the instance of `other`.
     pub fn equals(&self, other: &Self) -> crate::Result<bool> {
-        println!("equals");
+        tracing::trace!("equals");
         // TODO: LOL this &* below is tricky! Have a function to do this, and make instance not
         // pub.
         (self.class.equality_check)(&*self.instance, &*other.instance)
--- a/languages/rust/oso/src/host/class.rs
+++ b/languages/rust/oso/src/host/class.rs
@@ -7,8 +7,10 @@ use std::collections::HashMap;
 use std::fmt;
 use std::sync::Arc;
 
+use crate::errors::{OsoError, OsoResult};
 use crate::FromPolar;
 
+use super::downcast;
 use super::class_method::{ClassMethod, Constructor, InstanceMethod};
 use super::method::{Function, Method};
 use super::to_polar::ToPolarIter;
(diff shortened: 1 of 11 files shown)


languages/rust/oso/src/host/class.rs
```diff
@@ -56,34 +73,38 @@ impl Default for Class {
     }
 }
 
-impl<T> Class<T> {
+// TODO seems like the name is based on fully qualified name, so we may want to
+// require this to be specified.
```
saolsen, line 77: Yeah that's a good idea, it's defaulted for the proc macro so should be fine to error if it's not specified.

languages/rust/oso/src/host/class.rs
```diff
@@ -56,34 +73,38 @@ impl Default for Class {
...
+            equality_check: Arc::from(equality_not_supported(name)),
             ty: std::marker::PhantomData,
             type_id: TypeId::of::<T>(),
         }
     }
 
     pub fn with_default() -> Self
     where
-        T: std::default::Default + 'static,
+        T: std::default::Default,
```
saolsen, line 101: @samscott89 Why was static part of all of these before if we don't need it?

languages/rust/oso/src/host/class.rs
```diff
@@ -94,14 +115,37 @@ impl<T> Class<T> {
...
+            println!("{:?} == {:?}", a, b);
+
+            Ok((f)(a, b))
+        });
+
+        self
+    }
+
+    pub fn with_equality_check(self) -> Self
+    where T: PartialEq<T> + std::fmt::Debug {
```
saolsen, line 145: This is really cool, but why is debug needed?

languages/rust/oso/src/host/mod.rs
```diff
@@ -99,6 +111,11 @@ impl Host {
         id: u64,
     ) -> crate::Result<()> {
         // @TODO: Handle the error if the class doesn't exist.
+        println!("name: {:?}", name);
```
saolsen, line 114: Note to clean all the debug printing up before merging.

languages/rust/oso/tests/test_polar.rs
```diff
@@ -357,3 +370,47 @@ fn test_macros() {
 
     test.qvar_one(r#"new Baz().world() = x"#, "x", "goodbye world".to_string());
 }
+
+// TODO: dhatch see if there is a relevant test to port.
+#[test]
+fn test_unify_externals() {
+    let mut test = OsoTest::new();
+
+    #[derive(PartialEq, Clone, Debug)]
```
saolsen, line 379: I think we should use the derive macro in most cases for tests, since that's likely the main way people will use the library.

polar-core/src/messages.rs
```diff
@@ -36,7 +36,12 @@ impl MessageQueue {
 
     pub fn push(&self, kind: MessageKind, msg: String) {
         let mut messages = self.messages.lock().unwrap();
-        messages.push_back(Message { kind, msg });
+        messages.push_back(Message { kind, msg: msg.clone() });
+
+        // TODO hack to actually get logs.
```
saolsen, line 41: This isn't the worst idea, I've needed it as well.

saolsen (commented):
Looks pretty good to me, just having it work for the same type seems like all we need.

languages/rust/oso/src/host/class.rs
samscott89: It's needed, but moved up to the top:

```
impl<T> Class<T>
where T: 'static
{
```

languages/rust/oso/src/host/class.rs
```diff
@@ -94,14 +115,37 @@ impl<T> Class<T> {
...
         self
     }
 
+    pub fn set_equality_check<F>(mut self, f: F) -> Self
+    where
+        F: Fn(&T, &T) -> bool + 'static,
+        T: std::fmt::Debug
+    {
+        self.equality_check = Arc::new(move |a, b| {
+            println!("equality check");
```
samscott89, line 131: tracing::trace! ?

languages/rust/oso/src/host/class.rs
dhatch: Have a note to follow up on this as a separate PR, it will be easier if we refactor Class to have a builder.

languages/rust/oso/src/host/class.rs
dhatch: Removed, was just temporary.  Separate follow up for actually being able to debug print.

languages/rust/oso/src/host/mod.rs
dhatch: Should be done

languages/rust/oso/tests/test_polar.rs
dhatch: I'm not sure.  I like the explicitness (also in this case I haven't implemented anything to derive the equality function using the proc macro.

dhatch: Also feel the test surface area is smaller if it doesn't use the proc macro and uses this interface directly.

polar-core/src/messages.rs
dhatch: Fixed this in a separate PR #431 .

dhatch:
Ok, this is ready to go.

saolsen (approved):

PR #428, opened 2020-09-16
Rust quickstart example and trait bounds + hacks to make it work.

saolsen:


languages/rust/oso/src/host/class.rs
```diff
@@ -215,3 +215,5 @@ pub struct Instance {
     pub attributes: Arc<InstanceMethods>,
     pub methods: Arc<InstanceMethods>,
 }
+
+unsafe impl Send for Instance {}
```
samscott89, line 219:
```suggestion
// @TODO: This is very unsafe.
// Temporary workaround. We need to differentiate between instances which 
// _do_ need to be `Send` (e.g. registered as constants on the base `Oso` objects)
// and instances which don't need to be Send (e.g. created/accessed on a single thread for
// just one query).
unsafe impl Send for Instance {}
```

languages/rust/oso/src/oso.rs
```diff
@@ -13,7 +12,7 @@ use crate::ToPolar;
 
 #[derive(Clone)]
 pub struct Oso {
-    inner: Rc<polar_core::polar::Polar>,
+    inner: Arc<polar_core::polar::Polar>,
     host: Arc<Mutex<Host>>,
```
samscott89, line 16: Follow up: I don't think we need this to be shared between all queries - typically we clone the host and pass to the query. Would remove most of the contention in exchange for a more expensive clone up front.

dhatch: Have this noted down in my ticket.

dhatch (approved):
This looks good, minus the Send hack.  I will add that to the list on the catch all hard stuff ticket (Mutexes + Semaphores).  

I haven't checked that it is actually sufficient to write the docs, or compared it against any of the other examples.

PR #435, opened 2020-09-17
rust methods that return multiple values

saolsen:


languages/rust/oso/src/host/to_polar.rs
```diff
@@ -172,3 +172,19 @@ impl<C: ToPolarIter> ToPolarIter for Option<C> {
         )
     }
 }
+
+pub struct PolarValues<I, Iter>
+where
+    I: ToPolarIter + 'static,
```
dhatch, line 178: Fun stuff. Should we call `ToPolarIter` something else?

Could revisit this separately.  If we never use it for more than one result, it seems like a trait along the lines of `to_polar_result() -> oso::Result<Option<Value>>` would be all we need?

languages/rust/oso/src/host/class.rs
```diff
@@ -175,6 +175,20 @@ where
         self
     }
 
+    pub fn add_values_method<F, Args, I>(mut self, name: &str, f: F) -> Self
```
dhatch, line 178: `add_iterator_method`? With a docstring indicating the generation of choices behavior.

dhatch (approved):
Some naming questions. We can revisit implementation when we improve memory model.

languages/rust/oso/src/host/to_polar.rs
saolsen: It is used for more than one thing now with the iterator stuff. I changed the names to hopefully be better.

languages/rust/oso/src/host/class.rs
saolsen: :+1:

PR #437, opened 2020-09-18
rust: allow more args

saolsen:


languages/rust/oso/src/host/from_polar.rs
```diff
@@ -213,3 +214,482 @@ where
...
+where
+    A: FromPolar,
+    B: FromPolar,
+    C: FromPolar,
+{
+    fn from_polar(_term: &Term, _host: &mut Host) -> crate::Result<Self> {
+        Err(crate::OsoError::FromPolar)
+    }
+
+    fn from_polar_list(terms: &[Term], host: &mut Host) -> crate::Result<Self> {
```
dhatch, line 227: Remove default impl of this.

languages/rust/oso/src/host/from_polar.rs
```diff
@@ -213,3 +214,482 @@ where
...
     }
 }
+impl<A, B, C> FromPolar for (A, B, C)
+where
+    A: FromPolar,
+    B: FromPolar,
+    C: FromPolar,
+{
+    fn from_polar(_term: &Term, _host: &mut Host) -> crate::Result<Self> {
+        Err(crate::OsoError::FromPolar)
```
dhatch, line 224: Why are these the same trait if the implementation is an error for every term?

languages/rust/oso/src/host/method.rs
```diff
@@ -83,3 +267,345 @@ where
...
+    ) -> Self::Result {
+        (self)(
+            receiver, args.0, args.1, args.2, args.3, args.4, args.5, args.6, args.7, args.8,
+            args.9, args.10, args.11, args.12, args.13, args.14, args.15,
+        )
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
dhatch, line 480: add #[ignore]

languages/rust/oso/src/host/from_polar.rs
saolsen: Yeah probably don't have to be, could do it like ToPolar and to ToPolarResults where one feeds into the other. Don't really wanna mess with it now though.

languages/rust/oso/src/host/from_polar.rs
dhatch: Skip nvm. too hard.

languages/rust/oso/src/host/from_polar.rs
dhatch: On follow up list.

dhatch (approved):
love me some 16 arg functions.

languages/rust/oso/src/host/method.rs
```diff
@@ -83,3 +267,346 @@ where
...
+    // This would be great if it was a proc macro but I don't wanna set up another crate.
```
samscott89, line 481: This seems like a decent solution:
https://docs.rs/impl-trait-for-tuples/0.1.3/impl_trait_for_tuples/

samscott89: Might not work for method though since the trait isn't on the tuple itself, it's the assoc type.

saolsen:
No real time to mess with this today but I'll add it as part of the clean up list.

PR #438, opened 2020-09-18
rust string methods

saolsen:


saolsen:
This could use some factoring I don't have time for right now.

languages/rust/oso/src/builtins.rs
```diff
@@ -29,7 +29,114 @@ fn dictionary() -> Class<HashMap<Symbol, Value>> {
 fn string() -> Class<String> {
     Class::<String>::with_default()
         .name("String")
+        .add_method("len", |s: &String| s.len() as i64)
+        .add_method("is_empty", |s: &String| s.is_empty())
```
samscott89, line 33: These would be better as `String::is_empty`, etc.

languages/rust/oso/src/builtins.rs
```diff
@@ -29,7 +29,114 @@ fn dictionary() -> Class<HashMap<Symbol, Value>> {
 fn string() -> Class<String> {
     Class::<String>::with_default()
         .name("String")
+        .add_method("len", |s: &String| s.len() as i64)
```
samscott89, line 32: Needed because we don't implement `From/ToPolar` for `usize`? We should add those, then this can be `String::len`

languages/rust/oso/src/builtins.rs
samscott89: Oh, but its implemented on `str`.

languages/rust/oso/src/builtins.rs
saolsen: Since we use an `i64` for ints, a `u64` (or `usize`) doesn't actually fit. Seems better to just have these methods return an int instead of *say* we support `usize` when we don't exactly.

languages/rust/oso/src/builtins.rs
saolsen: I think this is actually the only one that can be written that way.

languages/rust/oso/src/builtins.rs
samscott89: Yeah, separate issue for us to fix with polar-core.

samscott89 (approved):
A few more follow ups. But looks good for now.

PR #439, opened 2020-09-18
rust messages

saolsen:


languages/rust/oso/src/macros.rs
```diff
@@ -8,5 +8,15 @@ macro_rules! lazy_error {
 }
 
 macro_rules! check_messages {
-    ($core_obj:expr) => {};
+    ($core_obj:expr) => {
+        while let Some(message) = $core_obj.next_message() {
+            match message.kind {
+                ::polar_core::messages::MessageKind::Print => ::tracing::debug!("{}", &message.msg),
+                ::polar_core::messages::MessageKind::Warning => {
+                    ::tracing::info!("{}", &message.msg)
```
samscott89, line 16: This could probably be `warn!` to match up with it being a warning :)
```suggestion
                    ::tracing::warn!("{}", &message.msg)
```

saolsen:
I still kinda dislike this. It means that if you're just trying out oso like doing the quickstart or something you won't get any prints or warnings. Is there any way to detect if they have the logging library set up and if not use print directly? Some compile flag?

samscott89:
> 
> 
> I still kinda dislike this. It means that if you're just trying out oso like doing the quickstart or something you won't get any prints or warnings. Is there any way to detect if they have the logging library set up and if not use print directly? Some compile flag?

It still feels to me like a pretty bad violation (and something I think we should fix in the other libraries) for an imported library to print stuff to wherever. 

A couple of options:

- If `POLAR_LOG` is set, we init the tracing subscriber + set level to whatever the variable says (I think this could even be compile-time flag)
- There's a method `print_to_stdout()` on oso which enables printing to stdout
- We make sure all the quickstarts include some amount of logging (doesn't need to be a tracing one, can be anything that supports `log`, like `env_logger`)

samscott89:
Also, we can get rid of the use of `log` in polar-core. I don't think it's achieving anything with this fix in place.

dhatch (approved):
lgtm

PR #440, opened 2020-09-18
Support kwargs in method calls

leina05:
PR checklist:

- [x] Added changelog entry.
- [x] Ruby tests
- [x] Rust integration tests
- [x] Check for no kwargs in Java calls
- [x] Check for no kwargs in Javascript calls
- [x] Check for no kwargs in Rust calls


polar-core/src/polar.lalrpop
```diff
@@ -264,7 +256,7 @@ Exp10<T>: Term = {
 }
 
 CallTerm: Value = {
-    <SimpleCall>,
+    <Call>,
```
plotnick, line 259: Nice.

plotnick (approved):
Modulo fixing the tests and the minor suggestions, this looks great.

polar-core/src/rewrites.rs
```diff
@@ -294,60 +294,60 @@ mod tests {
     #[test]
     fn rewrite_nested_literal() {
         let mut kb = KnowledgeBase::new();
-        let mut term = parse_query("new Foo { x: bar.y }");
-        assert_eq!(term.to_polar(), "new Foo{x: bar.y}");
+        let mut term = parse_query("new Foo ( x: bar.y )");
```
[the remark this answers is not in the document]
samscott89: Does it still parse with whitespace? I think it should, right? (agree with the change, just curious)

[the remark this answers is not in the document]
plotnick: Yeah, it's fine with the whitespace.

polar-core/src/vm.rs
```diff
@@ -1168,6 +1178,7 @@ impl PolarVirtualMachine {
         self.log_with(
             || {
                 let mut msg = format!("LOOKUP: {}.{}", instance.to_string(), field_name);
+                // TODO (leina): add kwargs to log
```
gj, line 1181: Seems like a good TODO to get in before shippin'

polar-core/src/polar.lalrpop
```diff
@@ -181,14 +181,6 @@ BuiltinOperator: Operator = {
 };
 
 New: Value = {
-    // Keyword argument constructor.
-    "new" <literal:Spanned<InstanceLiteralTerm>> => {
```
gj, line 185: Can we rm the definition of `InstanceLiteralTerm` on 166?

polar-core/src/polar.lalrpop
```diff
@@ -264,7 +256,7 @@ Exp10<T>: Term = {
 }
 
 CallTerm: Value = {
-    <SimpleCall>,
```
gj, line 267: Looks like there might be some additional cleanup of `SimpleCall` stuff we could do in this file. I'm not sure why `Value` still accepts a `SimpleCall` on 491:
```
pub Value: Value = {
    <Number>,
    <PolarString>,
    <Boolean>,
    <Variable>,
    <DictionaryTerm>,
    <BuiltinOperation>,
    <RewrittenOperation>,
    <SimpleCall>,
    <List<"Term">>,
};
```
or why we couldn't just collapse the declaration of `SimpleCall` into `Call` now that we don't need to refer to `SimpleCall` separately

gj: oh wait is the `SimpleCall` stuff still necessary since we parse Polar predicates (rule heads) the same as method/constructor calls?

gj: but wait that wouldn't make sense either since (I think) the `param: Specializer` fields in a Polar predicate require different semantics than method calls, which require all kwargs to be at the end of the arglist.

polar-core/src/polar.lalrpop
plotnick: Yes.

polar-core/src/polar.lalrpop
plotnick: We can kill `SimpleCall` and fold its cases into `Call`. Thanks for catching this.

polar-core/src/polar.lalrpop
samscott89: Wait... doesn't that mean we can completely get rid of the whole `<"Pattern">`? (Not for this PR, to verify later!). That would be nice.

polar-core/src/polar.lalrpop
samscott89: Word of caution: we do support prolog-style `person("sam") = person(name)` style unification, which is possibly what this syntax is referring to.

languages/ruby/lib/oso/polar/query.rb
```diff
@@ -104,28 +106,17 @@ def handle_call(attribute, call_id:, instance:, args:)
...
-                   else
-                     Hash[kwargs.map { |k, v| [k.to_sym, host.to_ruby(v)] }]
-                   end
-        else
-          raise InvalidConstructorError
-        end
+        raise InvalidConstructorError unless constructor.key? 'Call'
+
+        cls_name = constructor['Call']['name']
+        args = constructor['Call']['args'].map { |arg| host.to_ruby(arg) }
```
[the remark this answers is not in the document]
plotnick: `args` will always be an array here.

languages/java/oso/src/main/java/com/osohq/oso/Query.java
```diff
@@ -109,23 +109,21 @@ private void handleCall(
...
-          JSONArray initargs;
-          if (constructor.has("InstanceLiteral")) {
-            // Keyword initargs are not supported in Java.
-            className = constructor.getJSONObject("InstanceLiteral").getString("tag");
-            throw new Exceptions.InstantiationError(className);
-          } else if (constructor.has("Call")) {
+          if (constructor.has("Call")) {
             className = constructor.getJSONObject("Call").getString("name");
-            initargs = constructor.getJSONObject("Call").getJSONArray("args");
+            JSONArray initargs = constructor.getJSONObject("Call").getJSONArray("args");
```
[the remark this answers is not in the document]
leina05: I don't think `args` can be null, as it's not an `Option`

polar-core/src/polar.lalrpop
leina05: `SimpleCall` is used for Polar predicates... I don't think it can be folded into `Call`

leina05: We need it to be separate because of specializers, as Gabe pointed out.

languages/java/oso/src/main/java/com/osohq/oso/Query.java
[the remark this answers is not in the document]
gj: I think it is an Option, but it won't be null in MakeExternal because the `new` operator always requires parens (and thus args)

polar-core/src/vm.rs
plotnick: Done.

polar-core/src/polar.lalrpop
gj: @plotnick and I talked about this -- we use `ParameterList` to capture the `param: Specializer` case in rule heads. We folded `SimpleCall` into `Call` and everything seems kosher.

languages/js/src/Polar.test.ts
```diff
@@ -149,6 +149,14 @@ Application error: Foo { a: 'A' }.a is not a function at line 1, column 1`
     expect(await qvar(p, 'try(new X(), x)', 'x')).toStrictEqual([]);
   });
 
+  test('rejects keyword arguments in method calls', async () => {
```
gj, line 152:
```suggestion
  test('rejects keyword arguments in method calls', () => {
```

languages/js/src/Polar.test.ts
```diff
@@ -466,6 +474,14 @@ describe('#makeInstance', () => {
     const instance = p.__host().getInstance(1);
     expect(instance).toStrictEqual(new ConstructorArgs(1, 2));
   });
+
+  test('rejects keyword args', async () => {
```
gj, line 478:
```suggestion
  test('rejects keyword args', () => {
```

gj (approved):

gj:
@leina05 to update docs as a follow-up

docs/changelogs/vNEXT.rst
```diff
@@ -7,32 +7,53 @@ NEXT
...
+
+  .. code-block:: polar
+
+    new Foo()
+
+    new Foo{}
 
-- summary of breaking change
+The former accepted positional arguments, and the latter accepted keyword
+arguments. In this release, the two forms have combined their powers, and
```
samscott89, line 25: image

PR #445, opened 2020-09-22
clear() -> clear_rules()

leina05:
PR checklist:

- [x] Added changelog entry.
- [x] All non-python libraries

### DIFF
--- a/languages/python/oso/polar/ffi.py
+++ b/languages/python/oso/polar/ffi.py
@@ -27,6 +27,11 @@ class Polar:
         process_messages(self.next_message)
         check_result(result)
 
+    def clear_rules(self):
+        """Clear all rules from the Polar KB"""
+        result = lib.polar_clear_rules(self.ptr)
+        check_result(result)
+
     def new_query_from_str(self, query_str):
         new_q_ptr = lib.polar_new_query(self.ptr, to_c_str(query_str), 0)
         process_messages(self.next_message)
--- a/languages/python/oso/polar/ffi.py
+++ b/languages/python/oso/polar/ffi.py
@@ -30,6 +30,7 @@ class Polar:
     def clear_rules(self):
         """Clear all rules from the Polar KB"""
         result = lib.polar_clear_rules(self.ptr)
+        process_messages(self.next_message)
         check_result(result)
 
     def new_query_from_str(self, query_str):
--- a/languages/python/oso/tests/parity/test_api.py
+++ b/languages/python/oso/tests/parity/test_api.py
@@ -247,12 +247,12 @@ def test_iter_fields(polar, load_policy, query):
 
 
 @pytest.mark.xfail(EXPECT_XFAIL_PASS, reason="Test relies on internal classes.")
-def test_clear(polar, load_policy, query):
+def test_clear_rules(polar, load_policy, query):
     old = Path(__file__).parent / "policies" / "load.pol"
     fails = Path(__file__).parent / "policies" / "reload_fail.pol"
     new = Path(__file__).parent / "policies" / "reload.pol"
 
-    polar.clear()
+    polar.clear_rules()
     polar.load_file(old)
 
     actor = Actor(name="milton", id=1)
@@ -264,11 +264,11 @@ def test_clear(polar, load_policy, query):
 
     # raises exception because new policy file specifies on a class defined in the old file,
     # but not in the new file
-    polar.clear()
+    polar.clear_rules()
     with pytest.raises(PolarRuntimeError):
         polar.load_file(fails)
 
-    polar.clear()
+    polar.clear_rules()
     polar.load_file(new)
     assert query(Predicate(name="allow", args=[actor, "make", resource]))
     assert not query(Predicate(name="allow", args=[actor, "get", resource]))
--- a/languages/ruby/lib/oso/polar/ffi/polar.rb
+++ b/languages/ruby/lib/oso/polar/ffi/polar.rb
@@ -11,6 +11,7 @@ module Oso
 
           attach_function :new, :polar_new, [], FFI::Polar
           attach_function :load, :polar_load, [FFI::Polar, :string, :string], :int32
+          attach_function :clear_rules, :polar_clear_rules, [FFI::Polar], :int32
           attach_function :next_inline_query, :polar_next_inline_query, [FFI::Polar, :uint32], FFI::Query
           attach_function :new_id, :polar_get_external_id, [FFI::Polar], :uint64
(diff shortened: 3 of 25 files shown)


dhatch:
looks good so far

languages/ruby/lib/oso/polar/ffi/polar.rb
```diff
@@ -39,6 +40,13 @@ def load(src, filename: nil)
           raise FFI::Error.get if loaded.zero?
         end
 
+        # @raise [FFI::Error] if the FFI call returns an error.
+        def clear_rules
+          cleared = Rust.clear_rules(self)
+
```
gj, line 46: Do we need to `process_messages` here?

languages/rust/oso/src/oso.rs
```diff
@@ -64,8 +64,9 @@ impl Oso {
     }
 
     /// Clear out all files and rules that have been loaded.
-    pub fn clear(&mut self) {
-        *self = Self::new();
+    pub fn clear_rules(&mut self) -> crate::Result<()> {
+        self.inner.clear_rules()?;
```
gj, line 68: `check_messages!(self.inner);` ?

polar-core/src/polar.rs
```diff
@@ -180,6 +180,18 @@ impl Polar {
         self.load(src, None)
     }
 
+    /// Clear rules from the knowledge base
+    pub fn clear_rules(&self) -> PolarResult<()> {
+        let mut kb = self.kb.write().unwrap();
+        kb.rules = HashMap::new();
+        kb.sources = Sources::default();
+        kb.inline_queries = vec![];
```
gj, line 188: Why do these three reassign and the following two call `clear()` on the existing data structure? Why not call `clear()` on all five (or reassign all five)?

polar-core/src/polar.rs
```diff
@@ -180,6 +180,18 @@ impl Polar {
         self.load(src, None)
     }
 
+    /// Clear rules from the knowledge base
+    pub fn clear_rules(&self) -> PolarResult<()> {
```
gj, line 184: Does this need to return a result? what's the error case that's getting propagated up? Unless I'm missing something, I only see panics (via the `unwrap()` calls) and success cases.

languages/python/oso/polar/ffi.py
```diff
@@ -27,6 +27,11 @@ def load(self, string, filename=None):
         process_messages(self.next_message)
         check_result(result)
 
+    def clear_rules(self):
+        """Clear all rules from the Polar KB"""
+        result = lib.polar_clear_rules(self.ptr)
```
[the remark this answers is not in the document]
gj: ETA: I'm not saying we need to -- just that it seems like if we need to in one lib it seems reasonable that we should need to in all of them

gj (changes requested):
Excellent work; few questions

languages/python/oso/polar/ffi.py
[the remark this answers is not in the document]
gj: Reading the Rust code again -- are you sure we need to check messages here? I'm not seeing which messages would be queued by `clear_rules()` on the Rust side and in need of checking

polar-core/src/polar.rs
leina05: I changed this so it doesn't return a result. The library functions still return 1 or 0 for failure, as panics are converted to failures.

polar-core/src/polar.rs
leina05: Updated to be consistent. `Sources` still needs to be reassigned to the default.

languages/python/oso/tests/parity/test_api.py
```diff
@@ -247,12 +247,12 @@ def test_iter_fields(polar, load_policy, query):
 
 
 @pytest.mark.xfail(EXPECT_XFAIL_PASS, reason="Test relies on internal classes.")
```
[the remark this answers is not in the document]
leina05: This is a super old test that uses old syntax. I wasn't using it to test this, but I can try to get it working again. The "relies on internal classes" refers to the fact that it uses classes defined in Polar.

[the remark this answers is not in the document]
leina05: I added a new python test to explicitly test `clear_rules()`

polar-wasm-api/src/polar.rs
```diff
@@ -24,6 +24,12 @@ impl Polar {
             .map_err(Error::into)
     }
 
+    #[wasm_bindgen(js_class = Polar, js_name = clearRules)]
+    pub fn wasm_clear_rules(&self) -> JsResult<()> {
+        self.0.clear_rules();
+        Ok(())
+    }
```
gj, line 31:
```suggestion
    pub fn wasm_clear_rules(&self) {
        self.0.clear_rules()
    }
```

leina05: What will happen now if `clear_rules()` panics?

gj: It'll blow up with a pretty stack trace courtesy of https://github.com/rustwasm/console_error_panic_hook

languages/rust/oso/src/oso.rs
```diff
@@ -64,8 +64,9 @@ impl Oso {
     }
 
     /// Clear out all files and rules that have been loaded.
-    pub fn clear(&mut self) {
-        *self = Self::new();
+    pub fn clear_rules(&mut self) -> crate::Result<()> {
+        self.inner.clear_rules();
+        Ok(())
```
gj, line 69:
```suggestion
    pub fn clear_rules(&mut self) {
        self.inner.clear_rules()
```

languages/python/oso/polar/ffi.py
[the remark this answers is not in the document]
dhatch: In general I think we should check messages after most API calls.  We could easily add logging later that would occur during clear_rules, and shouldn't have to go back in the libraries and add a call to remember to check them.

It's not particularly expensive & clear_rules would likely not occur in the critical path, so an extra call doesn't seem like an issue.

gj (approved):

dhatch (commented):
Commented inline - not a huge deal, but I don't agree with the rationale for removing `check_messages`.  It would be best to do that check as frequently as possible so that messages are logged near the time they are generated in the core.  We may add messages that log during clear_rules in the future, and it doesn't hurt to have the extra check there.

languages/python/oso/polar/ffi.py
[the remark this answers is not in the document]
leina05: Yeah that makes sense. Rebasing to add message checks back in...

PR #448, opened 2020-09-22
add app method call syntax docs

leina05:
PR checklist:

- [ ] Added changelog entry.

### DIFF
--- a/docs/using/polar-syntax.rst
+++ b/docs/using/polar-syntax.rst
@@ -279,14 +279,36 @@ Dictionary Key Access
 ^^^^^^^^^^^^^^^^^^^^^
 
 The dot ``.`` operator can be used to access the value associated with
-a key in a dictionary or class instance. For example, the rule::
+a key in a dictionary. For example::
 
-  first_name(dict, x) if
-    dict = new Person() and
-    x = dict.first_name;
+    dict = { hello: "world" } and
+    dict.hello = "world"
 
-will access the value of the field named ``"first_name"`` in ``dict``,
-and unify it with ``x``.
+A string stored in a variable can be used as the key in a dot lookup using the following syntax::
+
+    dict = { hello: "world" } and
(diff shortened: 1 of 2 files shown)


docs/using/polar-syntax.rst
```diff
@@ -279,14 +279,36 @@ Dictionary Key Access
...
+
+Accessing a field on an application instance looks like::
+
+    person = new Person() and
+    first_name = person.first_name
+
+Calling an application method from a policy looks like this::
+
+    person = new Person() and
+    person.set_address("12345 Broadway", city: "New York", state: "NY");
```
samscott89, line 310: One more example for class methods?

docs/using/polar-syntax.rst
```diff
@@ -279,14 +279,36 @@ Dictionary Key Access
...
+
+A string stored in a variable can be used as the key in a dot lookup using the following syntax::
+
+    dict = { hello: "world" } and
+    key = "hello" and
+    dict.(key) = "world"
+
+.. _application-methods:
+
+Application Method Calls
```
gj, line 295: Maybe a more generic title here since this is talking about both field access and method calls?

docs/using/polar-syntax.rst
leina05: Wow we have no docs on constants... we should make a task. I'll add a short bit here but I think it needs its own section.

leina05: The syntax for the actual call isn't different though, so I'm not even sure if it makes sense to put here. The reason I have a different example for `field` is to show the no parens syntax.

samscott89: Sorry to be annoying. But this shows mutating an instance, which is not ideal. Maybe "check_address"?

samscott89 (approved):
LGTM

PR #452, opened 2020-09-24
Uniform ExternalOp errors

gj:
PR checklist:

- [x] Added changelog entry.

### DIFF
--- a/docs/changelogs/vNEXT.rst
+++ b/docs/changelogs/vNEXT.rst
@@ -41,7 +41,5 @@ Link to relevant documentation section
 Other bugs & improvements
 =========================
 
-- bulleted list
-- improvements
-- of smaller
-- potentially with doc links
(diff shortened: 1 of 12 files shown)


dhatch:
Is this based on the Rust `UnimplementedOperation` error?

gj:
> Is this based on the Rust `UnimplementedOperation` error?

Yes. Is that bad?

dhatch:
Not really, just the Rust errors are definitely not finalized, so I wouldn't try to mirror every library off of them yet.  But I think this particular one (UnimplementedOperation) is reasonable to have more broadly.

dhatch:
So if I'm reading this properly, we only support these in Python? Could you make sure we have a follow up in notion for this (not certain we do)?

dhatch (approved):
lgtm

PR #453, opened 2020-09-24
Rust error handling

dhatch:
So far, have gone through and enumerated all things that *could* be errors.  The Rust library is a bit different from other libraries in the sense that there are more issues that could come up that may need an error that are not problems in other libraries.

Our current error handling logic for application errors (that go back into oso & get checked depending on the current operation) allows some operations to raise errors when invalid attributes or methods are accessed, and others to silently fail.  The idea behind this is to prevent policies with rules that cannot match to fail in a prolog-style (backtrack), instead of blowing up execution.

We've made the call that invalid attribute access (`x.attr`) should raise an error & be protected by specializers.  Attribute access during a match `x matches {attr: y}` should not.

Error cases:

- External unify on type that does not support it:

  ```
  x = y
  ```

  Where one of `x` or `y` is an external instance.

  In the Rust library, not all types are comparable (do not implement `PartialEq`), and due to limitations in the class implementation (no generics) we cannot compare mixed types.

  In other libraries, a failed unification does not result in an error.  I could see a case for an error in Rust so that the user knows to call `ClassBuilder::with_equality_check()` to make the equality available to Rust.

- Attribute doesn't exist

  Raise an error in Rust library & other libraries.

- Method does not exist.

  Raise error in rust & other libraries.

- Class method does not exist

  Raise error in Rust & other libraries.

- Wrong argument types

  Raise type error in Rust library.  In other libraries, the language itself would have this error.

- Match with attribute that does not exist.

  Fail with backtrack instead (since this is used to detect if a class has an attribute)

- Match with non-existent class

  Class not found error (not sure what happens for other libraries)

- Wrong argument arity for method and constructor

  Error in Rust library, language would raise in others

- Class does not exist for construction

  Class does not exist error

- Use of keyword arguments

  Not supported, raise unsupported operation.

- Operators

  Unsupported operation

PR checklist:

- [ ] Added changelog entry.

### DIFF
--- a/languages/rust/oso/src/errors.rs
+++ b/languages/rust/oso/src/errors.rs
@@ -1,5 +1,7 @@
 use thiserror::Error;
 
+// TODO stack traces????
+
 /// oso errors
 ///
 /// TODO: fill in other variants
@@ -33,6 +35,9 @@ pub enum OsoError {
     #[error("{operation} are unimplemented in the oso Rust library")]
     UnimplementedOperation { operation: String },
 
+    #[error(transparent)]
+    InvalidCallError(#[from] InvalidCallError),
+
     #[error("failed to convert type to Polar")]
     ToPolar,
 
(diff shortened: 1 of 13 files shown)


dhatch:
Tests for errors & the error cases listed out above. (No need to look at the code).  Everything is relatively straightforward, besides external unification. I’m not sure whether failed unifications in the Rust library should cause an error or fail with a backtrack like in the other libraries.

In other libraries, if an external unification fails we backtrack.  This seems fine IMO since there is nothing special needed to get that unification to succeed if the equality operation is implemented on the type (in python we just compare).

But, in Rust the user must call `ClassBuilder::with_equality_check()` or probably add something to the proc macro (when we implement this).  So, it could be easier to have a type you expect to compare equal, but forget this call.

At the same time, having types that can’t be compared throw an exception if unification is attempted could break some behavior…. thinking of a policy like:

```
allow(actor, action, new PostResource())
```

for example.  

If you called `allow` with resource as an instance of another type, this unification would fail.  If it caused an exception, it would prevent rules that should actually match from matching, and terminate the query.

samscott89:
Thanks for writing that up, Dave.

> On the other hand, having types that can’t be compared throw an exception if unification is attempted could break some behavior…. thinking of a policy like:

The example given kinda feels like a win to have it be a hard error. The workaround would be:

```
allow(actor, action, res: PostResource) if res = PostResource;
```

or whatever. I think "Hey! Your policy wanted me to compare a `PostResource` with a `CompletelyDifferentResource` and I don't know how to do that!" would be fine.

dhatch:
Implemented above, now unification is passing.  Turns out we only use the `polar_core::Polar::application_error` function for "InvalidCallErrors" (a type in the Ruby, Java & JS libraries).  InvalidCallError means a method or attribute doesn't exist (I think in Ruby it also means the wrong number of arguments).

ApplicationError sounds like it should mean any error that occurs during execution of an application method, but in reality those errors are just raised directly in other languages (& it looks like maybe TypeErrors and other errors sporadically depending on the language).

I think we need to come back and revisit this.  Even if we keep the mechanism of passing certain errors back into Polar,
we should probably have specific types for those errors so it's clear which ones need to be handled that way.

saolsen (approved):

PR #454, opened 2020-09-25
Rust value type

saolsen:
At first I was going to have everything go through PolarValue but that doesn't work because it doesn't represent things like instances. That means it's really just the type that is used for passing things in from rust and getting back results. I added another trait that is FromPolarValue to handle conversions between it and rust types.

With this you should be able to get back mixes of PolarValue and types too, which is cool.
`results.get_typed::<HashMap<String, Vec<PolarValue>>("foo");`


languages/rust/oso/src/host/value.rs
```diff
@@ -0,0 +1,163 @@
+use polar_core::terms::*;
+use std::collections::hash_map::HashMap;
+use std::convert::TryFrom;
+
+// Should we call it something else?
+#[derive(Clone, Debug, PartialEq)]
+pub enum PolarValue {
```
dhatch, line 7: What about Predicate?

languages/rust/oso/src/host/value.rs
```diff
@@ -0,0 +1,163 @@
...
+}
+
+macro_rules! polar_to_int {
+    ($i:ty) => {
+        impl FromPolarValue for $i {
+            fn from_polar_value(val: PolarValue) -> crate::Result<Self> {
+                if let PolarValue::Integer(i) = val {
+                    <$i>::try_from(i).map_err(|_| crate::OsoError::FromPolar)
+                } else {
+                    Err(crate::OsoError::FromPolar)
```
dhatch, line 91: I think these should raise type errors instead which can be more descriptive (i.e. expected integer, got ... whatever it is).

languages/rust/oso/src/host/to_polar.rs
```diff
@@ -166,6 +167,12 @@ impl<T: ToPolar> ToPolar for HashMap<String, T> {
     }
 }
 
+impl ToPolar for PolarValue {
+    fn to_polar_value(self, _host: &mut Host) -> Value {
+        self.to_term().value().clone()
```
dhatch, line 172: Should `to_term()` instead be `to_value()`? It looks like this is the only usage & we are dropping the Term part.

languages/rust/oso/src/host/from_polar.rs
```diff
@@ -126,7 +127,7 @@ impl<T: FromPolar> FromPolar for Vec<T> {
 
 impl<T: FromPolar> FromPolar for HashMap<String, T> {
```
dhatch, line 128: Do avoid redundancy, could we instead have this implementation be something like:

```
impl <T: FromPolarValue> FromPolar for T {
    fn from_polar(term: &Term, _: &Host) -> {
        T::from_polar_value(PolarValue::from_term(term))
    }
}
```

languages/rust/oso/tests/common/mod.rs
```diff
@@ -74,7 +74,7 @@ impl OsoTest {
 
     pub fn qvar_one<T>(&mut self, q: &str, var: &str, expected: T)
     where
-        T: oso::FromPolar + PartialEq<T> + std::fmt::Debug,
```
dhatch, line 77: Why did this change to `FromPolarValue` instead of `FromPolar`?

languages/rust/oso/src/host/value.rs
```diff
@@ -0,0 +1,163 @@
...
+                }
+                Value::List(list)
+            }
+            PolarValue::Variable(s) => Value::Variable(Symbol(s.clone())),
+        };
+        Term::new_from_ffi(value)
+    }
+}
+
+pub trait FromPolarValue: Clone + Sized + 'static {
```
dhatch, line 74: Should we implement `Into<T>` as well for all the types that support `FromPolarValue`?

languages/rust/oso/src/query.rs
```diff
@@ -265,18 +265,16 @@ impl ResultSet {
         self.bindings.is_empty()
     }
 
-    pub fn get(&self, name: &str) -> Option<crate::Value> {
+    pub fn get(&self, name: &str) -> Option<crate::PolarValue> {
```
dhatch, line 268: If this only returns PolarValue now, how would an external instance be returned?

dhatch (commented):
Left some comments... on one hand, I like that `FromPolarValue` stands alone and doesn't require a host class.  On the other... it seems like it prevents us from returning instances from `ResultSet`.

languages/rust/oso/src/host/value.rs
```diff
@@ -0,0 +1,163 @@
...
+            }
+            Value::List(l) => {
+                let mut list = vec![];
+                for t in l {
+                    list.push(PolarValue::from_term(t)?);
+                }
+                PolarValue::List(list)
+            }
+            Value::Variable(Symbol(sym)) => PolarValue::Variable(sym.clone()),
+            _ => return Err(crate::OsoError::FromPolar),
```
dhatch, line 41: This should be a type error as well.

languages/rust/oso/tests/common/mod.rs
saolsen: So I could write the tests with PolarValue instead of Value

languages/rust/oso/src/host/value.rs
saolsen: punt

languages/rust/oso/src/host/to_polar.rs
saolsen: There's recursive calls that would have to be wrapped in term constructors, only the top level is unwrapped like this.

languages/rust/oso/src/host/from_polar.rs
saolsen: ah yeah that cleans it up a ton

dhatch (approved):
lgtm

PR #455, opened 2020-09-28
Check native Polar types match against fields

samscott89:
PR checklist:

- [x] Added changelog entry.


languages/java/oso/src/main/java/com/osohq/oso/Query.java
```diff
@@ -71,6 +71,10 @@ private void handleCall(
       ffiQuery.applicationError(e.getMessage());
       ffiQuery.callResult(callId, null);
       return;
+    } catch (Exceptions.InvalidAttributeError e) {
```
samscott89, line 74: This was causing uncaught runtime exceptions when specializing on non-existing fields.

E.g. `x matches { non_existent: 1 }` 

We rely on the `check_errors` flag to determine whether we want this to be a hard failure.

Incidentally... maybe we _do_ want this to be a hard error if specializing on a type. E.g. `x matches Foo { non_existent: 1}` should fail if `non_existent` doesn't exist. But not for dicts?

languages/rust/oso/src/host/class.rs
```diff
@@ -301,7 +301,7 @@ impl Instance {
     pub fn class<'a>(&self, host: &'a Host) -> crate::Result<&'a Class> {
         host.get_class_by_type_id(self.inner.as_ref().type_id())
             .map_err(|_| OsoError::MissingClassError {
-                name: self.name(&host).to_owned(),
+                name: self.debug_type_name.to_string(),
```
samscott89, line 304: Was overflowing stack prev

polar-core/src/vm.rs
```diff
@@ -905,57 +905,31 @@ impl PolarVirtualMachine {
...
-
-                // For each field on the right, isa its value against the corresponding value on
-                // the left.
-                for (k, v) in right.fields.iter() {
-                    let left = left
-                        .fields
-                        .get(&k)
-                        .expect("left fields should be a superset of right fields")
-                        .clone();
+            (Value::Variable(symbol), _) => {
```
samscott89, line 908: Reordering to make sure we deref first

polar-core/src/vm.rs
```diff
@@ -905,57 +905,31 @@ impl PolarVirtualMachine {
...
+                        right: right.clone(),
+                    })?;
                 }
             }
 
-            (Value::InstanceLiteral(_), _) => {
-                panic!("How did an instance literal get here???");
-            }
-
-            (Value::ExternalInstance(_), Value::Pattern(Pattern::Dictionary(right))) => {
```
samscott89, line 942: This case was generalised to `(_, Value::Pattern(...))`, since we want it to do the same for _any_ type (except dictionaries which are handled first)

polar-core/src/vm.rs
```diff
@@ -1007,41 +981,64 @@ impl PolarVirtualMachine {
...
+                    };
+                    self.append_goals(vec![lookup, isa])?;
                 }
             }
 
-            (
-                Value::ExternalInstance(left_instance),
-                Value::Pattern(Pattern::Instance(right_literal)),
-            ) => {
+            (_, Value::Pattern(Pattern::Instance(right_literal))) => {
```
samscott89, line 1038: Same idea: if the RHS is an instance literal, we're always going to check types and recurse, so combine these two expressions.

languages/java/oso/src/main/java/com/osohq/oso/Query.java
gj: Maybe it would make sense to match the semantics of the host language to determine what type of error it should be? E.g.,

```js
> class Foo {}
undefined
> new Foo().foo
undefined
```

vs.

```rb
[1] pry(main)> class Foo; end
=> nil
[2] pry(main)> Foo.new.bar
NoMethodError: undefined method `bar' for #<Foo:0x007ffa38031300>
from (pry):2:in `<main>'
```

languages/ruby/spec/oso/language_parity_spec.rb
```diff
@@ -101,6 +101,12 @@ def self.sum(args)
 oso.query('builtinSpecializers("foo", "String")').next
 raise unless oso.query('builtinSpecializers("bar", "String")').to_a.empty?
 
+oso.query('builtinSpecializers(1, "IntegerWithFields")')
```
gj, line 104:
```suggestion
oso.query('builtinSpecializers(1, "IntegerWithFields")').next
```

languages/ruby/spec/oso/language_parity_spec.rb
```diff
@@ -101,6 +101,12 @@ def self.sum(args)
...
+raise unless oso.query('builtinSpecializers(2, "IntegerWithGarbageFields")').to_a.empty?
+raise unless oso.query('builtinSpecializers({}, "DictionaryWithFields")').to_a.empty?
+raise unless oso.query('builtinSpecializers({z: 1}, "DictionaryWithFields")').to_a.empty?
+oso.query('builtinSpecializers({y: 1}, "DictionaryWithFields")')
```
gj, line 108:
```suggestion
oso.query('builtinSpecializers({y: 1}, "DictionaryWithFields")').next
```

test/test.polar
```diff
@@ -15,6 +15,12 @@ builtinSpecializers(x: Float, "Float") if x > 0.0;
 builtinSpecializers(x: List, "List") if x = ["foo", *_rest];
 builtinSpecializers(x: Dictionary, "Dictionary") if x.foo = "foo";
 builtinSpecializers(x: String, "String") if x = "foo";
+# This should not match
```
gj, line 18: Maybe it should for JS, re: previous review comment about JS's permissiveness in returning `undefined` for nonexistent properties.

gj (changes requested):
JS parity test?

test/test.polar
dhatch: It also checks that the value is 2.  Why would that match?

gj: true

languages/java/oso/src/main/java/com/osohq/oso/Query.java
samscott89: I think probably JS is the only one that won't care about accessing an attribute that doesn't exist?

samscott89:
> 
> 
> JS parity test?

Thanks, good catch.

gj (approved):
Pushed three commits doing some further reorganization of the `match` arms in `isa()` + adding a changelog.

Everything LGTM (and @plotnick)

PR #462, opened 2020-10-08
Remove custom constructors

gj:
Custom constructors are a relic. They were useful for translating kwargs -> positional args before we supported positional args when constructing an instance via Polar's `new` keyword. They were also useful for specifying a find_or_create-style class method as a constructor, but that's been superseded by the introduction of calling methods directly on registered constants, including classes.

PR checklist:

- [x] Ruby
- [x] Java
- [x] Python
- [x] Added changelog entry.


samscott89 (commented):
This is really good, but I think it would be worth doing this for Python too and removing from the decorator.

plotnick (approved):
This is strictly an improvement, now that we now have features that make custom constructors unnecessary. No specific comments, since most of it is deletions, but thanks for sprucing up the docs & comments along the way.

plotnick:
> ... I think it would be worth doing this for Python too ...

I'd forgotten they were still there. Is there any reason we need them anywhere anymore?

samscott89:
> 
> 
> > ... I think it would be worth doing this for Python too ...
> 
> I'd forgotten they were still there. Is there any reason we need them anywhere anymore?

Discussed with @gj - I don't think so. They are being used (via the `@polar_class` decorator) in some examples that would be substantially better off using class methods.

plotnick (approved):
Woot. Thanks.

PR #463, opened 2020-10-08
Partial Eval

dhatch:
Parts to review:

- [x] Constraints & VM (Leina)
- [x] Runnable (Gabe)
- [x] Python lib & Django query conversion (Steve)

...

- simplifier


languages/python/django-oso/django_oso/auth.py
```diff
@@ -33,6 +38,65 @@ def authorize(request, resource, *, actor=None, action=None):
...
+    for result in results:
+        resource_partial = result["bindings"]["resource"]
+        if filter is None:
+            filter = Q()
+
+        next_filter = partial_to_query_filter(resource_partial, resource_type)
+        if next_filter == Q():
+            return next_filter
+
+        filter |= next_filter
```
saolsen, line 92: This is super cool if I understand it correctly.
If you had two different rules about why you could access something, those both become filters that are then or'd together in django?
Like if there's one rule that says you can see a widget if it's color is blue and another rule that says you can see a widget if it's color is green, you end up with a sql query that's like `select * from widgets where color = green or color = blue`?
Is that right?

languages/python/django-oso/django_oso/partial.py
```diff
@@ -0,0 +1,99 @@
...
+                    Dot, [
+                        Variable('_this'),
+                        'is_private'])])])
+
+    Output::
+
+        Q(is_private=False)
+    """
+
+    q = and_expr(partial, type_name)
```
saolsen, line 33: Is this always an and?

languages/python/django-oso/django_oso/partial.py
```diff
@@ -0,0 +1,99 @@
...
+        elif expression.operator == "And":
+            q = q & and_expr(expression, type_name)
+        elif expression.operator == "Isa":
+            try:
+                assert expression.args[1].tag == type_name
+            except (AssertionError, IndexError, AttributeError, TypeError):
+                raise UnsupportedError(
+                    f"Unimplemented partial isa operation {expression}."
+                )
+        else:
```
saolsen, line 65: What about `not` and `or`. Those could make a lot of sense for filtering.

languages/python/oso/tests/test_polar.py
```diff
@@ -752,3 +762,79 @@ def kwarg_method(self, x=1, y=2):
...
+def unwrap_and(x):
+    assert isinstance(x, Expression)
+    assert x.operator == "And"
+    if len(x.args) == 1:
+        return x.args[0]
+    else:
+        return x.args
+
+
+def test_partial(polar):
```
saolsen, line 776: Can you have more than one partial in a query? Can those partials reference each other?
```python
f(a,b) if a.x = b.y
```
```python
polar.query_rule("f", (Partial("A"), Partial("B"))
```

polar-core/src/terms.rs
```diff
@@ -207,10 +212,11 @@ pub enum Value {
     Variable(Symbol),
     RestVariable(Symbol),
     Expression(Operation),
+    Partial(Constraints),
```
saolsen, line 215: Do we want to add a way to create these from polar? (other than I guess hacking it and registering the python Partial type as a class)

saolsen (commented):
Dope stuff. Django filtering implementation seems pretty limited in that it only supports a few operations. Probably want to flesh that out before we announce "the Django integration supports partial eval". Really cool proof of concept though. I'm excited to try it out on some other things.

languages/python/oso/tests/test_polar.py
saolsen: Super extra bonus points for somehow coming up with a join in the django integration.

polar-core/src/terms.rs
plotnick: I want to add the syntax `∂x` to mean `Partial(Variable(x))`, but maybe that's going to far.

polar-core/src/vm.rs
```diff
@@ -2467,6 +2489,139 @@ impl PolarVirtualMachine {
...
+            right: Term::new_temporary(Value::Boolean(true)),
+        })?;
+
+        Ok(QueryEvent::Run { runnable, call_id })
+    }
+}
+
+impl Runnable for PolarVirtualMachine {
+    /// Run the virtual machine. While there are goals on the stack,
+    /// pop them off and execute them one at at time until we have a
```
gj, line 2508:
```suggestion
    /// pop them off and execute them one at a time until we have a
```

gj (approved):
Runnable LGTM

Went over it synchronously with @plotnick last week and added some comments to the code.

languages/python/django-oso/django_oso/auth.py
plotnick: Yup, that is exactly right.

languages/python/django-oso/django_oso/partial.py
plotnick: Yes, right now. Partial constraints are interpreted conjunctively, so the natural interpretation is as an `And` expression. But when the simplifier improves we should be able to, e.g., simplify a one-argument `And` to just the arg.

languages/python/django-oso/django_oso/partial.py
plotnick: Agreed. The semantics of `not` could be difficult, but I agree is worth doing. `or` should be easy, it just hasn't been done yet.

languages/python/oso/polar/host.py
```diff
@@ -204,5 +214,20 @@ def to_python(self, value):
             )
         elif tag == "Variable":
             return Variable(value[tag])
+        elif tag == "Expression":
```
samscott89, line 217: Is there a way we can make sure these don't get handled for regular values?

languages/python/oso/tests/test_polar.py
plotnick: You can have multiple partials, and they can reference each other, but it doesn't currently work on the way out to Python. Agree that a join is the right interpretation.

languages/python/oso/polar/host.py
samscott89: What happens if you do:
```
❯ python -m oso
query> x = (1 and 2)
UnexpectedPolarTypeError
Expression
```

languages/python/django-oso/django_oso/partial.py
dhatch: `or` actually isn't needed because the results come out as two separate bindings since it generates choice points.

languages/python/oso/polar/host.py
plotnick: Not exactly sure what you mean here, but `x = (1 and 2)` isn't a well-formed query. The fact that it gets an error in Python but not, apparently, in some of the other languages (Ruby, JS) is (I think) a non-partial-eval-related bug.

polar-core/src/partial/constraints.rs
```diff
@@ -0,0 +1,501 @@
...
+        if let Value::Pattern(Pattern::Instance(instance)) = right.value() {
+            let call_id = counter.next();
+            self.last_call_id = call_id;
+
+            // is_subclass check of instance tag against proposed
+            return Some(QueryEvent::ExternalIsSubclass {
+                call_id,
+                left_class_tag: self.proposed_tag.clone().unwrap(),
+                right_class_tag: instance.tag.clone(),
+            });
```
leina05, line 169: Had a larger offline convo with @dhatch about this. tl;dr is that it's overly limiting to assume the only way two `isa` constraints are compatible is if the newer constraint class is a subclass of the older one.
Two examples that conflict:
1. The newer class is a superclass of the older one. We would still want this to match. Thinking of two rules that specialize on a superclass and a subclass. Even if the instance matches the subclass rule first, we would still want it to match the superclass rule (unless there is a `cut`). 
2. Multiple inheritance. E.g., in Python two unrelated classes could both be a superclass of a third class. Comparing an `isa` for these two unrelated classes would say they aren't compatible, even though the third class would match both.

polar-core/src/partial/constraints.rs
```diff
@@ -0,0 +1,501 @@
...
+
+        None
+    }
+}
+
+impl Runnable for IsaConstraintCheck {
+    fn run(&mut self, counter: Counter) -> PolarResult<QueryEvent> {
+        if self.proposed_tag.is_none() {
+            return Ok(QueryEvent::Done { result: true });
+        }
```
leina05, line 182: When would this happen?

languages/python/oso/polar/host.py
samscott89: Oh that's weird.

Yeah, the example is contrived, but just worried that by now permitting expressions to be converted to Python we might lose an error case.

polar-core/src/partial/constraints.rs
```diff
@@ -0,0 +1,501 @@
...
+
+        if let Some(result) = self.result.take() {
+            if !result {
+                return Ok(QueryEvent::Done { result: false });
```
leina05, line 186: What does it mean for the result to be `false` before the `run()` has executed?

polar-core/src/partial/simplify.rs
```diff
@@ -0,0 +1,217 @@
+use std::collections::HashSet;
```
leina05, line 1: Not going to do a thorough review of this in the interest of time and because it's being taken as a follow-up.

leina05 (commented):
Leaving this for now because I need to take a short break. 
Still need to look at:

- [x] polar.rs
- [ ] terms.rs
- [ ] vm.rs

polar-core/src/vm.rs
```diff
@@ -1976,6 +1941,29 @@ impl PolarVirtualMachine {
             right_value = self.value(right_sym).cloned();
         }
 
+        // Rebind the previous name of `term` to a variable pointing to
+        // `var`. `term` must be a partial.
+        //
+        // Return: A new partial with the name updated to `var`.
+        let mut rebind_partial = |var: &Symbol, term: &Term| {
```
leina05, line 1948: Is there a reason this has to be a closure vs. defining a method on `Partial`?

languages/rust/oso/src/query.rs
```diff
@@ -106,8 +109,8 @@ impl Query {
         Ok(self.inner.call_result(call_id, None)?)
     }
 
-    fn application_error(&mut self, error: crate::OsoError) {
-        self.inner.application_error(error.to_string())
+    fn application_error(&mut self, error: crate::OsoError) -> crate::Result<()> {
+        Ok(self.inner.application_error(error.to_string())?)
```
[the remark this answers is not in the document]
plotnick: It's related cleanup, I think. When we factored out the `Runnable` trait, we added results to the methods that were missing them (`external_question_result` and something else, maybe). I think this is just threading those through.

polar-core/benches/bench.rs
```diff
@@ -320,7 +322,7 @@ impl Runner {
     }
 
     fn handle_external_isa(&mut self, call_id: u64) {
-        self.query.question_result(call_id, true)
+        self.query.question_result(call_id, true).unwrap();
```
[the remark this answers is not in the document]
plotnick: Yes, in this context. `question_result` didn't previously return a result, but now it does. If it were to fail for some reason, we'd want to know about it but the benchmark wouldn't care why.

polar-core/src/vm.rs
dhatch: Yes, it calls methods on the vm.

polar-core/src/counter.rs
```diff
@@ -0,0 +1,51 @@
+use std::sync::atomic::{AtomicU64, Ordering};
```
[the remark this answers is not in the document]
plotnick: No, this is factoring the counter aspect of the KB into its own first-class thing so that we can use them in runnables without passing around the whole KB. The behavior of the counter is the same as before (e.g., wrapping at 2**53).

polar-core/src/partial/constraints.rs
dhatch: `self.result` would only be false after an external is subclass check.  This means that the proposed new constraint was not a subclass of an existing constraint.

polar-core/src/partial/constraints.rs
dhatch: If the proposed constraint is not of the form `x matches Type{fields ?}`.  Instead it could be `x matches { }`, which we don't constrain right now.

polar-core/src/partial/constraints.rs
dhatch: Thanks, will handle as follow up.

polar-core/src/events.rs
```diff
@@ -1,16 +1,30 @@
...
+    Done {
+        result: bool,
+    },
+
+    /// Run the runnable and place its result in `call_id` when it
+    /// completes.
+    #[serde(skip)]
+    Run {
+        runnable: Box<dyn Runnable>,
+        call_id: u64,
```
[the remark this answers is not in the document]
plotnick: The `call_id` is used to identify the runnable to its parent; the result is passed back to the parent for this `call_id`. The result is a `bool`, and can't go directly in here.

polar-core/src/partial/constraints.rs
```diff
@@ -0,0 +1,501 @@
...
+    fn check_constraint(
+        &mut self,
+        mut constraint: Operation,
+        counter: &Counter,
+    ) -> Option<QueryEvent> {
+        if constraint.operator != Operator::Isa {
+            return None;
+        }
+
+        let right = constraint.args.pop().unwrap();
```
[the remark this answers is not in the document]
dhatch: Yeah - it's everywhere in the codebase.  We have talked about a new Operation enum that would avoid the need for this.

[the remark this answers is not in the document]
dhatch: Would like to handle this separately.

polar-core/src/partial/constraints.rs
```diff
@@ -0,0 +1,501 @@
...
+    /// Returns: None if compatible, QueryEvent::Done { false } if incompatible,
+    /// or QueryEvent to ask for compatibility.
+    fn check_constraint(
+        &mut self,
+        mut constraint: Operation,
+        counter: &Counter,
+    ) -> Option<QueryEvent> {
+        if constraint.operator != Operator::Isa {
+            return None;
+        }
```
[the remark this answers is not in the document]
dhatch: Yes, it could.  But, these constraints are approximate.  It's okay if conflicting constraints are pushed.  It is incorrect if constraints that don't conflict are marked as conflicting.

polar-core/src/partial/constraints.rs
```diff
@@ -0,0 +1,501 @@
...
+        };
+
+        self.operations.push(op);
+    }
+
+    /// Add lookup of `field` assigned to `value` on `self.
+    ///
+    /// Returns: A partial expression for `value`.
+    pub fn lookup(&mut self, field: Term, value: Term) -> Term {
+        // Note this is a 2-arg lookup (Dot) not 3-arg. (Pre rewrite).
```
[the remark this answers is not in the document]
plotnick: Right.

polar-core/src/partial/constraints.rs
```diff
@@ -0,0 +1,501 @@
...
+
+        Self {
+            existing,
+            proposed_tag,
+            result: None,
+            last_call_id: 0,
+        }
+    }
+
+    /// Check if constraint is compatible with proposed.
```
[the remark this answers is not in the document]
dhatch: yes

polar-core/src/partial/constraints.rs
```diff
@@ -0,0 +1,501 @@
...
+        new.operations = operations;
+        new
+    }
+
+    pub fn name(&self) -> &Symbol {
+        &self.variable
+    }
+
+    fn variable_term(&self) -> Term {
+        Term::new_temporary(Value::Variable(sym!("_this")))
```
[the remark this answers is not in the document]
dhatch: Why?

polar-core/src/partial/constraints.rs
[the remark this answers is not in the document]
dhatch: Not that there is no support, just that we represent lookups as 2-arg lookup, not the post-rewrite 3-arg with an output variable.

polar-core/src/partial/constraints.rs
plotnick: If you try to match a dictionary, not an instance; e.g., `x matches {...}`.

polar-core/src/partial/constraints.rs
```diff
@@ -0,0 +1,501 @@
...
+        assert!(matches!(field.value(), Value::String(_)));
+
+        self.operations.push(op!(
+            Unify,
+            value.clone(),
+            term!(op!(Dot, self.variable_term(), field))
+        ));
+
+        let name = value.value().as_symbol().unwrap();
+        Term::new_temporary(Value::Partial(Constraints::new(name.clone())))
```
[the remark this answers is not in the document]
dhatch: This assumes that `value` is a variable, which I believe is guaranteed due to how the 2 => 3 arg rewrite works.

In the example you brought up, the unify would be applied to the temporary variable.  Recall that `partial1.field = value1` gets rewritten into `partial1.field = _temp_1 and _temp_1 = value`.

The returned partial is bound to `_temp_1` in the VM, so that when `_temp_1 = value` is evaluated, the constraint is added there.

Does this make sense to you?

polar-core/src/partial/constraints.rs
plotnick: That a child runnable returned `false`, in which case we should, too. But this probably will not happen before `run` has executed; it's re-entered after each event.

languages/python/django-oso/django_oso/auth.py
saolsen: 🎆

polar-core/src/partial/constraints.rs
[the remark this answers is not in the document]
plotnick: Yes, there is no consistent or good interface for dealing with these args. We've talked about reworking the expression representations entirely, but some interface clean-ups could be a nice intermediate step.

saolsen (approved):

gj (changes requested):
- [ ] changelog entry

polar-core/src/events.rs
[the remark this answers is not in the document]
leina05: Okay, I was confused by the comment. After reading more I understand that the result goes in a `Done` event.

polar-core/src/partial/constraints.rs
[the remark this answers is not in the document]
leina05: Eh never mind.

polar-core/src/partial/constraints.rs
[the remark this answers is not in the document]
leina05: Oh yeah, I forgot about rewrites.

polar-core/src/events.rs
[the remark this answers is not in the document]
plotnick: Yes, that is quite unclear.

PR #466, opened 2020-10-12
multithreading test

saolsen:
Adds our multithreaded bug reproducer script as a test. This along with the tests we added when doing the rust integration should give us some confidence everything works in a multithreaded scenario.


gj (commented):
Do you think it's worth adding a similar test for each language lib?

saolsen:
> Do you think it's worth adding a similar test for each language lib?

Would be nice. I don't really have time to do that today but also feel ok because java and ruby call the code similarly to python, rust has it's own static assertions of multithreaded safety (which came up and were fixed during the rust work) and wasm is single threaded.

dhatch (approved):

PR #469, opened 2020-10-12
debugger

saolsen:


saolsen:
```
debug> stack
5: a(1)
  in query at line 1, column 1
4: debug() and b(x) and c()
  in rule a at line 1, column 9 in file test.polar
3: b(x)
  in rule a at line 1, column 21 in file test.polar
2: (y = 1 and x = y) and y = 1
  in rule b at line 2, column 9 in file test.polar
1: y = 1 and x = y
  in rule b at line 2, column 10 in file test.polar
0: x = y
  in rule b at line 2, column 20 in file test.polar

debug> query 0
QUERY: _x_5 = _y_6, BINDINGS: {_x_5 = 1, _y_6 = 1}
debug> query 1
QUERY: _y_6 = 1 and _x_5 = _y_6, BINDINGS: {_x_5 = 1, _y_6 = 1}
debug> query 2
QUERY: _y_6 = 1 and _x_5 = _y_6 and _y_6 = 1, BINDINGS: {_x_5 = 1, _y_6 = 1}
debug> query 3
QUERY: b(_x_2), BINDINGS: {_x_2 = 1}
debug> query 4
QUERY: debug() and b(_x_2) and c(), BINDINGS: {_x_2 = 1}
debug> query 5
QUERY: a(1), BINDINGS: {}
debug> query 6
Error: level is out of range
debug> 
```
You can check the query for each level of the stack.
The stack is also used for over/out stepping so you can think of an over step as going to the next sibling in 1 and an out step as going to the next sibling in 2.

polar-core/src/debugger.rs
```diff
@@ -35,89 +48,14 @@ impl PolarVirtualMachine {
...
-        snapshot: Queries,
-    },
+    /// Step **over** the current query. Will break on the next query where the trace stack is at the level
+    /// as the current one.
+    Over { level: usize },
+    /// Step **out** of the current query. Will break on the next query where the trace stack is at a lower
+    /// level than the current one.
+    Out { level: usize },
+    /// Step **in**. Will break on the next query.
+    InTo,
```
gj, line 58: nit: Can we capitalize this variant as `Into`?

polar-core/src/debugger.rs
```diff
@@ -35,89 +48,14 @@ impl PolarVirtualMachine {
...
-    /// - Store a slice of the current query stack *without the last three queries* (current query,
-    ///   body of current rule, and head of current rule).
-    /// - Evaluate goals until the current query stack *without the current query* matches the
-    ///   stored slice.
-    Out {
-        /// Snapshot of the current query stack sans its last three queries.
-        snapshot: Queries,
-    },
+    /// Step **over** the current query. Will break on the next query where the trace stack is at the level
+    /// as the current one.
```
gj, line 52:
```suggestion
    /// Step **over** the current query. Will break on the next query where the trace stack is at the same
    /// level as the current one.
```

polar-core/src/debugger.rs
saolsen: I did this because vscode was hi-lighting it differently because Into is a builtin rust thing.

polar-core/src/formatting.rs
```diff
@@ -60,6 +60,7 @@ impl Trace {
 /// Traverse a [`Source`](../types/struct.Source.html) line by line until `offset` is reached,
 /// and return the source line containing the `offset` character as well as `num_lines` lines
 /// above and below it.
+// @TODO: Can we have the carrot under the whole range of the expression instead of just the beginning.
```
samscott89, line 63:
```suggestion
// @TODO: Can we have the caret under the whole range of the expression instead of just the beginning.
```

🥕 

We should be able to, you can get the full span of a term with `term.span()`. Need to check what calls this though

saolsen: Yeah, lets have it be a followup. Think it's just stack traces and warnings other than this. Both seem like they'd also be improved by hi lighting the whole expression.

polar-core/src/debugger.rs
saolsen: updated it to `Into`

gj (changes requested):
- [x] changelog entry

docs/changelogs/vNEXT.rst
```diff
@@ -55,6 +55,13 @@ Note that custom constructors are still supported for the Rust library since
 specifying a static ``new`` method for a type is nothing more than a
 convention.
 
+Improvements to the debugger
+----------------------------
+Three ways to step through queries, step, over and out.
```
gj, line 60: None of these commands are new, but the implementation is. Maybe add a blurb about how each has changed (+ optionally link to the updated documentation)?

docs/changelogs/vNEXT.rst
```diff
@@ -55,6 +55,13 @@ Note that custom constructors are still supported for the Rust library since
...
+Inspect the stack.
```
gj, line 61: Which stack?

docs/changelogs/vNEXT.rst
```diff
@@ -55,6 +55,13 @@ Note that custom constructors are still supported for the Rust library since
...
+Inspect queries at every level of the stack.
```
gj, line 62: Are these two lines talking about the same thing?

docs/changelogs/vNEXT.rst
saolsen: Yeah, I don't really wanna call it the trace stack or the query stack. Since the differences are subtle and mostly implementation details. If you want something more specific we should call it the "call stack".

docs/changelogs/vNEXT.rst
saolsen: No, one is the `stack` command which gives you a summary and the other are the `query n` commands which shows the query at the nth level.

polar-core/src/debugger.rs
```diff
@@ -229,18 +207,94 @@ impl Debugger {
...
-                })
+            "query" => {
+                let mut level = parts.get(1).and_then(|s| s.parse().ok()).unwrap_or(0);
+                let mut trace_stack = vm.trace_stack.clone();
+
+                let mut term = vm.trace.last().and_then(|t| t.term());
+                while level > 0 {
+                    if let Some(trace) = trace_stack.pop().map(|ts| ts.as_ref().clone()) {
+                        if let Some(t) = trace.last() {
+                            if let Trace{node: Node::Term(t), ..} = &**t {
```
gj, line 218: `&**another_one!`

polar-core/src/debugger.rs
```diff
@@ -229,18 +207,94 @@ impl Debugger {
...
+                        }
+                        Node::Term(t) => {
+                            if matches!(t.value(), Value::Expression(Operation { operator: Operator::And, args}) if args.len() == 1)
+                            {
+                                continue;
+                            }
+
+
+                            let _ = write!(st, "{}: {}", i-1, vm.term_source(t, false));
+                            i-= 1;
```
gj, line 271: rustfmt taking the day off

docs/more/dev-tools/debugger.rst
```diff
@@ -52,22 +51,22 @@ Print the debugger command reference.
...
-                            query and stop at its next sibling (if one exists).
-    over                    Evaluate goals until reaching the next sibling of the
-                            current query (if one exists).
-    queries                 Print current query stack.
-    q[uit]                  Alias for 'continue'.
-    stack                   Alias for 'queries'.
-    s[tep]                  Evaluate one goal.
-    var [<name> ...]        Print available variables. If one or more arguments
-                            are provided, print the value of those variables.
+  bindings                Print current binding stack.
```
gj, line 54: This help output doesn't seem accurate given the changes in `debugger.rs`

polar-core/src/debugger.rs
```diff
@@ -172,23 +111,51 @@ impl Debugger {
...
+    }
+
+    /// Produce the `Goal::Debug` for breaking on a Query (as opposed to breaking on a Goal).
+    /// This is used to implement the `step`, `over`, and `out` debug commands
+    pub fn break_query(&self, vm: &PolarVirtualMachine) -> Option<Rc<Goal>> {
+        let message = vm.trace.last().and_then(|trace| {
+            if let Trace {
+                node: Node::Term(q),
+                ..
+            } = &**trace
```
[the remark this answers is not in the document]
saolsen: Yeah... This is preferable to making this like three nested if-lets which is the only other way I managed to write it. `trace` is an `&RC<Trace>` and there's just not a great way to match the trace part of that other than this.

polar-core/src/debugger.rs
saolsen: updated

polar-core/src/debugger.rs
```diff
@@ -172,23 +111,51 @@ impl Debugger {
...
+                    self.break_query(vm)
+                }
+                (Step::Over { level }, DebugEvent::Query) if vm.trace_stack.len() == *level => {
+                    self.break_query(vm)
+                }
+                _ => None,
+            }
+        } else {
+            None
+        }
```
gj, line 132: nit: The previous format (an `.and_then()` instead of an `if let Some(step)`) obviates this trailing `else` block

polar-core/src/debugger.rs
```diff
@@ -172,23 +111,51 @@ impl Debugger {
...
-                        || "".to_string(),
-                        |query| self.query_source(&query, &vm.kb.read().unwrap().sources, 0),
-                    ),
-                }))
+                match q.value() {
+                    Value::Expression(Operation {
+                        operator: Operator::And,
+                        args,
+                    }) if args.len() == 1 => None,
+                    _ => {
```
gj, line 149: nit: Elsewhere in this file, we use a `matches!` macro for this same check. Not sure which I prefer; just pointing out the dissonance.

```rust
if matches!(t.value(), Value::Expression(Operation { operator: Operator::And, args}) if args.len() == 1)
```

polar-core/src/debugger.rs
```diff
@@ -229,18 +207,97 @@ impl Debugger {
...
+                } else {
+                    return Some(Goal::Debug {
+                        message: "".to_owned()
+                    })
+                }
             }
-            "out" => {
-                self.step = Some(Step::Out {
-                    snapshot: vm.queries[..vm.queries.len().saturating_sub(3)].to_vec(),
+            "stack" | "trace" => {
```
gj, line 240: nit: same as above, this is a very long match arm that might be better split out into a separate function

gj (approved):
few nits; don't need to be addressed prior to shipping

PR #471, opened 2020-10-12
Fix macOS CI

saolsen:
Not sure why but `lalrpop` can't pick up `serde_derive` from the `derive` feature on `serde`. Maybe something to do with one being a dev dep and one a normal dep? IDK all I know is adding the `serde_derive` crate explicitly fixed the osx build.


saolsen:
Well that's not the whole story, there's something wrong with caching too that interacts with this. Turned off caching for the osx jobs for now.

samscott89 (approved):

PR #472, opened 2020-10-12
Improve post-rewrite expression formatting

plotnick:
Three-arg variants of operations like `.` and `+` were previously formatted poorly, either as `.(x, y, z)` or just `a + b`, both of which are confusing and/or misleading. We now format them as `a.b = c` or `a + b = c`.

### DIFF
--- a/polar-core/src/formatting.rs
+++ b/polar-core/src/formatting.rs
@@ -409,17 +409,22 @@ pub mod to_polar {
                         )
                     }
                 }
-                // `Dot` sometimes formats as a predicate
+                // Lookup operator
                 Dot => {
-                    if self.args.len() == 2 {
-                        let call_term = if let Value::String(s) = self.args[1].value() {
-                            s.to_string()
-                        } else {
-                            self.args[1].to_polar()
-                        };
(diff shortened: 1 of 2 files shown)


polar-core/src/formatting.rs
```diff
@@ -430,12 +435,22 @@ pub mod to_polar {
                 ),
                 // Binary operators
                 Mul | Div | Add | Sub | Eq | Geq | Leq | Neq | Gt | Lt | Unify | Isa | In
-                | Assign => format!(
-                    "{} {} {}",
-                    to_polar_parens(self.operator, &self.args[0]),
-                    self.operator.to_polar(),
-                    to_polar_parens(self.operator, &self.args[1])
-                ),
+                | Assign => match self.args.len() {
```
gj, line 438: Can all of these operators appear in a three-arg formulation?

plotnick: Probably not `Unify`, `Isa`, `Assign`. I don't know that it's worth splitting them out, though.

gj (commented):
It looks like the tests that were updated only test the dot + unify operators. Do you think it's worth adding tests for additional operators?

gj (approved):

PR #474, opened 2020-10-13
Remove old quickstart files

gj:


leina05 (approved):
LGTM, taking adding tests to quickstarts + submodules as a follow up

PR #479, opened 2020-10-15
Test Quickstarts

gj:
Sample output:
```
[nodejs] Installing dependencies...
[nodejs] Starting server...
[nodejs] Testing with no rules...
[nodejs] Checking that Alice cannot see their own expense...
[nodejs] Restarting server...
[nodejs] Testing string matching rule...
[nodejs] Checking that alice@example.com can see any expense...
[nodejs] Checking that alice@foo.bar cannot see any expense...
[nodejs] Restarting server...
[nodejs] Testing application data rule...
[nodejs] Checking that Alice can see their own expense...
[nodejs] Checking that Bhavik can see their own expense...
[nodejs] Checking that Alice cannot see Bhavik's expense...
[nodejs] Checking that Bhavik cannot see Alice's expense...
[python] Installing dependencies...
[python] Starting server...
etc...
```

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -97,7 +97,10 @@ nodejs-test:
 	cd examples/inheritance/nodejs && yarn && yarn jest
 	cd examples/user_types/nodejs && yarn && yarn jest
 
-test: doctest
+test-quickstarts:
+	$(MAKE) -C examples/quickstart test
+
+test: doctest test-quickstarts
 	$(MAKE) python-test
 	$(MAKE) ruby-test
 	$(MAKE) java-test
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -100,7 +100,7 @@ nodejs-test:
 test-quickstarts:
 	$(MAKE) -C examples/quickstart test
 
-test: doctest test-quickstarts
+test: doctest
 	$(MAKE) python-test
 	$(MAKE) ruby-test
 	$(MAKE) java-test
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -26,7 +26,7 @@ JS_FILES := $(call rwildcard ../languages/js/src, *.ts)
 help:
 	@$(SPHINXBUILD) -M help "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
 
-.PHONY: help Makefile html doctest
+.PHONY: help Makefile html doctest test-quickstarts
 
 _api_docs/ruby: $(RUBY_FILES)
 	mkdir -p _api_docs
--- a/docs/examples/quickstart/test_quickstarts.rb
+++ b/docs/examples/quickstart/test_quickstarts.rb
@@ -11,6 +11,7 @@ CURL_EMPTY = "curl: (52) Empty reply from server\n"
 
 quickstarts = [
 #   { lang: 'nodejs', setup: 'npm i', server: 'npm start' },
+  { lang: 'java', setup: 'make build', server: 'make run' },
   { lang: 'python', setup: 'pip install --upgrade -r requirements.txt', server: 'python server.py' },
   { lang: 'ruby', setup: 'bundle', server: 'bundle exec ruby server.rb' },
   { lang: 'rust', setup: 'cargo build', server: 'cargo run' }
--- a/docs/examples/quickstart/test_quickstarts.rb
+++ b/docs/examples/quickstart/test_quickstarts.rb
@@ -32,7 +32,8 @@ quickstarts.each do |qs|
       Timeout.timeout 30 do
         begin
           puts "#{prefix} Starting server..."
-          server = spawn qs[:server], %i[out err] => '/dev/null'
+          #   server = spawn qs[:server], %i[out err] => '/dev/null'
+          server = spawn qs[:server]
           received = CURL_ERROR
           while received == CURL_ERROR || received == CURL_EMPTY
             sleep 0.5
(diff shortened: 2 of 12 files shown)


gj:
*[Knocking furiously on wood]* I *think* we have conquered the test flakiness.

The Rust quickstart takes ages since it requires nightly (b/c of Rocket), so we have to recompile everything (including oso) from the ground up on nightly.

@leina05 suggested that we might only need to run this when releasing since that's the only time the outcome will change (b/c the quickstarts rely on the latest released version).

docs/Makefile
```diff
@@ -97,7 +97,10 @@ nodejs-test:
 	cd examples/inheritance/nodejs && yarn && yarn jest
 	cd examples/user_types/nodejs && yarn && yarn jest
 
-test: doctest
+test-quickstarts:
```
dhatch, line 100: Add to .PHONY

docs/examples/quickstart/test_quickstarts.rb
```diff
@@ -0,0 +1,143 @@
...
+  puts "#{e}: #{server}"
+end
+
+# rubocop:disable Metrics/BlockLength
+
+quickstarts.each do |qs|
+  lang = qs[:lang]
+  qs_dir = "oso-#{lang}-quickstart"
+  Bundler.with_unbundled_env do
+    Dir.chdir(qs_dir) do
```
dhatch, line 54: Can we make this chdir relative to the test file so that the test works when run from a different directory?

docs/examples/quickstart/test_quickstarts.rb
```diff
@@ -0,0 +1,143 @@
...
+  until (server = `lsof -ti :5050 2>&1`.split.first.to_i).zero?
+    sleep 0.5
+    Process.kill 'TERM', server
+  end
+rescue Errno::ESRCH => e
+  puts "#{e}: #{server}"
+end
+
+def kill_server(server)
+  return if server.nil?
```
dhatch, line 40: Should this be an assert? When would `server` be nil?

dhatch (commented):
Cool - a few small questions.

Looks like this adds 4m to the docs test? It would be nice if these could run in parallel.  Could we make the script take an input parameter that specifies the language to run and have separate make targets for each so that `make -jN` will run these concurrently?

Also you mentioned that everything needs to be rebuilt in nightly, but I'm not seeing much time for that in the logs.  Is that still true?

docs/examples/quickstart/test_quickstarts.rb
```diff
@@ -0,0 +1,143 @@
...
+  received = CURL_ERROR
+  while [CURL_ERROR, CURL_EMPTY].include? received
+    sleep 0.5
+    received = `curl -sSH "user: #{user}" localhost:5050/expenses/#{expense_id} 2>&1`
+  end
+  [server, received]
+end
+
+def ensure_port_5050_is_open
+  until (server = `lsof -ti :5050 2>&1`.split.first.to_i).zero?
```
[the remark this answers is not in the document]
gj: In Ruby, assignments are expressions, so the value of `x = 1` is `1`. This line is saying "keep running `lsof -ti :5050 2>&1` until parsing the first line of the output as an integer returns `0`, which will never be a pid. If it returns a non-zero int, it means there's at least one pid still listening on port 5050. `.to_i` returns `0` if it's passed something it can't reasonably parse into an integer -- in this case, an empty string b/c `lsof -ti :5050 2>&1` returned nothing b/c there's no longer anything listening on port 5050.

Not the most intuitive code, but it gets the job done and I didn't spend much effort cleaning it up.

gj:
> Looks like this adds 4m to the docs test? It would be nice if these could run in parallel. Could we make the script take an input parameter that specifies the language to run and have separate make targets for each so that `make -jN` will run these concurrently?

Is this concern addressed if we only run it on release as a separate step instead of as part of the docs tests?

> Also you mentioned that everything needs to be rebuilt in nightly, but I'm not seeing much time for that in the logs. Is that still true?

Not sure, but also not a concern if we only run this as a job in the Create Release workflow on merges to main.

docs/examples/quickstart/test_quickstarts.rb
```diff
@@ -0,0 +1,143 @@
...
+rescue Errno::ESRCH => e
+  puts "#{e}: #{server}"
+end
+
+# rubocop:disable Metrics/BlockLength
+
+quickstarts.each do |qs|
+  lang = qs[:lang]
+  qs_dir = "oso-#{lang}-quickstart"
+  Bundler.with_unbundled_env do
```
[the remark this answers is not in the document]
gj: This is a workaround for the fact that we're using Ruby to orchestrate these tests and one of the examples under test is a Ruby project. Without this, building and running the Ruby quickstart guide would use an unclean environment, e.g., would use the local version of the `oso-oso` gem (in the `languages/ruby` path) instead of the latest release.

docs/examples/quickstart/test_quickstarts.rb
gj: It bit me once, so I added the check. Not sure if it's still necessary (reading the code it doesn't seem like it should be possible), but I at least moved the check to a more appropriate place in 5ff8521.

docs/examples/quickstart/test_quickstarts.rb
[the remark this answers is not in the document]
dhatch: Thanks.

dhatch (approved):
lgtm. I guess we can't test until we merge it really.

PR #481, opened 2020-10-19
mod + rem operations

gj:
PR checklist:

- [ ] Added changelog entry.


polar-core/src/lexer.rs
```diff
@@ -108,6 +110,8 @@ impl ToString for Token {
             Token::Bang => "!".to_owned(),          // !
             Token::Mul => "*".to_owned(),           // *
             Token::Div => "/".to_owned(),           // /
+            Token::Mod => "mod".to_owned(),         // mod
+            Token::Rem => "rem".to_owned(),         // rem
```
saolsen, line 114: Why not `%` instead of `rem`?

polar-core/src/polar.lalrpop
```diff
@@ -27,42 +27,44 @@ extern {
...
+        "cut" => lexer::Token::Cut,         // cut
+        "debug" => lexer::Token::Debug,     // debug
+        "print" => lexer::Token::Print,     // print
+        "in" => lexer::Token::In,           // in
+        "forall" => lexer::Token::ForAll,   // forall
+        "if" => lexer::Token::If,           // if
+        "and" => lexer::Token::And,         // and
+        "or" => lexer::Token::Or,           // or
+        "not" => lexer::Token::Not,         // not
+        "matches" => lexer::Token::Matches, // matches
```
saolsen, line 67: 📏

saolsen (approved):

polar-core/src/lexer.rs
plotnick: Because then it's ambiguous whether it means `rem` or `mod`. They agree on nonnegative integers, but for other cases they behave differently. This has been a perennial problem in languages; e.g., see [this crazy Wiki chart](https://en.wikipedia.org/wiki/Modulo_operation#In_programming_languages). I agree that `%` is slightly nicer, but this way we can avoid users asking, "ok, but does `%` mean `mod` or `rem`?"

PR #484, opened 2020-10-21
Fail on all unsupported partial operations except Not

gj:
We now error on:

- [x] comparison of two partials
- [x] unification of two partials
- [x] partial as the RHS of `matches` (not currently parse-able)
- [x] method calls on a partial
- [x] arithmetic ops
- [x] cut in the presence of 1+ partials
- [x] dot lookups where the field is a partial

Added tests for:

- [x] `in`
- [x] `==` (additional comparison op test)
- [x] multiple partials passed into the same rule
- [x] the two trivial partials (`f(x);` and `g(x) if false;`) reducing to `true`/`false`
- [x] some uses of `cut`, which are all currently `#[ignore]`-ed pending future `cut` work.
- [x] assignment to and from a partial

### DIFF
--- a/polar-core/src/partial/constraints.rs
+++ b/polar-core/src/partial/constraints.rs
@@ -488,6 +488,32 @@ mod test {
         assert!(matches!(error, PolarError {
             kind: ErrorKind::Runtime(RuntimeError::Unsupported { .. }), ..}));
 
+        // Multiple interacting query partials.
+        // Unification.
+        polar.load_str(r#"h(x, y) if x = y;"#)?;
+        let mut query = polar.new_query_from_term(
+            term!(call!(
+                "h",
+                [Constraints::new(sym!("a")), Constraints::new(sym!("b"))]
+            )),
+            false,
+        );
+        let error = query.next_event().unwrap_err();
+        assert!(matches!(error, PolarError {
+            kind: ErrorKind::Runtime(RuntimeError::Unsupported { .. }), ..}));
+
+        // Comparison.
+        polar.load_str(r#"i(x, y) if x > y;"#)?;
+        let mut query = polar.new_query_from_term(
+            term!(call!(
+                "i",
+                [Constraints::new(sym!("a")), Constraints::new(sym!("b"))]
+            )),
+            false,
+        );
+        let error = query.next_event().unwrap_err();
+        assert!(matches!(error, PolarError {
+            kind: ErrorKind::Runtime(RuntimeError::Unsupported { .. }), ..}));
         Ok(())
     }
 }
--- a/polar-core/src/partial/constraints.rs
+++ b/polar-core/src/partial/constraints.rs
@@ -452,6 +452,23 @@ mod test {
     }
 
     #[test]
+    fn test_multiple_partials() -> TestResult {
+        let polar = Polar::new();
+        polar.load_str(r#"f(x, y) if x = 1 and y = 2;"#)?;
+        let mut query = polar.new_query_from_term(
+            term!(call!(
+                "f",
+                [Constraints::new(sym!("a")), Constraints::new(sym!("b"))]
+            )),
+            false,
+        );
+        let next = next_binding(&mut query)?;
+        assert_eq!(next[&sym!("a")], term!(1));
+        assert_eq!(next[&sym!("b")], term!(2));
+        Ok(())
+    }
+
+    #[test]
     fn test_unsupported_ops_on_partial() -> TestResult {
         let polar = Polar::new();
 
--- a/polar-core/src/partial/constraints.rs
+++ b/polar-core/src/partial/constraints.rs
@@ -537,4 +537,45 @@ mod test {
             kind: ErrorKind::Runtime(RuntimeError::TypeError { .. }), ..}));
         Ok(())
     }
+
+    #[test]
+    fn test_cut_with_partial() -> TestResult {
+        let polar = Polar::new();
+        polar.load_str(
+            r#"
+            f(x) if x = 1;
+            f(x) if x = 2 and cut;
+            f(x) if x = 3;
+        "#,
+        )?;
+        let mut query =
+            polar.new_query_from_term(term!(call!("f", [Constraints::new(sym!("a"))])), false);
+        let next = next_binding(&mut query)?;
+        assert_eq!(next[&sym!("a")], term!(1));
+        let next = next_binding(&mut query)?;
+        assert_eq!(next[&sym!("a")], term!(2));
+        assert!(matches!(query.next_event()?, QueryEvent::Done { .. }));
+        Ok(())
+    }
+
+    #[test]
+    fn test_method_sorting_with_cut_and_partial() -> TestResult {
+        let polar = Polar::new();
+        polar.load_str(
+            r#"
+            f(x, y) if cut and x = 1;
+            f(x, y: 2) if x = 2;
+        "#,
+        )?;
+        let mut query = polar.new_query_from_term(
+            term!(call!("f", [Constraints::new(sym!("a")), value!(2)])),
+            false,
+        );
+        let next = next_binding(&mut query)?;
+        assert_eq!(next[&sym!("a")], term!(2));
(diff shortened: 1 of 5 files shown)


polar-core/src/vm.rs
```diff
@@ -223,6 +227,31 @@ impl Default for PolarVirtualMachine {
     }
 }
 
+#[allow(clippy::ptr_arg)]
+fn query_contains_partial(goals: &Goals) -> bool {
+    if goals.len() == 1 {
```
dhatch, line 232: This relies on the VM always being instantiated with a single goal.  If there are other goals that contain partials, or a query that occurs later with a partial, it won't be identified.

What if we instead set this flag in the beginning of the `query` function? (The handler for the Query goal.). That could walk the term in the same way, and wouldn't have the same short comings.

polar-core/src/vm.rs
```diff
@@ -892,6 +923,8 @@ impl PolarVirtualMachine {
         );
 
         match (&left.value(), &right.value()) {
+            (_, Value::Partial(_)) => unreachable!("cannot match against a partial"),
```
dhatch, line 926: Should this be an error, not a panic?

polar-core/src/vm.rs
```diff
@@ -1490,6 +1523,15 @@ impl PolarVirtualMachine {
                 ])?;
             }
             Operator::Cut => {
+                if self.query_contains_partial {
+                    return Err(self.set_error_context(
+                        &term,
+                        error::RuntimeError::Unsupported {
+                            msg: "cannot use cut with a partial".to_string(),
```
dhatch, line 1530:
```suggestion
                            msg: "cannot use cut with a query containing a partial".to_string(),
```

polar-core/src/vm.rs
```diff
@@ -2007,14 +2059,16 @@ impl PolarVirtualMachine {
     /// This is sort of a "sub-goal" of `Unify`.
     fn unify_partial(&mut self, partial: &partial::Constraints, right: &Term) -> PolarResult<()> {
         let mut partial = partial.clone();
-        if let Value::Partial(right_partial) = right.value() {
```
dhatch, line 2010: Was this code wrong?

polar-core/src/partial/simplify.rs
```diff
@@ -30,6 +30,10 @@ pub struct Simplifier;
 impl Folder for Simplifier {
     fn fold_term(&mut self, t: Term) -> Term {
         match t.value() {
+            Value::Partial(Constraints { operations, .. }) if operations.is_empty() => {
```
dhatch, line 33: A partial with no constraints is replaced with a boolean true.  I'm not sure I understand this simplification. Could you explain further?

Couldn't it be any value?

polar-core/src/partial/constraints.rs
```diff
@@ -447,4 +438,199 @@ mod test {
...
+    #[test]
+    fn test_trivial_partials() -> TestResult {
+        let polar = Polar::new();
+        polar.load_str(
+            r#"f(x);
+               g(x) if false;"#,
+        )?;
+        let mut query = polar.new_query_from_term(term!(call!("f", [partial!("a")])), false);
+        let next = next_binding(&mut query)?;
+        assert_eq!(next[&sym!("a")], term!(true));
```
dhatch, line 520: I don't understand why this should be simplified to true. Related to other comment.

polar-core/src/partial/constraints.rs
```diff
@@ -447,4 +438,199 @@ mod test {
...
+        assert!(matches!(query.next_event()?, QueryEvent::Done { .. }));
+
+        let mut query = polar.new_query_from_term(term!(call!("g", [partial!("a")])), false);
+        assert!(matches!(query.next_event()?, QueryEvent::Done { .. }));
+        Ok(())
+    }
+
+    #[test]
+    fn test_in_with_partial() -> TestResult {
```
dhatch, line 529: What about `1 in x` where `x` is a partial? That should fail, right?

dhatch (changes requested):
Mostly good. Small comments / questions.

dhatch:
What about failing isa with fields?

polar-core/src/vm.rs
gj: I believe this is currently an unparseable production, which is why we set it as `unreachable!()`.

polar-core/src/vm.rs
gj: Is this necessary? The only time users will encounter this error is when they query with a partial for a rule containing a `cut`.

polar-core/src/vm.rs
plotnick: It doesn't rely on it, it tests for it. The problem with testing in `query` is that it gets hit for every sub-query, of which there are potentially many. This way we do a walk up front, and we're done.

polar-core/src/vm.rs
plotnick: I think so. I think that if you unify two partials, we should make their constraint sets identical thereafter, either by sharing a ref, or adding constraints to the other whenever we add them to one, or something. That seems medium hard, so we just made it an error for now.

polar-core/src/partial/simplify.rs
plotnick: Constraints are conjunctive truth conditions. If there are no constraints, the truth conditions are vacuous, and we have a zero-argument `and`, which evaluates to `true`.

You could substitute any trivially true expression, such as `1 = 1`, but only `true` can't be simplified any further.

polar-core/src/partial/constraints.rs
plotnick: Indeed. Tested in `e63cd15e`.

plotnick:
> What about failing isa with fields?

I knew we were forgetting something. Thank you; fixed in `4757bf7b`.

dhatch (approved):

polar-core/src/vm.rs
dhatch: Didn't think of the sub-query point.  Perhaps I should have phrased this differently.

If the VM is created with more than one goal, we never check for partials.  This could result in incorrect behavior if the assumption that we always start the VM with one goal is broken.

Instead, we could iterate all initial goals and walk over all query goals to find partials.

polar-core/src/vm.rs
dhatch: What if we have:

```
f(x, y) if x matches y;
```

And the user queries with `y` as a partial?

polar-core/src/vm.rs
dhatch: this is fine nvm.

polar-core/src/partial/simplify.rs
dhatch: What about this knowledge base:

```
f(x);
```

query: `f(partial(x))`

Output `x = true`

query: `f(x)` (x is a variable)
output: `x = Variable('_x_2')`

These seem inconsistent.

polar-core/src/vm.rs
plotnick: The parser treats the RHS of `x matches y` as an instance literal with the class `y`, not as the variable `y`. So I don't believe this case can currently arise, but we made it an error anyway.

polar-core/src/vm.rs
plotnick: Ok, done.

polar-core/src/partial/simplify.rs
plotnick: A possibly more consistent answer would be a partial with an empty constraint set. But we've decided that such partials, and ones that unify against ground terms, should be maximally simplified.

polar-core/src/vm.rs
gj: 47dfd0e27bbb6cd7b3540f84a42ed1106fd112bc

polar-core/src/vm.rs
gj: b85270fb6f69a58fd7bdcfcafe8bfaa946c4708e

PR #485, opened 2020-10-24
No longer automatically iterate on lookups

samscott89:
PR checklist:

- [x] Python
- [x] Ruby
- [x] Java
- [x] Node.js
- [x] Rust
- [x] Update docs examples
- [x] Doc entry
- [x] Added changelog entry.
- [x] Update FFI support doc to add NextExternal
- [x] New tests for in behaviour
- [ ] Update (out of repository) examples - do last

### DIFF
--- a/docs/changelogs/vNEXT.rst
+++ b/docs/changelogs/vNEXT.rst
@@ -13,12 +13,28 @@ Breaking changes
 .. warning:: This release contains breaking changes. Be sure
    to follow migration steps before upgrading.
 
-Breaking change 1
------------------
+Explicit iterator syntax
+------------------------
 
-- summary of breaking change
+Calling methods and accessing attributes on application instances no longer
+returns multiple values. Instead, to iterate over values of an application
+instance the explicit ``in`` operator must be used.
 
-Link to migration guide
+**Old syntax**:
+
+   amount = recipe.ingredients().quantity
+
+**New syntax**:
+
+  item in recipe.ingredients() and
+  amount = item.quantity
+
+With this change, any methods on the iterable instance itself can now be used::
+
+  recipe.ingredients().total_cost()
+
+The above would call the ``total_cost`` method on the iterable returned by
+``recipe.ingredients()``.
 
 New features
 ============
--- a/docs/using/polar-syntax.rst
+++ b/docs/using/polar-syntax.rst
@@ -405,10 +405,11 @@ arguments must come before keyword arguments::
 In (List Membership)
 ^^^^^^^^^^^^^^^^^^^^
 
-The ``in`` operator can be used to iterate over a list or
-an iterable application instance.
+The ``in`` operator can be used to iterate over elements of
+builtin and application types. Iterable builtin types are
+``List``, ``String``, and ``Dictionary``.
 
-The first operand will be unified with each element of the list. If the
+The first operand will be unified with each element. If the
 second operand is not iterable, the operation will fail.
 
 For example::
@@ -422,6 +423,30 @@ The left-hand side does not need to be a variable, for example::
     1 in [1, 2, 3, 1]
 
 Will succeed *twice*: 1 is in the first and fourth position.
+
+Iterating over a ``String`` returns each character (as another string)::
+
+    hexstring(s) if
+      forall(c in s, c in "0123456789abcdef");
+
+Iterating over  ``Dictionary`` returns a list with two elements for the key
+(as a string)
+and the value::
+
+    x in {a: 1, b: 2}
+    [key, _] in {a: 1, b: 2}
+    [_, value] in {a: 1, b: 2}
+
+The above returns::
+
+    x => ["a", 1]
+    x => ["b", 2]
+
+    key => "a"
+    key => "b"
+
+    value => 1
+    value => 2
 
 .. _operator-forall:
 
--- a/docs/using/polar-syntax.rst
+++ b/docs/using/polar-syntax.rst
@@ -423,8 +423,6 @@ The left-hand side does not need to be a variable, for example::
 
 Will succeed *twice*: 1 is in the first and fourth position.
 
-The ``in`` operator generates *alternatives* for each element.
-
 .. _operator-forall:
 
 For All
--- a/docs/using/polar-syntax.rst
+++ b/docs/using/polar-syntax.rst
@@ -405,10 +405,11 @@ arguments must come before keyword arguments::
 In (List Membership)
 ^^^^^^^^^^^^^^^^^^^^
 
-The ``in`` operator can be used to iterate over a list. If the second operand is a list, the first operand will
-be unified with each element of the list. If the
-second operand is not a list (or variable bound to a list),
-the operation will fail.
+The ``in`` operator can be used to iterate over a list or
+an iterable application instance.
+
+The first operand will be unified with each element of the list. If the
+second operand is not iterable, the operation will fail.
 
 For example::
 
@@ -416,8 +417,13 @@ For example::
 
 Will bind ``x`` to ``1``, ``2``, ``3``, in turn, and check that ``x = 1``
 for each. This expression will only succeed for the first item (``1``).
+The left-hand side does not need to be a variable, for example::
 
-The ``in`` operator generates *alternatives* for each element of the list.
+    1 in [1, 2, 3, 1]
+
+Will succeed *twice*: 1 is in the first and fourth position.
+
+The ``in`` operator generates *alternatives* for each element.
 
 .. _operator-forall:
 
--- a/languages/java/oso/src/main/java/com/osohq/oso/Query.java
+++ b/languages/java/oso/src/main/java/com/osohq/oso/Query.java
@@ -58,7 +58,8 @@ public class Query implements Enumeration<HashMap<String, Object>> {
   }
 
   /** Helper for `ExternalCall` query events */
-  private void handleCall(String attrName, Optional<JSONArray> jArgs, JSONObject polarInstance, long callId)
+  private void handleCall(
+      String attrName, Optional<JSONArray> jArgs, JSONObject polarInstance, long callId)
       throws Exceptions.OsoException {
     Optional<List<Object>> args = Optional.empty();
     if (jArgs.isPresent()) {
@@ -67,7 +68,11 @@ public class Query implements Enumeration<HashMap<String, Object>> {
     try {
       Object instance;
       if (polarInstance.getJSONObject("value").has("ExternalInstance")) {
-        long instanceId = polarInstance.getJSONObject("value").getJSONObject("ExternalInstance").getLong("instance_id");
+        long instanceId =
+            polarInstance
+                .getJSONObject("value")
+                .getJSONObject("ExternalInstance")
+                .getLong("instance_id");
         instance = host.getInstance(instanceId);
       } else {
         instance = host.toJava(polarInstance);
@@ -77,8 +82,11 @@ public class Query implements Enumeration<HashMap<String, Object>> {
       try {
         Class<?> cls = instance instanceof Class ? (Class<?>) instance : instance.getClass();
         if (args.isPresent()) {
-          Class<?>[] argTypes = args.get().stream().map(a -> a.getClass()).collect(Collectors.toUnmodifiableList())
-              .toArray(new Class[0]);
+          Class<?>[] argTypes =
+              args.get().stream()
+                  .map(a -> a.getClass())
+                  .collect(Collectors.toUnmodifiableList())
+                  .toArray(new Class[0]);
           Method method = MethodUtils.getMatchingAccessibleMethod(cls, attrName, argTypes);
           if (method == null) {
             throw new Exceptions.InvalidCallError(cls.getName(), attrName, argTypes);
@@ -228,8 +236,7 @@ public class Query implements Enumeration<HashMap<String, Object>> {
           System.out.print("debug> ");
           try {
             String input = br.readLine();
-            if (input == null)
-              break;
+            if (input == null) break;
             String command = host.toPolarTerm(input).toString();
             ffiQuery.debugCommand(command);
           } catch (IOException e) {
@@ -253,10 +260,9 @@ public class Query implements Enumeration<HashMap<String, Object>> {
     }
   }
 
-  /**
-   * Get the next JSONified Polar result of a cached method call (enumeration).
-   */
-  protected JSONObject nextCallResult(long callId) throws NoSuchElementException, Exceptions.OsoException {
+  /** Get the next JSONified Polar result of a cached method call (enumeration). */
+  protected JSONObject nextCallResult(long callId)
+      throws NoSuchElementException, Exceptions.OsoException {
     return host.toPolarTerm(getCall(callId).nextElement());
   }
 }
--- a/languages/java/oso/src/main/java/com/osohq/oso/Query.java
+++ b/languages/java/oso/src/main/java/com/osohq/oso/Query.java
@@ -112,7 +112,7 @@ public class Query implements Enumeration<HashMap<String, Object>> {
     }
   }
 
-  /** Helper for `ExternalCall` query events */
+  /** Helper for `NextExternal` query events */
   private void handleNextExternal(long callId, JSONObject term) throws Exceptions.OsoException {
     if (!calls.containsKey(callId)) {
       Object result = host.toJava(term);
--- a/languages/java/oso/src/main/java/com/osohq/oso/Query.java
+++ b/languages/java/oso/src/main/java/com/osohq/oso/Query.java
@@ -58,15 +58,49 @@ public class Query implements Enumeration<HashMap<String, Object>> {
   }
 
   /** Helper for `ExternalCall` query events */
-  private void handleCall(
-      String attrName, Optional<JSONArray> jArgs, JSONObject polarInstance, long callId)
+  private void handleCall(String attrName, Optional<JSONArray> jArgs, JSONObject polarInstance, long callId)
       throws Exceptions.OsoException {
     Optional<List<Object>> args = Optional.empty();
     if (jArgs.isPresent()) {
       args = Optional.of(host.polarListToJava(jArgs.get()));
     }
     try {
-      registerCall(attrName, args, callId, polarInstance);
+      Object instance;
+      if (polarInstance.getJSONObject("value").has("ExternalInstance")) {
+        long instanceId = polarInstance.getJSONObject("value").getJSONObject("ExternalInstance").getLong("instance_id");
+        instance = host.getInstance(instanceId);
+      } else {
+        instance = host.toJava(polarInstance);
+      }
+      // Select a method to call based on the types of the arguments.
+      Object result = null;
+      try {
+        Class<?> cls = instance instanceof Class ? (Class<?>) instance : instance.getClass();
+        if (args.isPresent()) {
+          Class<?>[] argTypes = args.get().stream().map(a -> a.getClass()).collect(Collectors.toUnmodifiableList())
+              .toArray(new Class[0]);
+          Method method = MethodUtils.getMatchingAccessibleMethod(cls, attrName, argTypes);
+          if (method == null) {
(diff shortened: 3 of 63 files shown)


languages/python/oso/polar/query.py
```diff
@@ -88,47 +86,33 @@ def handle_make_external(self, data):
 
     def handle_external_call(self, data):
         call_id = data["call_id"]
-        if call_id not in self.calls:
```
leina05, line 91: Can we get rid of `self.calls` entirely now?

languages/python/oso/polar/query.py
```diff
@@ -88,47 +86,33 @@ def handle_make_external(self, data):
...
-            value = next(self.calls[call_id])
-            self.ffi_query.call_result(call_id, self.host.to_polar(value))
-        except StopIteration:
+            attr = getattr(instance, attribute)
+        except AttributeError as e:
+            self.ffi_query.application_error(str(e))
             self.ffi_query.call_result(call_id, None)
+            return
+        if (
+            callable(attr) and not data["args"] is None
```
leina05, line 101: I know this isn't related to your changes, but... 
So `data["args"] is None` is how we check whether or not this was parsed as a field or method? If it's a method with no args, is it an empty list? Should we add an explicit flag here, rather than relying on the `None` check?

leina05: It's especially weird for that to be the check for methods, since we now have `kwargs`, which will be `None` if you don't have `kwargs`, even if the call is a method.

samscott89: > I know this isn't related to your changes, but... 
> So `data["args"] is None` is how we check whether or not this was parsed as a field or method? If it's a method with no args, is it an empty list? Should we add an explicit flag here, rather than relying on the `None` check?

All of this is correct.

- Yes that's how we check
- Yes empty list for call instead of none
- a flag might be better. You would still want/need to check the none/list case but might make it less error prone?

leina05: Yeah, would also improve readability.

polar-core/src/vm.rs
```diff
@@ -1418,6 +1425,27 @@ impl PolarVirtualMachine {
...
                         )?;
                     }
+                    // Push an `ExternalLookup` goal for external instances and built-ins.
+                    Value::Dictionary(_) | Value::ExternalInstance(_) | Value::String(_) => {
+                        // Generate symbol for next result and bind to `false` (default)
+                        let next = self.kb.read().unwrap().gensym("next_value");
+                        self.bind(&next, Term::new_temporary(Value::Boolean(false)));
+                        let next_term = Term::new_temporary(Value::Variable(next.clone()));
+                        let call_id = self.new_call_id(&next);
+
```
leina05, line 1435: Kind of confused about exactly what's happening here, but I think this is just setting up a variable to receive the next value from the iterator?

polar-core/src/vm.rs
```diff
@@ -1772,10 +1802,10 @@ impl PolarVirtualMachine {
                 self.bind(&answer, Term::new_temporary(Value::Boolean(false)));
 
                 // append unify goal to be evaluated after external op result is returned & bound
-                self.append_goals(vec![Goal::Unify {
+                self.push_goal(Goal::Unify {
```
leina05, line 1805: Reason for this change?

polar-core/benches/bench.rs
```diff
@@ -169,91 +167,10 @@ pub fn many_rules(c: &mut Criterion) {
     });
 }
```
leina05, line 169: I have no idea how bench tests work so not reviewing these changes.

polar-core/src/vm.rs
```diff
@@ -1418,6 +1425,27 @@ impl PolarVirtualMachine {
...
+                        self.bind(&next, Term::new_temporary(Value::Boolean(false)));
+                        let next_term = Term::new_temporary(Value::Variable(next.clone()));
+                        let call_id = self.new_call_id(&next);
+
+                        // append unify goal to be evaluated after
+                        // next result is fetched
+                        self.append_goals(vec![
+                            Goal::NextExternal {
+                                call_id,
+                                term: self.deep_deref(&list),
```
leina05, line 1441: I think we should rename `list` to reflect that it could be a non-list. Maybe `iterator` would be better.

polar-core/src/events.rs
```diff
@@ -88,4 +88,9 @@ pub enum QueryEvent {
         operator: Operator,
         args: TermList,
     },
+
+    NextExternal {
+        call_id: u64,
+        term: Term,
```
leina05, line 94: The name `term` is very vague. Would be nice if it provided more context, e.g. `iterable_term` or something.

languages/java/oso/src/main/java/com/osohq/oso/Query.java
```diff
@@ -76,7 +118,23 @@ private void handleCall(
       ffiQuery.callResult(callId, null);
       return;
     }
+  }
+
+  /** Helper for `NextExternal` query events */
+  private void handleNextExternal(long callId, JSONObject term) throws Exceptions.OsoException {
+    if (!calls.containsKey(callId)) {
+      Object result = host.toJava(term);
```
leina05, line 126: Isn't it possible that `term` is an instance, in which case we would need to use `host.getInstance()`, not `host.toJava()` (the same way we do on lines 70-80?

languages/java/oso/src/main/java/com/osohq/oso/Query.java
```diff
@@ -66,7 +66,49 @@ private void handleCall(
...
     try {
-      registerCall(attrName, args, callId, polarInstance);
+      Object instance;
+      if (polarInstance.getJSONObject("value").has("ExternalInstance")) {
+        long instanceId =
+            polarInstance
+                .getJSONObject("value")
+                .getJSONObject("ExternalInstance")
+                .getLong("instance_id");
+        instance = host.getInstance(instanceId);
```
leina05, line 76: Can this code be folded into `host.toJava()`?

leina05: After looking at `toJava()`, I think that this branch may be unnecessary because `toJava()` checks for `ExternalInstance` objects. Unless the structure of this event is a special case.

languages/java/oso/src/main/java/com/osohq/oso/Query.java
```diff
@@ -76,7 +118,23 @@ private void handleCall(
...
     }
+  }
+
+  /** Helper for `NextExternal` query events */
+  private void handleNextExternal(long callId, JSONObject term) throws Exceptions.OsoException {
+    if (!calls.containsKey(callId)) {
+      Object result = host.toJava(term);
+      Enumeration<Object> enumResult;
+      if (result instanceof Enumeration) {
+        // TODO: test this
```
leina05, line 129: Highlighting this TODO

languages/java/oso/src/main/java/com/osohq/oso/Query.java
```diff
@@ -66,7 +66,49 @@ private void handleCall(
...
+      if (polarInstance.getJSONObject("value").has("ExternalInstance")) {
+        long instanceId =
+            polarInstance
+                .getJSONObject("value")
+                .getJSONObject("ExternalInstance")
+                .getLong("instance_id");
+        instance = host.getInstance(instanceId);
+      } else {
+        instance = host.toJava(polarInstance);
+      }
```
leina05, line 79:
```suggestion
      Object instance = host.toJava(polarInstance);
```

I believe this works. Small clean-up.

languages/java/oso/src/test/java/com/osohq/oso/PolarTest.java
```diff
@@ -288,17 +288,6 @@ public void testNoKeywordArgs() throws Exception {
         () -> p.query("x = (new MyClass(\"test\", 1)).foo(\"test\", id: 1)"));
   }
 
-  @Test
```
leina05, line 291: Do we need a test to replace this? `testHandleCall`?

leina05 (commented):
## Review of only docs, examples,  java, and polar-core
Looks good! High-level, do we need to add tests for the new operand types for `in` (Dictionaries, Strings, and Iterable External Instances) the way we currently have specific tests for lists?

I will review the rest later this afternoon.

languages/java/oso/src/main/java/com/osohq/oso/Query.java
```diff
@@ -66,7 +66,49 @@ private void handleCall(
       args = Optional.of(host.polarListToJava(jArgs.get()));
     }
     try {
-      registerCall(attrName, args, callId, polarInstance);
+      Object instance;
```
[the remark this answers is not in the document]
samscott89: It's no longer registering anything. A call always returns a single value. We could keep it as a separate method to do the call part though if you think that would be better?

languages/java/oso/src/main/java/com/osohq/oso/Query.java
samscott89: Big cleanup! Nice! I was just moving the code that was already there.

docs/changelogs/vNEXT.rst
```diff
@@ -13,12 +13,28 @@ Breaking changes
...
 
-- summary of breaking change
+Calling methods and accessing attributes on application instances no longer
+returns multiple values. Instead, to iterate over values of an application
+instance the explicit ``in`` operator must be used.
 
-Link to migration guide
+**Old syntax**:
+
+   amount = recipe.ingredients().quantity
```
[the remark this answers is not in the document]
samscott89: Yep! Here's a test case where it was used previously: https://github.com/osohq/oso/pull/485/files#diff-347a4ce26269e34690f5914f1f082bccb9868128ebd7caf088a77ea9f5018ed0L23-R24

polar-core/src/vm.rs
```diff
@@ -357,8 +360,7 @@ impl PolarVirtualMachine {
                 call_id,
                 instance,
                 field,
-                check_errors,
```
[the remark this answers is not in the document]
samscott89: Previously we needed to push a check error goal after every new LookupExternal event we created. Since there is only ever one LookupExternal event, the CheckError goal can just be pushed directly by the goal creator.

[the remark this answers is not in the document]
samscott89: Which is what is happening here: https://github.com/osohq/oso/pull/485/files#diff-31c6da6cb2dee04c7bdefb96252140c683d8d697de90ec8ae7a6745bf46139d9R1635

polar-core/src/vm.rs
samscott89: Yup! Same pattern we've used elsewhere. We might want to extract this into a helper method.

polar-core/src/vm.rs
samscott89: It seemed wasteful to create a vector to push a single goal, when we have a method that accepts a single goal.

docs/using/polar-syntax.rst
```diff
@@ -402,19 +402,25 @@ arguments must come before keyword arguments::
...
 Will bind ``x`` to ``1``, ``2``, ``3``, in turn, and check that ``x = 1``
 for each. This expression will only succeed for the first item (``1``).
+The left-hand side does not need to be a variable, for example::
+
+    1 in [1, 2, 3, 1]
+
+Will succeed *twice*: 1 is in the first and fourth position.
 
-The ``in`` operator generates *alternatives* for each element of the list.
+The ``in`` operator generates *alternatives* for each element.
```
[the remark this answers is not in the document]
samscott89: Removing

polar-core/src/vm.rs
samscott89: did this `new_call_var`

polar-core/benches/bench.rs
samscott89: They were a mess and overly complicated. So I moved the one that benched externals to be a bench in the oso library instead, so that its way simpler.

polar-core/src/events.rs
samscott89: Sure

languages/python/oso/polar/query.py
samscott89: If its okay, I would prefer to deal with this in a follow up API cleanup.

languages/java/oso/src/main/java/com/osohq/oso/Query.java
samscott89: I think all of these are handled by your suggestion?

languages/java/oso/src/main/java/com/osohq/oso/Query.java
samscott89: Done

polar-core/src/vm.rs
samscott89: Done

languages/python/oso/polar/query.py
leina05: Cool, made a notion task

languages/python/oso/tests/test_polar.py
```diff
@@ -835,3 +836,23 @@ class Post:
 
     with pytest.raises(StopIteration):
         next(results)
+
+
+def test_iterators(polar, qeval, qvar):
+    assert qvar('x in "abc"', "x") == ["a", "b", "c"]
+    assert qvar("x in {a: 1, b: 2}", "x") == ["a", "b"]
```
leina05, line 843: It's unfortunate that these features are totally language-specific, even though they appear to be builtin.

languages/ruby/lib/oso/polar/query.rb
```diff
@@ -97,13 +70,33 @@ def application_error(message)
...
-      rescue InvalidCallError => e
+      rescue ArgumentError, NoMethodError => e
         application_error(e.message)
         call_result(nil, call_id: call_id)
+      end
+
+      def handle_next_external(call_id, iterable)
+        unless calls.key? call_id
+          value = host.to_ruby iterable
+          raise InvalidIteratorError unless value.is_a? Enumerable
```
leina05, line 93: Is `Enumerable` really the only check here? Not familiar enough with Ruby.

languages/rust/oso/tests/test_polar_rust.rs
```diff
@@ -489,6 +489,7 @@ fn test_unify_externals() {
         if m == "cannot bind expression 'x = new Foo(1)' to '_instance_23'"));
 }
 
+#[ignore = "Re-enable after updating iterator implementation"]
```
leina05, line 492: Does this need to be re-enabled?

leina05 (commented):
Looks great, thanks for adding tests. Things I'm thinking about (maybe as follow-ups):
- Super annoying comment, but `NextExternal` as an event name is a little hard to parse without context. `NextIterableItem` or something like that would be more clear, but this is pretty nit picky.
- Are there any public examples we need to update that aren't tested in CI? Thinking about Polar Adventure, Zenefits demo (I already have the updates for this locally), anything else? The `=` to `in` change is breaking, so let's make sure examples are up to date.
- The behavior of `in` for various pseudo-builtin types is pretty different depending on the host language. Not sure if this is something we need to document, or whether it's intuitive enough based on the host language behavior.

If we generate a list of any other repos/examples we need to update for this change and figure out who's responsible for making them before Monday, then I think this is good to go.

test/test.polar
```diff
@@ -81,3 +81,5 @@ testUnifyClass(A);
 testUnspecializedRuleOrder(_x, _y, z) if z = 3;
 testUnspecializedRuleOrder(_x, _y: String, z) if z = 2;
 testUnspecializedRuleOrder(_x: String, _y, z) if z = 1;
+
+testIterables() if 1 in new List([3, 2, 1]);
```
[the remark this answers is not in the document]
samscott89: Yep! This just seemed like one way to test the builtin version of list in each language

languages/ruby/lib/oso/polar/query.rb
samscott89: As far as I can tell, that's the main trait. @gj ?

languages/rust/oso/tests/test_polar_rust.rs
samscott89: After the Rust changes are implemented (follow up PR incoming)

samscott89:
> 
> 
> Looks great, thanks for adding tests. Things I'm thinking about (maybe as follow-ups):
> 
>   * Super annoying comment, but `NextExternal` as an event name is a little hard to parse without context. `NextIterableItem` or something like that would be more clear, but this is pretty nit picky.
> 

Agree, but would rather address as follow up... For now wanted it consistent

>   * Are there any public examples we need to update that aren't tested in CI? The `=` to `in` change is breaking, so let's make sure examples are up to date.
>       
>       * Polar Adventure
>       * oso-social app (definitely the branch we sent to Zenefits will break from this)
>       * anything else?
> 

Yes, want to do a review of all the repos as a follow up to this. The demo repo will definitely break

>     * The behavior of `in` for various pseudo-builtin types is pretty different depending on the host language. Not sure if this is something we need to document, or whether it's intuitive enough based on the host language behavior.

Agreed. This was annoying me. Just pushed a change to make it work in the core for strings + dicts.

> 
> 
> If we generate a list of any other repos/examples we need to update for this change and figure out who's responsible for making them before Monday, then I think this is good to go.

Sounds good.

polar-core/benches/bench.rs
samscott89: In case you're interested, this is how it looks now:
https://github.com/osohq/oso/pull/515/files#diff-ccf3c4f853abe6996692d38527d164a48366a51ed18810306292370442de2abdR73

Instead of all that fake external stuff, we can just create a new struct with whatever implementation we want

languages/java/oso/src/test/java/com/osohq/oso/PolarTest.java
```diff
@@ -562,4 +551,50 @@ public void testPathMapper() throws Exception {
...
+    }
+
+    // code for data structure
+    public Iterator<Integer> iterator() {
+      return list.iterator();
+    }
+  }
+
+  @Test
+  public void testIterators() throws Exception {
```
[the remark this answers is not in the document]
leina05: @samscott89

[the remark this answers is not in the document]
samscott89: Oh sorry forgot to reply. Enumerations are tested through the prior functionality in `testEnumerationCallResults`.

languages/ruby/lib/oso/polar/query.rb
leina05: Cursory google search says that's correct

leina05 (approved):

languages/ruby/lib/oso/polar/query.rb
gj: I think this is fine -- the call to `.lazy` on 95 will turn any `Enumerable` into a lazy enumerator. So this check is basically asking "is it possible to create an enumerator out of `value`, even if it isn't already an enumerator?"

languages/java/oso/src/main/java/com/osohq/oso/Query.java
```diff
@@ -76,7 +109,27 @@ private void handleCall(
       ffiQuery.callResult(callId, null);
       return;
     }
+  }
+
+  /** Helper for `NextExternal` query events */
+  private void handleNextExternal(long callId, JSONObject iterable) throws Exceptions.OsoException {
+    if (!calls.containsKey(callId)) {
+      Object result = host.toJava(iterable);
+      Enumeration<Object> enumResult;
```
gj, line 118: @samscott89 I see there was already some usage of `Enumeration<Object>` in the previous code, but did you think about using Iterator instead?

https://docs.oracle.com/javase/10/docs/api/java/util/Iterator.html

> An iterator over a collection. **Iterator takes the place of Enumeration in the Java Collections Framework.** Iterators differ from enumerations in two ways:
>    - Iterators allow the caller to remove elements from the underlying collection during the iteration with well-defined semantics.
>    - Method names have been improved.

PR #497, opened 2020-10-29
nil

plotnick:
- [x] Python: `nil` = `None`
- [x] Ruby: `nil` = `nil`
- [x] Java: `nil` = `null`
- [x] JS: `nil` = `null`
- [ ] ~Rust: `nil` = `Option<T>::None`~ To be handled as a follow-up PR

PR checklist:

- [x] Added changelog entry.
- [x] Docs


languages/python/django-oso/tests/test_django_oso.py
```diff
@@ -133,28 +133,38 @@ def test_route_authorization(client, settings, simple_policy):
...
-    Post(name="test_public", is_private=False, timestamp=1).save()
-    Post(name="test_private", is_private=True, timestamp=1).save()
-    Post(name="test_private_2", is_private=True, timestamp=1).save()
+    posts = [
+        Post(name="test", is_private=False, timestamp=1).save(),
+        Post(name="test_past", is_private=False, timestamp=-1).save(),
+        Post(name="test_public", is_private=False, timestamp=1).save(),
+        Post(name="test_private", is_private=True, timestamp=1).save(),
+        Post(name="test_private_2", is_private=True, timestamp=1).save(),
+        Post(name="test_option", is_private=True, timestamp=0, option=True).save(),
```
gj, line 142:
```suggestion
        Post(name="test_option", is_private=False, timestamp=1, option=True).save(),
```
wouldn't this assure us that this Post isn't turned up by the authorize_filter on 148 b/c of `option` not being `nil` (as opposed to `is_private` being `True` or `timestamp` not being greater than `0`)?

gj (commented):
Generally LGTM

languages/python/django-oso/tests/test_django_oso.py
plotnick: Yup, thanks.

docs/changelogs/vNEXT.rst
```diff
@@ -23,14 +23,21 @@ Link to migration guide
...
-
-- summary
-- of
-- user facing changes
-
-Link to relevant documentation section
+``nil``
+-------
+
+Polar now pre-defines `a constant named ``nil``
```
samscott89, line 29: Tiniest of nits, but isn't it "predefined" and not hyphenated? Also, is it predefined or just defining?

docs/getting-started/policies/application-types.rst
```diff
@@ -555,6 +555,20 @@ see the :doc:`/using/libraries/index` guides.
   Literals in Polar are constant, and any changes made to such objects
   by calling a method will not be persisted.
 
+``nil``
```
dhatch, line 558: Should we also have a section for this in each language library page? I think someone could be looking for it in either place.

languages/java/oso/pom.xml
```diff
@@ -71,6 +71,15 @@
 
   <build>
   <plugins>
+    <plugin>
+       <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
```
gj, line 76: Why's this here? Looks related to testing?

docs/changelogs/vNEXT.rst
gj: I think it's plain ole' defining

```suggestion
Polar now defines `a constant named ``nil``
```

docs/getting-started/policies/application-types.rst
gj: Yeah, I think we should add it to the language library pages since we mention other type translations there.

docs/changelogs/vNEXT.rst
plotnick: 312bb23

docs/getting-started/policies/application-types.rst
plotnick: f676bab

languages/java/oso/pom.xml
plotnick: Yes, it's to make the Java tests spit out stack traces instead of silently swallowing them so you can't see what actually failed.

docs/using/libraries/java/index.rst
```diff
@@ -212,18 +217,18 @@ Likewise, dictionaries constructed in Polar may be passed into Java methods.
...
 ^^^^^^^^^^^^
-Oso handles Java objects that implement the `Enumeration <https://docs.oracle.com/javase/10/docs/api/java/util/Enumeration.html>`_ interface by evaluating each of the
-object's elements one at a time:
+You may iterate over a Java `Enumeration <https://docs.oracle.com/javase/10/docs/api/java/util/Enumeration.html>`_
+using the Polar :ref:`operator-in` operator:
 
 .. code-block:: polar
    :caption: :fa:`oso` policy.polar
 
-   allow(actor, action, resource) if actor.getGroup = "payroll";
```
samscott89, line 221: Ah crap, good catch! My regex failed me since I was only looking for `<blah> = <x>.<y>`

docs/using/libraries/ruby/index.rst
```diff
@@ -34,7 +34,8 @@ the ``#register_class`` method. An example of this can be found :ref:`here <appl
 
 Numbers and Booleans
 ^^^^^^^^^^^^^^^^^^^^
-Polar supports both integer and floating point numbers, as well as booleans (see :ref:`basic-types`).
+Polar supports integer and floating point real numbers, as well as booleans (see :ref:`basic-types`).
+These map to the Ruby ``Integer``, ``Float``, ``TrueClass``, and ``FalseClass`` types.
```
gj, line 38: nit: I think this mirrors the "integer and floating point real numbers, as well as booleans" line above

```suggestion
These map to the Ruby ``Integer``, ``Float``, and ``TrueClass``/``FalseClass`` types.
```

docs/using/libraries/ruby/index.rst
```diff
@@ -113,6 +114,11 @@ Likewise, lists constructed in Polar may be passed into Ruby methods:
   user = User.new(["HR", "payroll"])
   raise "should be allowed" unless oso.allowed?(user, "foo", "bar")
 
+There is currently no syntax for random access to a list element within a policy;
+i.e., there is no Polar equivalent of the Ruby expression ``user.groups[1]``.
+To access the elements of a list, you may iterate over it with :ref:`operator-in`
+or destructure it with :ref:`pattern matching <patterns-and-matching>`.
```
gj, line 120: Might be worth mentioning that you could also access it as `user.groups.at(1)`?

docs/using/libraries/ruby/index.rst
```diff
@@ -140,43 +146,39 @@ Likewise, dictionaries constructed in Polar may be passed into Ruby methods.
 
 Enumerators
```
gj, line 147: This section (and the below comment) should probably talk about `Enumerable` instead of `Enumerator`. The changes in #485 (I think) make it such that you can enumerate over any enumerable (strings, lists, etc., in addition to explicit enumerators) with the `in` operator.

docs/using/libraries/ruby/index.rst
```diff
@@ -140,43 +146,39 @@ Likewise, dictionaries constructed in Polar may be passed into Ruby methods.
...
-  allow(actor, action, resource) if actor.get_group() = "payroll";
+  allow(actor, action, resource) if "payroll" in actor.get_groups();
 
 .. code-block:: ruby
   :caption: :fas:`gem` app.rb
 
   class User
-    def get_group(self)
+    def get_groups(self)
       ["HR", "payroll"].to_enum
```
gj, line 162: I don't think we need this `to_enum` call anymore. See: previous comment

docs/using/libraries/java/index.rst
```diff
@@ -212,18 +217,18 @@ Likewise, dictionaries constructed in Polar may be passed into Java methods.
 
 Enumerations
 ^^^^^^^^^^^^
-Oso handles Java objects that implement the `Enumeration <https://docs.oracle.com/javase/10/docs/api/java/util/Enumeration.html>`_ interface by evaluating each of the
-object's elements one at a time:
+You may iterate over a Java `Enumeration <https://docs.oracle.com/javase/10/docs/api/java/util/Enumeration.html>`_
```
gj, line 220: Not sure how the Java bits of #485 are implemented, but (assuming they're the same as the Ruby bits I looked at) can't we now enumerate over anything we can convert into an enumeration? I.e., the below `return Collections.enumeration(List.of("HR", "payroll"));` could be `return List.of("HR", "payroll");`?

gj (commented):
Same comments from the Ruby + Java docs updates apply to the other language library docs updates -- just didn't want to keep repeating myself 😄

docs/using/libraries/ruby/index.rst
gj: This is a general comment that we're saying "there's no syntax for index access, but you can iterate over all elements of a list" instead of ", ...but you can use List methods defined in the host language to access an element at a particular position," which seems much more analogous to index access than does enumerating with `in`.

Applies to all language library docs -- not just Ruby.

docs/using/libraries/ruby/index.rst
plotnick: > nit: I think this mirrors the "integer and floating point real numbers, as well as booleans" line above

Sorry, what line where?

docs/using/libraries/ruby/index.rst
plotnick: bd8bea72

docs/using/libraries/ruby/index.rst
plotnick: bd8bea72

docs/using/libraries/java/index.rst
plotnick: There are several cases: collections, iterables, enumerations. As far as I can tell, the others reduce to the last, so I think it's fine.

plotnick: eb8ec10

docs/using/libraries/ruby/index.rst
gj: Line 37: integer and floating point real numbers, as well as booleans
Line 38: These map to the Ruby `Integer`, `Float`, and `TrueClass`/`FalseClass` types.

gj (approved):

PR #498, opened 2020-10-30
Clean up / standardize integration tests

gj:
All of this is purely cosmetic -- no actual tests have been harmed.

- Moved `assert[_eq]!()` calls inside of `q[null,eval,var,etc]()` to avoid repeating it a million times.
- Moved the "logic" of `q[null,eval,var,etc]` into separate `[null,eval,var,etc]` functions. The `q-` variants call the bare ones and then `assert[_eq]!()` as appropriate. This is because there were a few nonstandard uses of the `q-` variants.
- Passed the result len into `qext()` since that's all we ever assert on anyway.
- Used the raw string (`r#""#`) syntax every time we have multiple rules instead of `putting(); them(); all(); on(); one(); line();`.
- Added a `values!()` macro that turns `values![1, 2, 3]` into `vec![value!(1), value!(2), value!(3)]` and `values![[1, 2], [3, 4]]` into `vec![vec![value!(1), value!(2)], vec![value!(3), value!(4)]]`. A ton of the "expected" values in our test assertions are of this form.
- Added `qruntime!()` to assert that running a query raises a certain `RuntimeError` and `qparse!()` to assert that loading a Polar snippet raises a certain `ParseError`.
- Return `Result<(), PolarError>` from tests that `.unwrap()` `PolarResult`s so we can use the `?` operator.

### DIFF
--- a/polar-core/tests/integration_tests.rs
+++ b/polar-core/tests/integration_tests.rs
@@ -197,15 +197,27 @@ fn query_results_with_externals(query: Query) -> (QueryResults, MockExternal) {
 }
 
 #[track_caller]
-fn qeval(p: &mut Polar, query_str: &str) {
+#[must_use = "test results need to be asserted"]
+fn eval(p: &mut Polar, query_str: &str) -> bool {
     let q = p.new_query(query_str, false).unwrap();
-    assert!(!query_results!(q).is_empty());
+    !query_results!(q).is_empty()
+}
+
+#[track_caller]
+fn qeval(p: &mut Polar, query_str: &str) {
+    assert!(eval(p, query_str));
+}
+
+#[track_caller]
+#[must_use = "test results need to be asserted"]
+fn null(p: &mut Polar, query_str: &str) -> bool {
+    let q = p.new_query(query_str, false).unwrap();
+    query_results!(q).is_empty()
 }
 
 #[track_caller]
 fn qnull(p: &mut Polar, query_str: &str) {
-    let q = p.new_query(query_str, false).unwrap();
-    assert!(query_results!(q).is_empty());
+    assert!(null(p, query_str));
 }
 
 #[track_caller]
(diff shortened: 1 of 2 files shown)


polar-core/tests/integration_tests.rs
```diff
@@ -249,34 +264,78 @@ fn qvars(polar: &mut Polar, query_str: &str, vars: &[&str]) -> Vec<Vec<Value>> {
...
-    assert!(qnull(&mut polar, "k(1)"));
-    assert!(qeval(&mut polar, "k(2)"));
-    assert!(qnull(&mut polar, "k(3)"));
-    assert_eq!(qvar(&mut polar, "k(a)", "a"), vec![value!(2)]);
+macro_rules! qruntime {
+    ($query:tt, $err:pat) => {
+        assert!(matches!(_qruntime(&mut Polar::new(), $query), ErrorKind::Runtime($err)));
+    };
+
+    ($query:tt, $err:pat, $cond:expr) => {
```
samscott89, line 285: TIOLI: You could make these conditional tokens, something like:

```suggestion
    ($query:tt, $err:pat, $($cond:expr)?) => {
```

polar-core/tests/integration_tests.rs
```diff
@@ -249,34 +264,78 @@ fn qvars(polar: &mut Polar, query_str: &str, vars: &[&str]) -> Vec<Vec<Value>> {
...
-fn test_jealous() {
-    let polar = Polar::new();
-    polar
-        .load_str(
-            r#"loves("vincent", "mia");
-               loves("marcellus", "mia");
-               jealous(a, b) if loves(a, c) and loves(b, c);"#,
-        )
-        .unwrap();
+fn test_functions() -> TestResult {
```
samscott89, line 311: I never knew you could do this and then use `?`. Sick

samscott89: Probably not adding much to the functions that dont need it though.

samscott89 (approved):
This looks like a huge improvement in the test writing experience.

PR #499, opened 2020-10-30
Error on binding expressions & patterns to variables

gj:
PR checklist:

- [ ] Added changelog entry.

### DIFF
--- a/polar-core/tests/integration_tests.rs
+++ b/polar-core/tests/integration_tests.rs
@@ -1787,3 +1787,19 @@ fn test_list_matches() {
         vec![value!([3, Value::RestVariable(Symbol::new("ys"))])],
     );
 }
+
+#[test]
+fn error_on_binding_expressions_and_patterns_to_variables() -> TestResult {
+    qruntime!("x = (1 and 2)", RuntimeError::TypeError { msg: m, .. }, m == "cannot bind expression '1 and 2' to 'x'");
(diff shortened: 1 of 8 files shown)


polar-core/tests/integration_tests.rs
```diff
@@ -1787,3 +1787,19 @@ fn test_list_matches() {
...
+    qruntime!("x matches y", RuntimeError::TypeError { msg: m, .. }, m == "cannot bind pattern 'y' to 'x'");
+    let mut p = Polar::new();
+    p.load_str(
+        r#"f(x: y) if x = 1;
+           g(x: {}) if x = 1;"#,
+    )?;
+    qruntime!(&mut p, "f(x)", RuntimeError::TypeError { msg: m, .. }, m == "cannot bind pattern 'y' to '_x_1'");
+    qruntime!(&mut p, "g(x)", RuntimeError::TypeError { msg: m, .. }, m == "cannot bind pattern '{}' to '_x_2'");
+    Ok(())
+}
```
saolsen, line 1805: Did you try any with the var on the right hand side of the unify? I think it'll be fine because of the way args are sorted before calling unify_var but it would be good to check.

saolsen (approved):

polar-core/tests/integration_tests.rs
gj: f428853

gj:
@samscott89 holding off on the changelog for now b/c there will likely be some messaging overlap between this and #503.

PR #500, opened 2020-10-30
Use GTM

gj:


samscott89 (approved):
Worked locally. Verified with chrome + google tag assitant

PR #501, opened 2020-10-30
Update GitHub Actions triggers

gj:
Development + Rust benchmark workflows now run on:
- [re-]opening a new PR (including from a fork)
- pushing code to a branch associated with an open PR (including from a fork)
- pushing a commit containing `[test]` and/or `[bench]` to a base repo branch (no forks) not associated with an open PR

Note that if you have an open PR in the base repo and you push a commit to it containing `[test]` and/or `[bench]`, CI will run twice (in parallel) for the two matching events (`synchronize` of a `pull_request_target` and matching `[test]`/`[bench]` in the commit message of a `push` event).

I took 5 minutes to try to figure out if we could avoid that duplication entirely, but we would have to check if at least one open PR contains the commit in question (e.g., something like https://github.com/jwalton/gh-find-current-pr). Since that Action wouldn't do it out of the box, it seems easier to just collectively remember that you don't need to add `[test]` / `[bench]` if you already have an open PR.

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -1,7 +1,5 @@
 name: Rust benchmark
-on:
-  push:
-    branches:
+on: [pull_request_target]
 
 jobs:
   benchmark:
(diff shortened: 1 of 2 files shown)


samscott89 (approved):
🌎

PR #503, opened 2020-10-30
Parser macro: value or logical

samscott89:
PR checklist:

- [ ] Added changelog entry.

### DIFF
--- a/languages/rust/oso/tests/test_polar_rust.rs
+++ b/languages/rust/oso/tests/test_polar_rust.rs
@@ -4,8 +4,7 @@
 use maplit::hashmap;
 use thiserror::Error;
 
-use oso::{ClassBuilder, OsoError, PolarClass};
-use polar_core::error::{ErrorKind, PolarError, RuntimeError};
+use oso::{ClassBuilder, PolarClass};
 
 mod common;
 
--- a/languages/rust/oso/tests/test_polar_rust.rs
+++ b/languages/rust/oso/tests/test_polar_rust.rs
@@ -453,13 +453,6 @@ fn test_unify_externals() {
 
     test.oso.register_class(bar_class).unwrap();
 
-    let mut query = test.oso.query("x = new Bar(1) = new Bar(2)").unwrap();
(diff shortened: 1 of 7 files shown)


languages/rust/oso/tests/test_polar_rust.rs
```diff
@@ -454,13 +454,6 @@ fn test_unify_externals() {
 
     test.oso.register_class(bar_class).unwrap();
 
-    let mut query = test.oso.query("x = new Bar(1) = new Bar(2)").unwrap();
```
samscott89, line 457: Deleting these tests because they are now covered by parser tests: `x = y = z` parses to `(x = y)  = z` and you cannot unify a variable with an expression.

polar-core/src/polar.lalrpop
```diff
@@ -475,16 +519,18 @@ pub Pattern: Value = {
...
-    <BuiltinOperation>,
-    <RewrittenOperation>,
-    <Call>,
-    <List<"Term">>,
+
+pub Value: ValueOrLogical = {
+    <IsLogical<BuiltinOperation>>,
+    <IsAny<Boolean>>,
+    <IsAny<Variable>>,
+    <IsAny<Call>>,
```
saolsen, line 527: Why are these `IsAny`? Wouldn't a variable always be a value and a call always be a logical?

saolsen (commented):
This is pretty cool, really nice to catch this kind of error early. Just had one question.

polar-core/src/polar.lalrpop
samscott89: Variable can be a boolean which can be queried. E.g. 
```
x = true and x
```
You might be right for `Call`, although I think we maybe support doing `f(x) = f(1)` to unify to `x = 1`?

saolsen: ah yeah, that call unify thing keeps coming up, I don't like it but I guess that's ok

saolsen (approved):

PR #505, opened 2020-10-31
Update dependencies

samscott89:

### DIFF
--- a/polar-c-api/Cargo.toml
+++ b/polar-c-api/Cargo.toml
@@ -11,8 +11,7 @@ bench = false
 
 [dependencies]
 polar-core = { path = "../polar-core", version = "0.7.1" }
-serde = { version = "1.0.116", features = ["derive", "rc"] }
-serde_json = "1.0.58"
+serde_json = "1.0.59"
 
 [build_dependencies]
 cbindgen = "0.15.0"
--- a/polar-core/Cargo.toml
+++ b/polar-core/Cargo.toml
@@ -17,17 +17,14 @@ name = "bench"
 harness = false
 
 [dependencies]
-anyhow = "1.0.33"
-js-sys = "0.3.45"
 lalrpop-util = "0.19.1"
-lazy_static = "1.4.0"
-regex = "1.4.0"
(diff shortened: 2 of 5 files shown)


polar-core/Cargo.toml
```diff
@@ -17,17 +17,14 @@ name = "bench"
...
-regex = "1.4.0"
-serde = { version = "1.0.116", features = ["derive", "rc"] }
-serde_json = "1.0.58"
+regex = "1.4.1"
+serde = { version = "1.0.117", features = ["derive", "rc"] }
+serde_json = "1.0.59"
 
 [build_dependencies]
 serde_derive = "1.0"
-lalrpop = { version = "0.19.1", features = ["lexer"] }
```
samscott89, line 30: We use our own lexer.

Which also means we no longer need lazy static for the parser, since there are no regexes to compile on init.

polar-core/Cargo.toml
```diff
@@ -38,5 +35,8 @@ pretty_assertions = "0.6.1"
 maplit = "1.0.2"
 indoc = "1.0.3"
 
+[target.wasm32.dependencies]
+js-sys = "0.3.45"
```
samscott89, line 39: We were building 20-30 dependencies for this on non-wasm builds

polar-c-api/Cargo.toml
```diff
@@ -11,8 +11,7 @@ bench = false
 
 [dependencies]
 polar-core = { path = "../polar-core", version = "0.7.1" }
-serde = { version = "1.0.116", features = ["derive", "rc"] }
-serde_json = "1.0.58"
+serde_json = "1.0.59"
```
gj, line 14: were we not using the `derive` + `rc` features in this crate? trying to square this change w/ the change in `polar-core/Cargo.toml`, where we kept `serde` and `serde_json` as explicit deps.

gj (approved):
One little question, but LGTM

polar-c-api/Cargo.toml
gj: Talked about this offline. We weren't using serde at all.

PR #507, opened 2020-11-02
Minor follow-up to #480 & #484

plotnick:
* Rename `fold_symbol` back to `fold_name`, used for non-variable symbols in name position.
* Scan all goals on VM init for queries.
* Make a currently-unreachable case an error.

### DIFF
--- a/polar-core/src/folder.rs
+++ b/polar-core/src/folder.rs
@@ -39,8 +39,9 @@ pub trait Folder: Sized {
     fn fold_instance_id(&mut self, i: u64) -> u64 {
         fold_instance_id(i, self)
     }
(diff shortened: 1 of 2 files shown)


dhatch (approved):
looks good, but I still want to discuss this comment from last time: https://github.com/osohq/oso/pull/484#discussion_r514412204

Maybe we can chat about it quickly tomorrow AM.

gj:
> looks good, but I still want to discuss this comment from last time: [#484 (comment)](https://github.com/osohq/oso/pull/484#discussion_r514412204)
> 
> Maybe we can chat about it quickly tomorrow AM.

This change is in #487

PR #509, opened 2020-11-04
Add benchmarks for not

gj:


samscott89 (approved):
LGTM.

Lets see if it works fine on pushing to main... (I have no reason to believe it wont)

PR #511, opened 2020-11-05
Remove csp from not implementation.

dhatch:
PR checklist:

- [ ] Added changelog entry.

### DIFF
--- a/polar-core/src/partial/constraints.rs
+++ b/polar-core/src/partial/constraints.rs
@@ -178,12 +178,7 @@ impl IsaConstraintCheck {
 }
 
 impl Runnable for IsaConstraintCheck {
-    fn run(
-        &mut self,
-        _: Option<&mut BindingStack>,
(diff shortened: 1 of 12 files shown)


polar-core/src/partial/constraints.rs
```diff
@@ -802,7 +816,55 @@ mod test {
...
+        // This output only has _this matches User{}, because we immediately judge x matches NotUser{} to be
+        // incompatible (so the not fails before reaching the and).
+        assert_partial_expression!(next_binding(&mut query)?, "a", "_this matches User{}");
+
+        Ok(())
+    }
+
+    #[test]
+    #[ignore]
+    // TODO (dhatch): remove ignore after IN can be inverted.
```
gj, line 851: Is this comment still applicable? I don't see any `in`s in the below test, and we do have tests that ensure we can invert `in`s.

PR #513, opened 2020-11-06
Fix inconsistent removal of single item ands / ors

dhatch:
- Simplifier always operates over partials
- A separate stage converts partials to expressions
- Constraints::into_expression doesn't add an AND if there is
  only one constraint.

### DIFF
--- a/polar-core/src/partial/simplify.rs
+++ b/polar-core/src/partial/simplify.rs
@@ -15,10 +15,9 @@ pub fn simplify_bindings(bindings: Bindings) -> Bindings {
         .into_iter()
         .map(|(var, value)| match value.value() {
             Value::Partial(_) => {
-                let mut simplified = simplify_partial(value);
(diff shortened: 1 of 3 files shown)


polar-core/src/partial/simplify.rs
```diff
@@ -28,8 +27,20 @@ pub fn simplify_bindings(bindings: Bindings) -> Bindings {
 
 pub struct Simplifier;
 impl Folder for Simplifier {
-    fn fold_term(&mut self, t: Term) -> Term {
-        fn maybe_unwrap_operation(o: &Operation) -> Option<Term> {
+    /// Deduplicate constraints.
+    fn fold_constraints(&mut self, c: Constraints) -> Constraints {
```
dhatch, line 31: This diff looks like a bigger change than it is.  I just moved maybe_unwrap_operation into `fold_operation` and removed the case for eliding a single item partial.

dhatch:
Covering in #514

PR #514, opened 2020-11-06
Add in & contains operations to Constraints.

dhatch:


polar-core/src/partial/constraints.rs
```diff
@@ -97,6 +97,40 @@ impl Constraints {
         constraint_check
     }
 
+    /// Add a constraint that this must contain some known value.
+    ///
+    /// From OTHER in THIS where other is not a partial or variable.
+    pub fn contains(&mut self, other: Term) {
+        // TODO for now, this is represented as an in operation, but we may
+        // want some other representation eventually.
+        // TODO what about non-ground compound terms like [x, 1] in THIS
```
gj, line 106: Why would we treat `[x, 1] in THIS` any differently from other non-ground terms?

polar-core/src/partial/constraints.rs
```diff
@@ -701,18 +753,90 @@ mod test {
...
+
+    #[test]
+    fn test_contains_partial() -> TestResult {
+        let p = Polar::new();
+        p.load_str(
+            r#"
+            contains(x, y) if x in y;
+            contains_dot(x, y) if x in y.foo;
+            contains_dot_dot(x, y) if x in y.foo.bar and y.foo = 2;
+        "#,
```
gj, line 774: matches the formatting elsewhere

```suggestion
            r#"contains(x, y) if x in y;
               contains_dot(x, y) if x in y.foo;
               contains_dot_dot(x, y) if x in y.foo.bar and y.foo = 2;"#,
```

polar-core/src/partial/constraints.rs
```diff
@@ -701,18 +753,90 @@ mod test {
...
+
+    #[test]
+    fn test_in_partial() -> TestResult {
+        let p = Polar::new();
+        p.load_str(
+            r#"
+            f(x) if y in x.values;
+            g(x, y) if y in x.values;
+            h(x) if y in x.values and (y.bar = 1 and y.baz = 2) or y.bar = 3;
+        "#,
```
gj, line 805:
```suggestion
            r#"f(x) if y in x.values;
               g(x, y) if y in x.values;
               h(x) if y in x.values and (y.bar = 1 and y.baz = 2) or y.bar = 3;"#,
```

polar-core/src/partial/constraints.rs
```diff
@@ -701,18 +753,90 @@ mod test {
...
+        )?;
+
+        let mut q = p.new_query_from_term(term!(call!("f", [partial!("a")])), false);
+        // TODO (dhatch): This doesn't work now, but ultimately this should have
+        // no constraints since nothing is added to `y`.
+        assert_partial_expressions!(
+            next_binding(&mut q)?,
+            "a" => "() in _this.values",
```
gj, line 813: Where is this `()` coming from?

polar-core/src/partial/constraints.rs
```diff
@@ -701,18 +753,90 @@ mod test {
...
+        );
+        assert_query_done!(q);
+
+        let mut q = p.new_query_from_term(term!(call!("h", [partial!("a")])), false);
+        assert_partial_expressions!(
+            next_binding(&mut q)?,
+            "a" => "(_this.bar = 1 and _this.baz = 2) in _this.values"
+        );
+
+        // TODO (dhatch): Why is this an expression, but above is a partial?
```
gj, line 835: Is this comment still relevant?

polar-core/src/partial/constraints.rs
dhatch: Discussed offline - the reason is translation is different on the adapter side, so I want this to be a separate concept.  The representation of constraints as operations is potentially only temporary, and once we have a better concept we can encode this as a separate constraint type which will make writing adaptors easier.

The downside of having a different type for constraints is translating back into a Polar expression would be difficult.  But, the VM can't evaluate the expressions we work with after simplification as is, so I don't think this is much different than our current state.

gj (approved):

PR #515, opened 2020-11-06
Implement iterator + nil for Rust

samscott89:
This also includes some refactoring of the `PolarValue` interface, and makes it so that fewer of the polar-core details leaks into the Rust oso implementation.

### `2765d50 `

Made `PolarValue` the intermediate type consistently. So now `ToPolar` and `FromPolar` both convert to/from `PolarValue` and Rust types.

This means the part which needs the `host`, which does `PolarValue -> polar_core::Term` mostly happens within `query.rs`.

Which means that we can purge uses of `Term` throughout the internal Rust oso implementation.

### `96fc39b`

Implemented the iterator changes from core.

Classes now have an explicit method to convert to an iterator. So we can revert to just returning one result for every lookup. Using the `in` operator will attempt to convert the type into an iterator.


languages/rust/oso/src/host/to_polar.rs
```diff
@@ -31,40 +31,41 @@ use crate::PolarValue;
...
-            instance_id: instance,
-        })
+    fn to_polar(self) -> PolarValue {
+        PolarValue::new_from_instance(self)
     }
+}
+
+impl<C: crate::PolarClass + Send + Sync> ToPolar for C {
+    fn to_polar(self) -> PolarValue {
+        DEFAULT_CLASSES
```
samscott89, line 41: This is the worst part of the change. We used to be able to just in time register the class on the host. But now we have this `T -> PolarValue -> Term` two step process, we lose the type in step 1, but only have the host in step 2.

This keeps the old behaviour (classes work without registering them), but introduces this global lazy static to make that work.

dhatch: What was wrong with passing host into each of these conversion functions? I haven't spend that much time reasoning through this, but would a global behave correctly if there are multiple instances of oso? We have tried to remove uses of globals in most of the other language libraries.

dhatch:
@samscott89 can you summarize the changes to make this easier to review?

languages/rust/oso/src/host/class.rs
```diff
@@ -184,6 +194,34 @@ where
         self
     }
 
+    /// Set an equality function to be used for polar `==` statements.
```
dhatch, line 197: Comment incorrect.

languages/rust/oso/src/host/from_polar.rs
```diff
@@ -30,52 +33,168 @@ use crate::errors::TypeError;
 ///
 /// `FromPolar` implementors must also be concrete, sized types without
```
dhatch, line 34: This comment is no longer correct.

languages/rust/oso/src/host/class.rs
```diff
@@ -231,8 +269,9 @@ where
         Args: FromPolarList,
```
dhatch, line 269: Should we remove `add_iterator_method`?  We no longer allow this multiple result behavior without `in`, right?

dhatch: Other question, should we remove the ability for single calls to return multiple results without the use of the `NextExternal` event.

languages/rust/oso/src/host/class.rs
```diff
@@ -73,12 +87,7 @@ impl Class {
     /// Call class method `attr` on `self` with arguments from `args`.
     ///
     /// Returns: An iterable of results from the method.
-    pub fn call(
```
dhatch, line 76: Comment is now incorrect.

languages/rust/oso/src/query.rs
```diff
@@ -133,34 +138,39 @@ impl Query {
...
-            } else {
-                Box::new(std::iter::once(instance.get_attr(&name.0, &mut self.host)))
-            };
-            self.calls.insert(call_id, results);
+            let instance = Instance::from_polar(PolarValue::from_term(&iterable, &self.host)?)?;
+            let iter = instance
+                .downcast::<crate::host::PolarIterator>(Some(&self.host))
+                .map(|i| Ok(i.clone()))
+                .unwrap_or_else(|_| instance.as_iter(&self.host))?;
+            self.calls.insert(call_id, iter);
```
dhatch, line 163: So `self.calls` is now only used for iterators, right? I haven't read the implementation in other language libraries, but should we rename these instance variables so that the name reflects the functionality?

languages/rust/oso/src/host/class.rs
samscott89: Not sure I follow, here's what's happening here:

`add_iterator_method` ensures that the result is iterable, and wraps it in the `PolarIterator` struct. This struct acts as a registered class, and `in` works on this class.

(Mid-sentence realised this could be better. Will make it so that `PolarIterator` also implements `as_iter`, so that we can remove the manual downcasting.)

It's already true that multiple results can only be retrieved using `in`. You can do `x = foo.iter_method()` to get back the PolarIterator, but all you can do with this class is basically `y in x or z in x`.

languages/rust/oso/src/query.rs
samscott89: Correct. Rename what instance variables?

samscott89: Ohh, like `instance`? Yeah, we can make it `iterable` for example, although that is what `as_iter` is checking.

languages/rust/oso/src/host/to_polar.rs
samscott89: I think it was fine, as long as this was a fully internal trait. But I think the really cool thing we get by having an intermediate `PolarValue` type, is that it's actually reasonable for somebody to implement `ToPolar` who doesn't need to understand any of the internal details. E.g. you might have a newtype `NonZero(u32)` that you want to implement to/from polar for. Not needing to know what the hell `Host` is seems like a big win.

dhatch (approved):
lgtm, I don't follow why we still need the `add_iterator_method`, but can handle offline / after this PR

PR #524, opened 2020-11-13
A Partial has constraints

gj:
- Rename `Constraints` -> `Partial`
- Rename `partial.operations` -> `partial.constraints`


dhatch:
This was the original naming scheme when we started.   I'm trying to recall why we moved to constraints -- I think it's inline with the idea of having `Value::Partial(Constraints)` and `Value::Variable(Symbol)`.  There may be other types that ultimately use constraints besides a partial value.  One example is the discussion around unbound variables having constraints, which would make partials part of the normal execution path.

I'm not necessarily opposed to this, just adding some color.  Could you explain your rationale?

Has @plotnick seen / discussed this?

dhatch (approved):
lgtm

PR #530, opened 2020-11-17
Add before_compile hook API for sqlalchemy.

dhatch:
PR checklist:

- [ ] Added changelog entry.

### DIFF
--- a/languages/python/sqlalchemy-oso/sqlalchemy_oso/auth.py
+++ b/languages/python/sqlalchemy-oso/sqlalchemy_oso/auth.py
@@ -6,13 +6,13 @@ from sqlalchemy.orm.session import Session
 from sqlalchemy.orm.util import class_mapper
 from sqlalchemy.sql import expression as sql
 
-from sqlalchemy_oso.partial import partial_to_query
+from sqlalchemy_oso.partial import partial_to_filter
 
 
-def null_query(session: Session) -> Query:
+def null_query(session: Session, model) -> Query:
     """Return an intentionally empty query."""
     # TODO (dhatch): Make this not hit the database.
-    return session.query(sql.false()).filter(sql.false())
+    return session.query(model).filter(sql.false())
 
 
 def register_models(oso: Oso, base):
@@ -38,24 +38,46 @@ def authorize_model(oso: Oso, actor, action, session: Session, model) -> Query:
     :param session: The SQLAlchemy session.
     :param model: The model to authorize, must be a SQLAlchemy model.
     """
-    # TODO (dhatch): More robust name mapping?
+    filters = authorize_model_filter(oso, actor, action, session, model)
+    if filters is None:
+        return session.query(model)
 
+    return session.query(model).filter(filters)
+
+
+def authorize_model_filter(oso: Oso, actor, action, session: Session, model):
+    """Return SQLAlchemy expression that applies the policy to ``model``.
+
+    Executing this query will return only authorized objects. If the request is
+    not authorized, a query that always contains no result will be returned.
+
+    :param oso: The oso class to use for evaluating the policy.
+    :param actor: The actor to authorize.
+    :param action: The action to authorize.
+
+    :param session: The SQLAlchemy session.
+    :param model: The model to authorize, must be a SQLAlchemy model.
+    """
+    # TODO (dhatch): Check that model is a model.
+    # TODO (dhatch): More robust name mapping?
     assert class_mapper(model), f"Expected a model; received: {model}"
 
     partial_resource = Partial("resource", TypeConstraint(model.__name__))
     results = oso.query_rule("allow", actor, action, partial_resource)
 
-    combined_query = None
+    combined_filter = None
+    has_result = False
     for result in results:
+        has_result = True
+
         resource_partial = result["bindings"]["resource"]
-        query = partial_to_query(resource_partial, session, model)
-        if combined_query is None:
-            combined_query = query
-        else:
-            # TODO (dhatch): Use OR here?
-            combined_query = combined_query.union(query)
+        filter = partial_to_filter(resource_partial, session, model)
+        if combined_filter is None:
+            combined_filter = filter
+        elif filter is not None:
+            combined_filter = combined_filter | filter
 
-    if combined_query is None:
-        return null_query(session)
+    if not has_result:
+        return sql.false()
 
-    return combined_query
+    return combined_filter
--- a/languages/python/sqlalchemy-oso/sqlalchemy_oso/hooks.py
+++ b/languages/python/sqlalchemy-oso/sqlalchemy_oso/hooks.py
@@ -1,0 +1,154 @@
+"""SQLAlchemy hooks that transparently enable oso on SQLAlchemy operations.
+
+There are several potential interfaces to integrate oso with SQLAlchemy:
+
+    - :py:func:`authorized_sessionmaker`: (**recommended**) Session factory that
+       creates a session that applies authorization on every query.
+    - :py:func:`enable_hooks`: Globally enable oso on all queries.
+    - :py:func:`make_authorized_query_cls`: Make a query class that is
+       authorized before execution.
+
+.. note::
+
+    If using any API besides :py:func:`authorized_sessionmaker`, ensure you set
+    ``enable_baked_queries=False`` on the session. Query caching can interfere
+    with authorization.
+
+    It is recommended to scope authorization context (the oso instance, user and
+    action) to a single session.  Otherwise, the identity map (SQLAlchemy's
+    cache of retrieved objects) may contain objects that were authorized for a
+    previous user. This could cause incorrect behavior.
+
+    :py:func:`authorized_sessionmaker` will enforce this.  If the authorization
+    context changes during the session, an Exception will be raised.
+"""
+import functools
+from typing import Any, Callable
+
+from sqlalchemy.event import listen, remove
+from sqlalchemy.orm.query import Query
+from sqlalchemy.orm import aliased, sessionmaker
+
+from oso import Oso
+
+from sqlalchemy_oso.auth import authorize_model, authorize_model_filter
+
+
+def enable_hooks(
+    get_oso: Callable[[], Oso],
+    get_user: Callable[[], Any],
+    get_action: Callable[[], Any],
+    target=None,
+):
+    """Enable all SQLAlchemy hooks."""
+    if target is None:
+        target = Query
+
+    return enable_before_compile(target, get_oso, get_user, get_action)
+
+
+def enable_before_compile(
+    target,
+    get_oso: Callable[[], Oso],
+    get_user: Callable[[], Any],
+    get_action: Callable[[], Any],
+):
+    """Enable before compile hook."""
+    auth = functools.partial(
+        authorize_query, get_oso=get_oso, get_user=get_user, get_action=get_action
+    )
+
+    listen(target, "before_compile", auth, retval=True)
+
+    return lambda: remove(target, "before_compile", auth)
+
+
+def authorize_query(query: Query, get_oso, get_user, get_action) -> Query:
+    """Authorize an existing query with an oso instance, user and action."""
+    oso = get_oso()
+    action = get_action()
+    actor = get_user()
+
+    # TODO (dhatch): This is necessary to allow ``authorize_query`` to work
+    # on queries that have already been made.  If a query has a LIMIT or OFFSET
+    # applied, SQLAlchemy will by default throw an error if filters are applied.
+    # This prevents these errors from occuring, but could result in some
+    # incorrect queries. We should remove this if possible.
+    query = query.enable_assertions(False)
+
+    entities = {column["entity"] for column in query.column_descriptions}
+    for entity in entities:
+        # Only apply authorization to columns that represent a mapper entity.
+        if entity is None:
+            continue
+
+        authorized_filter = authorize_model_filter(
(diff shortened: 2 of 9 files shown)


languages/python/sqlalchemy-oso/sqlalchemy_oso/auth.py
```diff
@@ -5,23 +5,25 @@
...
 from sqlalchemy.sql import expression as sql
 
-from sqlalchemy_oso.partial import partial_to_query
+from sqlalchemy_oso.partial import partial_to_query, partial_to_filter
 
-def null_query(session: Session) -> Query:
+
+def null_query(session: Session, model) -> Query:
     """Return an intentionally empty query."""
     # TODO (dhatch): Make this not hit the database.
```
gj, line 13: Do you have a plan for this?

languages/python/sqlalchemy-oso/sqlalchemy_oso/auth.py
```diff
@@ -58,6 +60,50 @@ def authorize_model(oso: Oso, actor, action, session: Session, model) -> Query:
...
 
     if combined_query is None:
-        return null_query(session)
+        return null_query(session, model)
 
     return combined_query
+
+
+# TODO (dhatch): Api might not be viable.
+def authorize_model_filter(oso: Oso, actor, action, session: Session, model):
```
gj, line 69: Do we need both this + `authorize_model`?

languages/python/sqlalchemy-oso/sqlalchemy_oso/hooks.py
```diff
@@ -0,0 +1,158 @@
+"""SQLAlchemy hooks that transparently enable oso on SQLAlchemy operations.
+
+There are several potential interfaces to integrate oso with SQLAlchemy:
+
+    - :py:func:`authorized_sessionmaker`: (**recommended**) Session factory that
+       creates a session that applies authorization on every query.
+    - :py:func:`enable_hooks`: Globally enable oso on all queries.
+    - :py:func:`make_authorized_query_cls`: Make an query class that is
```
gj, line 8:
```suggestion
    - :py:func:`make_authorized_query_cls`: Make a query class that is
```

languages/python/sqlalchemy-oso/sqlalchemy_oso/hooks.py
```diff
@@ -0,0 +1,158 @@
...
+        target = Query
+
+    disable_before_compile = enable_before_compile(
+        target, get_oso, get_user, get_action
+    )
+
+    def disable_hooks():
+        disable_before_compile()
+
+    return disable_hooks
```
gj, line 54: Does this work?
```suggestion
    return enable_before_compile(
        target, get_oso, get_user, get_action
    )
```

languages/python/sqlalchemy-oso/sqlalchemy_oso/hooks.py
```diff
@@ -0,0 +1,158 @@
...
+
+
+def authorize_query(query: Query, get_oso, get_user, get_action) -> Query:
+    """Authorize an existing query with an oso instance, user and action."""
+    oso = get_oso()
+    action = get_action()
+    actor = get_user()
+
+    # TODO (dhatch): Super hack to get this to work in flask... fix this.
+    query = query.enable_assertions(False)
```
gj, line 80: If you aren't fixing this prior to shipping, could you expand the comment to explain what it's doing?

languages/python/sqlalchemy-oso/sqlalchemy_oso/hooks.py
```diff
@@ -0,0 +1,158 @@
...
+
+    entities = {column["entity"] for column in query.column_descriptions}
```
gj, line 82: Add `if column["entity"]` guard into comprehension instead of the `if entity is None:` check below?

languages/python/sqlalchemy-oso/sqlalchemy_oso/hooks.py
```diff
@@ -0,0 +1,158 @@
...
+            continue
+
+        authorized_filter = authorize_model_filter(
+            oso, actor, action, query.session, entity
+        )
+        if authorized_filter is not None:
+            query = query.filter(authorized_filter)
+
+    # TODO can't print query because printing query causes infinite recursion.
+    #print(f"authorize_query: {query.statement.compile()}")
```
gj, line 95: Was this just for debugging or was it intended as an `INFO`-level log output of the library?

languages/python/sqlalchemy-oso/sqlalchemy_oso/hooks.py
```diff
@@ -0,0 +1,158 @@
...
+
+    return AuthorizedQuery
+
+
+def authorized_sessionmaker(get_oso, get_user, get_action, *args, **kwargs):
+    """Session factory for sessions with oso authorization applied.
+
+    :param get_oso: Callable that return oso instance to use for authorization.
+    :param get_user: Callable that returns user for an authorization request.
+    :param get_action: Callable that returns user for the action.
```
gj, line 113: Is this comment correct?

languages/python/sqlalchemy-oso/sqlalchemy_oso/hooks.py
```diff
@@ -0,0 +1,158 @@
...
+
+    The ``query_cls`` parameter cannot be used with ``authorize_sessionmaker``.
+
+    Baked queries will be disabled for this session, because they are incompatible
+    with authorization.
+
+    All other positional and keyword arguments are passed through to
+    :py:func:`sqlalchemy.orm.session.sessionmaker` unchanged.
+    """
+    # TODO (dhatch): Should be possible with additional wrapping.
```
gj, line 123: Is this comment still relevant now that you have `make_authorized_query_cls` as an alternate API?

Or is this comment talking about removing `make_authorized_query_cls` as a public API and rolling the functionality into here, dispatching based on whether a `query_cls` param is provided?

languages/python/sqlalchemy-oso/tests/test_sqlalchemy.py
```diff
@@ -0,0 +1,331 @@
...
+    oso.load_str('allow("user", "read", post: Post) if post.access_level = "public";')
+    oso.load_str('allow("user", "write", post: Post) if post.access_level = "private";')
+    oso.load_str('allow("admin", "read", post: Post);')
+    oso.load_str(
+        'allow("moderator", "read", post: Post) if '
+        '(post.access_level = "private" or post.access_level = "public") and '
+        "post.needs_moderation = true;"
+    )
+
+    query = session.query(Post.contents, Post.id)
```
gj, line 79: Could this test + the one above be extracted into a shared setup where we pass in the query? Since that seems to be the only difference between the two tests

languages/python/sqlalchemy-oso/tests/test_sqlalchemy.py
```diff
@@ -0,0 +1,331 @@
...
+        subquery = session.query(Post).subquery()
+        query = session.query(subquery)
+
+        # Fine with hooks if you don't authorize it.
+        assert query.count() == 1
+    finally:
+        disable()
+
+
+@pytest.mark.xfail(reason="No good, aliases don't work right now.")
```
gj, line 233: How do subqueries work but explicit aliases don't?

languages/python/sqlalchemy-oso/sqlalchemy_oso/auth.py
dhatch: To make it not hit the database? Not precisely. Could have a special sentinel object we return that indicates to the calling code to return no results.  But ultimately it would require a special SQLAlchemy query subclass that doesn't hit the db. Might be tough.

languages/python/sqlalchemy-oso/sqlalchemy_oso/auth.py
dhatch: No, I want to drop authorize_model ultimately.  The original reason authorize_model worked over Query was because I thought we may need to add JOINs (which cannot just be constructed as filter expressions).  That turned out not to be necessary, which is why we now have `authorize_model_filter`.

languages/python/sqlalchemy-oso/sqlalchemy_oso/hooks.py
dhatch: Probably, I just wrote it this way to leave a pattern for adding more hooks.

languages/python/sqlalchemy-oso/sqlalchemy_oso/hooks.py
dhatch: Good shout yeah

languages/python/sqlalchemy-oso/sqlalchemy_oso/auth.py
dhatch: Combined this slightly more.

languages/python/sqlalchemy-oso/sqlalchemy_oso/hooks.py
dhatch: I think it's more readable as is with the comment.

languages/python/sqlalchemy-oso/sqlalchemy_oso/hooks.py
dhatch: Removed

languages/python/sqlalchemy-oso/sqlalchemy_oso/hooks.py
dhatch: Thanks fixed

languages/python/sqlalchemy-oso/sqlalchemy_oso/hooks.py
dhatch: This is talking about making it possible to use a custom query class when using oso.  Right now you cannot because of how this is implemented (another query class).

languages/python/sqlalchemy-oso/tests/test_sqlalchemy.py
dhatch: Fixed up 510ff36dc3b5eac8c530aa9cf55643321ece987b

languages/python/sqlalchemy-oso/sqlalchemy_oso/auth.py
dhatch: 510ff36dc3b5eac8c530aa9cf55643321ece987b

dhatch: 547a55bf47c17125ecbecbe2ee1fa1406f3d7c6d

languages/python/sqlalchemy-oso/tests/test_sqlalchemy.py
dhatch: Subqueries don't necessarily use an alias.

languages/python/sqlalchemy-oso/tests/test_sqlalchemy.py
gj: nicely done

languages/python/sqlalchemy-oso/sqlalchemy_oso/auth.py
```diff
@@ -38,24 +38,46 @@ def authorize_model(oso: Oso, actor, action, session: Session, model) -> Query:
...
+    not authorized, a query that always contains no result will be returned.
+
+    :param oso: The oso class to use for evaluating the policy.
+    :param actor: The actor to authorize.
+    :param action: The action to authorize.
 
+    :param session: The SQLAlchemy session.
+    :param model: The model to authorize, must be a SQLAlchemy model.
+    """
+    # TODO (dhatch): Check that model is a model.
```
gj, line 61:
```suggestion
```

languages/python/sqlalchemy-oso/sqlalchemy_oso/auth.py
```diff
@@ -38,24 +38,46 @@ def authorize_model(oso: Oso, actor, action, session: Session, model) -> Query:
...
+    # TODO (dhatch): More robust name mapping?
```
gj, line 62: You removed this comment from `authorize_model` -- does it still apply here?

languages/python/sqlalchemy-oso/sqlalchemy_oso/partial.py
```diff
@@ -16,15 +16,9 @@
...
-def partial_to_query(expression: Expression, session: Session, model) -> Query:
-    """Convert constraints in ``partial`` to a query over ``model``."""
-    query = session.query(model)
-
-    expr = translate_expr(expression, session, model)
-    if expr is not None:
-        return query.filter(expr)
-
-    return query
+def partial_to_filter(expression: Expression, session: Session, model):
```
gj, line 19: Is this function a useful abstraction?

gj (approved):

languages/python/sqlalchemy-oso/sqlalchemy_oso/auth.py
dhatch: Yes, authorize model calls this, so we only need it in one place.

languages/python/sqlalchemy-oso/sqlalchemy_oso/partial.py
dhatch: Yeah, we might rename & change the internals of this to match other frameworks.

PR #532, opened 2020-11-17
Fix bug with ordering of comparison operands

gj:


polar-core/src/terms.rs
```diff
@@ -358,6 +358,12 @@ impl Term {
     }
 }
 
+/// Operand chirality.
```
dhatch, line 361: I don't know what chirality means.  Can we use a term that is widely understood?

dhatch: Also is there a use for this concept outside of the argument to the partial `compare` function? If not, what about just moving it into the partial module and calling it ArgumentPosition?

polar-core/src/partial/partial.rs
```diff
@@ -143,7 +143,10 @@ impl Partial {
 
         let op = Operation {
             operator,
-            args: vec![self.variable_term(), other],
+            args: match operand {
```
dhatch, line 146: Would it be beneficial to normalize operations so that `this` is always on the LHS?

That could eliminate the need for branches for both in all adapters.

dhatch (changes requested):
Nice catch! Some comments

polar-core/src/terms.rs
gj: no fun

(I didn't know what it meant either)

dhatch: 😎

gj: Moved it into partial module and updated naming a bit

polar-core/src/partial/partial.rs
gj: Done in e8654d2. Good suggestion!

polar-core/src/partial/partial.rs
```diff
@@ -130,23 +130,25 @@ impl Partial {
...
+        // Normalize comparison operations so this is always on LHS.
+        let args = vec![self.variable_term(), operand.term];
+        let mut operation = Operation { operator, args };
+        if operand.side == Side::Left && asymmetric_op {
+            operation.operator = match operation.operator {
+                Gt => Lt,
+                Geq => Leq,
+                Lt => Gt,
+                Leq => Geq,
+                o => o,
```
dhatch, line 148: Should this final case be `unreachable!()`? Due to L135.

dhatch (approved):
lgtm

polar-core/src/partial/partial.rs
gj: 091ad77

PR #534, opened 2020-11-18
Generalize sub_this to handle levels of nesting > 2.

dhatch:
Expressions like `x in y.a.b.c` were not translating properly.  This PR fixes that by changing `sub_this` to operate recursively over operations.

I also did some re-organization and moved each simplification method onto `Simplifier`.  I started this to make it easier to unit test each in isolation, but didn't end up needing to write those tests to figure out the issue.

PR checklist:

- [ ] Added changelog entry.

### DIFF
--- a/polar-core/src/partial/partial.rs
+++ b/polar-core/src/partial/partial.rs
@@ -857,11 +857,11 @@ mod test {
     #[test]
     fn test_nested_dot_in() -> TestResult {
         let p = Polar::new();
-        p.load_str("f(x, y) if x in y.a.b.c;");
+        p.load_str("f(x, y) if x in y.a.b.c;")?;
 
-        let mut q = p.new_query_from_term(term!(call!("f", [1, partial!("a")])), false)?;
+        let mut q = p.new_query_from_term(term!(call!("f", [1, partial!("a")])), false);
 
-        assert_partial_expression!(next_binding(&mut q)?, "a", "_this");
+        assert_partial_expression!(next_binding(&mut q)?, "a", "1 in _this.a.b.c");
         Ok(())
     }
 
@@ -869,10 +869,18 @@ mod test {
     fn test_nested_dot_lookup() -> TestResult {
         let p = Polar::new();
         p.load_str("f(x, y) if x = y.a.b.c;")?;
+        p.load_str("f(x, y) if x > y.a.b.c and x < y.a.b and y.a.b.c > x;")?;
 
         let mut q = p.new_query_from_term(term!(call!("f", [1, partial!("a")])), false);
 
-        assert_partial_expression!(next_binding(&mut q)?, "a", "_this");
+        assert_partial_expression!(next_binding(&mut q)?, "a", "_this.a.b.c = 1");
+        assert_partial_expression!(
+            next_binding(&mut q)?,
+            "a",
+            "_this.a.b.c < 1 and _this.a.b > 1 and _this.a.b.c > 1"
+        );
         Ok(())
     }
+
+    // TODO test for nested dot with compare.
 }
--- a/polar-core/src/partial/partial.rs
+++ b/polar-core/src/partial/partial.rs
@@ -858,9 +858,7 @@ mod test {
     fn test_nested_dot_in() -> TestResult {
         let p = Polar::new();
         p.load_str("f(x, y) if x in y.a.b.c;")?;
-
         let mut q = p.new_query_from_term(term!(call!("f", [1, partial!("a")])), false);
-
         assert_partial_expression!(next_binding(&mut q)?, "a", "1 in _this.a.b.c");
         Ok(())
     }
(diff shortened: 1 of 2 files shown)


polar-core/src/partial/simplify.rs
```diff
@@ -149,6 +87,90 @@ impl Folder for PartialToExpression {
...
+                operator: Operator::Dot,
+                args,
+            }) => {
+                assert!(args[0].value().as_symbol().unwrap().is_this_var());
+                arg.clone_with_value(Value::Expression(Operation {
+                    operator: Operator::Dot,
+                    args: vec![dot_op.clone(), args[1].clone()],
+                }))
+            },
+            Value::Expression(Operation {
```
dhatch, line 148: This new branch is the only actual change in the PR.

polar-core/src/partial/simplify.rs
```diff
@@ -149,6 +87,84 @@ impl Folder for PartialToExpression {
...
+            dot_op.value(),
+            Value::Expression(Operation {
+                operator: Operator::Dot,
+                ..
+            })
+        ));
+
+        match arg.value() {
+            Value::Variable(v) if v.is_this_var() => dot_op.clone(),
+            Value::Expression(Operation { operator, args }) => {
```
dhatch, line 141: This is the only actual change.

polar-core/src/partial/simplify.rs
```diff
@@ -149,6 +87,84 @@ impl Folder for PartialToExpression {
...
+            .iter()
+            .filter(|o| seen.insert(o))
+            .cloned()
+            .collect();
+        partial.clone_with_constraints(constraints)
+    }
+
+    /// If ``operation`` is an AND or OR operation with 1 argument, return the 1st argument.
+    ///
+    /// Returns: Some(op) if a rewrite occured, or None.
```
gj, line 105:
```suggestion
    /// If `operation` is a 1-arg AND or OR operation, return its argument.
    ///
    /// Returns: Some(op) if a rewrite occurred; otherwise None.
```

polar-core/src/partial/simplify.rs
```diff
@@ -149,6 +87,84 @@ impl Folder for PartialToExpression {
...
+            // Unwrap a single-arg And or Or expression and fold the inner term.
+            Operation {
+                operator: Operator::And,
+                args,
+            }
+            | Operation {
+                operator: Operator::Or,
+                args,
+            } if args.len() == 1 => {
+                if let Value::Expression(op) = args[0].value() {
```
gj, line 117: What could this be if not a `Value::Expression`? Is this just for getting at the inner operation? (I recognize that I'm asking you to diagnose code that I originally wrote 😆)

polar-core/src/partial/simplify.rs
```diff
@@ -149,6 +87,84 @@ impl Folder for PartialToExpression {
...
+                    Some(op.clone())
+                } else {
+                    None
+                }
+            }
+            _ => None,
+        }
+    }
+
+    /// Subsitute the this variable in a constraint with a dot operation.
```
gj, line 127:
```suggestion
    /// Substitute the this variable in a constraint with a dot operation.
```

polar-core/src/partial/simplify.rs
```diff
@@ -149,6 +87,84 @@ impl Folder for PartialToExpression {
...
+                arg.clone_with_value(Value::Expression(Operation {
+                    operator: *operator,
+                    args: args.iter().map(|arg| Self::sub_this(arg, dot_op)).collect(),
+                }))
+            }
+            _ => arg.clone(),
+        }
+    }
+
+    /// Substitute the _this variable in a list of constraints with a dot operation path.
```
gj, line 151: Just to standardize
```suggestion
    /// Substitute the this variable in a list of constraints with a dot operation path.
```

polar-core/src/partial/partial.rs
```diff
@@ -853,4 +853,42 @@ mod test {
...
         Ok(())
     }
+
+    #[test]
+    fn test_nested_dot_in() -> TestResult {
+        let p = Polar::new();
+        p.load_str("f(x, y) if x in y.a.b.c;")?;
+
+        let mut q = p.new_query_from_term(term!(call!("f", [1, partial!("a")])), false);
+
```
gj, line 863:
```suggestion
        let mut q = p.new_query_from_term(term!(call!("f", [1, partial!("a")])), false);
```

polar-core/src/partial/partial.rs
```diff
@@ -853,4 +853,42 @@ mod test {
...
+    #[test]
+    fn test_nested_dot_lookup() -> TestResult {
+        let p = Polar::new();
+        p.load_str("f(x, y) if x = y.a.b.c;")?;
+        p.load_str("f(x, y) if x > y.a.b.c and x < y.a.b and y.a.b.c > x;")?;
+        p.load_str("f(x, y) if x = y.a;")?;
+        p.load_str("f(x, y) if x = y.a.b;")?;
+        p.load_str("f(x, y) if x = y.a.b.c.d;")?;
+        p.load_str("f(x, y) if x = y.a.b.c.d.e;")?;
+        p.load_str("f(x, y) if x = y.a.b.c.d.e.f;")?;
```
gj, line 877:
```suggestion
        p.load_str(
            r#"f(x, y) if x = y.a.b.c;
               f(x, y) if x > y.a.b.c and x < y.a.b and y.a.b.c > x;
               f(x, y) if x = y.a;
               f(x, y) if x = y.a.b;
               f(x, y) if x = y.a.b.c.d;
               f(x, y) if x = y.a.b.c.d.e;
               f(x, y) if x = y.a.b.c.d.e.f;"#
        )?;
```

gj (commented):
test formatting nits

gj (approved):
Core change LGTM

polar-core/src/partial/simplify.rs
dhatch: Yeah... I had the same thought.  I think it's invalid Polar if this isn't an expression.  Seems okay like this though.

PR #535, opened 2020-11-19
Bidirectional class compatibility checks w/ partial matches

gj:
PR checklist:

- [ ] Added changelog entry.

### DIFF
--- a/polar-core/src/partial/isa_constraint_check.rs
+++ b/polar-core/src/partial/isa_constraint_check.rs
@@ -31,11 +31,21 @@ impl IsaConstraintCheck {
         }
     }
 
-    /// Check if the existing constraints set is compatible with the proposed
-    /// matches class.
+    /// Check if existing constraints are compatible with the proposed type constraint (`_this
+    /// matches Post{}`).
     ///
-    /// Returns: None if compatible, QueryEvent::Done { false } if incompatible,
-    /// or QueryEvent to ask for compatibility.
+    /// If the existing constraint is also a type constraint , we return a pair of
+    /// `QueryEvent::ExternalIsSubclass`es to check whether the type constraints are compatible.
+    /// The constraints are compatible if they are the same class or if either is a subclass of the
+    /// other.
+    ///
+    /// If the existing constraint is not a type constraint, there's no external check required,
+    /// and we return `None` to indicate compatibility.
+    ///
+    /// Returns:
+    /// - `None` if compatible.
+    /// - A pair of `QueryEvent::ExternalIsSubclass` checks if compatibility cannot be determined
+    /// locally.
     fn check_constraint(
         &mut self,
         mut constraint: Operation,
@@ -80,14 +90,17 @@ impl Runnable for IsaConstraintCheck {
 
         if let Some(result) = self.result.take() {
             if result {
+                // If the primary check succeeds, there's no need to check the alternative.
                 self.alternative_check = None;
             } else if self.alternative_check.is_none() {
+                // If both checks fail, we fail.
                 return Ok(QueryEvent::Done { result: false });
             }
         }
 
         let counter = counter.expect("IsaConstraintCheck requires a Counter");
         loop {
+            // If there's an alternative waiting to be checked, check it.
             if let Some(alternative) = self.alternative_check.take() {
                 return Ok(alternative);
             } else if let Some(constraint) = self.existing.pop() {
--- a/polar-core/src/partial/isa_constraint_check.rs
+++ b/polar-core/src/partial/isa_constraint_check.rs
@@ -9,6 +9,7 @@ pub struct IsaConstraintCheck {
     existing: Vec<Operation>,
     proposed_tag: Option<Symbol>,
     result: Option<bool>,
+    alternative_check: Option<QueryEvent>,
     last_call_id: u64,
 }
 
@@ -25,6 +26,7 @@ impl IsaConstraintCheck {
             existing,
             proposed_tag,
             result: None,
+            alternative_check: None,
             last_call_id: 0,
         }
     }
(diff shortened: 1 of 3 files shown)


polar-core/src/partial/isa_constraint_check.rs
```diff
@@ -69,20 +89,25 @@ impl Runnable for IsaConstraintCheck {
...
-            if let Some(constraint) = next {
-                if let Some(event) = self.check_constraint(constraint, &counter) {
-                    return Ok(event);
+            // If there's an alternative waiting to be checked, check it.
+            if let Some(alternative) = self.alternative_check.take() {
+                return Ok(alternative);
+            } else if let Some(constraint) = self.existing.pop() {
+                if let Some((primary, alternative)) = self.check_constraint(constraint, &counter) {
+                    self.alternative_check = Some(alternative);
+                    return Ok(primary);
```
leina05, line 109: Don't really understand the mechanics of this so not able to fully review it.

polar-core/src/partial/partial.rs
```diff
@@ -407,6 +411,37 @@ mod test {
...
+        };
+        assert_partial_expression!(
+            next_binding(),
+            "a",
+            "_this matches PostSubclass{} and _this matches Post{}"
+        );
+        assert_query_done!(q);
+        Ok(())
+    }
+
```
leina05, line 444: Would it be redundant to test with the opposite order of subclass/superclass as well (i.e. it hits the superclass constraint first, then the subclass)?

polar-core/src/partial/isa_constraint_check.rs
```diff
@@ -25,20 +26,31 @@ impl IsaConstraintCheck {
...
+    /// The constraints are compatible if they are the same class or if either is a subclass of the
+    /// other.
+    ///
+    /// If the existing constraint is not a type constraint, there's no external check required,
+    /// and we return `None` to indicate compatibility.
+    ///
+    /// Returns:
+    /// - `None` if compatible.
+    /// - A pair of `QueryEvent::ExternalIsSubclass` checks if compatibility cannot be determined
+    /// locally.
```
leina05, line 48: One other case that this doesn't account for, which comes up in languages that support multiple inheritance:

- If you have these three classes:
```python
class Doctor:
    pass

class Administrator:
    pass

class DepartmentHead(Doctor, Administrator):
    pass
```

And then rules that looked like this:

```python
f(x: Doctor);
g(x: Administrator);
```
Those constraints would be incompatible by this implementation, although there exists a class (`DepartmentHead`) for which they are compatible. 

I don't think that this is a common use case AT ALL, but we should probably document the behavior and say that we won't recognize multiple inheritance structures with Partials?

gj: I don't quite follow -- what's the actual rule that's incompatible under the current implementation?

polar-core/src/partial/partial.rs
gj: That gets tested in `test_partial_isa_two_rule`, above 🙂

polar-core/src/partial/isa_constraint_check.rs
leina05: Updated my comment to try to be more clear

leina05: I may be misunderstanding how the engine evaluates it though. If it only compares each `isa` constraint to the initial `Partial` type constraint, and not to each other, than it would be fine.

gj: The two `matches` constraints would exist on separate branches of an `or`. The partial's constraints would look something like this:
```python
_this matches DepartmentHead{}
and ((_this matches Doctor{} and _this.approver = true)
or (_this matches Administrator{} and _this.approver = true))
```
I added parens to hopefully make it clearer.

leina05: Ah yeah, I meant for them to be on the same `and` branch, my bad. E.g., 

```
is_department_head(x: DepartmentHead) if is_doctor(x) and is_administrator(x);
is_doctor(x: Doctor);
is_administrator(x: Administrator);
```

Wondering if this would cause the constraints to clash or not.

gj: Ah, you're right! We'll add a failing test.

Right now we would get to the point of adding the `_this matches Administrator{}` constraint, check it against the existing `_this matches DepartmentHead{}` constraint (which would succeed), and then check it against the existing `_this matches Doctor{}` constraint, which _should_ succeed but will currently fail.

gj: (This PR is still a step in the direction of correctness, though)

dhatch (approved):
LGTM, although I'm not sure our library code always handles the narrowing case:

```
f(x: Post) if g(x);
g(x: PostSubclass);
```

if `x` isn't actually a post subclass, our isa assert probably fails.  I think you may have fixed this with https://github.com/osohq/oso/pull/533/commits/f10ff9ac1782b7c6706a2ef37c95dbbd07807c26#diff-a88e4d38f71841b81cb8ff7aa71124ca648ac2fb2231ac27b83f4424f0db1d3fR66 in Django.

polar-core/src/partial/isa_constraint_check.rs
leina05: Thanks for checking that! Yeah again, I think this is a very niche case, especially since most of our mainstream use cases don't seem to use complex multi-dispatch.

leina05 (approved):

gj:
> I think you may have fixed this with [f10ff9a#diff-a88e4d38f71841b81cb8ff7aa71124ca648ac2fb2231ac27b83f4424f0db1d3fR66](https://github.com/osohq/oso/commit/f10ff9ac1782b7c6706a2ef37c95dbbd07807c26#diff-a88e4d38f71841b81cb8ff7aa71124ca648ac2fb2231ac27b83f4424f0db1d3fR66) in Django.

Yep, we'll port that to the other adapters.

PR #536, opened 2020-11-19
Add field constraints to partials

gj:


polar-core/src/partial/partial.rs
```diff
@@ -84,16 +86,46 @@ impl Partial {
...
     }
 
-    pub fn isa(&mut self, other: Term) -> Box<dyn Runnable> {
-        let isa_op = op!(Isa, self.variable_term(), other);
-
-        let constraint_check = Box::new(IsaConstraintCheck::new(
-            self.constraints.clone(),
-            isa_op.clone(),
-        ));
+    pub fn isa(&mut self, other: Term) -> Option<Box<dyn Runnable>> {
```
dhatch, line 89: Comment indicating what `None` means.

polar-core/src/partial/partial.rs
```diff
@@ -84,16 +86,46 @@ impl Partial {
...
+                // Construct field-less Isa operation since field constraints will be added
+                // separately.
+                let tag_pattern = term!(pattern!(instance!(tag.clone())));
+                let type_constraint = op!(Isa, self.variable_term(), tag_pattern);
+                let existing = self.constraints.clone();
+                let check = Box::new(IsaConstraintCheck::new(existing, type_constraint.clone()));
+
+                self.add_constraint(type_constraint);
+
+                // Add field constraints.
```
dhatch, line 112: As an alternative, could we call `self.isa(Pattern::Dictionary)` here? Would save a small amount of code duplication. I can make this change & test, just lmk what you think.

polar-core/src/partial/partial.rs
```diff
@@ -353,11 +385,70 @@ mod test {
     #[test]
     fn test_partial_isa_with_fields() -> TestResult {
         let p = Polar::new();
-        p.load_str("f(x: Post{id: 1});")?;
+        p.load_str(
+            r#"f(x: Post{id: 1});
+               f(x: Post{id: 1}) if x matches {id: 2};
+               f(x: Post{id: 1}) if x matches Post{id: 2};
```
dhatch, line 391: We are missing a case for a instance literal with more than one field. Will add.

gj: Line 394, below?

dhatch: That one fails, so we never actually check that the code worked since no binding is returned.

dhatch (approved):
lgtm

PR #539, opened 2020-11-20
Fix new clippy issues.

samscott89:


languages/rust/oso/src/repl.rs
```diff
@@ -112,21 +112,20 @@ pub fn main() -> anyhow::Result<()> {
...
 
-        if input.starts_with("%def") {
-            if let Err(e) = oso.load_str(&input[4..]) {
+        if let Some(define) = input.strip_prefix("%def") {
+            if let Err(e) = oso.load_str(define) {
                 println!("{}", e);
             }
             continue;
         }
-        input.pop(); // remove trailing `;`
```
samscott89, line 129: (this was incorrect, I went to try out the REPL after making the above change to confirm it works)

gj (approved):

PR #542, opened 2020-11-23
Rust release job

saolsen:
`cargo package` as part of release job and `cargo publish` as part of publish job.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -113,3 +113,39 @@ jobs:
         run: echo '//registry.npmjs.org/:_authToken=${NPM_AUTH_TOKEN}' > .npmrc
       - run: npm publish
         working-directory: "oso-js-node-${{ github.event.inputs.version }}"
+  rust:
+    name: Publish Rust crates to Crates.io
+    runs-on: ubuntu-18.04
+    env:
+      CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/cache@v2
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            target
+          key: ${{ runner.os }}-cargo-release-${{ hashFiles('**/Cargo.lock') }}
+      - name: Install Rust stable toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+          override: true
+      - name: package polar-core
+        working-directory: polar-core
+        run: |
+          cargo publish
+      - name: package polar-core
+        working-directory: polar-core
+        run: |
+          cargo publish
+      - name: package oso-derive
+        working-directory: languages/rust/oso-derive
+        run: |
+          cargo publish
+      - name: package oso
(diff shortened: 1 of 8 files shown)


.github/workflows/publish.yml
```diff
@@ -113,3 +113,39 @@ jobs:
...
+      CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/cache@v2
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            target
+          key: ${{ runner.os }}-cargo-release-${{ hashFiles('**/Cargo.lock') }}
```
samscott89, line 129: Does this work as a cache key?

.github/workflows/publish.yml
```diff
@@ -113,3 +113,39 @@ jobs:
...
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+          override: true
+      - name: package polar-core
+        working-directory: polar-core
+        run: |
+          cargo publish
+      - name: package polar-core
```
samscott89, line 140: polar-core is here twice

polar-capo and polar-wasm-api instead?

.github/workflows/release.yml
```diff
@@ -294,6 +294,40 @@ jobs:
...
+        with:
+          profile: minimal
+          toolchain: stable
+          override: true
+      - name: package polar-core
+        working-directory: polar-core
+        run: |
+          cargo package
+      - name: package polar-core
+        working-directory: polar-core
```
samscott89, line 320: Same as other comment

.github/workflows/publish.yml
saolsen: why wouldn't it?

.github/workflows/publish.yml
saolsen: Oh oops, and I dunno if we want capi or wasm since we don't publish those as crates.

.github/workflows/publish.yml
samscott89: (I.e. does the package step use the same files as `cargo build --release`?

.github/workflows/publish.yml
samscott89: Oh probably not then.

.github/workflows/publish.yml
```diff
@@ -113,3 +113,35 @@ jobs:
...
+      - name: Install Rust stable toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+          override: true
+      - name: package polar-core
+        working-directory: polar-core
+        run: |
+          cargo publish
```
samscott89, line 139: Super minor TIOLI nit: these don't need to be multiline

samscott89 (approved):

.github/workflows/publish.yml
samscott89: Sometimes it has to rebuild a bunch of things anyway

saolsen: Actually I think cargo publish always does a full rebuild anyway so it doesn't matter. I'll just drop the cacheing for these.

PR #543, opened 2020-11-23
Trigger release jobs in other repos.

saolsen:
Adds a step to the release workflow that notifies other repos that a release is being built for a specific commit.
This lets us trigger release workflows on other repos so we test all our examples when we test our main repo.

It should work for tagged releases and the nightly main release job.
Proof of concept is the oso-flask-integration repo.

Here's the workflow that is triggered by this event.
https://github.com/osohq/oso-flask-integration/blob/main/.github/workflows/test_release.yml

It's not  really possible to main the release workflow dependent on all these other release workflows succeeding so I think we need to add success/failure messages in slack for all the release workflows so we see if a nightly or tag build breaks anything and as part of the end of sprint release process make sure all these downstream release jobs succeeded before publishing.


samscott89:
> It's not really possible to main the release workflow dependent on all these other release workflows succeeding so I think we need to add success/failure messages in slack for all the release workflows so we see if a nightly or tag build breaks anything and as part of the end of sprint release process make sure all these downstream release jobs succeeded before publishing.

Did you consider doing the same process (cloning repo + running the tests), but where the osohq/oso repo clones the example repository? We _could_ run that process on every merge to main, for example. Especially if we add them all as submodules like we have done with the quickstarts? Then they would all be accessible through this repo.

OTOH, I quite like this approach. It prevents us from adding too much bloat to this repo. And the trigger + slack notify seems like a fine approach to me.

samscott89 (approved):

PR #546, opened 2020-11-24
Revert release workflow change of "Rust release job (#542)"

gj:
This reverts commit 96eeb86ba1e3fbf10c977eeae011417ed4539703.


saolsen:
We need the publish one, just drop the release one.

saolsen:
The one in publish.yml is the one we need to keep and the one in release.yml is the one to revert.

saolsen (approved):

PR #553, opened 2020-11-25
List filtering docs

gj:
Co-authored-by: Alex Plotnick <alex@osohq.com>


docs/getting-started/list-filtering/index.rst
```diff
@@ -1,9 +1,87 @@
-==============
-List Filtering
-==============
+=======================
+Authorizing Collections
```
gj, line 2:
```suggestion
List Filtering
```

samscott89 (commented):
This overall looks really good. A few specific bugs communicated in slack (markdown links + headers, TOC too big).

Main feedback which can be follow up:

- Docs intro needs to get to the point sooner. E.g. first sentence: "oso supports applying authorization logic at the ORM layer so that you can efficiently authorize entire data sets. For example, ...". 

- Could be nice in the django example to turn on SQL logging and instruct people to check it out to see whats happening? E.g. https://github.com/osohq/oso-django-integration/blob/sam/partial-expenses/oso_tutorial/oso_tutorial/settings.py#L143-L150

samscott89 (approved):

PR #560, opened 2020-11-30
Fix Rust benchmarks

samscott89:
This was causing benchmarks to fail halfway

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -73,8 +73,8 @@ jobs:
           alert-comment-cc-users: "@osohq/eng"
           alert-threshold: "150%"
           # Only push and save on pushes to main.
-          auto-push: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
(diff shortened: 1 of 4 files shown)


.github/workflows/bench.yml
```diff
@@ -73,8 +73,8 @@ jobs:
           alert-comment-cc-users: "@osohq/eng"
           alert-threshold: "150%"
           # Only push and save on pushes to main.
-          auto-push: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
+          auto-push: ${{ github.event_name == 'push' && github.repository == 'osohq/oso' && github.ref == 'refs/heads/main' }}
```
gj, line 76: Isn't the repo always going to be `osohq/oso` due to the event restrictions at the top of this file?

gj (approved):

.github/workflows/bench.yml
samscott89: No, the workflow executes in whatever repo created the pull request. So if someone creates a PR from _their_ main branch this triggers. Which is what was happening here: https://github.com/osohq/oso/pull/562

PR #562, opened 2020-11-30
Update django-oso with User model registration, logging vs print

devmonkey22:
- Update `init_oso`:
  - Register (dynamically set) user model, default as `auth::User`.
  - Register AnonymousUser as `auth::AnonymousUser` polar class to match User.
- Update policy file initialization to log policy files loaded via logging not `print`.
- Remove unused imports from `apps.py`.

### DIFF
--- a/languages/python/django-oso/django_oso/oso.py
+++ b/languages/python/django-oso/django_oso/oso.py
@@ -40,22 +40,13 @@ def init_oso():
         for model in app.get_models():
             register_class(model, polar_model_name(model))
 
-    # Custom registration for auth (AnonymousUser, User)
+    # Custom registration for auth (AnonymousUser)
     if apps.is_installed("django.contrib.auth"):
         from django.contrib.auth.models import AnonymousUser
 
         # Register under `auth` app_label to match default User model, but also fully-qualified name
         register_class(AnonymousUser, "auth::AnonymousUser")
         register_class(AnonymousUser, "django::contrib::auth::AnonymousUser")
-
-        # Register the user model, possibly customized
-        try:
-            from django.contrib.auth import get_user_model
-
-            User = get_user_model()
-            register_class(User, polar_model_name(User))
-        except ImproperlyConfigured:
-            pass
 
     # Register request
     register_class(HttpRequest)
--- a/languages/python/django-oso/django_oso/oso.py
+++ b/languages/python/django-oso/django_oso/oso.py
@@ -1,11 +1,15 @@
+import logging
 import os.path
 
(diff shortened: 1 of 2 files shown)


devmonkey22:
Note: I didn't create a changelog entry yet because I wasn't sure what version this would go into, and the vNext template didn't seem my place to use and create a 0.8.3.  Let me know though.

devmonkey22:

languages/python/django-oso/django_oso/oso.py
```diff
@@ -36,11 +40,22 @@ def register_class(model, name=None):
...
         from django.contrib.auth.models import AnonymousUser
 
-    register_class(AnonymousUser, name=f"django::contrib::auth::AnonymousUser")
+        # Register under `auth` app_label to match default User model, but also fully-qualified name
+        register_class(AnonymousUser, "auth::AnonymousUser")
+        register_class(AnonymousUser, "django::contrib::auth::AnonymousUser")
+
+        # Register the user model, possibly customized
+        try:
+            from django.contrib.auth import get_user_model
```
dhatch, line 53: This shouldn't be necessary, because the registration on L41 (https://github.com/osohq/oso/pull/562/files#diff-f0f374f317fb12ecab1cae1843229bf549c8f2d6c1efdf882ed4a72f0f031996R41) should also register the built in User model if it's part of an installed app.

Were you having trouble using the built-in User model in the policy?

languages/python/django-oso/django_oso/oso.py
```diff
@@ -36,11 +40,22 @@ def register_class(model, name=None):
...
             register_class(model, polar_model_name(model))
 
-    # Custom registration for auth (AnonymousUser)
+    # Custom registration for auth (AnonymousUser, User)
     if apps.is_installed("django.contrib.auth"):
         from django.contrib.auth.models import AnonymousUser
 
-    register_class(AnonymousUser, name=f"django::contrib::auth::AnonymousUser")
+        # Register under `auth` app_label to match default User model, but also fully-qualified name
+        register_class(AnonymousUser, "auth::AnonymousUser")
```
dhatch, line 48: Nice, good idea here.

dhatch (commented):
Looks pretty good. Left some comments inline. We may need to update documentation for the AnonymousUser name change.  I can help with that.

dhatch (approved):
LGTM! Thanks for the PR Mike!  I will do the documentation & changelog updates as a separate follow up PR.

Currently github actions seems to be down, so will need to wait to merge this.

PR #566, opened 2020-12-01
Add support for subfield isa to SQLAlchemy.

dhatch:
PR checklist:

- [x] Added changelog entry.


docs/changelogs/vNEXT.rst
```diff
@@ -37,7 +37,5 @@ contribution!
 Other bugs & improvements
 =========================
 
-- bulleted list
-- improvements
-- of smaller
-- potentially with doc links
+- Matches operations on fields of partials are now handled correctly in the
```
gj, line 40:
```suggestion
- ``matches`` operations on fields of partials are now handled correctly in the
```

languages/python/sqlalchemy-oso/sqlalchemy_oso/partial.py
```diff
@@ -16,41 +17,55 @@
...
+    else:
+        for field_name in dot_path(left):
+            _, model, __ = get_relationship(model, field_name)
+
+    assert not right.fields, "Unexpected fields in isa expression"
+    constraint_type = get_model(right.tag)
+    if not issubclass(model, constraint_type):
+        return sql.false()
+    else:
+        return None
```
gj, line 66: Also is there a `sql.true()` (or equivalent) that we could call here instead of `None`? @plotnick and I just made the same change in the Odoo adapter: https://github.com/osohq/oso_odoo/pull/4/files#diff-6d3cd0165f6f824122a73df7d8e31e00e5479c533ed967833761eb1888858023L66-R70
```suggestion
    if issubclass(model, constraint_type):
        return None
    else:
        return sql.false()
```

languages/python/django-oso/django_oso/partial.py
```diff
@@ -73,6 +73,8 @@ def isa_expr(expr: Expression, model: Model, **kwargs):
     for attr in dot_path(left):
         model = getattr(model, attr).field.related_model
     constraint_type = apps.get_model(django_model_name(right.tag))
+    assert not right.fields, "Unexpected fields in isa expression"
```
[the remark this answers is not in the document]
dhatch: Just a check, but I think we should make a practice of checking stuff like this. I want to be very careful we don't miss checking any constraints.

gj, line 76:
```suggestion
    assert not right.fields, "Unexpected fields in matches expression"
```

languages/python/sqlalchemy-oso/sqlalchemy_oso/auth.py
```diff
@@ -9,6 +9,16 @@
...
 
 
+def polar_model_name(model) -> str:
+    """Return polar class name for SQLAlchemy model."""
+    return model.__name__
+
+
+def sqlalchemy_model_name(tag: str) -> str:
+    """Return sqlalchemy model name for polar class name."""
+    return tag
```
gj, line 19: What's the point of this function?

gj, line 19: Where are these functions used?

languages/python/sqlalchemy-oso/sqlalchemy_oso/partial.py
```diff
@@ -16,41 +17,51 @@
 EmitFunction = Callable[[Session, Any], Any]
 
 
-def partial_to_filter(expression: Expression, session: Session, model):
+def partial_to_filter(expression: Expression, session: Session, model, get_model):
```
gj, line 20: Maybe thread some of these arguments around as `**kwargs`? Would make future changes to the arity less bigly.

languages/python/sqlalchemy-oso/sqlalchemy_oso/partial.py
```diff
@@ -16,41 +17,51 @@
...
+def translate_and(expression: Expression, session: Session, model, get_model):
     assert expression.operator == "And"
     expr = and_()
     for expression in expression.args:
-        translated = translate_expr(expression, session, model)
-        if translated is None:
-            continue
-
+        translated = translate_expr(expression, session, model, get_model)
         expr = expr & translated
```
gj, line 45:
```suggestion
        expr = expr & translate_expr(expression, session, model, get_model)
```

languages/python/sqlalchemy-oso/sqlalchemy_oso/auth.py
dhatch: Thanks. Removed one, used the other.

dhatch: Gone.

languages/python/sqlalchemy-oso/sqlalchemy_oso/auth.py
```diff
@@ -62,7 +64,7 @@ def authorize_model_filter(oso: Oso, actor, action, session: Session, model):
     # TODO (dhatch): More robust name mapping?
     assert class_mapper(model), f"Expected a model; received: {model}"
 
-    partial_resource = Partial("resource", TypeConstraint(model.__name__))
+    partial_resource = Partial("resource", TypeConstraint(polar_model_name(model.__name__)))
```
gj, line 67:
```suggestion
    partial_resource = Partial("resource", TypeConstraint(polar_model_name(model)))
```

gj (approved):
Migrate to **kwargs or just :shipit: -- up 2 u

dhatch:
This PR seems messed up... checks aren't running.

PR #567, opened 2020-12-01
Run PR CI on all branches.

samscott89:


samscott89:
False alarm, turns out GH Actions are just having a tough one today

PR #568, opened 2020-12-01
Implement comparisons in sqlalchemy.

dhatch:
PR checklist:

- [x] Added changelog entry.

### DIFF
--- a/languages/python/django-oso/django_oso/partial.py
+++ b/languages/python/django-oso/django_oso/partial.py
@@ -92,17 +92,23 @@ def and_expr(expr: Expression, model: Model, **kwargs):
     return q
 
 
-def compare_expr(expr: Expression, _model: Model, path=(), **kwargs):
+def compare_expr(expr: Expression, model: Model, path=(), **kwargs):
     q = Q()
     (left, right) = expr.args
     left_path = dot_path(left)
     if left_path:
         return COMPARISONS[expr.operator](q, "__".join(path + left_path), right)
     else:
-        if isinstance(right, Model):
+        if isinstance(right, model):
             right = right.pk
         else:
+            # QUESTION: Should this just return an empty set? Similar to how invalid isa
+            # returns empty set.
             raise UnsupportedError(f"Unsupported comparison: {expr}")
+
+        if not expr.operator in ('Eq', 'Unify'):
+            raise UnsupportedError(f"Comparing model only supported with Eq or Unify not: {expr}")
+
         return COMPARISONS[expr.operator](q, "__".join(path + ("pk",)), right)
 
 
--- a/languages/python/django-oso/django_oso/partial.py
+++ b/languages/python/django-oso/django_oso/partial.py
@@ -104,9 +104,11 @@ def compare_expr(expr: Expression, model: Model, path=(), **kwargs):
         else:
             return Q(pk__in=[])
 
-        if expr.operator not in ('Eq', 'Unify'):
-            raise UnsupportedError(f"Unsupported comparison: {expr}. Models can only be compared"
-                                   " with `=` or `==`")
+        if expr.operator not in ("Eq", "Unify"):
+            raise UnsupportedError(
+                f"Unsupported comparison: {expr}. Models can only be compared"
+                " with `=` or `==`"
+            )
 
         return COMPARISONS[expr.operator](q, "__".join(path + ("pk",)), right)
 
--- a/languages/python/django-oso/django_oso/partial.py
+++ b/languages/python/django-oso/django_oso/partial.py
@@ -102,12 +102,11 @@ def compare_expr(expr: Expression, model: Model, path=(), **kwargs):
         if isinstance(right, model):
             right = right.pk
         else:
-            # QUESTION: Should this just return an empty set? Similar to how invalid isa
-            # returns empty set.
-            raise UnsupportedError(f"Unsupported comparison: {expr}")
+            return Q(pk__in=[])
 
-        if not expr.operator in ('Eq', 'Unify'):
-            raise UnsupportedError(f"Comparing model only supported with Eq or Unify not: {expr}")
+        if expr.operator not in ('Eq', 'Unify'):
+            raise UnsupportedError(f"Unsupported comparison: {expr}. Models can only be compared"
+                                   " with `=` or `==`")
 
         return COMPARISONS[expr.operator](q, "__".join(path + ("pk",)), right)
 
--- a/languages/python/sqlalchemy-oso/sqlalchemy_oso/partial.py
+++ b/languages/python/sqlalchemy-oso/sqlalchemy_oso/partial.py
@@ -72,9 +72,10 @@ def translate_compare(expression: Expression, session: Session, model, get_model
         assert inspect(right)
         assert isinstance(right, model)
 
-        if not expression.operator in ("Eq", "Unify"):
+        if expression.operator not in ("Eq", "Unify"):
             raise UnsupportedError(
-                f"Comparing model only supported with Eq or Unify not: {expression.operator}"
+                f"Unsupported comparison: {expression}. Models can only be compared"
+                " with `=` or `==`"
             )
 
         primary_keys = [pk.name for pk in inspect(model).primary_key]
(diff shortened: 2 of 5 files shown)


languages/python/django-oso/django_oso/partial.py
```diff
@@ -91,17 +91,23 @@ def and_expr(expr: Expression, model: Model, **kwargs):
...
     left_path = dot_path(left)
     if left_path:
         return COMPARISONS[expr.operator](q, "__".join(path + left_path), right)
     else:
-        if isinstance(right, Model):
+        if isinstance(right, model):
             right = right.pk
         else:
+            # QUESTION: Should this just return an empty set? Similar to how invalid isa
+            # returns empty set.
```
gj, line 105: Feels different to me -- the semantics of returning an empty set would be "this is a valid query that returns no results," which I don't think is an accurate representation of what's happening.

languages/python/django-oso/django_oso/partial.py
```diff
@@ -91,17 +91,23 @@ def and_expr(expr: Expression, model: Model, **kwargs):
...
             raise UnsupportedError(f"Unsupported comparison: {expr}")
+
+        if not expr.operator in ('Eq', 'Unify'):
+            raise UnsupportedError(f"Comparing model only supported with Eq or Unify not: {expr}")
```
gj, line 109: Why is this necessary?

languages/python/django-oso/django_oso/partial.py
```diff
@@ -91,17 +91,23 @@ def and_expr(expr: Expression, model: Model, **kwargs):
...
-def compare_expr(expr: Expression, _model: Model, path=(), **kwargs):
+def compare_expr(expr: Expression, model: Model, path=(), **kwargs):
     q = Q()
     (left, right) = expr.args
     left_path = dot_path(left)
     if left_path:
         return COMPARISONS[expr.operator](q, "__".join(path + left_path), right)
     else:
-        if isinstance(right, Model):
+        if isinstance(right, model):
```
gj, line 101: I don't think this is correct. All the previous code is doing is translating any Django model into its primary key. We pass the authorized model around as `model`, so this change would cause comparisons on subfields that don't inherit from `model` to fail.

dhatch: I should add a test for this.  This branch (from my understanding) only runs for something like `this = post` where `post` is a variable with the `Post` type.  I changed the check because conceivably you could write `this = user`.  This doesn't do any type checking so if the types don't match, we still extract a PK, and then we'd be trying to find a Post that has the PK of the user.

languages/python/django-oso/django_oso/partial.py
dhatch: Not sure... see my example above.  Unifying something with a type that doesn't match would end up in this branch, which is very similar to checking isa with a type that doesn't match.

languages/python/django-oso/django_oso/partial.py
dhatch: Because it doesn't make sense otherwise.  If you write `this < post`, you would probably expect to use the natural ordering of post, which isn't something we know about, not to get all posts that have a primary key less than that of the right hand side.

languages/python/django-oso/django_oso/partial.py
gj: I see where you're coming from -- if `left_path` is falsy then there's no dot operation and the LHS must be `_this`, so the only reasonably comparable thing is `model`.

dhatch: Exactly

dhatch:
Enumerated tests to add for this here: https://www.notion.so/osohq/TEST-Test-for-model-comparison-this-in-SQLAlchemy-others-68aad30b405e484aa6d058407a52289d

languages/python/django-oso/django_oso/partial.py
gj: In light of the above realization, I agree that this should probably return `Q(pk__in=[])`, which is Django's version of the "always false" SQL clause. (See usage in `isa_expr()`, above.)

languages/python/django-oso/django_oso/partial.py
gj: Cool, can we use the error message from above (`Unsupported comparison: {expr}`) once we replace that with the `Q(pk__in=[])` thing?

languages/python/sqlalchemy-oso/sqlalchemy_oso/partial.py
```diff
@@ -61,6 +61,10 @@ def translate_compare(expression: Expression, session: Session, model):
     else:
         assert left == Variable("_this")
         assert inspect(right)
+        assert isinstance(right, model)
+
+        if not expression.operator in ('Eq', 'Unify'):
+            raise UnsupportedError(f"Comparing model only supported with Eq or Unify not: {expression.operator}")
```
gj, line 67:
```suggestion
            raise UnsupportedError(f"Unsupported comparison: {expression}")
```

languages/python/django-oso/django_oso/partial.py
gj, line 109:
```suggestion
            raise UnsupportedError(f"Unsupported comparison: {expr}")
```

languages/python/sqlalchemy-oso/sqlalchemy_oso/partial.py
```diff
@@ -61,6 +61,12 @@ def translate_compare(expression: Expression, session: Session, model):
     else:
         assert left == Variable("_this")
         assert inspect(right)
+        assert isinstance(right, model)
+
+        if not expression.operator in ("Eq", "Unify"):
+            raise UnsupportedError(
+                f"Comparing model only supported with Eq or Unify not: {expression.operator}"
+            )
```
gj, line 69:
```suggestion
        if expression.operator not in ("Eq", "Unify"):
            raise UnsupportedError(f"Unsupported comparison: {expression}")
```

languages/python/django-oso/django_oso/partial.py
```diff
@@ -91,17 +91,25 @@ def and_expr(expr: Expression, model: Model, **kwargs):
...
             right = right.pk
         else:
+            # QUESTION: Should this just return an empty set? Similar to how invalid isa
+            # returns empty set.
             raise UnsupportedError(f"Unsupported comparison: {expr}")
+
+        if not expr.operator in ("Eq", "Unify"):
+            raise UnsupportedError(
+                f"Comparing model only supported with Eq or Unify not: {expr}"
+            )
```
gj, line 111:
```suggestion
        if expr.operator not in ("Eq", "Unify"):
            raise UnsupportedError(f"Unsupported comparison: {expr}")
```

languages/python/django-oso/django_oso/partial.py
```diff
@@ -91,17 +91,25 @@ def and_expr(expr: Expression, model: Model, **kwargs):
...
     if left_path:
         return COMPARISONS[expr.operator](q, "__".join(path + left_path), right)
     else:
-        if isinstance(right, Model):
+        if isinstance(right, model):
             right = right.pk
         else:
+            # QUESTION: Should this just return an empty set? Similar to how invalid isa
+            # returns empty set.
             raise UnsupportedError(f"Unsupported comparison: {expr}")
```
gj, line 106: we should probably create a "constant" like `always_false = Q(pk__in=[])` at the top of the file and then use it in both places
```suggestion
			return Q(pk__in=[])
```

languages/python/django-oso/django_oso/partial.py
dhatch: I used a different message because it is more specific and tells you that a Unification or Equality comparison would be valid, not just that the comparison is unsupported.

dhatch: Revised to match other error more closely.

languages/python/django-oso/django_oso/partial.py
dhatch: I saw you guys added this constant in another PR, we can replace it then.

gj (approved):
~Align SQLAlchemy error w/ Django error (https://github.com/osohq/oso/pull/568/files#r534404138)~ and then :shipit:

PR #569, opened 2020-12-01
Improve high level API used by SQLAlchemy adapter.

dhatch:
- The oso object, user and action to authorize are now attached to sessions.  There is no way for the user, action or oso object to change for an individual session. This means the identity map (internal SQLAlchemy cache) will only contain objects from one authorization context, and will never contain unauthorized objects (a risk of the prior approach). The `AuthorizedSession` class handles this, and the changes to `make_authorized_query_cls` to accept `oso`, `action` and `user` instead of `get_oso`, `get_action`, and `get_user`.
- We have a new `scoped_session` factory that behaves like the SQLAlchemy [scoped session](https://docs.sqlalchemy.org/en/13/orm/contextual.html#unitofwork-contextual).  Test added [here](https://github.com/osohq/oso/pull/569/files#diff-016043b481920e3042a3cb543b568a4687d9b744532117a637893dfcdb8e3888R217) showing how that works. This will be useful for the Flask integration.

PR checklist:

- [x] Added changelog entry.
- [x] Update documentation
- [x] Make members private (like `enable_hooks`)
- [x] Test for query property.
- [x] Add optional `flask_sqlalchemy.SQLAlchemy` subclass (this could be a separate PR)

### DIFF
--- a/languages/python/sqlalchemy-oso/sqlalchemy_oso/session.py
+++ b/languages/python/sqlalchemy-oso/sqlalchemy_oso/session.py
@@ -14,7 +14,7 @@ from sqlalchemy_oso.auth import authorize_model_filter
 
 
 @event.listens_for(Query, "before_compile", retval=True)
-def before_compile(query):
+def _before_compile(query):
     """Enable before compile hook."""
     return _authorize_query(query)
 
@@ -63,7 +63,7 @@ def authorized_sessionmaker(get_oso, get_user, get_action, class_=None, **kwargs
     :param get_action: Callable that returns action for the authorization request.
     :param class_: Base class to use for sessions.
 
-    All other positional and keyword arguments are passed through to
+    All other keyword arguments are passed through to
     :py:func:`sqlalchemy.orm.session.sessionmaker` unchanged.
     """
     if class_ is None:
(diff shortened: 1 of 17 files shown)


languages/python/sqlalchemy-oso/sqlalchemy_oso/session.py
```diff
@@ -0,0 +1,155 @@
...
+
+from oso import Oso
+
+from sqlalchemy_oso.auth import authorize_model_filter
+
+
+@event.listens_for(Query, "before_compile", retval=True)
+def _before_compile(query):
+    """Enable before compile hook."""
+    return _authorize_query(query)
```
gj, line 15: TODO: test that returning `None` here (e.g., the check on line 24, below) does nothing to the provided query.

languages/python/sqlalchemy-oso/sqlalchemy_oso/session.py
```diff
@@ -0,0 +1,155 @@
...
+    :param class_: Base class to use for sessions.
+
+    All other keyword arguments are passed through to
+    :py:func:`sqlalchemy.orm.session.sessionmaker` unchanged.
+    """
+    if class_ is None:
+        class_ = Session
+
+    # oso, user and action must remain unchanged for the entire session.
+    # If they change before a query runs, an error is thrown.
```
gj, line 69:
```suggestion
```

languages/python/sqlalchemy-oso/tests/test_flask.py
```diff
@@ -0,0 +1,89 @@
...
+
+    authorized = sqlalchemy.session.query(TestModel).all()
+    assert len(authorized) == 1
+    assert authorized[0].id == 1
+
+    authorized = TestModel.query.all()
+    assert len(authorized) == 1
+    assert authorized[0].id == 1
+
+    oso.load_str("allow(_, _, tm: TestModel) if tm.id = 2;")
```
gj, line 84: Not an issue for now, but once we enable `cut` w/ partials we will want to test whether you can add a more specific rule that has a `cut` that should authorize a subset of a less-specific rule that was previously cached by SQLAlchemy.

gj (approved):
Add the one test we talked about and :shipit:

PR #570, opened 2020-12-01
Add test for duplicate results in Django adapter

gj:
Co-authored-by: Alex Plotnick <alex@osohq.com>
Co-authored-by: Sam Scott <sam@osohq.com>


samscott89 (approved):
Wow, this is some of your best work @gj. Great stuff.

On a serious note. I notice that you used `objects.authorize` so that it would do the filtering. Even though it fails with `authorize_model`. Seems a bit suspicious that they differ.

gj:
> On a serious note. I notice that you used `objects.authorize` so that it would do the filtering. Even though it fails with `authorize_model`. Seems a bit suspicious that they differ.

`authorize_model` returns a Django filter, not a QuerySet, which is the thing you can call `.distinct()` on.

PR #572, opened 2020-12-02
Improve handling of 'in' operation with empty constraints

gj:
PR checklist:

- [ ] Added changelog entry.

### DIFF
--- a/languages/python/django-oso/django_oso/partial.py
+++ b/languages/python/django-oso/django_oso/partial.py
@@ -9,7 +9,6 @@ from .oso import django_model_name
 
 
 FALSE_FILTER = Q(pk__in=[])
-TRUE_FILTER = ~Q(pk__in=[])
 
 COMPARISONS = {
     "Unify": lambda f, v: Q(**{f: v}),
@@ -77,7 +76,7 @@ def isa_expr(expr: Expression, model: Model, **kwargs):
         model = getattr(model, attr).field.related_model
     constraint_type = apps.get_model(django_model_name(right.tag))
     assert not right.fields, "Unexpected fields in matches expression"
-    return TRUE_FILTER if issubclass(model, constraint_type) else FALSE_FILTER
+    return None if issubclass(model, constraint_type) else FALSE_FILTER
 
 
 def and_expr(expr: Expression, model: Model, **kwargs):
--- a/languages/python/django-oso/django_oso/partial.py
+++ b/languages/python/django-oso/django_oso/partial.py
@@ -1,4 +1,4 @@
-from django.db.models import Q, Model
+from django.db.models import Q, Model, Count, Subquery
 from django.apps import apps
 
 from polar.expression import Expression
@@ -112,8 +112,7 @@ def in_expr(expr: Expression, model: Model, path=(), **kwargs):
 
     if isinstance(left, Expression):
         if left.operator == "And" and not left.args:
-            from django.db.models import Count, Subquery
-
+            # An unconstrained partial is in a list if the list is non-empty.
             count = Count("__".join(right_path))
             filter = COMPARISONS["Gt"]("__".join(right_path + ("count",)), 0)
             subquery = Subquery(
--- a/languages/python/django-oso/django_oso/partial.py
+++ b/languages/python/django-oso/django_oso/partial.py
@@ -41,14 +41,14 @@ def partial_to_query_filter(partial: Expression, model: Model, **kwargs):
 
 
 COMPARISONS = {
-    "Unify": lambda q, f, v: Q(**{f: v}),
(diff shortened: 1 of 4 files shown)


languages/python/django-oso/django_oso/partial.py
```diff
@@ -41,28 +41,29 @@ def partial_to_query_filter(partial: Expression, model: Model, **kwargs):
...
-    "Leq": lambda q, f, v: Q(**{f"{f}__leq": v}),
-    "Lt": lambda q, f, v: Q(**{f"{f}__lt": v}),
+    "Unify": lambda f, v: Q(**{f: v}),
+    "Eq": lambda f, v: Q(**{f: v}),
+    "Neq": lambda f, v: ~Q(**{f: v}),
+    "Geq": lambda f, v: Q(**{f"{f}__gte": v}),
+    "Gt": lambda f, v: Q(**{f"{f}__gt": v}),
+    "Leq": lambda f, v: Q(**{f"{f}__leq": v}),
+    "Lt": lambda f, v: Q(**{f"{f}__lt": v}),
+    "In": lambda f, v: Q(**{f"{f}__in": v}),
```
dhatch, line 51: Why is `in` in here if it's not handled by the `compare_expr` function?

languages/python/django-oso/django_oso/partial.py
```diff
@@ -91,28 +92,37 @@ def and_expr(expr: Expression, model: Model, **kwargs):
...
     assert expr.operator == "In"
-    q = Q()
     (left, right) = expr.args
     right_path = dot_path(right)
     assert right_path, "RHS of in must be a dot lookup"
     right_path = path + right_path
 
     if isinstance(left, Expression):
-        return translate_expr(left, model, path=right_path, **kwargs)
+        if left.operator == "And" and not left.args:
```
dhatch, line 117: Could we add a comment about this case? (Why it's needed - since it's nuanced)

languages/python/sqlalchemy-oso/tests/test_post_relationship.py
```diff
@@ -396,6 +396,49 @@ def test_partial_in_collection(session, oso, tag_nested_many_many_test_fixture):
     assert len(posts) == 1
 
 
+def test_empty_constraints_in_becomes_count_gt_0(
```
dhatch, line 399: Maybe rename this test in both SQLAlchemy & Django so the test name is agnostic to the query (`becomes_count_gt_0` portion).  SQLAlchemy doesn't translate to that.

dhatch (changes requested):
Looks pretty good - some small comments that should be quick.

languages/python/django-oso/django_oso/partial.py
plotnick: We use it in `in_expr` as `return COMPARISONS["In"]("pk", subquery)`. It's doing the same thing as the other comparison operators, so seemed odd to have a different mechanism. But we could consider weakening the name to something like `OPERATORS` or `FILTERS` to include non-comparisons. As I write this, I'm actually tempted to add entries with keys `True` and `False` that return our always-true filter `Q(pk__in=[])` and its negation.

languages/python/django-oso/django_oso/partial.py
gj: 2989407

languages/python/django-oso/django_oso/partial.py
plotnick: Gabe convinced me that constants were better: de4b867

languages/python/sqlalchemy-oso/tests/test_post_relationship.py
gj: afc1c66

languages/python/django-oso/django_oso/partial.py
dhatch: I think we should put this lambda somewhere else (just inside `in_expr`?) because of the `if expr.operator in COMPARISONS` branch in `translate_expr`.  That set of conditions is subtly ordering dependent.  If the `compare_expr` was moved before the `in_expr` branch, the program would be incorrect.

I also like the ability to have `assert expr.operation in COMPARISONS` in `compare_expr` to make sure the caller passed in a valid expression.  If we add the `In` key to `COMPARISONS` that assert is no longer valid (`In` is not a valid input to `compare_expr`).

plotnick: Ok, f17fdeee adds a top-level function *next to* `COMPARISONS`. I would very much like to keep all of the `Q`-constructing functions together and with the same interface.

dhatch (approved):
lgtm

PR #573, opened 2020-12-03
Mypy for sqlalchemy-oso

dhatch:


languages/python/oso/setup.py
```diff
@@ -67,7 +67,7 @@ def get_version(rel_path):
     #
     # If using Python 2.6 or earlier, then these have to be included in
     # MANIFEST.in as well.
-    # package_data={"polar": ["policies/*.polar", "policies/*.pol"]},  # Optional
+    package_data={"polar": ["py.typed"], "oso": ["py.typed"]},  # Optional
```
samscott89, line 70: Oh nice, thanks for working this out.

languages/python/sqlalchemy-oso/requirements.txt
```diff
@@ -1,2 +1,3 @@
 oso~=0.8.2
 sqlalchemy~=1.3.20
+sqlalchemy-stubs
```
samscott89, line 3: Is this needed here?

samscott89 (changes requested):
Looks good! Small question about the stubs dependency

languages/python/sqlalchemy-oso/requirements.txt
dhatch: Instead of a development dependency?

dhatch: Yeah might work that way.

dhatch: A user would need to install that to use type checking though if this library ships with types.

samscott89 (approved):
Looks good once its passing CI

languages/python/sqlalchemy-oso/setup.cfg
```diff
@@ -9,6 +9,12 @@ plugins = sqlmypy
 [mypy-pytest.*]
 ignore_missing_imports = True
 
+[mypy-oso.*]
+ignore_missing_imports = True
```
samscott89, line 13: Doesn't this defeat the purpose?

PR #574, opened 2020-12-03
Negate matches for Django

gj:
PR checklist:

- [ ] Added changelog entry.

### DIFF
--- a/languages/python/django-oso/django_oso/partial.py
+++ b/languages/python/django-oso/django_oso/partial.py
@@ -4,6 +4,7 @@ from django.apps import apps
 from polar.expression import Expression
 from polar.exceptions import UnsupportedError
 from polar.partial import dot_path
+from oso import Variable
 
 from .oso import django_model_name
 
@@ -97,9 +98,8 @@ def compare_expr(expr: Expression, model: Model, path=(), **kwargs):
     if left_path:
         return COMPARISONS[expr.operator]("__".join(path + left_path), right)
     else:
-        if isinstance(right, model):
-            right = right.pk
-        else:
+        assert left == Variable("_this")
+        if not isinstance(right, model):
             return FALSE_FILTER
 
         if expr.operator not in ("Eq", "Unify"):
@@ -108,7 +108,7 @@ def compare_expr(expr: Expression, model: Model, path=(), **kwargs):
                 " with `=` or `==`"
             )
 
-        return COMPARISONS[expr.operator]("__".join(path + ("pk",)), right)
+        return COMPARISONS[expr.operator]("__".join(path + ("pk",)), right.pk)
 
 
 def in_expr(expr: Expression, model: Model, path=(), **kwargs):
--- a/languages/python/django-oso/django_oso/partial.py
+++ b/languages/python/django-oso/django_oso/partial.py
@@ -51,11 +51,7 @@ def partial_to_query_filter(partial: Expression, model: Model, **kwargs):
         Q(is_private=False)
     """
 
-    q = translate_expr(partial, model, **kwargs)
-    if q is None:
-        return Q()
-
-    return q
+    return translate_expr(partial, model, **kwargs)
 
 
 def translate_expr(expr: Expression, model: Model, **kwargs):
@@ -90,9 +86,7 @@ def and_expr(expr: Expression, model: Model, **kwargs):
     assert expr.operator == "And"
     q = Q()
     for arg in expr.args:
-        expr = translate_expr(arg, model, **kwargs)
-        if expr:
-            q = q & expr
+        q &= translate_expr(arg, model, **kwargs)
     return q
 
 
--- a/languages/python/django-oso/django_oso/partial.py
+++ b/languages/python/django-oso/django_oso/partial.py
@@ -136,4 +136,5 @@ def in_expr(expr: Expression, model: Model, path=(), **kwargs):
 
 def not_expr(expr: Expression, model: Model, **kwargs):
     assert expr.operator == "Not"
+    assert expr.args[0].operator == "Isa"
     return ~translate_expr(expr.args[0], model, **kwargs)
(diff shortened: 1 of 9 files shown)


languages/python/django-oso/django_oso/partial.py
```diff
@@ -76,11 +79,10 @@ def isa_expr(expr: Expression, model: Model, **kwargs):
         model = getattr(model, attr).field.related_model
     constraint_type = apps.get_model(django_model_name(right.tag))
     assert not right.fields, "Unexpected fields in matches expression"
-    return None if issubclass(model, constraint_type) else FALSE_FILTER
+    return TRUE_FILTER if issubclass(model, constraint_type) else FALSE_FILTER
 
 
 def and_expr(expr: Expression, model: Model, **kwargs):
-    assert expr.operator == "And"
```
dhatch, line 83: Why remove these assertions? They are there since we don't have typed expressions for each operation. This ensures the input to this function is valid.

languages/python/django-oso/tests/test_django_oso.py
```diff
@@ -144,19 +144,32 @@ def test_partial(rf, partial_policy):
...
     assert (
         str(authorize_filter)
-        == "(AND: ('is_private', False), ('timestamp__gt', 0), ('option', None))"
+        == "(AND: (NOT (AND: ('pk__in', []))), ('is_private', False), ('timestamp__gt', 0), ('option', None))"
     )
 
     q = Post.objects.filter(authorize_filter)
+    assert (
+        str(q.query)
+        == 'SELECT "test_app_post"."id", "test_app_post"."is_private", "test_app_post"."name",'
```
dhatch, line 153: Where is the `NOT pk__in` in this SQL?

languages/python/django-oso/tests/test_django_oso.py
```diff
@@ -187,9 +200,18 @@ def test_partial_subfield_isa():
     authorize_filter = authorize_model(None, Post, actor="foo", action="bar")
     assert (
         str(authorize_filter)
-        == "(OR: (AND: ('pk__in', []), ('created_by__name', 'not alice')), ('created_by__name', 'alice'))"
+        == "(OR:"
+        + " (AND: (NOT (AND: ('pk__in', []))), ('pk__in', []), ('created_by__name', 'not alice')),"
```
dhatch, line 204: This is a `AND: True, False`? That seems contradictory.  Also I don't see these translated in the SQL.

languages/python/django-oso/django_oso/partial.py
```diff
@@ -76,11 +79,10 @@ def isa_expr(expr: Expression, model: Model, **kwargs):
         model = getattr(model, attr).field.related_model
     constraint_type = apps.get_model(django_model_name(right.tag))
     assert not right.fields, "Unexpected fields in matches expression"
-    return None if issubclass(model, constraint_type) else FALSE_FILTER
+    return TRUE_FILTER if issubclass(model, constraint_type) else FALSE_FILTER
```
dhatch, line 82: I know this makes this code slightly cleaner (can remove `None` checks in a few spots -- although I don't see that actually in this PR), but it pollutes the stringified output for filters, making the tests harder to read and understand.

Should we reconsider just keeping this as `None`?

dhatch (commented):
Approach looks good. Some comments.

languages/python/django-oso/django_oso/partial.py
gj: Didn't seem useful since we're already switching on the operator in `translate_expr()`, but I don't really mind them and have added them back in a3447049

languages/python/django-oso/tests/test_django_oso.py
gj: `(AND: ('pk__in', []))` translates to "always false"; `(NOT (AND: ('pk__in', [])))` translates to "always true", so there's no reason for it to appear in the SQL output. The Django ORM has optimized the trivial truth constraint away.

languages/python/django-oso/tests/test_django_oso.py
gj: Yeah, the entire thing is:

```
OR(
	AND(true, false, created_by__name="not alice"),
	AND(true, true, created_by__name="alice")
)
```
The Django optimizer removes the first disjunct b/c `AND(true, false, ...)` cannot be `true`. It also optimizes away both `true`s from the second disjunct b/c `AND(true, anything)` is the same as `AND(anything)`.

languages/python/django-oso/django_oso/partial.py
gj: We would both prefer to leave it like this. It's nice that all translation components now _always_ return filters. Using "always true" and "always false" representations that Django's ORM knows how to understand means we can lean on its boolean logic handling instead of duplicating that in our own code.

gj: FWIW, we tried using `None` both last night + again just now to see if it's possible, and it quickly became more complicated than this implementation. Agree that the filters look a little funky, but all of the funkiness is optimized away by Django's ORM so that the funkiness won't be visible to end users.

gj: Removed `None` checks: 7dfa0be

languages/python/django-oso/django_oso/partial.py
```diff
@@ -133,3 +132,8 @@ def in_expr(expr: Expression, model: Model, path=(), **kwargs):
             return translate_expr(left, model, path=right_path, **kwargs)
     else:
         return COMPARISONS["Unify"]("__".join(right_path), left)
+
+
+def not_expr(expr: Expression, model: Model, **kwargs):
+    assert expr.operator == "Not"
```
gj, line 138:
```suggestion
    assert expr.operator == "Not"
    assert expr.args[0].operator == "Isa"
```

dhatch (approved):
lgtm

PR #575, opened 2020-12-04
Fix `ground` in `string`

samscott89:


polar-core/tests/integration_tests.rs
```diff
@@ -1868,4 +1868,6 @@ fn test_builtin_iterables() {
         "x",
         vec![value!(1), value!(2)],
     );
+
+    qeval(&mut p, r#"forall(x in "abc", x in "abacus")"#);
```
gj, line 1872: Does this only match b/c all of `"a"`, `"b"`, and `"c"` are in `"abacus"`? Could we add a complementary `qnull` test when that doesn't hold?

samscott89: Yep, will add a few more tests. Just wanted to push this before checkin so people could see the problem.

samscott89: @gj added

polar-core/src/vm.rs
```diff
@@ -1485,7 +1485,7 @@ impl PolarVirtualMachine {
                             s.chars()
                                 .map(|c| c.to_string())
                                 .map(Value::String)
-                                .filter(|c| !is_ground || term.value() == c)
+                                .filter(|c| !is_ground || c == value)
```
gj, line 1488: To ensure I have this correct: `term.value()` is the entire `in` operation, right? Now we're comparing `c` (a character produced by iterating over the string) and `value`, the LHS of the `in` operation?

gj (approved):

gj:
Oh and please add a changelog entry for the bugfix.

polar-core/src/vm.rs
samscott89: Yep! The previous behaviour made no sense, but it never got past the `!is_ground` check.

PR #576, opened 2020-12-04
Add readline back in.

samscott89:
I was overeager in addressing lint warnings. The readline import has side effects and is indeed "used"


plotnick (approved):
Yes indeed. REPL users thank you.

gj (approved):

PR #578, opened 2020-12-07
Bump node dependencies.

samscott89:
Fix tiny helper type error.

### DIFF
--- a/languages/js/package.json
+++ b/languages/js/package.json
@@ -35,7 +35,7 @@
     "@types/lodash": "^4.14.165",
     "@types/node": "^14.14.10",
     "chokidar-cli": "^2.1.0",
-    "gts": "^3.0.3",
+    "gts": "^2.0",
     "jest": "^26.6.3",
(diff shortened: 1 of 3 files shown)


gj (changes requested):
Did you test this locally? TypeDoc was pinned to 0.17.0-3 because we're using "library" mode, which went away after that version. It's currently being worked on (https://github.com/TypeStrong/typedoc/issues/1364), but I was waiting to migrate us to that version until it stabilizes. The warning about highlight.js is not very dire -- we don't highlight any user input.

samscott89:
> 
> 
> Did you test this locally? TypeDoc was pinned to 0.17.0-3 because we're using "library" mode, which went away after that version. It's currently being worked on ([TypeStrong/typedoc#1364](https://github.com/TypeStrong/typedoc/issues/1364)), but I was waiting to migrate us to that version until it stabilizes. The warning about highlight.js is not very dire -- we don't highlight any user input.

Okay, sounds good. Thanks

PR #579, opened 2020-12-07
Builtin Roles for sqlalchemy

saolsen:
Still porting in the tests but you can see usage here https://github.com/osohq/oso-github-sqlalchemy.

https://github.com/osohq/oso-github-sqlalchemy/blob/main/app/models.py#L90-L115

PR Checklist:
- [x]  move unit tests for role helpers into sqlalchemy lib
- [x]  move `OsoSession` class over from test_app


languages/python/sqlalchemy-oso/sqlalchemy_oso/roles.py
```diff
@@ -0,0 +1,398 @@
...
+            inherits_role_helper(role.name, inherited_role_name, role_order) and
+            inherited_role = new {Role}(name: inherited_role_name, {Resource.lower()}: role.{Resource.lower()});
+        """
+
+    # @TODO: Group
+    oso.load_str(policy)
+
+
+def resource_role_class(
+    declarative_base, user_model, resource_model, roles, group_model=None
```
gj, line 152: Follow-up: Maybe take `user_model` or `group_model` instead of always the former and optionally the latter

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles.py
```diff
@@ -0,0 +1,398 @@
...
+from sqlalchemy import inspect
+
+
+ROLE_CLASSES: List[Any] = []
+
+
+def enable_roles(oso):
+    """Enable the SQLAlchemy Role-Based Access Control base policy. This method activates the following polar rules:
+
+        `role_allow(role, action, resource)`:
```
gj, line 16:
```suggestion
        ``role_allow(role, action, resource)``:
```

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles.py
```diff
@@ -0,0 +1,398 @@
...
+from sqlalchemy.ext.declarative import declared_attr
+from sqlalchemy.orm import relationship, backref
+from sqlalchemy import inspect
+
+
+ROLE_CLASSES: List[Any] = []
+
+
+def enable_roles(oso):
+    """Enable the SQLAlchemy Role-Based Access Control base policy. This method activates the following polar rules:
```
gj, line 14: Ensure that these doc strings get picked up in the auto-generated API documentation.

gj (approved):
:shipit: (and then do follow-ups)

leina05:
> :shipit: (and then do follow-ups)

Recording these follow-ups here: https://www.notion.so/osohq/SQLAlchemy-Roles-Lib-Follow-ups-abc0b02f81bd425189cfbaa0403e9045

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles.py
samscott89: `actor_model` might make sense for example

plotnick:
Bit late to the party, and this suggestion should not block this PR from merging.

Instead of generating Polar primarily for the specializers (that's my current understanding -- please correct if mistaken), could we not use fixed names for the classes in the policy and then just register the actual classes using those names?

languages/python/sqlalchemy-oso/sqlalchemy_oso/session.py
```diff
@@ -11,6 +11,29 @@
 from sqlalchemy_oso.auth import authorize_model_filter
 
 
+class OsoSession:
```
dhatch, line 14: Make this a private class. Either:

- define in `set_get_session` 
- name `_OsoSession`.

languages/python/sqlalchemy-oso/sqlalchemy_oso/session.py
```diff
@@ -11,6 +11,29 @@
...
+
+    @classmethod
+    def set_get_session(cls, get_session):
+        OsoSession._get = get_session
+
+
+def set_get_session(oso: Oso, get_session_func):
+    """Set the function that oso uses to expose a SQLAlchemy session to the policy
+
+    :param [oso]: [The Oso instance used to evaluate the policy.]
```
dhatch, line 27: Is this doc string format correct? I haven't seen brackets in it before.

dhatch: Can we include in here how to access the session this creates from the policy? Just a polar snippet?

leina05:
> Bit late to the party, and this suggestion should not block this PR from merging.
> 
> Instead of generating Polar primarily for the specializers (that's my current understanding -- please correct if mistaken), could we not use fixed names for the classes in the policy and then just register the actual classes using those names?

The problem is that these classes are ALSO generated, so no way to have their names be fixed. Also these are classes that our users will write their own rules over so they should have names that are familiar to them.

languages/python/sqlalchemy-oso/sqlalchemy_oso/session.py
leina05: idk this was a sphinx thing I think

leina05: https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles.py
```diff
@@ -0,0 +1,398 @@
...
+        Column(
+            f"{resource_model.__name__.lower()}_role_id",
+            Integer,
+            ForeignKey(f"{resource_model.__name__.lower()}_roles.id"),
+            primary_key=True,
+        ),
+        Column(
+            "user_id",
+            Integer,
+            ForeignKey(f"{user_model.__tablename__}.id"),
```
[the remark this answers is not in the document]
leina05: Documented in API docs

PR #580, opened 2020-12-07
Partial in partial (ORM semantics)

gj:
PR checklist:

- [ ] Added changelog entry.

### DIFF
--- a/docs/examples/list-filtering/django/example/app/policy/example.polar
+++ b/docs/examples/list-filtering/django/example/app/policy/example.polar
@@ -7,4 +7,4 @@ allow(user: app::User, "GET", post: app::Post) if
 
 allow(user: app::User, "GET", post: app::Post) if
     post.access_level = "private" and
-    post.creator in user.direct_reports.all();
+    post.creator in user.direct_reports;
--- a/languages/python/django-oso/django_oso/auth.py
+++ b/languages/python/django-oso/django_oso/auth.py
@@ -82,7 +82,6 @@ def authorize_model(request, model, *, actor=None, action=None) -> Q:
     filter = None
     for result in results:
         resource_partial = result["bindings"]["resource"]
-        breakpoint()
         if filter is None:
             filter = Q()
 
--- a/languages/python/django-oso/django_oso/auth.py
+++ b/languages/python/django-oso/django_oso/auth.py
@@ -82,6 +82,7 @@ def authorize_model(request, model, *, actor=None, action=None) -> Q:
     filter = None
     for result in results:
         resource_partial = result["bindings"]["resource"]
+        breakpoint()
         if filter is None:
             filter = Q()
 
--- a/languages/python/django-oso/django_oso/partial.py
+++ b/languages/python/django-oso/django_oso/partial.py
@@ -122,7 +122,11 @@ def in_expr(expr: Expression, model: Model, path=(), **kwargs):
     (left, right) = expr.args
 
     if isinstance(right, QuerySet):
-        return contained_in("pk", Subquery(right.values("pk")))
+        if isinstance(left, Expression):
+            item_filter = translate_expr(left, model, **kwargs)
+        else:
+            item_filter = Q()
+        return contained_in("pk", Subquery(right.values("pk").filter(item_filter)))
     else:
         right_path = dot_path(right)
         assert right_path, "RHS of in must be a dot lookup"
--- a/languages/python/django-oso/django_oso/partial.py
+++ b/languages/python/django-oso/django_oso/partial.py
@@ -133,7 +133,9 @@ def in_expr(expr: Expression, model: Model, path=(), **kwargs):
             item_filter = translate_expr(left, model, path=path, **kwargs)
         else:
             item_filter = Q()
-        return contained_in("pk", Subquery(right.values("pk").filter(item_filter)))
+        return contained_in(
+            "__".join(path + ("pk",)), Subquery(right.values("pk").filter(item_filter))
+        )
     else:
         right_path = dot_path(right)
         assert right_path, "RHS of in must be a dot lookup"
--- a/languages/python/django-oso/django_oso/partial.py
+++ b/languages/python/django-oso/django_oso/partial.py
@@ -93,10 +93,6 @@ def and_expr(expr: Expression, model: Model, **kwargs):
         # Short-circuit: if any expr is false, the whole AND is false.
         if expr == FALSE_FILTER:
             return FALSE_FILTER
-        elif expr == TRUE_FILTER:
-            continue
-        elif isinstance(expr, Subquery):
-            breakpoint()
         q &= expr
     return q
 
@@ -109,9 +105,7 @@ def compare_expr(expr: Expression, model: Model, path=(), **kwargs):
         return COMPARISONS[expr.operator]("__".join(path + left_path), right)
     else:
         assert left == Variable("_this")
-        if isinstance(right, QuerySet):
-            return Subquery(right.values("pk"))
-        elif not isinstance(right, model):
+        if not isinstance(right, model):
             return FALSE_FILTER
 
         if expr.operator not in ("Eq", "Unify"):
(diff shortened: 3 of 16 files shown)


docs/examples/list-filtering/django/example/app/policy/example.polar
```diff
@@ -7,4 +7,4 @@ allow(user: app::User, "GET", post: app::Post) if
 
 allow(user: app::User, "GET", post: app::Post) if
     post.access_level = "private" and
-    post.creator in user.direct_reports.all();
+    post.creator in user.direct_reports;
```
samscott89, line 10: 🎉

languages/python/django-oso/django_oso/partial.py
```diff
@@ -73,10 +74,10 @@ def translate_expr(expr: Expression, model: Model, **kwargs):
         raise UnsupportedError(f"Unsupported partial expression: {expr}")
 
 
-def isa_expr(expr: Expression, model: Model, **kwargs):
+def isa_expr(expr: Expression, model: Model, path=(), **kwargs):
     assert expr.operator == "Isa"
     (left, right) = expr.args
-    for attr in dot_path(left):
+    for attr in path + dot_path(left):
```
dhatch, line 80: Was this a bug?

languages/python/django-oso/django_oso/partial.py
```diff
@@ -119,24 +120,45 @@ def compare_expr(expr: Expression, model: Model, path=(), **kwargs):
...
+        left = Variable("_this")
+
+    if (
+        isinstance(right, Expression)
+        and right.operator == "Unify"
+        and right.args[0] == Variable("_this")
+    ):
+        right = right.args[1]
+
+    if isinstance(right, QuerySet):
```
dhatch, line 135: Is this the only new case?

languages/python/oso/polar/host.py
```diff
@@ -19,18 +19,21 @@
 class Host:
     """Maintain mappings and caches for Python classes & instances."""
 
-    def __init__(self, polar, classes={}, instances={}):
+    def __init__(self, polar, classes={}, instances={}, to_polar_hooks=[]):
```
dhatch, line 22: no default mutable

```suggestion
    def __init__(self, polar, classes=None, instances=None, to_polar_hooks=None):
```

languages/python/oso/polar/host.py
```diff
@@ -19,18 +19,21 @@
...
         assert polar, "no Polar handle"
         self.ffi_polar = polar  # a "weak" handle, which we do not free
         self.classes = classes.copy()
         self.instances = instances.copy()
+        self.to_polar_hooks = to_polar_hooks.copy()
```
dhatch, line 27:
```suggestion
        self.to_polar_hooks = (to_polar_hooks or []).copy()
```

polar-core/src/vm.rs
```diff
@@ -1529,16 +1530,35 @@ impl PolarVirtualMachine {
...
-                                &item_partial.value().as_partial().unwrap().name().clone(),
-                                item_partial,
-                            );
-                            self.bind(partial.name(), partial.clone().into_term());
-                        } else {
-                            partial.in_contains(item);
-                            self.bind(&partial.name().clone(), partial.into_term());
+                        let partial_name = partial.name().clone();
+
+                        let partial_in_partial = |mut item: Partial, mut collection: Partial| {
```
dhatch, line 1535: Would this second partial always be a collection?

polar-core/src/vm.rs
```diff
@@ -1529,16 +1530,35 @@ impl PolarVirtualMachine {
...
+                            if collection.constraints().is_empty() {
```
dhatch, line 1536: Why this branch?

dhatch (commented):
I haven't thoroughly thought about this but it doesn't seem as bad as you guys are describing.  That may be because I'm missing the rationale for some stuff... I left some clarifying comments.

What happens with something like this (a filter on the contained collection)?  Do we have a test for this?

```
# Whole user, partial resource
allow(user, actor, resource) if
    direct_reports = user.direct_reports and 
    report in direct_reports and
    report.role = 'employee' and
    resource.owner in direct_reports;
```

`direct_reports: (this.role = 'exployee')`

`resource.owner in (this.role = 'exployee')`

`x = resource.owner and x in direct_reports and
direct_reports.role = 'employee'
`

languages/python/django-oso/django_oso/auth.py
```diff
@@ -82,6 +82,7 @@ def authorize_model(request, model, *, actor=None, action=None) -> Q:
     filter = None
     for result in results:
         resource_partial = result["bindings"]["resource"]
+        breakpoint()
```
samscott89, line 85:
```suggestion
```

languages/python/django-oso/tests/test_post_relationship.py
```diff
@@ -581,11 +823,46 @@ def test_reverse_many_relationship(tag_nested_many_many_fixtures):
...
+    """Add constraints on partial."""
+    Oso.load_str(
+        """
+        allow(user: test_app2::User, "read", post: test_app2::Post) if
+            post.created_by in user.direct_reports and
+            # This constraint is invalid (cannot constrain the list directly)
+            # It never gets output because you get a *new* partial every time you
+            # access user.direct_reports
+            user.direct_reports.username = "foo" and
+            user.direct_reports.foo();
```
gj, line 841: @dhatch reading the comment on 838, it seems like this test is most concerned about the fact that we produce a *new* partial every time you access `user.direct_reports`, but I think that nuance is lost amidst the other issues with this example (namely, the `collection.username = "foo"` check is nonsensical and we don't support method calls on partials as the last line of the rule attempts to do).

polar-core/src/vm.rs
gj: Yes.

languages/python/django-oso/django_oso/partial.py
gj: Yes.

languages/python/oso/polar/query.py
```diff
@@ -80,6 +80,7 @@ def handle_make_external(self, data):
         self.host.make_instance(cls_name, args, kwargs, id)
 
     def handle_external_call(self, data):
+        self.host.contains_partial = data["query_contains_partial"]
```
gj, line 83: @dhatch / @samscott89 are we fine with this temporary kludge shipping?

samscott89: What does this do?

gj: Since we now create partials on the host side (when encountering a Django `Manager` object), we need to ensure that we only do that when the user has already opted in to partials by calling `authorize_model` with a partial. If we create a partial when the user has not passed one in, we might end up spitting out a partial when they expect a boolean answer.

polar-core/src/vm.rs
gj: I think you're right about partials coming from the Django adapter, but we do have a test for it in the core (`test_in_partial`).

languages/python/oso/polar/query.py
samscott89: I'm only seeing it used here in the Python code:
```
languages/python/oso/polar/query.py
83:        self.host.contains_partial = data["query_contains_partial"]

```
so I'm a little confused how it is doing anything?

gj: https://github.com/osohq/oso/pull/580/files#diff-f373f8ebee08aac2d42390cfdc23da6040052b9f6a5de87f57cca9d3ff245cc4R131

gj: https://github.com/osohq/oso/blob/89ac64c0567fb880456b2e89446e60eaef0694ea/languages/python/oso/polar/host.py#L130-L135

samscott89: Ignore me, missed that it was renamed

languages/python/django-oso/django_oso/partial.py
dhatch, line 135: We need to check that the queryset has the same type as model on the left.

dhatch:
In #727

PR #588, opened 2020-12-10
drop groups from builtin roles

saolsen:
not doing this right now


leina05 (approved):
LGTM. We will test adding teams as a custom attribute on the class in the GitHub sample app.

PR #593, opened 2020-12-14
single single test case.

dhatch:


dhatch:
Closing, will add more tests as we continue this work. Have this test tracked in the Supported Query Types document.

PR #594, opened 2020-12-16
Updated role schema + library cleanups

leina05:
PR checklist:

- [x] Added changelog entry.


languages/python/sqlalchemy-oso/sqlalchemy_oso/roles.py
```diff
@@ -1,16 +1,142 @@
...
+    :param resource_model: The SQLAlchemy model representing resources that \
+    the generated Role mixin will be scoped to. The Role mixin will \
+    have a many-to-one (ForeignKey) relationship with this resource model. \
+    A many-to-many relationship to ``user_model`` is added to ``resource_model``; \
+    the relationship is named ``users``.
+
+    :param roles: An order-independent list of the built-in roles for this resource-specific role type.
+    :type roles: List[str]
+
+        .. code-block:: python
```
samscott89, line 41: What is this code block showing?

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles.py
```diff
@@ -1,16 +1,142 @@
...
+    """Create a :ref:`resource-specific role<resource-specific-roles>` Mixin
+    for SQLAlchemy models. The role mixin is an
+    `Association Object <https://docs.sqlalchemy.org/en/13/orm/basic_relationships.html#association-object>`_
+    between the ``user_model`` and the ``resource_model``.
+
+
+    :param declarative_base: The SQLAlchemy declarative base model that \
+    the role model and all related models are mapped to.
+
+    :param user_model: The SQLAlchemy model representing users that the \
```
samscott89, line 26: I know we spoke about groups being a follow up. But is there any reason this couldn't be `actor_model` and be for groups too?

samscott89: Where we wouldn't provide anything out of the box for people to get groups working with users, but at least it would be consistent with writing `allow(actor, ...)` rules for them

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles.py
```diff
@@ -1,16 +1,142 @@
...
+from sqlalchemy.schema import Column, ForeignKey
 from sqlalchemy.ext.declarative import declared_attr
-from sqlalchemy.orm import relationship, backref
-from sqlalchemy import inspect
-
+from sqlalchemy.orm import relationship, validates
+from sqlalchemy import inspect, UniqueConstraint
+from .session import _OsoSession
 
 ROLE_CLASSES: List[Any] = []
```
samscott89, line 11: Can we add a comment saying what this is used for.

From what I can tell, this is just used to construct the policy helpers, right?

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles.py
```diff
@@ -1,16 +1,142 @@
...
+        def validate_name(self, key, name):
+            if name not in self.choices:
+                raise ValueError(
+                    f"{name} Is not a valid choice for {self.__class__.__name__}"
+                )
+            return name
+
+        @declared_attr
+        def user_id(self):
+            type = inspect(user_model).primary_key[0].type
```
samscott89, line 102: Nit: would this recompute this every time user_id is called? Would it be better to compute this outside the method definition?

samscott89: (Ditto for below)

samscott89: Oh nevermind, I think this is only computed once on defining the class

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles.py
```diff
@@ -1,16 +1,142 @@
...
+    @declared_attr
+    def resource_id(cls):
+        type = inspect(resource_model).primary_key[0].type
+        name = inspect(resource_model).primary_key[0].name
+        table_name = resource_model.__tablename__
+        return Column(type, ForeignKey(f"{table_name}.{name}"))
+
+    @declared_attr
+    def resource(cls):
+        return relationship(resource_model.__name__, backref="roles")
```
samscott89, line 120: What about if there are multiple role models for the resource? Then the backref would be ambiguous

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles.py
```diff
@@ -1,16 +1,142 @@
...
+
+    # Add the relationship between the user_model and the resource_model
+    resources = relationship(
+        resource_model.__name__,
+        secondary=tablename,
+        viewonly=True,
+        backref="users",
+        sync_backref=False,
+    )
+    # @Q: Do we try to pluralize this name correctly?
```
samscott89, line 133: Not for now, will add a task. Should document it somewhere though

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles.py
```diff
@@ -413,33 +389,16 @@ def add_user_role(session, user, resource, role_name):
...
-
-        role = role_model(**kwargs)
-        session.add(role)
-        session.commit()
+    # create and save role
+    resource_name = resource_model.__name__.lower()
+    kwargs = {"name": role_name, resource_name: resource, "user": user}
+    new_role = role_model(**kwargs)
+    session.add(new_role)
+    session.commit()
```
samscott89: Should we be doing commit here? Seems safer, but also prevents user from doing a bulk update in a single transaction.

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles.py
```diff
@@ -413,33 +389,16 @@ def add_user_role(session, user, resource, role_name):
...
-        role.users.append(user)
-    else:
-        resource_name = resource_model.__name__.lower()
-        kwargs = {"name": role_name, resource_name: resource, "users": [user]}
-
-        role = role_model(**kwargs)
-        session.add(role)
-        session.commit()
+    # create and save role
+    resource_name = resource_model.__name__.lower()
```
samscott89, line 398: These seems like something that would be worth turning into a method so we don't get it wrong somewhere

languages/python/sqlalchemy-oso/tests/test_roles.py
```diff
@@ -312,88 +320,239 @@ def test_add_user_role(test_db_session):
...
+
+    results = list(
+        oso.query_rule("inherits_role", org_owner_role, Variable("inherited_role"))
+    )
+    results.sort(key=lambda x: x.get("bindings").get("inherited_role").name)
+    assert len(results) == 2
+    assert results[0].get("bindings").get("inherited_role").name == "BILLING"
+    assert results[1].get("bindings").get("inherited_role").name == "MEMBER"
+
+    # TODO: test `role_allow`
```
samscott89, line 534: TODO Done?

languages/python/sqlalchemy-oso/tests/test_roles.py
```diff
@@ -312,88 +320,239 @@ def test_add_user_role(test_db_session):
...
+    # make sure this query fails before any rules are added
+    results = list(oso.query_rule("role_allow", john, "READ", abbey_road))
+    assert len(results) == 0
+
+    # test basic `role_allow` rule
+    role_allow_str = """role_allow(role: RepositoryRole{name: "READ"}, "READ", repo: Repository) if
+                            role.repository.id = repo.id;"""
```
samscott89, line 541: To confirm: is this condition double-checking? We _should_ already know the role applies to the repo, right?

samscott89 (commented):
This is looking good. I have a few questions, but mostly just to check if I've understood things

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles.py
leina05: This is used to generate the rbac base policy. It stores the resource and role model inputs for each role model as it is created, which we use to write the specializers in the base policy.

Just thinking about this more now, we don't check for the case in which a user creates multiple role models for the same resource. We should probably fix that.

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles.py
leina05: Sure, we could make it `actor_model`. But then the `actor` passed into the policy would need to be the type of whatever `actor_model` they pass in here.

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles.py
leina05: This is old, will clean up.

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles.py
leina05: Yes I think that's right. To make it more clear I'll change `self` to `cls`

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles.py
leina05: Realized this as well. I'm trying to think about whether it makes sense to support multiple roles types for the same resource. If not, we might want to just throw an error if someone tries to do that.

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles.py
leina05: In the docstring for `resource_role_class`. Added a comment to the docs PR so we add it there too.

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles.py
leina05: Agreed.

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles.py
leina05: True, we could add it as a flag?

languages/python/sqlalchemy-oso/tests/test_roles.py
leina05: Correct. I think I'm going to remove these.

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles.py
leina05: I updated `resource_role_class()` to throw an error for now. In the future we can think about if we need to support this.

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles.py
samscott89: Someone could write their own:

```
allow(user, action, resource) if
   group in user.groups and
    allow(group, action, resource);
```

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles.py
samscott89: The use case might be having roles for users and groups, so treating a group as an actor type.

docs/changelogs/vNEXT.rst
```diff
@@ -7,30 +7,21 @@ NEXT
...
+for various required constraints and things that can go wrong.
 
-Breaking change 1
------------------
+It also includes docs!
 
-- summary of breaking change
-
-Link to migration guide
+https://docs.oso.dev/getting-started/builtin-roles/index.html
```
samscott89, line 20:
```suggestion
https://docs.osohq.com/getting-started/builtin-roles/index.html
```

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles.py
samscott89: I'll add this as a follow up

samscott89 (approved):
LGTM!

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles.py
leina05: Oh just saw your request about a comment, will add now

PR #599, opened 2020-12-18
roles docs

saolsen:
PR checklist:

- [ ] Added changelog entry.


docs/getting-started/builtin-roles/index.rst
```diff
@@ -0,0 +1,52 @@
...
+
+Lets say you have a User class and a Widget class and you want to assign users a role for a widget like
+"OWNER" or "User". Using the roles library you can generate a WidgetRole model which allows you to assign
+a user a role for a Widget. The schema for this new model's table looks like this. 
+
+.. image:: /getting-started/builtin-roles/roles.svg
+
+It's a join table between
+User and Widget that contains some metadata like the name of the role. In SQLAlchemy library we add User.widgets
+and Widget.users as relationships you can query, as well as User.widget_roles and Widget.roles to get the roles directly.
```
leina05, line 24: We should include that the convention for naming these is `users` and `Widget.__name__ + "s"`/

leina05:
This is a really nice overview. We should add backlinks to the API docs, roles guide and sqlalchemy guides somewhere, and also might want to mention somewhere that you can still write normal allow rules alongside the allow_role rules.

Should we talk about the specific library methods (resource_role_class and enable_roles)? I'm thinking we could say something like:

The steps to use roles are:

import sqlalchemy_oso library and set up the Oso instance for your app
create role models
enable the role base policy and write role rules in the policy
For an in-depth guide, checkout link to guides
This is kinda redundant of the guides themselves though

leina05:
Closing because this was merged into #615

PR #616, opened 2020-12-19
Use django _meta to resolve related fields.

samscott89:
The previous behaviour didn't work when going in the _other_ direction on a m2m with a related name.

E.g.

```
class Post(models.Model):
       tags = models.ManyToManyField(Tag, related_name="posts")

class Tag(models.Model):
   pass
```

Then `getattr(Tag, "posts")` returns the `tags` field on `Post, so `field.related_model` returns back `Tag` instead of `Post`.

Using the `_meta` api seems like the correct way to do this: https://docs.djangoproject.com/en/3.1/ref/models/meta/

I tried to write a test for this, but it kinda relies on writing something like:

```
post in tag.posts and
post matches test_app2::Post
```

which doesn't work right (this is probably a separate bug).


dhatch (approved):
lgtm, added a test

PR #627, opened 2020-12-23
Add PolarClass implementation for uuid::Uuid v0.7 & v0.8

Guara92:
Following the work done in https://github.com/osohq/oso/pull/554 by @johnhalbert I implemented the PolarClass trait for uuid using the newer versions of uuid, this is required for projects using the [uuidv7](https://github.com/diesel-rs/diesel/blob/1.4.x/diesel/Cargo.toml#L27) feature of `diesel`.

PR checklist
- [x] Added changelog entry.


Guara92:

languages/rust/oso/Cargo.toml
```diff
@@ -42,6 +42,7 @@ rustyline-derive = { version = "0.3.1", optional = true }
 lazy_static = "1.4.0"
 
 uuid-06 = { package = "uuid", version = "0.6.5", optional = true }
+uuid-07 = { package = "uuid", version = ">=0.7.0, <0.9.0", optional = true }
```
samscott89, line 45: I'm not sure this will be able to support both version simultaneously?

samscott89:
Hey @Guara92! Thanks for opening a PR with this :)

languages/rust/oso/Cargo.toml
Guara92: Yes this will let the user decide which version use, the only requirement is that the version of `uuid` specified in the user's `Cargo.toml` is the same of the one which depends `oso`, specified in the user's `Cargo.lock`.

samscott89: If that's the case, do we even need a separate one for uuid 6 and 7/8?

Guara92: I think that this is the only way if you would retain the compability with uuid 0.6, but I would happy to be proved wrong.

samscott89: So as far as I can understand it. It would be fine for us to make the `uuid` feature `"*"`. And whatever version people are using it would be fine. However, if anyone is using multiple versions of uuid in one project, then PolarClass would only be implemented for one.

samscott89 (approved):
Okay, this looks good! Thanks for the contribution @Guara92!

We still need to work out a longer-term strategy for supporting 3rd party types without needing them to be added in extras, but this is good for now.

PR #628, opened 2020-12-24
[WIP] Support for async methods in Python

samscott89:
PR checklist:

- [ ] Added changelog entry.
- [x] Tests
- [ ] Docs
- [x] Support async generators like we do in Node
- [ ] Verify whether it's okay for us to be creating an event loop for the sync compat
- [ ] Fix it breaking django


samscott89:
Closing for now

PR #631, opened 2020-12-31
Ruby 3 support

gj:
PR checklist:

- [x] Added changelog entry.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -483,12 +483,12 @@ jobs:
         uses: ruby/setup-ruby@v1
         with:
           bundler-cache: true
-          ruby-version: 2.4.0
+          ruby-version: "2.4.0"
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -367,7 +367,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        ruby-version: [2.4, 2.5, 2.6, 2.7]
+        ruby-version: [2.4, 2.5, 2.6, 2.7, "3.0"]
```
samscott89, line 370: What's with the `"` around 3.0? Seems suspicious.

samscott89 (approved):
Nice

.github/workflows/release.yml
samscott89: Oh I see the commit now. Love it.

PR #634, opened 2020-12-31
Go support

samscott89:
This is a _very_ rough pass at Go support. Needs a lot of cleanup + review.
Many things I didn't really think too deeply about how to implement it, I just wanted to get it working. 

This PR also includes a few other ideas:

**Cross-language test suite**
Implemented for Python + Go (and partially for Java). However, I had to modify a bunch of the test cases to work with Go (e.g. making attributes/methods public by uppercasing the first character). So it's still a bit TBD whether this is a good approach.

**Codegen for API types**
This used [serde-reflection](https://docs.rs/serde-reflection/0.3.2/serde_reflection/index.html) to register all types that cross the FFI boundary into a generic format. That code is not part of this PR but is in #663.

Various outstanding Go questions:
- Go doesn't have a native idea of a constructor. Currently I instantiate a struct by reflection. Either by index for varags, or by field name for kwargs. Should this be done differently?
- How to support things like Iterators and Comparables? I've implemented a few interfaces in interface.go which I use in the tests. Is this reasonable? What about channels? Should those be treated like generators?
- Registering classes/classes as constants. Right now I just register the zero value for a type as a class. This seems to work okay
- Floats: Go serializes nan/inf as null. If we want to support nan and inf we'll need to define our own float type and serializes it differently.
- Inheritance etc. Not sure if this makes any sense in Go. Maybe for interfaces? issubclass if type implements the interface? 
- Pretty sure the way I'm doing `none{} ` for `nil` makes no sense.

PR checklist:

- [x] Break out cross-language test into separate PR
- [x] Break out codegen into separate PR
- [x] Comments
- [x] Review Go library
- [ ] Try using this in an example Go app
- [x] Packaging
- [ ] Docs (docs.osohq.com)
- [ ] Further tests
- [ ] Added changelog entry.


languages/go/pkg/ffi.go
```diff
@@ -0,0 +1,251 @@
+package oso
```
cobraz, line 1: I would probably split this out as a separate `ffi` pkg. e.g. `github.com/osohq/oso/languages/go/pkg/ffi`. (Potentially `github.com/osohq/oso/languages/go/internal/pkg/ffi`)

languages/go/pkg/host.go
```diff
@@ -0,0 +1,325 @@
+// """Translate between Polar and the host language (Python)."""
```
cobraz, line 1: I would probably split this out as a separate host `pkg`. e.g. `github.com/osohq/oso/languages/go/pkg/host`. (Potentially `github.com/osohq/oso/languages/go/internal/pkg/host`)

languages/go/pkg/host.go
```diff
@@ -0,0 +1,325 @@
...
+	// Idea is that if the right class is less specific
+	// then it can be assigned to the left class?
+	if (*rightClass).AssignableTo(*leftClass) {
+		return true, nil
+	}
+	return false, nil
+}
+
+func (h Host) toPolar(v interface{}) (*Value, error) {
+	switch v := v.(type) {
```
cobraz, line 169: Would probably replace this with a Map.

languages/go/pkg/query.go
```diff
@@ -0,0 +1,329 @@
...
+		if err != nil {
+			return nil, err
+		}
+		var event QueryEvent
+		err = json.Unmarshal([]byte(*ffiEvent), &event)
+		if err != nil {
+			return nil, err
+		}
+
+		switch ev := event.QueryEventVariant.(type) {
```
cobraz, line 40: Would probably replace this with a Map.

languages/go/pkg/query.go
```diff
@@ -0,0 +1,329 @@
+package oso
```
cobraz, line 1: I would probably split this out as a separate host pkg. e.g. github.com/osohq/oso/languages/go/pkg/query. (Potentially github.com/osohq/oso/languages/go/internal/pkg/query)

languages/go/pkg/reflect.go
```diff
@@ -0,0 +1,133 @@
...
+func String(s string) *string {
+	return &s
+}
+
+func setFieldTo(field reflect.Value, input interface{}) error {
+	if !field.CanSet() {
+		return fmt.Errorf("cannot set field")
+	}
+	fieldType := field.Type()
+	switch fieldKind := field.Kind(); fieldKind {
```
cobraz, line 17: Should probably replace this with a Map

cobraz (commented):
Looks very good! Much of my feedback is based on structure. Maybe consider moving code that shouldn't be used by users of the package into an `internal` folder.

languages/go/pkg/host.go
samscott89: What would the map key be in this case?

languages/go/pkg/host.go
```diff
@@ -0,0 +1,325 @@
...
+package oso
+
+import (
+	"fmt"
+	"reflect"
+)
+
+type Host struct {
+	ffiPolar  PolarFfi
+	classes   map[string]reflect.Type
```
samscott89, line 12: Are `classes` and `instances` the right interfaces?

languages/go/pkg/host.go
```diff
@@ -0,0 +1,328 @@
...
+	}
+	return nil, &UnregisteredClassError{name: name}
+}
+
+func (h Host) cacheClass(cls reflect.Type, name string) error {
+	if v, ok := h.classes[name]; ok {
+		return &DuplicateClassAliasError{name: name, cls: cls, existing: v}
+	}
+	h.classes[name] = cls
+	return nil
```
leina05, line 57: This returns "name" in Python

saolsen (commented):
TODOs found during review

- [ ] Oso type and methods.
- [ ] Repl.
- [ ] Work through the implications of registering classes as constants of the 0 value.
- [ ] Write some tests for class methods and classes to figure out how they work.
- [ ] Debug events.
- [ ] Probably don't need to handle queryeventnone
- [ ] Confusing logic for variadic function args and also for setFieldTo, needs refactor / explanation.
- [ ] Comparers for builtin types, tests of them, maybe a better way to do it.
- [ ] Decide on constructors (reflection thing vs register constructer vs don't allow new and just use class methods)
- [ ] Punting on any sort of inheritance, can get some feedback or think it through and add it later.

languages/go/polar.go
```diff
@@ -0,0 +1,198 @@
...
+//                             print(variable + " = " + repr(value))
+//                     else:
+//                         print(True)
+//             except PolarRuntimeError as e:
+//                 print_error(e)
+//                 continue
+//             if not result:
+//                 print(False)
+
+func (p Polar) RegisterClass(cls reflect.Type, name *string) error {
```
leina05, line 175: This is using a pointer to signify an optional argument, but it's pretty annoying to have to create a variable to hold the string here, and doesn't really make sense for it to be a pointer. Thinking we should just have `RegisterClassWithName` instead

languages/go/reflect.go
```diff
@@ -0,0 +1,68 @@
...
+	"fmt"
+	"reflect"
+)
+
+func String(s string) *string {
+	return &s
+}
+
+func SetFieldTo(field reflect.Value, input interface{}) error {
+	// todo: explain how this works and why.
```
plotnick, line 13: Best todo ever.

plotnick (changes requested):
I think it would be good to get some Go-experienced eyes on this, but the Polar parts all seem straightforward and fine, with the exception of the noted precision-losing casts.

The only major problem I found was in local testing. All the tests pass, but I couldn't reliably make them _fail_ by twiddling the `test/spec/*.yml` files. For instance, if I change the query in `basic.yml` to `1 = 2` and run `make -C languages/go test`, they all still pass. There seems to be some sort of caching going on, though I'm unsure exactly where. But I think that if we're going to use non-source test files, updates to them must be reflected in the next `make test` run.

languages/go/polar_types.go
```diff
@@ -0,0 +1,2735 @@
+package oso
```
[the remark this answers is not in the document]
samscott89: This is a great suggestion, but it reminded me there's actually a standard format for this:
`^// Code generated .* DO NOT EDIT.$`

https://golang.org/cmd/go/#hdr-Generate_Go_files_by_processing_source

polar-core/src/numerics.rs
```diff
@@ -46,7 +46,21 @@ where
         type Value = f64;
 
         fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
-            formatter.write_str("JSON encoded data")
+            formatter.write_str("An integer (42), a float (1.2), or a string (\"NaN\")")
```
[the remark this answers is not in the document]
samscott89: We should probably be implementing a visitor for `Numeric` instead of just for float. I think that is closer to how JSON represents them - numbers are either float or ints.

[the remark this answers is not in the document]
samscott89: https://docs.serde.rs/serde_json/struct.Number.html is basically what we're working with

samscott89:
> The only major problem I found was in local testing. All the tests pass, but I couldn't reliably make them _fail_ by twiddling the `test/spec/*.yml` files. For instance, if I change the query in `basic.yml` to `1 = 2` and run `make -C languages/go test`, they all still pass. There seems to be some sort of caching going on, though I'm unsure exactly where. But I think that if we're going to use non-source test files, updates to them must be reflected in the next `make test` run.

The slightly unbelievable answer:

> To disable test caching, use any test flag or argument other than the cacheable flags. The idiomatic way to disable test caching explicitly is to use -count=1.

Cargo.lock
```diff
@@ -2,9 +2,9 @@
 # It is not intended for manual editing.
 [[package]]
 name = "aho-corasick"
-version = "0.7.15"
+version = "0.7.13"
```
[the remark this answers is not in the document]
saolsen: fixed

languages/go/ffi.go
```diff
@@ -0,0 +1,252 @@
...
+func (q *QueryFfi) delete() {
+	C.query_free(q.ptr)
+	q = nil
+}
+
+func (q QueryFfi) nextMessage() *C.char {
+	return C.polar_next_query_message(q.ptr)
+}
+
+func (q QueryFfi) callResult(callID int, term *Term) error {
```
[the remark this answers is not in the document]
saolsen: Ah yeah, I fixed this for instance_id but not call_id, will update

plotnick (approved):

languages/go/host.go
```diff
@@ -0,0 +1,289 @@
...
+		case uint:
+			intVal = int64(vv)
+		case uint8:
+			intVal = int64(vv)
+		case uint16:
+			intVal = int64(vv)
+		case uint32:
+			intVal = int64(vv)
+		case uint64:
+			intVal = int64(vv)
```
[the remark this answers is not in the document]
saolsen: :+1:

languages/go/tests/parity_test.go
```diff
@@ -0,0 +1,550 @@
...
+		return Result{result}
+	case []interface{}:
+		result := make([]Result, len(inputVal))
+		for idx, v := range inputVal {
+			result[idx] = NewResult(v)
+		}
+		return Result{result}
+	case uint64:
+		// standardise uints to ints
+		return Result{inner: int64(inputVal)}
```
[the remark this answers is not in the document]
saolsen: yeah, this was a hack because we didn't handle u64 right, removing now that we do

PR #637, opened 2021-01-07
Partial unification

dhatch:
### 
PR checklist:

- [ ] Added changelog entry.

PR #638, opened 2021-01-08
Partial unification

plotnick:
Replace `Partial` with `Expression`, and use operations on unbound (or cyclically bound) variables to construct constraint expressions. Adds the ability to unify partials, as well as track partials "entangled" through other relations. Detects (some) inconsistencies at query time, and will ground out partials it can prove are consistent.

New & improved version of #637.

PR checklist:
- [ ] Docs.
- [ ] Changelog entry.


polar-core/src/partial/isa_constraint_check.rs
```diff
@@ -42,36 +52,90 @@ impl IsaConstraintCheck {
...
+                            left_class_tag: existing.tag.clone(),
+                            right_class_tag: proposed.tag.clone(),
+                        }),
+                    )
+                }
+                _ => (None, None),
+            }
+        } else if constraint_path.len() > proposed_path.len() {
+            // comparing existing x.a.b matches B{} vs. proposed x.a matches A{}
+            panic!("AAAAAAAAAAAAAAAAAAAA");
```
dhatch, line 110: fix panic

polar-core/src/partial/isa_constraint_check.rs
```diff
@@ -4,20 +4,30 @@ use crate::events::QueryEvent;
 use crate::runnable::Runnable;
 use crate::terms::{Operation, Operator, Pattern, Term, Value};
 
+fn path(x: &Term) -> Vec<Term> {
```
dhatch, line 7: haven't thoroughly reviewed this file yet

polar-core/tests/integration_tests.rs
```diff
@@ -314,7 +314,7 @@ fn test_functions() -> TestResult {
     qnull(&mut p, "k(1)");
     qeval(&mut p, "k(2)");
     qnull(&mut p, "k(3)");
-    qvar(&mut p, "k(a)", "a", values![2]);
+    // qvar(&mut p, "k(a)", "a", values![2]);
```
dhatch, line 317: Why is this commented out?

polar-core/tests/integration_tests.rs
```diff
@@ -597,11 +597,11 @@ fn test_not() -> TestResult {
 
     // Negate And with unbound variable.
     p.load_str("i(x,y) if not (y = 2 and x = 1);")?;
-    qvar(&mut p, "i(2,y)", "y", values![sym!("_y_44")]);
+    qvar(&mut p, "i(2,y)", "y", values![sym!("y")]);
```
dhatch, line 600: So a variable being bound to itself in the output now means it is unbound? I don't think the previous output (bound to a temporary) was better, but I could see this being a source of confusion.

dhatch (commented):
Comments inline - perhaps we can try to address together.

VM has been reviewed thoroughly, simplifier & inverter less so.  Haven't read isa constraint check completely.

polar-core/src/partial/simplify.rs
```diff
@@ -1,270 +1,361 @@
...
+    this_var: Symbol,
 }
 
-pub struct Simplifier;
 impl Folder for Simplifier {
-    /// Deduplicate constraints.
-    fn fold_partial(&mut self, partial: Partial) -> Partial {
-        fold_partial(self.deduplicate_constraints(partial), self)
+    fn fold_term(&mut self, t: Term) -> Term {
+        fold_term(self.deref(&t), self)
```
samscott89, line 140: This recurses infinitely for me with a sample policy. Reproed test case:
```
    #[test]
    fn test_partial_matches_with_cycle() -> TestResult {
        let p = Polar::new();
        p.load_str(
            r#"
                f(x) if g(x) and h(x);
                g(x) if x.foo = x;
                h(x: Integer);
            "#,
        )?;

        let mut q = p.new_query_from_term(term!(call!("f", [sym!("x")])), false);
        assert_query_done!(q); // ignoring for the sake of brevity what the result _should_ be
        Ok(())
    }
```

The use of matches is so that it hits the code path in `isa_expr` on vm.rs line 1252.

polar-core/src/partial/partial.rs
```diff
@@ -244,7 +244,10 @@ impl Operation {
     pub fn clone_with_new_constraint(&self, constraint: Term) -> Self {
         assert_eq!(self.operator, Operator::And);
```
dhatch, line 245: Check that `constraint` is actually an expression?

polar-core/src/partial/partial.rs
```diff
@@ -772,10 +994,30 @@ mod test {
                 e => panic!("not bindings: {:?}", e),
             }
         };
+
+        // Register `x` as a partial.
+        p.register_constant(
```
dhatch, line 999: Should we use `bind` here now? (And elsewhere that `register_constant` is used)

polar-core/src/partial/partial.rs
```diff
@@ -998,6 +1240,31 @@ mod test {
...
+        p.load_str(r#"f(x) if not (x.y = 1 and x.b = 2);"#)?;
+
+        let mut q = p.new_query_from_term(term!(call!("f", [sym!("x")])), false);
+        assert_partial_expression!(next_binding(&mut q)?, "x", "1 != _this.y or 2 != _this.b");
+        assert_query_done!(q);
+
+        Ok(())
+    }
+
+    #[ignore]
```
dhatch, line 1255: Is this test intentionally ignored? If so, can we leave a comment and record this TODO somewhere?

polar-core/src/partial/partial.rs
```diff
@@ -1095,6 +1363,49 @@ mod test {
...
+        p.load_str(
+            r#"f(x) if x > 1 and not (x < 0);
+               g(x) if x > 1 and not (x = 2);"#,
+        )?;
+
+        let mut q = p.new_query_from_term(term!(call!("f", [sym!("x")])), false);
+        assert_partial_expression!(next_binding(&mut q)?, "x", "_this > 1 and _this >= 0");
+        assert_query_done!(q);
+
+        // TODO(ap): The below fails because of interaction between the simplifier and inverter.
```
dhatch, line 1404: Note: record bug somewhere

polar-core/src/partial/simplify.rs
```diff
@@ -95,11 +93,11 @@ pub fn simplify_partial(var: &Symbol, term: Term, vm: &PolarVirtualMachine) -> T
 /// Simplify the values of the bindings to be returned to the host language.
 ///
 /// - For partials, simplify the constraint expressions.
-/// - For non-partials, deep deref.
-pub fn simplify_bindings(bindings: Bindings, vm: &PolarVirtualMachine) -> Option<Bindings> {
+/// - For non-partials, deep deref. TODO(ap/gj): deep deref.
```
dhatch, line 96: what's this todo

polar-core/src/vm.rs
```diff
@@ -1029,6 +1038,7 @@ impl PolarVirtualMachine {
     }
 
     /// Comparison operator that essentially performs partial unification.
+    #[allow(clippy::many_single_char_names)]
```
dhatch, line 1041: 😭

polar-core/tests/integration_tests.rs
```diff
@@ -1601,6 +1604,9 @@ fn test_cut() -> TestResult {
 fn test_forall() -> TestResult {
     let mut p = Polar::new();
     p.load_str("all_ones(l) if forall(item in l, item = 1);")?;
+    // not (item in l and not item = 1) _item != 2
```
dhatch, line 1607: remove this comment

polar-core/src/partial/simplify.rs
```diff
@@ -1,270 +1,390 @@
...
+                    .expect("expression")
+                    .clone(),
+                self,
+            ),
 
-#[cfg(test)]
-mod test {
-    use super::*;
-    use crate::terms::*;
+            // Non-trivial conjunctions. Choose a unification constraint to
```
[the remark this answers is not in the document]
dhatch: The reason I'm asking is I'd prefer we encode this in the comment string.  This way, the reader doesn't need to know what "trivial" means - which may vary upon the context.  In this case, it means non-empty conjunctions, right (with > 0 arguments).

polar-core/tests/integration_tests.rs
plotnick: No current reason, fixed in 7cb32750.

polar-core/tests/integration_tests.rs
plotnick: 7cb32750

PR #642, opened 2021-01-13
Expose `get_allowed_actions` method in the Python API

leina05:
PR checklist:

- [x] Added changelog entry.


languages/python/oso/oso/oso.py
```diff
@@ -38,3 +38,24 @@ def is_allowed(self, actor, action, resource) -> bool:
...
+                # TODO: is this the correct behavior?
+                if not allow_unbound:
+                    raise exceptions.OsoError(
+                        """get_allowed_actions() found an unbound action. To get allowed actions, specify allowed actions in the policy.
+    E.g., instead of `allow(actor: User, _action, resource: MyResource);`,
+    use `allow(actor: User, action, resource: MyResource) if action in ["CREATE", "READ", "WRITE", "DELETE"];`
+                        """
+                    )
+                else:
+                    return ["UNBOUND"]
```
samscott89, line 58: Minor: probably best to avoid the term "unbound". Maybe `allow_wildcard`, and returning `"*"` for example might be a bit clearer?

leina05: Yeah that's better.

languages/python/oso/oso/oso.py
```diff
@@ -38,3 +38,24 @@ def is_allowed(self, actor, action, resource) -> bool:
...
         except StopIteration:
             return False
+
+    def get_allowed_actions(self, actor, resource, allow_wildcard=False):
+        # Get allowed actions on the resource
+        results = self.query_rule("allow", actor, Variable("action"), resource)
+        actions = set()
+        for result in results:
+            action = result.get("bindings").get("action")
+            if type(action) == Variable:
```
samscott89, line 48: Some Python linters/IDEs will complain that this should be `isinstance(action, Variable)`.

languages/python/oso/oso/oso.py
```diff
@@ -38,3 +38,24 @@ def is_allowed(self, actor, action, resource) -> bool:
             return True
         except StopIteration:
             return False
+
+    def get_allowed_actions(self, actor, resource, allow_wildcard=False):
```
samscott89, line 42: Needs a proper docstring. Especially to describe the wildcard behaviour.

languages/python/oso/oso/oso.py
```diff
@@ -38,3 +38,24 @@ def is_allowed(self, actor, action, resource) -> bool:
...
+        # Get allowed actions on the resource
+        results = self.query_rule("allow", actor, Variable("action"), resource)
+        actions = set()
+        for result in results:
+            action = result.get("bindings").get("action")
+            if type(action) == Variable:
+                # TODO: is this the correct behavior?
```
samscott89, line 49: Yes. Although I would change the error message slightly. Maybe:

The result of `get_allowed_actions()` contained an "unconstrained" action that could represent any action, but `allow_wildcard` was set to true. To fix, set `allow_wildcard` to true and compare with the `"*"` string.

I _think_ if someone gets back an unbound this is probably what they are looking to achieve? I like it being off-by-default so that people need to "learn" this behaviour instead of being surprised by the result, however.

languages/python/oso/tests/test_oso.py
```diff
@@ -129,5 +130,26 @@ def test_allow_model(test_oso):
...
 
+def test_get_allowed_actions(test_oso):
+    rule = """allow(_actor: test_oso::Actor{name: "Sally"}, action, _resource: test_oso::Widget{id: "1"}) if
+        action in ["CREATE", "READ"];"""
+
+    test_oso.load_str(rule)
+    user = Actor(name="Sally")
+    resource = Widget(id="1")
+    assert set(test_oso.get_allowed_actions(user, resource)) == set(
+        ["get", "CREATE", "READ"]
```
samscott89, line 141: where did "get" come from?

leina05: From a different rule that already existed in the test policy

languages/python/oso/oso/oso.py
leina05: Done.

languages/python/oso/oso/oso.py
leina05: Done.

languages/python/oso/oso/oso.py
leina05: Yeah that's a good suggestion, done.

samscott89 (approved):
LGTM!

languages/python/oso/tests/test_oso.py
samscott89: Slightly unfortunate. If there's no easy way to make the test independent, then it's not worth trying to address.

leina05: It's not that hard to change, but would rather take it as a follow-up.

PR #643, opened 2021-01-13
Fix field unification bugs

gj:


gj (commented):
@plotnick says smash it

PR #644, opened 2021-01-15
Anonymous *rest vars

gj:
Without the fix, `a` and `b` are linked through the variable `*_`, and `a` is bound to `[1]`. After the change, `a` is (correctly) bound to `[*_236]` and has no relationship with `b`.

PR checklist:

- [ ] Added changelog entry.


polar-core/src/rewrites.rs
```diff
@@ -160,6 +160,14 @@ impl<'kb> Folder for Rewriter<'kb> {
         }
     }
 
+    fn fold_rest_variable(&mut self, v: Symbol) -> Symbol {
```
saolsen, line 163: Uh, I don't really get this, this looks like a function that isn't called anywhere?

saolsen: oh oh n/m I get it, lol was odd at first look didn't realize this was a method overload.

saolsen (approved):

PR #646, opened 2021-01-15
Address remaining VM todos

gj:


gj (commented):
@plotnick approves merging this into plotson/field-unification

PR #648, opened 2021-01-16
Make django filter builder use subqueries + existence to handle partials with in

samscott89:


languages/python/django-oso/tests/test_post_relationship.py
```diff
@@ -512,21 +512,21 @@ def test_empty_constraints_in(tag_nested_many_many_fixtures):
...
-        + ' (SELECT U0."id"'
-        + ' FROM "test_app2_post" U0 LEFT OUTER JOIN "test_app2_post_tags" U1 ON (U0."id" = U1."post_id")'
-        + ' GROUP BY U0."id", U0."contents", U0."access_level", U0."created_by_id", U0."needs_moderation"'
-        + ' HAVING COUNT(U1."tag_id") > 0)'
-    )
+    # assert str(authorize_filter).startswith(
+    #     "(AND: (NOT (AND: ('pk__in', []))), ('pk__in', <django.db.models.expressions.Subquery object at "
+    # )
+    posts = Post.objects.filter(authorize_filter).distinct()
+    # assert (
```
samscott89, line 519: TODO

languages/python/django-oso/tests/test_post_relationship.py
```diff
@@ -542,22 +542,22 @@ def test_in_with_constraints_but_no_matching_objects(tag_nested_many_many_fixtur
...
-        + ' "test_app2_post"."created_by_id", "test_app2_post"."needs_moderation"'
-        + ' FROM "test_app2_post"'
-        + ' INNER JOIN "test_app2_post_tags"'
-        + ' ON ("test_app2_post"."id" = "test_app2_post_tags"."post_id")'
-        + ' INNER JOIN "test_app2_tag"'
-        + ' ON ("test_app2_post_tags"."tag_id" = "test_app2_tag"."id")'
-        + ' WHERE "test_app2_tag"."name" = bloop'
-    )
+    # assert (
+    #     str(posts.query)
```
samscott89, line 551: TODO

languages/python/django-oso/django_oso/partial.py
```diff
@@ -22,14 +24,193 @@
...
+    "Eq": COMPARISONS["Eq"],
+    "Neq": COMPARISONS["Neq"],
+    "Lt": COMPARISONS["Gt"],
+    "Leq": COMPARISONS["Geq"],
+    "Gt": COMPARISONS["Lt"],
+    "Geq": COMPARISONS["Leq"],
+}
 
 
+# TODO: put in comparisons dict above?
```
gj, line 40: We didn't want to include `In` in the comparisons dict because `in` operations aren't really comparisons. We thought about renaming `COMPARISONS` to a few broader alternatives (that I can't remember at the moment), but ultimately decided against it.

samscott89: Yeah that's what I thought

languages/python/django-oso/tests/test_post_relationship.py
```diff
@@ -541,23 +545,19 @@ def test_in_with_constraints_but_no_matching_objects(tag_nested_many_many_fixtur
...
+    expected = """
+        SELECT "test_app2_post"."id", "test_app2_post"."contents", "test_app2_post"."access_level", "test_app2_post"."created_by_id", "test_app2_post"."needs_moderation"
+        FROM "test_app2_post"
+        WHERE "test_app2_post"."id" IN (SELECT DISTINCT V0."id"
+                                        FROM "test_app2_post" V0
+                                        LEFT OUTER JOIN "test_app2_post_tags" V1
+                                        ON (V0."id" = V1."post_id")
+                                        WHERE EXISTS(SELECT U0."id", U0."name", U0."created_by_id", U0."is_public"
+                                                     FROM "test_app2_tag" U0
+                                                     WHERE (U0."id" = V1."tag_id" AND U0."name" = bloop)))
```
gj, line 558: Would need to read into it a bit more, but at first blush going from a pair of `INNER JOIN`s to an `IN` + a `LEFT OUTER JOIN` + an `EXISTS` seems like a regression.

samscott89: I sort of agree, and it would be straightforward to revert that change. The reason I liked it, is because:
-  it matches the Polar semantics way closer
- there's no need to differentiate between an in with constraints, and one without
- the intersection case will "just work" (hopefully)
- and when we do negation, you can just negate the exists and you're done.

languages/python/oso/polar/partial.py
```diff
@@ -18,15 +23,9 @@ def dot_path(expr):
     _this.created_by => ('created_by',)
     _this.created_by.username => ('created_by', 'username')"""
 
-    if isinstance(expr, Variable) and expr != Variable("_this"):
+    if isinstance(expr, Variable):
```
dhatch, line 26: Are SQLAlchemy tests still passing with these changes? Is the doc comment above accurate still?

gj: I doubt it, and no

languages/python/django-oso/tests/test_post_relationship.py
```diff
@@ -579,18 +579,138 @@ def test_reverse_many_relationship(tag_nested_many_many_fixtures):
...
+        WHERE (EXISTS(SELECT U0."id"
+                      FROM "test_app2_user" U0
+                      WHERE (U0."id" = "test_app2_user_posts"."user_id"
+                             AND U0."id" = "test_app2_user_posts"."user_id"
+                             AND U0."id" = "test_app2_user_posts"."user_id"
+                             AND U0."id" > 1
+                             AND U0."id" <= 2))
+               AND EXISTS(SELECT U0."id"
+                          FROM "test_app2_user" U0
+                          WHERE (U0."id" = "test_app2_user_posts"."user_id"
```
dhatch, line 663: We are generating 3 identical clauses in this where.

samscott89: We _are_ asking it to do this three times, so kinda makes sense.

languages/python/oso/polar/partial.py
gj: updated

languages/python/django-oso/tests/test_post_relationship.py
gj: updated to not do this

PR #653, opened 2021-01-18
Support webpack builds of oso.

samscott89:
I think this needs more testing + docs before making this officially released. But good enough for us to use I think.

PR checklist:

- [ ] Added changelog entry.


samscott89:
Example webpack config using this here: https://github.com/osohq/oso/pull/654/files#diff-68dc684ad238439cb6066f6fffddfd6c259fd01cc3162a1c5a377f24bdaad9ca

languages/js/test/parity.ts
```diff
@@ -103,7 +103,7 @@ oso.registerClass(E);
   await oso.loadStr('?= MyMath.acos(1.0) = 0;');
 
   // test iterables work
-  if ((await oso.queryRule('testIterables').next()).done) throw new Error();
+  // if ((await oso.queryRule('testIterables').next()).done) throw new Error();
```
samscott89, line 106: This line was failing. Opened a new task to address this, and another to make sure CI actually catches these failures.

saolsen (approved):
This is cool! I seems like it published the browser version to a different place. Are we gonna try to make that it's own node package we publish? Maybe for this release we could just have it in the github assets and let people download and try it. Or is there a way to make a universal npm package that works for both?

Either way I think this PR is good.

samscott89:
> This is cool! I seems like it published the browser version to a different place. Are we gonna try to make that it's own node package we publish? Maybe for this release we could just have it in the github assets and let people download and try it. Or is there a way to make a universal npm package that works for both?
> 
> Either way I think this PR is good.

It's actually all in the same package. The package.json supports different entry points. So the `main` entry is for node, and `browser` for browser.

saolsen:
So dope

PR #654, opened 2021-01-18
New docs site with hugo

samscott89:
PR checklist:

- [x] Address further TODOs in docs. Hide content where not ready
- [ ] ~Make nav JS a little more robust - currently throwing exceptions + some console logging~ punted
- [x] Configure baseURL for new docs to `docs.osohq.com/v2/` and setup s3 sync to push there (without --delete flag!)
- [x] asyncify loading of the wasm bundle so it doesn't slow everything down
- [x] Added changelog entry.


gj (approved):
Shipping now since it's already been live for a couple days. Will address issues identified during a thorough review in a follow-up PR.

PR #655, opened 2021-01-18
SQLAlchemy adapter

dhatch:
PR checklist:

- [ ] Added changelog entry.


languages/python/sqlalchemy-oso/sqlalchemy_oso/preprocess.py
```diff
@@ -0,0 +1,114 @@
+"""Convert expressions from oso into a format that the SQLAlchemy translation can use."""
```
gj, line 1: This feels fine for now. I know we're probably in agreement about this, but to state the obvious: this + the Django change feel like a significant step backward from our previous goals of keeping complexity in the core so that new adapters will be easy to write and, following from that, keeping the adapter implementations roughly similar / standardized so that they're easier to keep in-sync.

PR #663, opened 2021-01-20
Polar cross-language code generation

leina05:
**Codegen for API types**
This uses serde-reflection to register all types that cross the FFI boundary into a generic format. We then walk over all the values from the registry, and pass them into a handlebars template. The idea is that we only need to write a short template for each kind of output value, to keep these consistent across languages. The result was great with the Go library. Since this is both statically typed, and doesn't have enum variants that equate to Rust types, this would have been a bit of a pain to do manually.

PR checklist:

- [ ] Added changelog entry.


samscott89:
Revisiting at a later date.

PR #665, opened 2021-01-20
cross platform parity tests

saolsen:
yml backed parity tests


samscott89:
Revisiting at a later date.

PR #666, opened 2021-01-20
Avoid combining Exist objects as filters to support django 2.2. 

samscott89:


languages/python/django-oso/Makefile
```diff
@@ -13,6 +13,10 @@ test-requirements: .make.test-requirements-install
 test: test-requirements
 	pytest tests
 
+test22: test-requirements
+	pip install django==2.2
```
gj, line 17: Good enough for now, but we should look into tox (or something similar) as a follow-up task

languages/python/django-oso/django_oso/auth.py
```diff
@@ -90,8 +90,8 @@ def authorize_model(request, model, *, actor=None, action=None) -> Q:
             filter = Q()
 
         next_filter = partial_to_query_filter(resource_partial, model)
-        if next_filter == Q():
-            return next_filter
+        if next_filter == TRUE_FILTER:
```
gj, line 93: Are there no cases where we return an empty `Q()` filter? I think that used to happen in some paths, in which case we would want to check if `next_filter` is `TRUE_FILTER` or `Q()`.

languages/python/django-oso/django_oso/models.py
```diff
@@ -32,7 +33,10 @@ def authorize(self, request, *, actor=None, action=None):
 
         # SELECT DISTINCT on inner query to support chaining methods on
         # returned QuerySet.
-        return self.filter(pk__in=self.filter(filter).distinct())
+        if filter == TRUE_FILTER:
+            return self.filter(filter)
```
gj, line 37: If it's just the `TRUE_FILTER`, what's the point of calling `self.filter` at all?

languages/python/django-oso/django_oso/partial.py
```diff
@@ -10,7 +10,10 @@
 
 from .oso import django_model_name
 
-
+# A query filter that is always true, but can still be combined
+# with other filters to narrow the scope.
+# See: https://forum.djangoproject.com/t/improving-q-objects-with-true-false-and-none/851/1
```
gj, line 15: Thanks for adding this

languages/python/django-oso/django_oso/partial.py
```diff
@@ -230,11 +232,16 @@ def not_expr(self, expr: Expression):
...
+        """For every subquery, construct a filter to make sure the result set is non-empty"""
+        if len(self.variables) == 0:
+            return self.filter
+        objects = self.model.objects.all()
         for subq in self.variables.values():
             filtered = subq.model.objects.filter(subq.finish()).values("pk")
             exists = Exists(filtered)
-            self.filter = exists & self.filter  # This _has_ to be this way around
+            name = f"{self.name}__exists"
+            objects = objects.annotate(**{name: exists}).filter(**{name: True})
```
gj, line 243: This is cute. For anyone happening upon this line in the future: https://docs.djangoproject.com/en/2.2/ref/models/expressions/#filtering-on-a-subquery-expression

languages/python/django-oso/tests/conftest.py
```diff
@@ -22,3 +23,24 @@ def pytest_configure():
...
+        return variable
+    else:
+        return f"({variable})"
+
+
+def is_true():
+    if VERSION >= (3, 1):
+        return ""
+    else:
+        return " = True"
```
gj, line 46: Oof. Not a huge fan. Would love to revisit as part of setting up tox or something similar.

languages/python/django-oso/tests/test_post_relationship.py
```diff
@@ -765,20 +790,22 @@ def test_redundant_in_on_same_field(tag_nested_many_many_fixtures):
...
-                          WHERE (U0."id" = "test_app2_post_tags"."tag_id"
-                                 AND U0."name" = random)))
+        WHERE "test_app2_post"."id" IN (SELECT V0."id"
+                                        FROM "test_app2_post" V0
+                                        LEFT OUTER JOIN "test_app2_post_tags" V1 ON (V0."id" = V1."post_id")
+                                        WHERE (EXISTS(SELECT U0."id"
+                                                      FROM "test_app2_tag" U0
+                                                      WHERE (U0."id" = {parenthesize('V1."tag_id"')}
+                                                      AND U0."name" = random)){is_true()}
+                                                      AND EXISTS(SELECT U0."id"
```
gj, line 805: This should line up with the `EXISTS` on line 801, right?

languages/python/django-oso/tests/test_post_relationship.py
```diff
@@ -765,20 +790,22 @@ def test_redundant_in_on_same_field(tag_nested_many_many_fixtures):
...
+                                                                 FROM "test_app2_tag" U0
+                                                                 WHERE (U0."id" = {parenthesize('V1."tag_id"')}
+                                                                 AND U0."is_public"{is_true()})){is_true()}))
```
gj, line 808: This is nice

languages/python/flask-oso/requirements.txt
```diff
@@ -1,2 +1,2 @@
-oso~=0.10.0
+oso==0.11.0-alpha
```
gj, line 1: Is this intentional?

languages/python/django-oso/Makefile
samscott89: Yep

languages/python/django-oso/django_oso/auth.py
samscott89: None that the tests hit, and line 90 above is now the only place we create the empty filter. I thought it would be better to remove all uses to be consistent.

languages/python/django-oso/django_oso/models.py
samscott89: I think it needs to be something, and I didn't want to return `.all()` in case it overwrites previous filters. And this gets dropped in the SQL anyway.

languages/python/django-oso/django_oso/partial.py
samscott89: Will add as a comment

languages/python/django-oso/tests/conftest.py
samscott89: I don't think it changes the need for something like this if we're going to be checking against the raw sql

languages/python/flask-oso/requirements.txt
samscott89: Yeah, tests were failing without. Didn't investigate much deeper though

languages/python/django-oso/tests/test_post_relationship.py
samscott89: Maybe?

languages/python/django-oso/tests/conftest.py
gj: Right, and we decided not to check against the Django filters because stringifying the top-level filter can't see through the nested Subquery/Exists objects...

Not sure there's a better strategy than this.

languages/python/django-oso/tests/test_post_relationship.py
gj: Oh no I'm off — where it is is 👍

languages/python/django-oso/Makefile
gj: created follow-up task

languages/python/django-oso/django_oso/partial.py
gj: @samscott89 ping

gj: done

gj (approved):

PR #672, opened 2021-02-01
updated model record field guide

saolsen:
Rework our existing guide to be more about model record and field authorization patterns instead of data access patterns.

The data access stuff was all pre list filtering so was mostly outdated but the model / record / field stuff was all good so cut out most of the data access stuff and expanded on model record field things.


docs/getting-started/application/patterns.rst
```diff
@@ -53,122 +53,61 @@ The above policy controls access to the Expense model on the **record level**.
...
-data**.  It relies on other data to make the decision: the submitter of the
-expense (``expense.submitted_by``), and the manager of the submitter
-(``submitted_by.manager``).  This *other* data is called **secondary
-authorization data**.  An important class of secondary authorization data is
-**actor data**.  This data includes properties of the actor, like their role, or
-team membership that is often used in controlling access regardless of whether
-it is over rows, columns or fields.
-
-Where the Policy is Evaluated
------------------------------
```
leina05, line 69: I actually think this content is useful. I don't think we need to make a fuss about defining terms like "policy evaluation point" unless we decide to use them consistently throughout the docs. But I do think it's worth touching on where the policy is evaluated. I think the best thing would be to have some kind of diagram that explains this better. E.g., a diagram showing the flow of the request and where the policy check is happening in this case.

docs/getting-started/application/patterns.rst
```diff
@@ -53,122 +53,61 @@ The above policy controls access to the Expense model on the **record level**.
...
 ==========================
 
-Sometimes, access control does not rely on properties of the primary data.  This
-type of access control is called **model level**.
+Sometimes, access control does not depend on the properties of the resource but just
+on it's type. This is called **model level** can be used for coarse level authorization.
+This is useful if you want to restrict access to a whole page if the user cannot see any
+of the resources of the kind the page is about. Since the rules depend only on the type
+of the model and not on the data it makes it easier to add these checks upstream in
+middleware.
```
leina05, line 64: This sentence is a little vague. Is this referring to some kind of request middleware?

docs/getting-started/application/patterns.rst
```diff
@@ -53,122 +53,61 @@ The above policy controls access to the Expense model on the **record level**.
...
+    allow(actor, "modify", "Project") if actor.role = "hr_admin";
+    allow(actor, "modify", "Organization") if actor.role = "hr_admin";
 
 This brief policy shows an example of model level access control:
 
 - An accountant can view expenses.
 - HR admins can modify teams, projects, and organizations.
 
-Notice that this policy does not rely on any **primary authorization data**.
-Therefore it can be evaluated either before or after the primary data fetch.
```
leina05, line 136: Think it might be worth keeping this in to reinforce the point, but replace "primary authorization data" with "resource attributes" or something like that.

docs/getting-started/application/patterns.rst
```diff
@@ -53,122 +53,61 @@ The above policy controls access to the Expense model on the **record level**.
...
             return NotAuthorizedResponse()
+        # Process request
 
-.. note::
-
-    You may have noticed that this policy still accesses **actor data**.  This
-    is fine, since usually this data will be fetched prior to authorization as
-    part of the authentication flow.
+Model level is also sometimes a good way to write rules about creating new resources.
+You can write a rule saying if a user can create an Expense or not.
```
leina05, line 92: It would be good to have more explanation here of why model-level access control is good for this.

docs/getting-started/application/patterns.rst
```diff
@@ -53,122 +53,61 @@ The above policy controls access to the Expense model on the **record level**.
...
+Record Level Access Control
+===========================
 
 Our :ref:`first example <first-record-level>` was an example of record level
-access control. In general, record level access control must be performed
-**after primary data fetch**. This holds true for actions that fetch, edit, or
-delete primary data. (Our example above used the ``"approve"`` action, which
-would result in an edit). An exception to this rule is actions that create
-new data.
+access control. Record level is about authorizing an action on a specific instance
```
leina05, line 98:
```suggestion
access control. Record level is about authorizing access to a specific instance
```

docs/getting-started/application/patterns.rst
```diff
@@ -53,122 +53,61 @@ The above policy controls access to the Expense model on the **record level**.
...
+of a resource. Is the user allowed to edit **this** Expense. Record level
```
leina05, line 99:
```suggestion
of a resource. It asks the question: is the actor allowed to edit **this** Expense? Record level
```

docs/getting-started/application/patterns.rst
```diff
@@ -53,122 +53,61 @@ The above policy controls access to the Expense model on the **record level**.
...
+rules are passed the instance and can check properties on it.
+
+Record level checks can also be used for create. When you're creating a new record you
+can create the instance, then check if it's an instance you are allowed to create and then
+save it to the database if it is.
```
leina05, line 104: This feels a little random without more set-up.

docs/getting-started/application/patterns.rst
```diff
@@ -282,303 +176,26 @@ We can combine field access control with our record level access control
     We use the ``oso.query`` method in this example to query a rule other than
```
leina05, line 176: We should link to the method for `oso.query`

docs/getting-started/application/patterns.rst
```diff
@@ -282,303 +176,26 @@ We can combine field access control with our record level access control
...
-        # Query oso for all fields allowed for this user.
-        allowed_fields = oso.query_rule("allow_field",
-                                   user, "view", expense, Variable("field"))
-        # Convert the returned query response into a list of fields
-        allowed_fields = [r["field"] for r in allowed_fields]
-        allowed_fields_sql = db.sql_escape(allowed_fields.join(", "))
-
-        expense = db.fetch(
-            f"SELECT {allowed_fields_sql} FROM expenses WHERE id = %",
-            expense_id)
```
leina05, line 308: This is actually very cool but definitely out of place without more explanation of stuff like `Variable`.

saolsen:
This is a post-it somewhere on dave's extended backlog, making a draft so it stops yelling in slack.

docs/getting-started/application/patterns.rst
```diff
@@ -183,51 +122,6 @@ Create Requests
             # Not authorized.
             return NotAuthorizedResponse()
 
-Here, we were able to evaluate the policy **before data fetch** (modification in
```
dhatch, line 186: Why did you want to remove this?

docs/getting-started/application/patterns.rst
```diff
@@ -282,303 +176,26 @@ We can combine field access control with our record level access control
...
-
-    Notice that we didn't have to change our policy file at all to make this
-    change from the previous example. We passed in a ``Variable`` which
-    instructs oso to find all values of ``field`` that match the rules we
-    defined in our policy.  This flexibility derives directly from writing a
-    :doc:`declarative policy in Polar </more/language/polar-foundations>`!
-
-.. _authorizing-list-endpoints:
-
-Authorizing List Endpoints
```
dhatch, line 331: I think we should keep most of this with maybe some wording updates and more context about how this compares to data filtering

dhatch: We should probably split it into a separate page about data filtering

dhatch (commented):
I am sorry this has languished for eternity.

Left some inline comments.  My main feedback is:

- The wording changes & improvements here are mostly all good
- Getting rid of primary / secondary authorization data is a good idea. that was confusing.
- We should consider keeping the content about where authorization is performed relative to data access. It's an important consideration and impacts how you integrate Oso with the application.
- The authorization access to list endpoints content is still important since we don't have universal support for data filtering. We should probably combine all this with the data filtering content on the internals page and have that be a separate page in our docs to address some of the recent feedback where people have missed data filtering in our docs.

As far as actually next steps on this... the new docs has this guide but none of the updates.

We'll need to port all these changes to the new docs and figure out the stuff I mentioned above. This might be at a decent point for Lito to pick up, I'll check w/ him / Sam on that.

PR #673, opened 2021-02-01
Go comments for API docs + changelog

leina05:
PR checklist:

- [x] Added changelog entry.

### DIFF
--- a/docs/changelogs/vNEXT.rst
+++ b/docs/changelogs/vNEXT.rst
@@ -12,7 +12,7 @@ Release 2021-2-3
 
 This week we are releasing ``go-oso`` as an alpha release!
 Download the package `here<NEED LINK>`_, and check out our `quick start<NEED LINK>`_!
-``go-oso`` supports Go versions 1.11+.
+``go-oso`` supports Go versions 1.12+.
 
 .. Include more details on why it's alpha?
 
--- a/docs/changelogs/vNEXT.rst
+++ b/docs/changelogs/vNEXT.rst
@@ -1,10 +1,20 @@
-.. title:: Changelog for Release DATE
+.. title:: Changelog for Release 2021-2-3
 .. meta::
-  :description: Changelog for Release DATE (RELEASED_VERSIONS) containing new features, bug fixes, and more.
+  :description: Changelog for Release 2021-2-3 (RELEASED_VERSIONS) containing new features, bug fixes, and more.
 
-############
-Release DATE
-############
+################
+Release 2021-2-3
+################
+
+==================================
+``go-oso`` VERSION
+==================================
+
+This week we are releasing ``go-oso`` as an alpha release!
+Download the package `here<NEED LINK>`_, and check out our `quick start<NEED LINK>`_!
+``go-oso`` supports Go versions 1.11+.
+
+.. Include more details on why it's alpha?
 
 ==================================
 ``RELEASED_PACKAGE_1`` NEW_VERSION
--- a/languages/go/interfaces/interfaces.go
+++ b/languages/go/interfaces/interfaces.go
@@ -1,10 +1,19 @@
 package interfaces
 
+/*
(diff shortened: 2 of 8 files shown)


docs/changelogs/vNEXT.rst
```diff
@@ -1,10 +1,20 @@
...
+Release 2021-2-3
+################
+
+==================================
+``go-oso`` VERSION
+==================================
+
+This week we are releasing ``go-oso`` as an alpha release!
+Download the package `here<NEED LINK>`_, and check out our `quick start<NEED LINK>`_!
+``go-oso`` supports Go versions 1.11+.
```
plotnick, line 15: I believe we need go 1.12+.

languages/go/interfaces/interfaces.go
```diff
@@ -1,10 +1,19 @@
 package interfaces
 
+/*
+Interface for values that can be compared.
+*/
 type Comparer interface {
+	// Should return `true` when values are equal; `false` when not equal.
 	Equal(other Comparer) bool
+	// Should return `true` when the comparer is less than `other`;`false` when not less than.
```
plotnick, line 9:
```suggestion
	// Should return `true` when the comparer is less than `other`; `false` when not less than.
```

languages/go/oso.go
```diff
@@ -4,12 +4,21 @@ import (
 	"reflect"
 )
 
+/*
+The central object to manage application policy state, e.g.
+the policy data, and verify requests.
```
plotnick, line 9:
```suggestion
The central object to manage policy state and verify requests.
```

languages/go/oso.go
```diff
@@ -19,26 +28,46 @@ func NewOso() (Oso, error) {
 	}
 }
 
+/*
+Load in polar policies from a ".polar" file, checking that all inline queries
+succeed.
```
plotnick, line 33:
```suggestion
Load Polar policy from a ".polar" file, checking that all inline queries succeed.
```

languages/go/oso.go
```diff
@@ -19,26 +28,46 @@ func NewOso() (Oso, error) {
...
+*/
 func (o Oso) LoadFile(f string) error {
 	return (*o.p).loadFile(f)
 }
 
+/*
+Load in polar policies from a string, checking that all inline queries succeed.
```
plotnick, line 40:
```suggestion
Load Polar policy from a string, checking that all inline queries succeed.
```

languages/go/query.go
```diff
@@ -63,6 +70,10 @@ func (q *Query) GetAllResults() ([]map[string]interface{}, error) {
 	return results, nil
 }
 
+/*
+Get the next query result (). Returns a pointer to a map of result bindings,
```
plotnick, line 74:
```suggestion
Get the next query result. Returns a pointer to a map of result bindings,
```

languages/python/oso/polar/polar.py
```diff
@@ -87,8 +87,7 @@ def __del__(self):
         del self.ffi_polar
 
     def load_file(self, policy_file):
-        """Load in polar policies. By default, defers loading of knowledge base
-        until a query is made."""
+        """Load in polar policies from a ".polar" file."""
```
plotnick, line 90:
```suggestion
        """Load Polar policy from a ".polar" file."""
```

plotnick (changes requested):
Suggested some minor tweaks to wording, but generally looks good. There seems to be a spurious library checked in, though: `languages/go/internal/ffi/native/macos/libpolar.a`.

saolsen:
> Suggested some minor tweaks to wording, but generally looks good. There seems to be a spurious library checked in, though: `languages/go/internal/ffi/native/macos/libpolar.a`.

Need to update the .gitignore since the native directory moved.

plotnick (approved):

PR #675, opened 2021-02-02
go quickstart

saolsen:


new-docs/content/any/getting-started/quickstart/index.md
```diff
@@ -68,7 +68,7 @@ allow an **actor** to perform an **action** on a **resource**.
 
   ```polar
   allow(actor: String, "GET", _expense: Expense) if
-      actor.{{< exampleGet "endswith" >}}("@example.com");
+      {{< exampleGet "endswith" >}}"@example.com");
   ```
 
   Note that the call to **{{< exampleGet "endswith" >}}** is actually calling
```
samscott89, line 74: This will look weird with the method changing above

leina05 (changes requested):
Stuff I found from trying it out:
- Server doesn't print out that it's running once it starts
- It's running on 8080 not 5050
- I don't know if this is a problem with this PR, but it is a problem:
Screen Shot 2021-02-02 at 9 47 01 AM

- This should be language-specific:
Screen Shot 2021-02-02 at 9 49 02 AM

- I know this is just an example, but you wouldn't be able to call `expense.amount` in Go because it would be private
Screen Shot 2021-02-02 at 9 45 18 AM

samscott89:
How about doing something like this in the Go code:

```
// User type
type User string

// EndsWith checks the user's email ends with a pattern
func (u User) EndsWith(pattern string) bool {
	return strings.HasSuffix(string(u), pattern)
}
```

Then you do `User(actor)` before passing to `IsAllowed`

And the policy is 

```
allow(actor: String, _, _expense: Expense) if
    actor.EndsWith("example.com");
```

The `String` specializer actually works here but is maybe confusing.

(Also, we should probably have a test somewhere that type aliases work with specializers)

samscott89:
> Then you do User(actor) before passing to IsAllowed

Okay, this works well for this example, but then doesn't work for comparing with the submitter. So you need to make `SubmittedBy` also a `User`, and implement the interface:

```

func (u User) Equal(other interfaces.Comparer) bool {
	if v, ok := other.(User); ok {
		return u == v
	} else {
		return false
	}
}
func (u User) Lt(other interfaces.Comparer) bool {
	return false
}
```

Which is a shame. I would expect the DeepEquals comparison check to work

leina05 (approved):

PR #683, opened 2021-02-02
Add docs footer.

samscott89:
Rendered:

image


new-docs/themes/tailwind/assets/css/postcss.config.js
```diff
@@ -18,6 +18,6 @@ module.exports = {
     require('autoprefixer')({
       path: [themeDir],
     }),
-    purgecss,
+    ...(process.env.HUGO_PURGECSS !== 'off' ? [purgecss] : [])
```
samscott89, line 21: @gj I'm getting the opposite problem to before - now new styles aren't being updated on changes. This seemed like a less risky workaround for now (agree that only running purge on prod build was not good).

gj: weird ¯\\_(ツ)\_/¯

PR #689, opened 2021-02-03
Remove code that shouldn't have been commited to syntax highlighting.

dhatch:


gj (commented):
What about the import of `monaco-editor`?

PR #696, opened 2021-02-05
Register constructors in Go

leina05:
Suggestion for how to implement the `new` operator in Go.

PR checklist:

- [x] Added changelog entry.


languages/go/query.go
```diff
@@ -174,6 +184,8 @@ func (q Query) handleExternalCall(event types.QueryEventExternalCall) error {
 			// construct callArgs by converting them to typed values, then call method to get results
 			for i := 0; i < end; i++ {
 				arg := args[i]
+				// callArg := reflect.ValueOf(arg)
+				// callArgs[i] = callArg
```
leina05, line 188:
```suggestion
```

languages/go/tests/parity_test.go
```diff
@@ -392,8 +392,15 @@ func String(s string) *string {
 }
 
 func (tc TestCase) setupTest(o oso.Oso, t *testing.T) error {
+	// TODO: how sketchy of a thing is this to do?
+	// is there a better way to pass around functions?
```
saolsen, line 396: I think you're good
```suggestion
```

saolsen (approved):

languages/go/oso.go
```diff
@@ -50,17 +50,19 @@ func (o Oso) ClearRules() error {
 
 /*
 Register a Go type so that it can be referenced within Polar files.
+Accepts a constructor function, or nil if no constructor is required.
 */
-func (o Oso) RegisterClass(cls reflect.Type) error {
-	return (*o.p).registerClass(cls, nil)
+func (o Oso) RegisterClass(cls reflect.Type, cptr interface{}) error {
```
dhatch, line 55: Is there a more specific type we can use for the constructor? `interface{}` accepts anything, right?

Something like a function pointer?

saolsen: Not really, there's no general function pointer type, there's just specific function types like `func (int, int) int`. We don't know any of those things since they're different for every constructor.

PR #697, opened 2021-02-05
Setup CI job for publishing docs.

samscott89:


dhatch:
Looks pretty good, I noticed that the webpack build is hard coded to `development` [here](https://github.com/osohq/oso/blob/sam/docs-ci-job/new-docs/themes/oso-webpack/webpack.config.js#L12). Is there anything we need to change with that?

samscott89:
> 
> 
> Looks pretty good, I noticed that the webpack build is hard coded to `development` [here](https://github.com/osohq/oso/blob/sam/docs-ci-job/new-docs/themes/oso-webpack/webpack.config.js#L12). Is there anything we need to change with that?

Oh damn, that would probably help with perf/file sizes.

samscott89:
> 
> 
> > Looks pretty good, I noticed that the webpack build is hard coded to `development` [here](https://github.com/osohq/oso/blob/sam/docs-ci-job/new-docs/themes/oso-webpack/webpack.config.js#L12). Is there anything we need to change with that?
> 
> Oh damn, that would probably help with perf/file sizes.

In short, yes. 8mb -> 2mb. But still large, because it's pulling in the highlighter from monaco.

gj (approved):

PR #699, opened 2021-02-08
golang windows support

saolsen:
cgo on windows is based on a MinGW toolchain, not an msvc one.

This adds a new mingw build of the static library that we can link to with cgo. Also cleans up some types and links the libs we need to link to make this work on all three platforms.

This also means that using oso on windows requires a MinGW toolchain (which I guess you wouldn't need if you didn't use any libraries that use cgo). I used [tdm-gcc](https://jmeubank.github.io/tdm-gcc/).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -111,6 +111,19 @@ jobs:
         with:
           name: oso_static_library
           path: polar-c-api/polar.h
+      - if: runner.os == 'Windows'
+        run: |
+          rustup target add x86_64-pc-windows-gnu
+          cargo build --target x86_64-pc-windows-gnu --release
+          ls target/x86_64-pc-windows-gnu
+          ls target/x86_64-pc-windows-gnu/release
+      - if: runner.os == 'Windows'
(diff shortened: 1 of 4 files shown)


saolsen:
Here's a test I did that only ran the go parts of the release job, tests pass on all platforms. https://github.com/osohq/oso/actions/runs/549455533

.github/workflows/release.yml
```diff
@@ -111,6 +111,17 @@ jobs:
         with:
           name: oso_static_library
           path: polar-c-api/polar.h
+      - if: runner.os == 'Windows'
```
samscott89, line 114: At this point I feel like we are better off splitting these into separate jobs.

saolsen: Yeah I had the same thought, I'm gonna do that in a followup.

samscott89 (changes requested):
Tried this locally and it works fine for me. (With install tdm-gcc)

What about updating the docs + changelog?

saolsen:
> Tried this locally and it works fine for me. (With install tdm-gcc)
> 
> What about updating the docs + changelog?

👍

samscott89 (approved):

PR #703, opened 2021-02-09
WIP authorize queries by action in SQLAlchemy

leina05:
WIP for using list filtering with unbound variable actions in order to return queries that are authorized by the allowed action. 

E.g., to understand the actions a user can take on the `Post` model, query for posts using SQLAlchemy: `session.query(Post)`, then call `authorize_query_by_action(query)`.  Returns a map of `action: query`, where the `query` has the appropriate authorization filter applied.

PR checklist:

- [ ] Added changelog entry.

### DIFF
--- a/languages/python/sqlalchemy-oso/sqlalchemy_oso/auth.py
+++ b/languages/python/sqlalchemy-oso/sqlalchemy_oso/auth.py
@@ -1,4 +1,4 @@
-from oso import Oso
+from oso import Oso, Variable
 from polar.partial import Partial, TypeConstraint
 
 from sqlalchemy.orm.query import Query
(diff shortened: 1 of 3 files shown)


dhatch (changes requested):
This is a cool idea for an API to add, but in SQLAlchemy it would require multiple sessions.

Each session has an internal cache that includes authorized objects.  A single session can only be used for one set of authorization parameters (user, action, oso instance). If any of those parameters are changed but the same session object is used, there is a risk of accessing data that should not be authorized.

As an example, suppose I have two models, `Post` and `Tag`. Posts have a single tag at the `post.tag` attribute. My policy is:

```
allow(_, "read", _: Post);
allow(_, "read_for_edit", _: Post);
allow(_, "read_for_edit", _: Tag);
```

If I use this API and issue a query for "read_for_edit" first, and then access some tags on the returned posts:

```
aq = authorize_query_by_action(...)
read_for_edit = aq['read_for_edit']

tags = []
for post in read_for_edit:
    tags.append(post.tag) # loads the tag into the session cache by running an authorized query for tag. this query will use the action set on the session object (which may not be read_for_edit)

read = aq['read'] # will include the same set of posts due to authorization rules
for post in read:
    assert post.tag is None # this assertion will fail because post.tag is available in the cache.
    # it should not be returned here because the "read" action cannot read posts.
```

Was this just an idea or was it for a specific user? I'm not really sure if it makes that much sense because I wouldn't expect list filtering to be used for non-read actions. Unless it's high priority we shouldn't do any follow up work in the near time, but maybe we could make sure a task is opened in the backlog with the idea & link to this PR for context?

PR #704, opened 2021-02-09
Refactor release workflow

saolsen:
I made it longer instead of shorter by splitting out the operating systems into separate jobs instead of using the matrix thing. There were just getting to be too many if os == whatever conditionals in them. I also have to save the static libs after the lib step (not just the dynamic libs) because go needs them but python also needs them and was just rebuilding so now I can re-use them and save 3 full rust builds.

I also was able to do the musl build in the linux job and the mingw build in the windows job to speed things up more.

Test run https://github.com/osohq/oso/actions/runs/552631323

I realize there's definitely some shared structure in the different platform jobs and we can pull some of those out into parameterized task definitions but I don't want to do that **yet**. I'm going to try and add in some 32 bit builds and other platforms (possibly mac M1 and linux Arm) and wanna get those in before abstracting it out.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -493,33 +493,6 @@ jobs:
         working-directory: go
 
   validate_python_linux:
-    name: Test python ${{ matrix.python-version }} on ${{ matrix.os }}
-    needs: [build_wheels]
-    runs-on: ${{ matrix.os }}
-    strategy:
-      matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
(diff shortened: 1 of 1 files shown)


saolsen:
It works https://github.com/osohq/oso/actions/runs/552631323

gj (approved):
Talked through this. Agree with general plan to add 32-bit, M1 Mac, etc. builds before we worry about creating bespoke, parameterizable GitHub Actions to abstract away a lot of this duplicated YAML.

PR #706, opened 2021-02-10
Version bump 0.10.2

gj:
We're going to release:

- `go-oso` 0.11.0-alpha.1
- `oso` 0.10.2 (JS package only)

No changes in other libraries, and it makes more sense to potentially jump from 0.10.1 -> 0.10.3 on the next changes to those libraries instead of releasing an 0.10.2 for all of them despite having zero changes.

PR checklist:

- [x] Added changelog entry.

PR #707, opened 2021-02-10
missed conf.py

saolsen:


gj:
Linking to #706 so we remember `conf.py` if we use that PR as a template for the next version bump PR.

saolsen:
Going to add the version check to the test job so we don't merge in something that will immediately break the release job.

gj (approved):
Merge when everything passes

PR #708, opened 2021-02-10
speed up test job

saolsen:
Runs all the steps of the test job in parallel.
They all depend on an initial build step that builds the rust code so it'll be in the cache. That way, all the individual language tests builds will pull the already compiled lib from the cache. Makes every individual step just as fast as doing them in series but lets us parallelize them.

Once we can pull the old docs build out of here it'll be even faster as that's the longest step.

Workflow run times vary throughout the day but to me this seems to take the test job from ~15 minutes to ~8 minutes and should be ~4 or ~5 minutes when we drop the old docs jobs.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -322,7 +322,10 @@ jobs:
       - name: Build python lib
         run: make python-build
       - name: Build js lib
-        run: make -C languages/js build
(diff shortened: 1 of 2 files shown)


dhatch (approved):
YES! YES! YES!

PR #709, opened 2021-02-10
golang repl and debugger

saolsen:
PR checklist:

- [ ] Added changelog entry.


languages/go/oso.go
```diff
@@ -150,3 +150,7 @@ func (o Oso) IsAllowed(actor interface{}, action interface{}, resource interface
 		return false, nil
 	}
 }
+
+func (o Oso) Repl() error {
```
leina05, line 154: Can we add a comment above for API docs?

languages/go/polar.go
```diff
@@ -135,7 +138,43 @@ func (p Polar) queryRule(name string, args ...interface{}) (*Query, error) {
...
+		}
+		if len(results) == 0 {
+			fmt.Println(false)
+		} else {
+			for _, bindings := range results {
+				for k, v := range bindings {
+					switch v := v.(type) {
+					// print strings with quotes but not variables or other types represented by strings
+					case string:
+						fmt.Printf("%v = %#v\n", k, v)
```
leina05, line 170: What does the `#` do?

languages/go/query.go
```diff
@@ -94,8 +97,7 @@ func (q *Query) Next() (*map[string]interface{}, error) {
 			defer q.ffiQuery.Delete()
 			return nil, nil
 		case QueryEventDebug:
-			// TODO
-			return nil, fmt.Errorf("Polar debugger is not yet implemented in Go.")
+			err = q.handleDebug(ev)
```
leina05, line 100: Do you need to return the error here?

leina05 (commented):
Looks good, a few questions. Have to try out the debugger still.

- Do we need a test for the debugger at the Go level (not sure if we have this for other libs)?
- The REPL doesn't print anything for a simple "true" with no bindings. Do we want to print "true"? Not sure what the standard behavior is here

languages/go/polar.go
saolsen: It prints the value as a go source literal, in this case that means adding quotes around it.

languages/go/query.go
saolsen: Ah yeah I think so.

languages/go/oso.go
saolsen: Yeah

languages/go/query.go
saolsen: No, I don't it gets returned below.

leina05 (approved):

PR #716, opened 2021-02-11
Move binding related methods into BindingManager.

dhatch:
PR checklist:

- [ ] Added changelog entry.


polar-core/src/bindings.rs
```diff
@@ -0,0 +1,313 @@
...
+        }
+
+        impl<'a> Folder for Derefer<'a> {
+            fn fold_term(&mut self, t: Term) -> Term {
+                match t.value() {
+                    Value::List(_) => fold_term(self.binding_manager.deref(&t), self),
+                    Value::Variable(_) | Value::RestVariable(_) => {
+                        let derefed = self.binding_manager.deref(&t);
+                        match derefed.value() {
+                            Value::Expression(_) => t,
```
samscott89, line 126: Is this possible? When does deref return an expression?

polar-core/src/bindings.rs
```diff
@@ -0,0 +1,313 @@
...
+    // TODO: Get rid of this, only used in inverter.
+    /// Check the state of `variable` at `bsp`.
+    pub fn variable_state_at_point(&self, variable: &Symbol, bsp: Bsp) -> VariableState {
+        let mut path = vec![variable];
+        while let Some(value) = self.value(path.last().unwrap(), bsp) {
+            match value.value() {
+                Value::Expression(e) => return VariableState::Partial(e.clone()),
+                Value::Variable(v) | Value::RestVariable(v) => {
+                    if v == variable {
+                        return VariableState::Cycle(path.into_iter().cloned().collect());
```
samscott89, line 152: Can't this just be `path.clone()`?

polar-core/src/bindings.rs
```diff
@@ -0,0 +1,313 @@
...
+                _ => return VariableState::Bound(value.clone()),
+            }
+        }
+        VariableState::Unbound
+    }
+
+    /// Add `term` as a constraint.
+    pub fn add_constraint(&mut self, term: &Term) -> PolarResult<()> {
+        let Operation { operator: op, args } = term.value().as_expression().unwrap();
+        assert!(
```
samscott89, line 166: Not for this PR, but just occurred to me that using assert here is not a great idea. We could have our own macro to return some PolarError

samscott89: Or maybe debug_assert plus return error in release

polar-core/src/bindings.rs
```diff
@@ -0,0 +1,313 @@
...
+            !matches!(*op, Operator::And | Operator::Or),
+            "Expected a bare constraint."
+        );
+        assert!(args.len() >= 2);
+
+        let (left, right) = (&args[0], &args[1]);
```
samscott89, line 172: Something suspicious here that we are checking `args >= 2` but we only look at two args.

There's an implicit assumption here that the third argument if it exists is always an unbound variable.
But I seem to recall hitting a case once where that's not true. Perhaps `isa` with a dictionary? 

`f(x: {y: 1})` I _think_ does `.(x, y, 1)` somewhere.

polar-core/src/bindings.rs
```diff
@@ -0,0 +1,313 @@
...
+        match (left.value(), right.value()) {
```
samscott89, line 173: Just an idea to break this up a little:

I think there are two things happening here:

1. "Promote" variables to partial, create cycle constraints where necessary
2. Constrain by the input term along with any cycle constraints created

Would it make sense to split along those lines?

polar-core/src/vm.rs
```diff
@@ -627,105 +615,21 @@ impl PolarVirtualMachine {
...
-            }
-        }
-        Ok(())
+        self.binding_manager.constrain(o)
     }
 
     /// Add a single constraint operation to the variables referenced in it.
     /// Precondition: Operation is either binary or ternary (binary + result var),
     /// and at least one of the first two arguments is an unbound variable.
     #[allow(clippy::many_single_char_names)]
```
samscott89, line 630:
```suggestion
```

polar-core/src/vm.rs
```diff
@@ -735,31 +639,19 @@ impl PolarVirtualMachine {
         for (var, value) in bindings.iter() {
             self.bind(var, value.clone());
         }
-        self.csp += bindings.len();
+        self.csp = self.bsp();
```
samscott89, line 642: Was this a bug before? It was incrementing it by the bsp?

samscott89 (commented):
I don't really have any blocking feedback. Just a few comments.

polar-core/src/bindings.rs
```diff
@@ -0,0 +1,313 @@
...
+
+#[derive(Clone, Debug)]
+pub struct Binding(pub Symbol, pub Term);
+
+// TODO This is only public for debugger and inverter.
+// Eventually this should be an internal interface.
+pub type BindingStack = Vec<Binding>;
+pub type Bindings = HashMap<Symbol, Term>;
+
+pub type Bsp = usize;
```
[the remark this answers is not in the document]
dhatch: Should be used throughout. The reason to use a type alias is to make it easy to change this interface later. We may change the internal representation of bindings in a way that makes `usize` an inconvenient marker for `bsp`.

polar-core/src/bindings.rs
```diff
@@ -0,0 +1,313 @@
...
+/// Variable binding state.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum VariableState {
+    Unbound,
+    Bound(Term),
+    Cycle(Vec<Symbol>),
+    Partial(Operation),
+}
+
+#[derive(Clone, Debug, Default)]
```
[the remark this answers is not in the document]
dhatch: Moved in future PR

polar-core/src/bindings.rs
```diff
@@ -0,0 +1,313 @@
...
+    bindings: BindingStack,
+}
+
+impl BindingManager {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Bind `var` to `val`.
+    pub fn bind(&mut self, var: &Symbol, val: Term) {
```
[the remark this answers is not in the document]
dhatch: See https://github.com/osohq/oso/pull/723/files#diff-cd1a8a2306ff7f13d326cbddac6d5d51a125e091b837906fbae6af2e18f0a99bR127

polar-core/src/bindings.rs
dhatch: This code is gone, see https://github.com/osohq/oso/pull/723/files#diff-cd1a8a2306ff7f13d326cbddac6d5d51a125e091b837906fbae6af2e18f0a99bR185

polar-core/src/bindings.rs
dhatch: Gone in https://github.com/osohq/oso/pull/723/files#diff-cd1a8a2306ff7f13d326cbddac6d5d51a125e091b837906fbae6af2e18f0a99bR185

polar-core/src/vm.rs
dhatch: This was only called during initialization once, so it was equivalent.

PR #717, opened 2021-02-11
Isa uses binding manager `add_constraint`.

dhatch:
PR checklist:

- [ ] Added changelog entry.

### DIFF
--- a/polar-core/src/bindings.rs
+++ b/polar-core/src/bindings.rs
@@ -170,7 +170,10 @@ impl BindingManager {
         assert!(args.len() >= 2);
 
         let (left, right) = (&args[0], &args[1]);
-        match (left.value(), right.value()) {
+        match (
+            extract_variable(left.value()),
+            extract_variable(right.value()),
+        ) {
             (Value::Variable(left_name), Value::Variable(right_name)) => {
                 match (
                     self.variable_state(left_name),
@@ -245,7 +248,11 @@ impl BindingManager {
                     }
                 }
             }
-            (_, _) => panic!("At least one side of a constraint expression must be a variable."),
+            (_, _) => panic!(
+                "At least one side of a constraint expression must be a variable. This is {} {}",
+                left.to_polar(),
+                right.to_polar()
+            ),
         }
 
         Ok(())
@@ -310,4 +317,20 @@ impl BindingManager {
     // relevant_bindings
     // variable_bindings
     // bindings
+}
+
+/// Get variable out of a term for ``add_constraint`` to determine where the
+/// constraint is stored.
+///
+/// If the term is a variable, the variable is returned.
+/// If the term is a dot expression, the VAR from VAR.field is returned.
+/// Otherwise, the term is returned.
+fn extract_variable(value: &Value) -> &Value {
+    match value {
+        Value::Variable(_) => value,
+        Value::Expression(expr) if expr.operator == Operator::Dot => {
+            extract_variable(expr.args[0].value())
+        }
+        _ => value,
+    }
 }
--- a/polar-core/src/vm.rs
+++ b/polar-core/src/vm.rs
@@ -130,6 +130,11 @@ pub enum Goal {
     BindBatch {
         bindings: Rc<RefCell<BindingStack>>,
     },
+
+    /// Add a new constraint
+    AddConstraint {
+        term: Term,
+    },
 }
 
 #[derive(Clone, Debug)]
@@ -496,6 +501,7 @@ impl PolarVirtualMachine {
                 .borrow_mut()
                 .drain(..)
                 .for_each(|Binding(var, value)| self.bind(&var, value)),
+            Goal::AddConstraint { term } => self.add_constraint(&term)?,
             Goal::Run { runnable } => return self.run_runnable(runnable.clone_runnable()),
         }
         Ok(QueryEvent::None)
@@ -947,6 +953,9 @@ impl PolarVirtualMachine {
             | (Value::RestVariable(l), Value::RestVariable(r)) => {
                 // Two variables.
                 match (self.variable_state(l), self.variable_state(r)) {
+                    // TODO (dhatch): This special case does not seem right to me.
+                    // Do we need this to preserve some functionality, or can we add
+                    // a constraint?
                     (VariableState::Unbound, VariableState::Unbound) => {
                         self.push_goal(Goal::Unify {
                             left: left.clone(),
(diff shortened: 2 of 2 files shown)


polar-core/src/bindings.rs
```diff
@@ -311,3 +318,19 @@ impl BindingManager {
...
     // bindings
 }
+
+/// Get variable out of a term for ``add_constraint`` to determine where the
+/// constraint is stored.
+///
+/// If the term is a variable, the variable is returned.
+/// If the term is a dot expression, the VAR from VAR.field is returned.
+/// Otherwise, the term is returned.
+fn extract_variable(value: &Value) -> &Value {
```
samscott89, line 328: When does this happen? Are there cases when we call `add_constraint` on a term like:
`x.foo <= 1` ? Why wouldn't that be rewritten

polar-core/src/vm.rs
```diff
@@ -947,6 +953,9 @@ impl PolarVirtualMachine {
             | (Value::RestVariable(l), Value::RestVariable(r)) => {
                 // Two variables.
                 match (self.variable_state(l), self.variable_state(r)) {
+                    // TODO (dhatch): This special case does not seem right to me.
```
samscott89, line 956: Agreed.

polar-core/src/vm.rs
```diff
@@ -961,41 +970,7 @@ impl PolarVirtualMachine {
...
-                        self.constrain(&e)?;
-                    }
-                    (VariableState::Partial(mut e), VariableState::Partial(f)) => {
-                        e.merge_constraints(f);
-                        let e = e.clone_with_new_constraint(
-                            op!(Isa, left.clone(), right.clone()).into_term(),
-                        );
-                        self.constrain(&e)?;
-                    }
+                    (_, _) => self.add_constraint(&term!(op!(Isa, left.clone(), right.clone())))?,
```
samscott89, line 973: Nice

polar-core/src/vm.rs
```diff
@@ -1007,10 +982,10 @@ impl PolarVirtualMachine {
                     left: x,
                     right: right.clone(),
                 })?,
-                VariableState::Cycle(c) => self.isa_expr(&cycle_constraints(c), left, right)?,
-                VariableState::Partial(e) => self.isa_expr(&e, left, right)?,
+                _ => self.isa_expr(left, right)?,
             },
             (_, Value::Variable(r)) | (_, Value::RestVariable(r)) => match self.variable_state(r) {
+                // TODO (dhatch): Is this case tested?
```
samscott89, line 988: Hmmm, yes, I think so. Look at `test_non_instance_specializers`

I think the "legitimate" use case for this is something like:

```
allow(user: User { id: id, is_staff: True }, ... ) if ...
```

I.e you want the isa to bind variables in the pattern.

polar-core/src/vm.rs
```diff
@@ -1096,48 +1070,43 @@ impl PolarVirtualMachine {
...
-                        }
-                        _ => unify,
-                    }
+                    let unify = op!(Unify, left, value);
+                    term!(unify)
                 };
 
-                let mut operation = operation.clone();
-
                 // Add all but the last field constraint to the operation.
```
samscott89, line 1087: Comment is no longer correct.

How come the so-called "bind dance"  isn't required any more?

polar-core/src/vm.rs
```diff
@@ -1096,48 +1070,43 @@ impl PolarVirtualMachine {
...
                 let var = left.value().as_symbol()?;
-                let simplified = simplify_partial(var, operation.clone().into_term());
+                // Get the existing partial on the LHS variable.
+                let partial = match self.variable_state(var) {
+                    VariableState::Partial(expr) => expr,
+                    VariableState::Cycle(c) => cycle_constraints(c),
+                    _ => panic!("Invariant violated. left must be a variable"),
+                };
+
+                let simplified = simplify_partial(var, partial.into_term());
```
samscott89, line 1105: I'm not sure if this is tested or correct.

polar-core/src/vm.rs
```diff
@@ -130,6 +130,11 @@ pub enum Goal {
     BindBatch {
         bindings: Rc<RefCell<BindingStack>>,
     },
+
+    /// Add a new constraint
+    AddConstraint {
```
samscott89, line 135: Should this be called `Constrain` to match the convention of the other goals?

polar-core/src/vm.rs
```diff
@@ -1171,50 +1138,29 @@ impl PolarVirtualMachine {
...
-                                VariableState::Partial(e) => {
-                                    unify = e.clone_with_new_constraint(unify.into_term());
-                                }
-                                VariableState::Unbound => (),
-                            }
-                        }
-                        _ => (),
-                    }
-                    operation = operation.clone_with_new_constraint(term!(unify));
+                    op!(Unify, left, v)
```
samscott89, line 1145: Should this be `Isa`?

polar-core/src/vm.rs
```diff
@@ -1517,15 +1463,16 @@ impl PolarVirtualMachine {
...
-                            self.push_goal(Goal::Unify { left, right })?;
-                        }
                         VariableState::Bound(value) => {
                             return Err(self.type_error(&left, format!("Can only assign to unbound variables, {} is bound to value {}.", var.to_polar(), value.to_polar())));
                         }
+                        // TODO should this one be allowed?
                         VariableState::Partial(e) => {
                             return Err(self.type_error(&left, format!("Can only assign to unbound variables, {} is bound to expression {}.", var.to_polar(), e.to_polar())));
                         }
+                        _ => {
```
samscott89, line 1473: Nit: Personally I think it's clearer to have the extra variant case spelled out.

samscott89 (approved):
Nice, this looks like a good improvement.

polar-core/src/bindings.rs
dhatch: Removed in new code

polar-core/src/vm.rs
dhatch: Alex removed in https://github.com/osohq/oso/pull/723/commits/0231cc74c6eaad57b362b40baffbd8026a2153c0

polar-core/src/vm.rs
dhatch: There are some integration tests, but it's not well tested.

polar-core/src/vm.rs
dhatch: Not sure, I have no idea what that comment was referring to.

polar-core/src/vm.rs
dhatch: Yeah, it should

dhatch: Looks like the non constraint code uses Isa, not Unify.

PR #719, opened 2021-02-11
add version check to test workflow

saolsen:
Add version check to the test workflow so we catch it before the release.

### DIFF
--- a/.github/actions/check_version/action.yml
+++ b/.github/actions/check_version/action.yml
@@ -5,29 +5,37 @@ runs:
   steps:
     - name: Set version env
       run: echo "oso_version=$(cat VERSION)" >> $GITHUB_ENV
+      shell: bash
     - name: Check python version matches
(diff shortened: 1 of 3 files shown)


.github/actions/check_version/action.yml
```diff
@@ -0,0 +1,41 @@
+name: "Check Version"
+description: "Checks that the oso version numbers all match"
+runs:
+  using: "composite"
```
gj, line 4: cool, hadn't seen this before

.github/workflows/test.yml
```diff
@@ -19,6 +19,9 @@ jobs:
             target
           key: ${{ runner.os }}-cargo-check-${{ hashFiles('**/Cargo.lock') }}
 
+      ## Check Version
+      - uses: ./.github/actions/check_version
```
gj, line 23: Not a huge deal, but this could go before the cache step

gj (approved):

PR #732, opened 2021-02-16
Workaround to make Go icon scale properly in code blocks.

samscott89:
Remove unused icons


plotnick (approved):
Indeed, makes Go icon scale properly.

PR #734, opened 2021-02-17
Alternate way to make mv work

samscott89:
`rm` works if the file exist, but not if the file _doesn't_ exist


plotnick (approved):
Thanks, Microsoft.

PR #741, opened 2021-02-17
Fix docs requirements 

samscott89:
Previously we ran tests and implicitly assumed deps were installed before

Then we made this better (in some ways) so that we would install these.

But installing these clashes with how we produce the built docs.

So this PR makes the split more explicit.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -122,7 +122,10 @@ deps-test:
 # in release mode, this means they should be installed
 # from the wheels
 check-deps:
-	python3 -c "import pkg_resources; pkg_resources.require(open('requirements-docs-test.txt'))"
+	pip3 show oso
+	pip3 show django-oso
+	pip3 show flask-oso
+	pip3 show sqlalchemy-oso
 
 submodules:
 	git submodule update --checkout --init --recursive --remote
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -50,10 +50,10 @@ _api_docs: _api_docs/ruby _api_docs/java _api_docs/js/node
 
 # Catch-all target: route all unknown targets to Sphinx using the new
 # "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).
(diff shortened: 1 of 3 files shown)


docs/Makefile
```diff
@@ -115,6 +115,18 @@ test: doctest
...
+	pip3 install -r requirements-docs-test.txt
+
+# Check that all the dependencies have been installed
+# in release mode, this means they should be installed
+# from the wheels
+check-deps:
+	pip3 show oso
+	pip3 show django-oso
+	pip3 show flask-oso
+	pip3 show sqlalchemy-oso
```
plotnick, line 128: Adapted from https://stackoverflow.com/a/65606063
```suggestion
	python3 -c "import pkg_resources; pkg_resources.require(open('requirements-docs-test.txt'))"
```
This way we only have to update deps in one place.

plotnick (approved):
One non-blocking suggestion (use Python to read the requirements in the makefile), otherwise looks good.

I'm not clear on why the oso-sqlalchemy-roles-guide is being updated, though. I don't think it's blocking since it seems purely additive, but maybe it should have a seperate PR?

samscott89:
> 
> 
> One non-blocking suggestion (use Python to read the requirements in the makefile), otherwise looks good.
> 
> I'm not clear on why the oso-sqlalchemy-roles-guide is being updated, though. I don't think it's blocking since it seems purely additive, but maybe it should have a seperate PR?

Thanks Alex! That's way better.

I updated the roles guide earlier this week or last week because it was missing dependencies. I noticed it was out of date because docs tests update the submodules, so figured I would include it here.

docs/Makefile
samscott89: Ah no this doesn't work, because they are all `-e .../` links

samscott89: This will all be gone soon, so its fine

samscott89:
Also checked the docs publish job works by deploying to docs-preview.

PR #743, opened 2021-02-18
update assignment op to work with partially bound variables

leina05:
PR checklist:

- [ ] Added changelog entry.


polar-core/src/vm.rs
```diff
@@ -1452,20 +1452,7 @@ impl PolarVirtualMachine {
...
-                            return Err(self.type_error(
-                                &left,
-                                format!(
-                                    "Can only assign to unbound variables, {} is not unbound.",
-                                    var.to_polar()
-                                ),
-                            ));
-                        }
-                    },
+                    (Value::Variable(var), _) => self.bind(var, right)?,
```
dhatch, line 1455: If the intention is to fail on bound values, we need to first check the variable state. If `right` is an unbound variable and `var` is bound `bind` will bind `right` to the value of `var`.

With two variable arguments `bind` binds them together and ensures they have the same value, it is bi-directional.

(This part of the interface was a bit up for debate, so if it's counterintuitive we could change that)

polar-core/tests/integration_tests.rs
```diff
@@ -1711,8 +1711,13 @@ fn test_assignment() {
...
-        s == "Can only assign to unbound variables, x is not unbound."
+        RuntimeError::IncompatibleBindings { msg: s },
+        s == "Cannot rebind Symbol(\"x\")"
+    );
+    qruntime!(
+        "x < 3 and x := 6",
+        RuntimeError::IncompatibleBindings { msg: s },
+        s == "Grounding failed"
     );
     qnull(&mut p, "x := 5 and x > 6");
```
dhatch, line 1722: Add a test case like `x = 1 and x := y`. This should fail still

dhatch (changes requested):

polar-core/src/vm.rs
leina05: oh, interesting. Okay

dhatch:
I am going to add a task to the next cycle to continue the conversation on how this should behave & get something merged to fix it if needed.

PR #744, opened 2021-02-18
Build local deps before making docs for preview.

samscott89:
Still need to install these for publishing the preview docs.

### DIFF
--- a/.github/workflows/docs-preview.yml
+++ b/.github/workflows/docs-preview.yml
@@ -53,7 +53,7 @@ jobs:
 
       ### Build old docs
       - name: Build docs
-        run: SPHINXOPTS="-W" make -C docs build
+        run: SPHINXOPTS="-W" make -C docs deps-test build
 
       ### Build new docs
       - name: Setup Hugo


.github/workflows/docs-preview.yml
```diff
@@ -53,7 +53,7 @@ jobs:
 
       ### Build old docs
       - name: Build docs
-        run: SPHINXOPTS="-W" make -C docs build
+        run: SPHINXOPTS="-W" make -C docs deps-test build
```
dhatch, line 56: What changed that meant we needed to install these here?

dhatch (approved):

.github/workflows/docs-preview.yml
samscott89: #741 

I'm loathe to spend too much time thinking about this, since we'll be deleting almost the entirety of this directory in a few weeks time.

samscott89: Which is me acknowledging this approach isn't great, but enough for now(?)

dhatch: Ok didn't realize this was building on something probably @plotnick should review since he reviewed that but also agreed this doesn't seem too important.

PR #750, opened 2021-02-22
Performance Improvements

saolsen:
This pr includes 4 different performance improvements that I did while profiling and digging into the slow parts of this query. 
https://github.com/saolsen/oso-casemgmt-django/blob/steve/perf/casemgmt/management/commands/slow_query.py

1) When simplifying a partial we get a lot of unifications of temp vars. This is because we bind a lot of new temp vars when we call methods and rules and those all get gathered up. 
Expressions end up looking like `a = b and b = c and c = d and d = e and e = f...`
Before in the simplifier we could only clean up one of these every loop which meant we looped a lot of times. I added a way to detect all these cycles in one pass so we can clean up big expressions like that in one iteration instead of n. This approximately halved the amount of times around the simplifier loop we went for the queries I was investigating.

2) When we simplify bindings we simplify ALL the bindings. There will be a binding for every temporary variable in the expressions (which, as we saw from the previous problem, is a lot). We only care about variables that are not temp, and temp variables that are used in expressions bound to non temp variables so instead of simplifying all the bindings I made it simplify bindings in dependency order, starting with the non temp ones and then simplifying any temp ones that are needed after that. This change reduced the total number of expressions simplified (for the query in question) from 80 to 37.

3) In the simplifier main loop we clone the current value, simplify the clone, check if they are equal and if they are we stop simplifying. This clone is expensive so now we just hash the term, simplify it, and compare the new hash to the old one to see if we're done. Not cloning the whole expression every step speeds things up.

4) When derefing vars (a thing that happens a lot!) the vm has to ask the binding manager what state a variable is in. The binding manager returns an enum of the variables state. The creation of that enum, in the partial case, contains a clone of an expression which is a pretty slow operation. In deref however, the vm only cares if the variable is bound or not, so it just throws away that cloned value. In the query I was looking at there were 67,000 clones of expressions that were not used. I added a way for the vm to ask its actual question (is this variable bound to anything?) so we can avoid those clones in all those cases where we don't need them.

These four things together speed up the query in question (for me on my computer) from ~0.24 seconds to ~0.09 seconds.

### DIFF
--- a/polar-core/src/bindings.rs
+++ b/polar-core/src/bindings.rs
@@ -239,19 +239,18 @@ impl BindingManager {
                         }
                     }
                 }
-
                 term.clone_with_value(Value::List(derefed))
             }
-            Value::Variable(v) => match self._variable_state(v) {
-                BindingManagerVariableState::Bound(value) => value,
-                _ => term.clone(),
+            Value::Variable(v) => match self._variable_is_bound(v) {
+                Some(value) => value,
+                None => term.clone(),
             },
-            Value::RestVariable(v) => match self._variable_state(v) {
-                BindingManagerVariableState::Bound(value) => match value.value() {
+            Value::RestVariable(v) => match self._variable_is_bound(v) {
+                Some(value) => match value.value() {
                     Value::List(l) if has_rest_var(l) => self.deref(&value),
                     _ => value,
                 },
-                _ => term.clone(),
+                None => term.clone(),
             },
             _ => term.clone(),
         }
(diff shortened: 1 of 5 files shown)


dhatch:
Thanks for writing up all those details @saolsen. How much of the time the query takes now (0.09) is spent in the simplifier vs. before? Trying to get a sense of how much is left in the simplifier that we could optimize compared to everything else.

saolsen:
Ok, redid this a bit. first three commits do work (though the third is very repetitive and verbose for now)

4th and 5th commits are attempts at the cycle thing but both break the tests.

PR #751, opened 2021-02-23
New "Write Oso Policies" tutorial

plotnick:
PR checklist:
- [x] Introduction
- [x] Summary
- [ ] Interlanguagization
- [ ] More (complex) examples


new-docs/content/any/getting-started/policies/index.md
```diff
@@ -9,325 +9,401 @@ description: Learn about writing Oso policies - the source of truth for authoriz
...
-    user.role = "accountant";
+allow(actor, "GET", resource) if
+    resource.submitted_by = actor;
 ```
 
-This would be fine, but if, for example, we wanted to allow the CFO to do
-whatever an accountant can do, we would need to duplicate all the rules. Or if
-we want to change how an application determines roles we would need to change
-all locations using this.
+The Polar term `resource.submitted_by` refers to the `submitted_by`
```
leina05, line 295: Would be good to explicitly say that we can pass application objects into queries, so we can write rules over them. I know we say it before, but good to reinforce what's happening here.

new-docs/content/any/getting-started/policies/index.md
```diff
@@ -9,325 +9,401 @@ description: Learn about writing Oso policies - the source of truth for authoriz
...
+Here `Expense` is the application model class, and `matches` checks
+that its left-hand side is an instance of its right (or any subclass
+thereof). If that type check fails, the next term won't be evaluated.
 
-```python
-from oso import Oso
+In order for Polar to use a class in a type check, it must be
+_registered_. If you're using an ORM adapter, this happens
+automatically for all model classes; otherwise, you'll have
+to register classes manually using `Oso.register_class`. If
```
leina05, line 319: We can link to the API docs for the appropriate language here.

new-docs/content/any/getting-started/policies/index.md
```diff
@@ -9,325 +9,401 @@ description: Learn about writing Oso policies - the source of truth for authoriz
...
+you forget, Polar will warn you about an "unknown specializer"
+at load time.
 
-class User:
-    def __init__(self, name, role):
-        self.name = name
-        self.role = role
+### Specializers
```
leina05, line 323: Great transition here and perfect way to build up to specializers.

new-docs/content/any/getting-started/policies/index.md
```diff
@@ -9,325 +9,401 @@ description: Learn about writing Oso policies - the source of truth for authoriz
...
+* `Float`
 
-{{% callout "Note" "green" %}}
-Type specializers automatically respect the **inheritance** hierarchy of
-application classes. See the [Resources with
-Inheritance](guides/more/inheritance) guide for an in-depth
-example of how this works.
-{{% /callout %}}
+These classes correspond to ones in the application language,
+e.g., in Python, `String` is actually the built-in `str` class.
```
leina05, line 360: We can link to appropriate reference docs here.

new-docs/content/any/getting-started/policies/index.md
```diff
@@ -9,325 +9,401 @@ description: Learn about writing Oso policies - the source of truth for authoriz
...
 
-Once a class is registered, class or static methods can also be called from Oso
-policies:
+## Method Calls
 
-{{< code file="policy.polar" >}}
-allow(actor: User, action, resource) if actor.name in User.superusers();
-{{< /code >}}
+In addition to accessing fields in a rule, you can also call methods
+on application objects. We'll demonstrate this using the built-in
```
leina05, line 365:
```suggestion
In addition to accessing the fields of application objects in a rule, you can also call methods on application objects. We'll demonstrate this using the built-in
```

new-docs/content/any/getting-started/policies/index.md
```diff
@@ -9,325 +9,401 @@ description: Learn about writing Oso policies - the source of truth for authoriz
...
+`String` class, but it applies to instances any registered class.
 
-{{% exampleGet "classMethodExample" %}}
+Suppose we decided to change our encoding of actions so that
+every "read" action started with the letter "r". We could allow
+all such actions using a rule like this:
 
-### Built-in Types
+```polar
```
leina05, line 372: this will be specific to Python, so we will need to parametrize it.

new-docs/content/any/getting-started/policies/index.md
```diff
@@ -9,325 +9,401 @@ description: Learn about writing Oso policies - the source of truth for authoriz
...
-{{% callout "Warning" "orange" %}}
-Do not attempt to mutate a literal using a method on it. Literals in Polar
-are constant, and any changes made to such objects by calling a method will
-not be persisted.
-{{% /callout %}}
+So far all of our examples have involved only `allow` rules
+with three parameters. That's a natural place to start with Oso,
+because the `is_allowed` function generates queries of the form
+`allow(actor, action, resource)`. But you can write rules for
+whatever you like, and query them from within your `allow` rules.
```
leina05, line 383: This is a great explanation, and we should keep it here. But I think as mentioned in my earlier comment, having something similar to this when we first introduce `allow` rules, to foreshadow this section, would be great.

leina05 (commented):
I think this is a great start. In general, I think it would be great to break down the concepts as much as possible, so you can tell people what you are about to explain to them (it also makes the doc less intimidating if it's broken into conceptual sections). 

I know this is meant to be a tutorial, and I really like how it builds up now. I think the way you build up to specializers is a great example of the style the guide is going for. I'd say the first section needs the most work, because it's the hardest to get right. Happy to talk through ideas live. And I might think about the organization a bit more, but want to get this review to you first.

new-docs/content/any/getting-started/policies/index.md
```diff
@@ -9,325 +9,401 @@ description: Learn about writing Oso policies - the source of truth for authoriz
...
+* A user may approve an `expense` instance if they manage the
+  user that submitted it and the amount is less than some maximum.
+
+The rest of this tutorial will be concerned with how to turn
+informal English rules like the above into concrete Polar rules
+that the Oso library can enforce.
+
+## Queries
+
+_FIXME: This suction sucks, but something like it is needed._
```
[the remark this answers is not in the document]
plotnick: It took me a day to get this joke, since I (obviously) didn't see the typo in the first place.

new-docs/content/any/getting-started/policies/index.md
```diff
@@ -9,325 +9,438 @@ description: Learn about writing Oso policies - the source of truth for authoriz
...
+it should be straightforward to turn into a Polar rule.
+
+If you're unsure of what _kind_ of authorization policy to write,
+you may want to head over to the [Conceptual Guides](learn) or
+start with a familiar pattern such as [roles](/guides/roles).
+This tutorial will not be concerned with a specific kind of policy,
+but rather with the language in which many different kinds of
+policies may be expressed.
+
+## Application Setup
```
leina05, line 68: This section is great. Perfect amount of context in my opinion.

new-docs/content/any/getting-started/policies/index.md
```diff
@@ -9,325 +9,438 @@ description: Learn about writing Oso policies - the source of truth for authoriz
...
-<!-- TODO(gj): link `Oso.is_allowed()` once API docs are hooked up. -->
-When we use `Oso.is_allowed()` (or equivalent), we are making a query that asks
-Oso to evaluate all rules that match *(a)* on the rule name (`allow`), and
-*(b)* on all the inputs.
+## Rules
 
-In the rule above, `actor`, `action`, and `resource` are simply the parameter
-names, i.e., they are variables that will match *anything*.
+The way you control _which_ boolean value is returned by
+`oso.is_allowed` is to define and load Polar rules that
```
leina05, line 93:
```suggestion
The way you control whether
`oso.is_allowed` returns `true` or `false` is to define and load Polar rules that
```

new-docs/content/any/getting-started/policies/index.md
```diff
@@ -9,325 +9,438 @@ description: Learn about writing Oso policies - the source of truth for authoriz
...
+_match_ only the desired set of `actor`, `action`, and `resource`
```
leina05, line 94:
```suggestion
**match** only the desired set of `actor`, `action`, and `resource`
```

new-docs/content/any/getting-started/policies/index.md
```diff
@@ -9,325 +9,438 @@ description: Learn about writing Oso policies - the source of truth for authoriz
...
+arguments. This matching may involve logical connectives,
+type checks, equality checks, variable bindings, field lookups,
+method calls, arithmetic, comparisons, etc. See the
+[Polar Syntax guide](polar-syntax) for a complete list
+of available operators.
 
-But if we replace `action` with a concrete type…
+Let's start with the basic syntax. A _rule definition_ in Polar
+has a _name_, a _parameter list_, and an optional _body_. Here's
```
leina05, line 102: In general I think it's more clear to use **bold face** for keyword-type emphasis (as here), and _italics_ for emphasis. When I see lots of italics, I read it as if someone is speaking very deliberately and emphasizing each word.

new-docs/content/any/getting-started/policies/index.md
```diff
@@ -9,325 +9,438 @@ description: Learn about writing Oso policies - the source of truth for authoriz
...
-allow(actor, "read", resource);
+allow("Zora", "read", "resource");
 ```
 
-…the rule will now only be evaluated if the second input exactly matches the
-string `"read"`.
+The name of this rule is `allow`. It has three string-valued
+parameters: `"Zora"`, `"read"`, and `"resource"`. The terminating
+semicolon comes right after the parameter list, so it has no body;
+we say that it is _unconditional_, or a _fact_.
```
leina05, line 112: Not sure if we want to throw "fact" into the mix. 
```suggestion
we say that it is _unconditional_, meaning no additional conditions must be met in order for it to be true.
```

new-docs/content/any/getting-started/policies/index.md
```diff
@@ -9,325 +9,438 @@ description: Learn about writing Oso policies - the source of truth for authoriz
...
-#### In a Rule Body
-
-The most common way to write an `if` statement in Polar is to add a body to a
-rule. The following rule allows **any** actor to approve **any** expense
-report:
+Here's a _conditional_ rule definition:
 
 ```polar
-allow(_actor, "approve", _report);
+allow("Zora", "read", "resource") if 1 = 0;
```
leina05, line 117: This might be less confusing, since we are using numerics.
```suggestion
allow("Zora", "read", "resource") if 1 == 0;
```

new-docs/content/any/getting-started/policies/index.md
```diff
@@ -9,325 +9,438 @@ description: Learn about writing Oso policies - the source of truth for authoriz
...
+so policies containing tens or hundreds of thousands of rules like
+that can still be used efficiently should your application and
+policy require it, but it's not really recommended.
+
+What we usually want instead is to write rules that match more
+than one argument by exploiting regularities or abstractions in
+our application objects and policy. For instance, each of the three
+rules above has the same second and third parameters, so we can
+collapse all three rules into one if we _conditionally_ match the
+first argument. We can do that by _binding_ a _variable_ called
```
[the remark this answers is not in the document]
leina05: I say this because I know the concept of creating variables has been a source of confusion in the past.

new-docs/content/any/getting-started/policies/index.md
```diff
@@ -9,325 +9,438 @@ description: Learn about writing Oso policies - the source of truth for authoriz
...
-interact with this directly. For example:
+Let's also assume also a simple `Resource` class, with an `owner`
+field that references a user ID:
 
 ```python
-from oso import Oso
+@dataclass
+class Resource:
+    owner: int
+```
```
leina05, line 270: Consistency check: if we change "resource" above to be "document:ID", we can change this class to be a `Document` class as well.

new-docs/content/any/getting-started/policies/index.md
```diff
@@ -9,325 +9,438 @@ description: Learn about writing Oso policies - the source of truth for authoriz
...
 
-oso = Oso()
-oso.load_str("role(user, role_name) if user.role = role_name;")
+Our application will be making calls like this:
 
-alice = User("alice", "accountant")
-assert list(oso.query_rule("role", alice, "accountant"))
+```python
+oso.is_allowed(User(id=0, admin=True), "read", Resource(owner=0))
 ```
```
leina05, line 281: For readability, take or leave

```suggestion
```python
user = User(id=0, admin=True)
resource = Resource(owner=0)
oso.is_allowed(user, "read", resource)
```
```

File: new-docs/content/any/getting-started/policies/index.md [hunk 18]
```diff
@@ -9,325 +9,438 @@ description: Learn about writing Oso policies - the source of truth for authoriz
...
-  [debug]   BACKTRACK
-  [debug]           LOOKUP: <__main__.User object at 0x105da6190>.role()
-  [debug]           => No more results.
-  [debug]           BACKTRACK
-  [debug]           HALT
-  ```
-{{% /callout %}}
+```polar
+allow(user, "read", resource) if
+    user.admin = true;
```
leina05, line 289:
```suggestion
```polar
# A user that is an administrator may read any resource. 
allow(user, "read", resource) if
    user.admin = true;
```

new-docs/content/any/getting-started/policies/index.md
```diff
@@ -9,325 +9,438 @@ description: Learn about writing Oso policies - the source of truth for authoriz
...
 
-### Summary
+allow(user, "read", resource) if
+    user.id = resource.owner;
+```
```
leina05, line 293:
```suggestion
# A user may read any resource that they own.
allow(user, "read", resource) if
    user.id = resource.owner;
```
```

File: new-docs/content/any/getting-started/policies/index.md [hunk 20]
```diff
@@ -9,325 +9,438 @@ description: Learn about writing Oso policies - the source of truth for authoriz
...
 
-We covered some of the basics of policies, how to represent conditional logic,
-and briefly touched on the core mechanic of pattern matching.
+Lookups happen at query time, on the live application objects;
+Polar uses a foreign-function interface to perform the lookup
```
leina05, line 296: Perhaps a link to Wikipedia for FFI? Or a better reference if you know one.

new-docs/content/any/getting-started/policies/index.md
```diff
@@ -9,325 +9,438 @@ description: Learn about writing Oso policies - the source of truth for authoriz
...
-alice = User("alice", "accountant")
-assert list(oso.query_rule("role", alice, "accountant"))
+```python
+oso.is_allowed(User(id=0, admin=True), "read", Resource(owner=0))
 ```
 
-{{% callout "Tip" "green" %}}
-  Try setting the `POLAR_LOG` environment variable before executing a polar
-  query to see a [trace](tracing) of how the query is evaluated:
+We can implement our policy rules by binding `user` and `resource`
```
leina05, line 283: Could consider adding a step to explain what the query is in between the library call and the rules, like you did earlier. Though I'm not exactly sure how we would want to represent the `user` and `resource` arguments if we were showing a Polar query. E.g., `allow(PythonInstance{tag: User, id: 1}, "read", PythonInstance{tag: Resource, id: 2})` seems weird. But it's helpful to separate the rules from the query mentally.

new-docs/content/any/getting-started/policies/index.md
```diff
@@ -9,325 +9,438 @@ description: Learn about writing Oso policies - the source of truth for authoriz
...
+allow(user, "read", resource) if
+    user matches User and
+    resource matches Resource and
+    user.id = resource.owner;
+```
 
-In addition to accessing attributes, you can also call methods on application
-instances in a policy:
+The `matches` operator succeeds when its left-hand side is an
+instance of (a subclass of) its right-hand side. For our example
```
leina05, line 340: I think(?) a class is technically a subclass of itself, but saying that "instance of" and "subclass of" are the same might throw some people off. Perhaps "is an instance of the type specified on its right-hand side. This means that the type of the instance on the left must be the type specified on the right-hand side, or a subclass of that type."

new-docs/content/any/getting-started/policies/index.md
```diff
@@ -9,325 +9,438 @@ description: Learn about writing Oso policies - the source of truth for authoriz
...
-The benefits of that extraction are that it makes it easier to keep logic
-consistent throughout and often results in a much more readable policy.
+The knowledge base _searches_ for a definition that matches the query.
+In this case all three rules match the name `allow`, but the
+first two rules' first parameters fail to match the first supplied
+argument, since (in Polar notation), `"Abagail" != "Johann"` and
+`"Carol" != "Johann"`. The third rule, however, matches each of
+the arguments with its corresponding parameter: `"Johann" = "Johann"`,
+`"read" = "read"`, and `"resource" = "resource"`. So the query
+succeeds, and `oso.is_allowed` returns `True`.
```
leina05, line 204: I think we need one sentence here explaining that only one rule must succeed for a query to succeed, to make it clear that rules are only additive and not subtractive. We could also state it as each rule definition represents an implicit "or". Just something to hit that home at the end of this.

new-docs/content/any/getting-started/policies/index.md
```diff
@@ -9,325 +9,438 @@ description: Learn about writing Oso policies - the source of truth for authoriz
...
+    resource matches Resource and
+    user.id = resource.owner;
+```
 
-In addition to accessing attributes, you can also call methods on application
-instances in a policy:
+The `matches` operator succeeds when its left-hand side is an
+instance of (a subclass of) its right-hand side. For our example
+query above, since `str` is not a subclass of either `User`
+or `Resource`, the `matches` would fail, and not raise an error.
```
leina05, line 342: Phrasing seems funny here.
```suggestion
or `Resource`, the `matches` would fail, causing Polar to stop evaluating the rule before attempting to access any non-existent fields (and raising an error). 
```

new-docs/content/any/getting-started/policies/index.md
```diff
@@ -9,325 +9,438 @@ description: Learn about writing Oso policies - the source of truth for authoriz
...
 
-{{< code file="policy.polar" >}}
-allow(actor, action, resource) if actor.{{% exampleGet "isAdminOf" %}}(resource);
-{{< /code >}}
+Here `User` and `Resource` refer to the application classes defined
+above. But in order for Polar to use a class in a type check, it must
+first be _registered_. If you're using an ORM adapter, this happens
+automatically for all model classes; otherwise, you can register
+classes manually using `Oso.register_class`. If you forget, Polar
```
leina05, line 348: Link to API docs here (but this will be language-specific, so adds a lot of overhead). I would do this last if time.

new-docs/content/any/getting-started/policies/index.md
```diff
@@ -9,325 +9,438 @@ description: Learn about writing Oso policies - the source of truth for authoriz
...
+will warn you about an "unknown specializer" at load time.
```
leina05, line 349: Readout of this warning for the above rules would be nice, but low priority.

new-docs/content/any/getting-started/policies/index.md
```diff
@@ -9,325 +9,438 @@ description: Learn about writing Oso policies - the source of truth for authoriz
...
+first and third parameters are _specialized_ on the `User`
+and `Resource` classes, respectively: they will only match
+when the supplied arguments are instances of (subclasses of)
+those classes.
 
-In our previous example, the **allow** rule expected the actor to be a `User`,
-but we couldn’t actually check that type assumption in the policy. If we
-register the `User` class, we can write the following rule:
+It's considered good practice to use a specializer on any
+parameter whose fields you access.
```
leina05, line 371: In the future, it would be great to have a "next level" guide that goes into cool stuff like field-level specializers. We don't have any great docs on this specific thing now, but maybe we can add a TODO here to link to them once we have it. 

Or, we could add one sentence explaining it here if you're up for it. E.g., "pattern matching can go beyond simple type checks as well. Field-level specializers can be especially useful for ensuring certain fields exist before entering a rule body:
```
allow(user: User{id: user_id}, "read", resource: Resource{owner: resource_owner}) if
    user_id = resource_owner;
```
For more information on pattern matching in Polar, see the [syntax guide](polar-syntax#rules)."

new-docs/content/any/getting-started/policies/index.md
leina05: Still think this would be good, but again it requires language-specific docs for each, so would do this last if time.

new-docs/content/any/getting-started/policies/index.md
```diff
@@ -9,325 +9,438 @@ description: Learn about writing Oso policies - the source of truth for authoriz
...
-{{% /callout %}}
+```polar
+allow(user, action, resource) if
+    resource_role_applies_to(resource, role_resource) and
+    user_in_role(user, role, role_resource) and
+    role_allow(role, action, resource);
+```
 
+Because all three parameters are variables, any three arguments
+will match the head, by binding. But queries for each of the body
```
leina05, line 426: This almost makes me want to have a subsection under the previous "Variables" section dedicated to binding, so that if the reader already forgot what "binding" means we could send them back there. Interested if you think this would be helpful.

new-docs/content/any/getting-started/policies/index.md
```diff
@@ -9,325 +9,438 @@ description: Learn about writing Oso policies - the source of truth for authoriz
...
-on strings regardless of whether they originated in your application or as a
-literal in your policy. This applies to all of Polar's [supported
-types](polar-syntax#primitive-types) in any supported application language. For
-examples using built-in types, see [the {{% exampleGet "langName" %}}
-library](reference/polar/classes) guide.
+So far all of our examples have involved only `allow` rules
+with three parameters. That's a natural place to start with Oso,
+because the `is_allowed` function generates queries of the form
+`allow(actor, action, resource)`. But you can write rules for
+whatever you like, and use them from within your `allow` rules.
```
leina05, line 414: I think it's worth adding one sentence somewhere in this section that lets the reader know they can also use the `query_rule()` method to query arbitrary rules directly from the application. Perhaps at the very end.

new-docs/content/any/getting-started/policies/index.md
plotnick: Ok. I have a very academic habit of emphasizing the first occurrence of every new keyword or technical term, but don't have strong feelings on what kind of emphasis to use. Switched to bold so we can use italics for ordinary emphasis.

new-docs/content/any/getting-started/policies/index.md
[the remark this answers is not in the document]
plotnick: Agreed. Added a `Binding` section as suggested later, but not entirely happy with it. LMK if it's an improvement, or how we can make it clear.er

new-docs/content/any/getting-started/policies/index.md
plotnick: Leave, but with reason: we have the same sort of one-liner later, and I think leaving them as one-liners makes it easier to visually spot the difference between them.

new-docs/content/any/getting-started/policies/index.md
```diff
@@ -427,6 +431,46 @@ those classes.
...
+```polar
+# A user may read any document that they own.
+allow(user: User{id: user_id}, "read", document: Document{owner: user_id});
+```
+
+The first specializer binds `user_id` to `user.id`. Then, since
+`user_id` is already bound, the second specializer compares its
+value against that of `document.owner`, making the explicit
+unification unnecessary.
+
```
leina05, line 473: Excellent explanation!

new-docs/content/any/getting-started/policies/index.md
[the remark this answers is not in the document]
leina05: I think it's great!

leina05 (approved):
All the changes look great! I know interlanguagization is next on your list, but I would publish this as is and take that as a follow-up. I think even with python-only examples this will add A LOT of helpful information to the docs so I'd rather not wait.

gj (commented):
Overall feedback (on the section that I reviewed): the writing is excellent, and the explanations are much clearer than what we had previously.

gj:
> All the changes look great! I know interlanguagization is next on your list, but I would publish this as is and take that as a follow-up. I think even with python-only examples this will add A LOT of helpful information to the docs so I'd rather not wait.

If we're going to publish this without fully inter-languagizerationing it, I think we should back out the `{{< lang >}}` shortcodes since there's currently some dissonance between the Oso API methods all being the Python versions and the text saying, e.g., "Java" if you have Java as your selected language.

PR #753, opened 2021-02-23
avoid cloning value being simplified

saolsen:
In the simplifier main loop we clone the current value, simplify the clone, check if they are equal and if they are we stop simplifying. This clone is expensive so now we just hash the term, simplify it, and compare the new hash to the old one to see if we're done. Not cloning the whole expression every step speeds things up.


dhatch:
If this actually helps perf seems good, but it seems surprising to me that the hash is faster than an Rc bump.

samscott89:
Already in from a prev PR

PR #754, opened 2021-02-23
only check if bound in deref

saolsen:
When derefing vars (a thing that happens a lot!) the vm has to ask the binding manager what state a variable is in. The binding manager returns an enum of the variables state. The creation of that enum, in the partial case, contains a clone of an expression which is a pretty slow operation. In deref however, the vm only cares if the variable is bound or not, so it just throws away that cloned value. In the query I was looking at there were 67,000 clones of expressions that were not used. I added a way for the vm to ask its actual question (is this variable bound to anything?) so we can avoid those clones in all those cases where we don't need them.

### DIFF
--- a/polar-core/src/bindings.rs
+++ b/polar-core/src/bindings.rs
@@ -239,19 +239,18 @@ impl BindingManager {
                         }
                     }
                 }
-
                 term.clone_with_value(Value::List(derefed))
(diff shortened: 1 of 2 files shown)


dhatch (approved):
lgtm

PR #755, opened 2021-02-23
simplify in dependency order

saolsen:
When we simplify bindings we simplify ALL the bindings. There will be a binding for every temporary variable in the expressions. We only care about variables that are not temp, and temp variables that are used in expressions bound to non temp variables so instead of simplifying all the bindings I made it simplify bindings in dependency order, starting with the non temp ones and then simplifying any temp ones that are needed after that. This change reduced the total number of expressions simplified (for the query I was looking at) from 80 to 37.

### DIFF
--- a/polar-core/src/partial/simplify.rs
+++ b/polar-core/src/partial/simplify.rs
@@ -1,4 +1,4 @@
-use std::collections::HashSet;
+use std::collections::{HashMap, HashSet, VecDeque};
 
 use crate::bindings::Bindings;
 use crate::folder::{fold_operation, fold_term, Folder};
@@ -94,7 +94,7 @@ pub fn simplify_partial(var: &Symbol, term: Term) -> Term {
 ///
 /// - For partials, simplify the constraint expressions.
 /// - For non-partials, deep deref. TODO(ap/gj): deep deref.
-pub fn simplify_bindings(bindings: Bindings) -> Option<Bindings> {
+pub fn simplify_bindings(bindings: Bindings, all: bool) -> Option<Bindings> {
     let mut unsatisfiable = false;
     let mut simplify = |var: Symbol, term: Term| {
         let simplified = simplify_partial(&var, term);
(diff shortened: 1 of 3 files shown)


polar-core/src/partial/simplify.rs
```diff
@@ -94,42 +94,117 @@ pub fn simplify_partial(var: &Symbol, term: Term) -> Term {
...
-    let bindings: Bindings = bindings
-        .iter()
-        .map(|(var, value)| match value.value() {
-            Value::Expression(o) => {
-                assert_eq!(o.operator, Operator::And);
-                (var.clone(), simplify(var.clone(), value.clone()))
+    let mut simplified_bindings = HashMap::new();
+    if all {
+        for (var, value) in &bindings {
+            match value.value() {
```
dhatch, line 113: This is just below but we throw away referenced_vars basically right?

polar-core/src/partial/simplify.rs
```diff
@@ -94,42 +94,117 @@ pub fn simplify_partial(var: &Symbol, term: Term) -> Term {
...
+                    simplified_bindings.insert(var.clone(), bindings[v].clone())
+                }
+                _ => simplified_bindings.insert(var.clone(), value.clone()),
+            };
+        }
+    } else {
+        let mut referenced_vars: VecDeque<Symbol> = VecDeque::new();
+        for (var, value) in &bindings {
+            if !var.is_temporary_var() {
+                match value.value() {
```
dhatch, line 136: This & the below could probably be a `simplify_var` function that just returns the simplified var & the referenced variables, right? Or returns the simplified var and getting the referenced vars is a separate step from it that only happens in the `!all` case.

dhatch (commented):
Looks good but I think it is worth doing the little cleanup to collapse these cases to make it easier to understand & change going forward.

dhatch (approved):
dope

PR #756, opened 2021-02-23
expressions dont need to be derefed

saolsen:
this cuts out 1/3 of the time in the query total


plotnick (approved):
Single-line optimizations are my very favorite kind. Nicely done.

PR #758, opened 2021-02-24
Ensure docs are built on every PR

gj:


.github/workflows/docs-preview.yml
```diff
@@ -61,9 +61,7 @@ jobs:
         with:
           hugo-version: '0.79.1'
       - name: Build Hugo docs
-        run: |
-          make build
-          HUGO_ENV=production hugo -b "https://docs-preview.oso.dev/"
+        run: make build
```
samscott89, line 64: Doesn't this already run on every PR?

gj: The YAML was off. Thanks for making me look twice at it.

samscott89: Ohhhh

.github/workflows/docs-preview.yml
```diff
@@ -3,8 +3,8 @@ on:
   push:
     branches:
       - main
-    pull_request:
-      types: [opened, synchronize, reopened]
+  pull_request:
```
samscott89, line 6: oof. my bad

.github/workflows/docs-preview.yml
```diff
@@ -16,6 +16,14 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
+      - uses: actions/cache@v2
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            target
+          key: ${{ runner.os }}-cargo-check-${{ hashFiles('**/Cargo.lock') }}
```
samscott89, line 25: This should have a different cache key. Otherwise, they race and if one does less work that gets cached.

samscott89: Although we do make python-build and make wasm-build in both so maybe its fine?

gj: Why would the one that does less work get cached? Wouldn't the tortoise win the race?

samscott89 (approved):

PR #760, opened 2021-02-24
External Comparison Operations for JS

Anniepoo:
This PR adds external comparison operator support to Polar for Javascript.  

Polar like   'new Foo() == new Foo()'  needs to be handled by the language adapter, not Polar core.
This is already done for Python. This PR adds support for Javascript.

PR checklist:

- [X] Added changelog entry.


languages/js/src/helpers.ts
```diff
@@ -217,6 +225,36 @@ function parseExternalIsa({
...
+    if (!isPolarTerm(arg)) {
+      throw new Error();
+    }
+  });
+  return {
+    kind: QueryEventKind.ExternalOp,
+    data: {
+      args,
+      callId: acallid,
+      operator,
```
gj, line 253: I think this function would be a good place to validate that `operator` is a variant of the `PolarOperator` enum since we're doing all of our other "did we receive bad data" checks here.

If we make that change, then we can remove the `default:` fallback in the switch statement in `src/Host.ts` and throw that `PolarError` here instead.

gj (changes requested):

languages/js/Makefile
```diff
@@ -1,46 +1,61 @@
-.PHONY: wasm install clean dev build fmtcheck typecheck lint parity test
+.PHONY: build test parity lint fmt fmtcheck typecheck docs repl clean install wasm
 
+# Build the WASM lib if it hasn't already been built; then build the TS lib.
 build: wasm
 	yarn build
 
+# Run the TS lib's test suite.
+test: install
```
gj, line 8: @Anniepoo I removed the new `jsbuildtest` Make target that we added, but `make test` should behave much better now. We don't actually need to build prior to running the tests (the `ts-jest` library handles that for us), and I implemented some smarter caching in the Makefile so that it isn't always building the WASM lib willy-nilly.

gj (commented):
TS changes look 👍 

There are probably some places to update the docs with this new functionality.

Let's wait for @dhatch to take a look at the various Makefile changes before we ship it

samscott89 (commented):
Also needs documentation for new functionality.

Redundant comment is redundant.

dhatch (approved):
Make changes look fine

gj (approved):

PR #761, opened 2021-02-25
Avoid double-escaping Polar code blocks in the docs

arusahni:
Use `innerText` when highlighting Polar code blocks to avoid double-escaping special characters.

This fixes an issue where the Polar code blocks in the docs were being
escaped twice - once by Hugo, and another time by Monaco. By using
`innerText` instead of `innerHTML`, the code blocks are unescaped, so Monaco
can format and re-escape it.

**Before**

Screenshot_2021-02-24 Polar Syntax

**After**

Screenshot_2021-02-24 Polar Syntax(1)

I think this is the right place to apply this, but I definitely was tracing my way through the build process.


arusahni:

gj (approved):
Nice catch, and well done tracking it down! Thanks @arusahni

PR #767, opened 2021-03-01
Additional tests for list filtering parity.

dhatch:
This includes tests linked to
https://www.notion.so/osohq/Supported-Query-Types-and-Features-435d7a998dc14db3a125c6e5ba5fe6ba.

I've implemented tests up to the "Partial collection intersects with other
ground list" in the Relationship table. Some tests fail, and are marked
as expected failing. These can serve as test cases when implementing new
operations.

### DIFF
--- a/languages/python/django-oso/tests/test_partial_parity.py
+++ b/languages/python/django-oso/tests/test_partial_parity.py
@@ -1,0 +1,34 @@
+"""Partial parity tests based on
+
+https://www.notion.so/osohq/Supported-Query-Types-and-Features-435d7a998dc14db3a125c6e5ba5fe6ba.
+"""
+import pytest
+
+from django_oso.models import authorize_model
+from django_oso.oso import Oso, reset_oso
+from test_app2.models import Post, Tag, User
+
+
+@pytest.fixture(autouse=True)
+def reset():
+    reset_oso()
+
+@pytest.mark.django_db
+def test_field_comparison():
+    post0 = Post(id=0, contents="private post", title="not private post")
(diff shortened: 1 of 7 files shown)


dhatch:
Opening this so we can get it reviewed, but this doesn't necessarily have to be in the release.

languages/python/django-oso/tests/test_post_relationship.py
```diff
@@ -549,7 +559,7 @@ def test_empty_constraints_in(tag_nested_many_many_fixtures):
                           WHERE U0."id" = {parenthesize('V1."tag_id"')}){is_true()})
     """
     assert str(posts.query) == " ".join(expected.split())
-    assert len(posts) == 4
+    assert len(posts) == 5
```
samscott89, line 562: Why did this change?

languages/python/django-oso/tests/test_partial_parity.py
```diff
@@ -0,0 +1,165 @@
+"""Partial parity tests based on
+
+https://www.notion.so/osohq/Supported-Query-Types-and-Features-435d7a998dc14db3a125c6e5ba5fe6ba.
```
gj, line 3: Don't love linking to private docs in public code

gj (approved):
Great work

languages/python/django-oso/tests/test_partial_parity.py
dhatch: Yeah, I hear that... Not sure of an easy workaround though unless I move this to a wiki page or something on GitHub.  I think it's valuable to have that reference when going through tests.

Issue #773, opened 2021-03-04
Rule querying requires `&mut self`, hampering multi-threaded querying


fisherdarling:
`query_rule` and `is_allowed` require a mutable borrow on an `Oso` instance. This requires me to create a new instance per route / thread, rather than reusing a global instance.

Question / Issue from the presentation on 3/3

samscott89:
Yep, you're right. Good catch! We've been unnecessarily wrapping things in mutexes.

gj:
This should be fixed as of 0.11.2. Thanks @fisherdarling!

PR #774, opened 2021-03-04
Querying `Oso` doesn't need a mutable reference.

samscott89:
Addresses #773 

PR checklist:

- [x] Added changelog entry.


gj (commented):
What about `clear_rules()`, `check_inline_queries()`, `load_file()`, and `load_str()`?

samscott89:
> 
> 
> What about `clear_rules()`, `check_inline_queries()`, `load_file()`, and `load_str()`?

My initial reaction was "no of course not, those mutate inner". But I think you're right. The locking happens in the core. So yes, these should be fine to be immutable refs too.

samscott89:
Also tested against the quickstart: https://github.com/osohq/oso-rust-quickstart/pull/1

gj (approved):
LGTM, but is this now a major version bump? Can't tell how "breaking" of a change it is. `rustc` will be able to tell end users that they can remove the `mut` designator, but will that be a warning or something that actually causes builds to fail?

samscott89:
> 
> 
> LGTM, but is this now a major version bump? Can't tell how "breaking" of a change it is. `rustc` will be able to tell end users that they can remove the `mut` designator, but will that be a warning or something that actually causes builds to fail?

It will be a warning, yes. Like in the quickstart we get:

```
warning: variable does not need to be mutable
  --> src/server.rs:59:13
   |
59 |         let mut guard = self.oso.lock().unwrap();
   |             ----^^^^^
   |             |
   |             help: remove this `mut`
   |
   = note: `#[warn(unused_mut)]` on by default
```

This would cause build to fail if they check for linter errors/deny warning (like we do), but I feel like this is reasonable for a minor bump

PR #775, opened 2021-03-04
Basic docs site search

gj:


samscott89 (changes requested):
All the mechanical changes look good - the HTML, the approach of spitting out JSON, and using lunr.

I think we need to spend a little more time on what documents get indexed. And if possible make it so that we have a separate index depending on what language is selected. It looks like we already spit out a separate index for each language anyway.

PR #776, opened 2021-03-05
Use language-specific isAllowed function name in quickstart guide.

phiggins:
While getting familiar with the company and product I ran through the Ruby tutorial and was momentarily confused when the quick start section of the docs referred to a call to `is_allowed()` which didn't appear in the sample code. Since the policy page appeared to handle this correctly for the different languages, I copied what was there and added it the to the quick start page. I manually checked the pages in my local env and they all appeared to be working as intended.

Signed-off-by: Pete Higgins <pete@peterhiggins.org>

### DIFF
--- a/new-docs/content/any/getting-started/quickstart/index.md
+++ b/new-docs/content/any/getting-started/quickstart/index.md
@@ -92,7 +92,7 @@ resource is an instance of the `Expense` class.
 
 Okay, so what just happened?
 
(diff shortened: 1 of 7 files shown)


phiggins:

phiggins:
recheck

samscott89 (approved):
Great catch! And thanks for doing the fix too.

gj:
🤦 thanks for catching this @phiggins !

Issue #746, opened 2021-02-18
Add `get_allowed_actions()` to non-Python libraries


leina05:
[`get_allowed_actions()`](https://github.com/osohq/oso/blob/b7b588cb32a405a1dfdf4c509e05b3e65d173a30/languages/python/oso/oso/oso.py#L43-L79) returns a list of all actions an actor is allowed to take on a resource based on the policy. Currently, this is only implemented in the Python library.

We'd like to add this to the other language libraries:

- [x] Node.js
- [x] Go
- [x] Ruby
- [x] Java
- [x] Rust

joshrotenberg:
- [X] Rust

samscott89:
- [x] Java

Thanks @joshrotenberg !

samscott89:
> 
> 
> -  [x]  Java
> 
> 
> Thanks @joshrotenberg !

 - [x] Go

Thanks @joshrotenberg again!

gj:
As of the 0.20 release that stabilizes tomorrow, `authorized_actions` (the new name for `get_allowed_actions`) has been added to [Ruby](https://github.com/osohq/oso/blob/e7160c3ed4f28e84a5ea6f8da2d761a889f5fe43/languages/ruby/lib/oso/oso.rb#L110-L140) and [Node.js](https://github.com/osohq/oso/blob/e7160c3ed4f28e84a5ea6f8da2d761a889f5fe43/languages/js/src/Oso.ts#L97-L137). 🎉 

Thanks again for all of your contributions @joshrotenberg !

PR #789, opened 2021-03-20
[Rust] implement get_allowed_actions

joshrotenberg:
Partial implementation for #746.

First crack at implementing `get_allowed_actions` as requested in #746 using the current Python implementation as a guide.

Items that still need attention:

* Is `Vec<String>` the correct return type here, or should it return the `PolarValue` itself, or ...?
* None of the current unit tests seem to have an action nor does the doc test for the module, so where should the tests go?
* The Python version supports an `allow_wildcard` parameter that is not yet supported here.


samscott89:
Thanks for making a start on this!

> Is `Vec<String>` the correct return type here, or should it return the `PolarValue` itself, or ...?

We could instead make it `Vec<T>` with an additional generic `T: FromPolar`. This would allow people to get whatever type they want, including `String`. Also, I noticed that you're using `BTreeSet` to ensure uniqueness, which is a good idea, so maybe we could just return that? Make it really obvious to the caller that we're doing that?

> None of the current unit tests seem to have an action nor does the doc test for the module, so where should the tests go?

We should have a `test_oso.rs` file in the `tests/` directory for this kind of thing. There are more unit tests missing that we should add (separately to this PR).

> The Python version supports an `allow_wildcard` parameter that is not yet supported here.

Maybe we could provide something like:

```
enum Action<T=String> {
    Any,
    Typed(T)
}
```

with an implementation:

```
impl<T: FromPolar> FromPolar for Action<T> {
    fn from_polar(val: PolarValue) -> crate::Result<Self> {
        if matches!(val, PolarValue::Variable(_)) {
            Ok(Action::Any)
        } else {
            T::from_polar(val).map(Action::Typed)
        }
    }
}
```

We could make this the suggested return type for above. E.g. `let actions: Vec<Action> = oso.get_allowed_actions(actor, resource)?;`.

joshrotenberg:
> > None of the current unit tests seem to have an action nor does the doc test for the module, so where should the tests go?
> 
> We should have a `test_oso.rs` file in the `tests/` directory for this kind of thing. There are more unit tests missing that we should add (separately to this PR).

I'll create a new PR for this first and then get back to this one once more tests are in place.

joshrotenberg:
@samscott89 here is a quick implementation. A few things:

* is `oso.rs` the right place for the Action struct definition? and should the `FromPolar` live with his buddies `from_polar.rs`?
* I see some docs in `lib.rs`, so I can add an example there. Where else does this need to be documented in the repo?
* I've tested `String` and `Action` return types. What else should be tested?

languages/rust/oso/tests/test_oso.rs
```diff
@@ -157,3 +158,29 @@ fn test_allow_model() -> oso::Result<()> {
...
+    assert!(actions.contains(&Action::Typed("get".to_string())));
+
+    let actor = Actor::new(String::from("sally"));
+    let resource = Widget::new(1);
+    let actions: HashSet<String> = oso.get_allowed_actions(actor, resource)?;
+
+    assert!(actions.len() == 3);
+    assert!(actions.contains(&"CREATE".to_string()));
+    assert!(actions.contains(&"READ".to_string()));
+    assert!(actions.contains(&"get".to_string()));
```
samscott89, line 183: is the `to_string` necessary? That would be way nicer if not

joshrotenberg: Fixed.

languages/rust/oso/src/oso.rs
```diff
@@ -62,6 +80,41 @@ impl Oso {
         }
     }
 
+    // Determine the actions actor is allowed to take on resource.
```
samscott89, line 83: This should be a doc comment (three `/`), and ideally would include an example.

samscott89:
> 
> 
> @samscott89 here is a quick implementation. A few things:
> 
>     * is `oso.rs` the right place for the Action struct definition? and should the `FromPolar` live with his buddies `from_polar.rs`?
> 

I would keep both in oso.rs for now.

>     * I see some docs in `lib.rs`, so I can add an example there. Where else does this need to be documented in the repo?
> 

The doc string for the function itself would be the best place for it.

>     * I've tested `String` and `Action` return types. What else should be tested?

If you feel up for it, testing with an enum, e.g.

```
enum ApiAction {
    Create,
    Read,
    Update,
    Delete, 
}
```
might be good. You would probably need to register a variant as a constant to make it work.

languages/rust/oso/src/oso.rs
```diff
@@ -62,6 +80,53 @@ impl Oso {
...
+                    if let Some(action) = result.get("action") {
+                        set.insert(T::from_polar(action).unwrap());
+                    }
+                }
+                Some(Err(_)) => continue,
+                None => break,
+            };
+        }
+
+        // Ok(set.into_iter().collect::<Vec<T>>())
```
samscott89, line 126:
```suggestion
```

languages/rust/oso/src/oso.rs
```diff
@@ -62,6 +80,53 @@ impl Oso {
...
+                (actor, PolarValue::Variable("action".to_owned()), resource),
+            )
+            .unwrap();
+
+        let mut set = HashSet::new();
+        loop {
+            match query.next() {
+                Some(Ok(result)) => {
+                    if let Some(action) = result.get("action") {
+                        set.insert(T::from_polar(action).unwrap());
```
samscott89, line 118:
```suggestion
                        set.insert(T::from_polar(action)?);
```

I think this should work

languages/rust/oso/src/oso.rs
```diff
@@ -62,6 +80,53 @@ impl Oso {
...
+                    }
+                }
+                Some(Err(_)) => continue,
```
samscott89, line 121:
```suggestion
                Some(Err(e)) => return Err(e),
```

Errors returned from Polar are usually pretty bad. It's better for us to bubble it up.

languages/rust/oso/src/oso.rs
```diff
@@ -62,6 +80,53 @@ impl Oso {
...
+        let mut query = self
+            .query_rule(
+                "allow",
+                (actor, PolarValue::Variable("action".to_owned()), resource),
+            )
+            .unwrap();
+
+        let mut set = HashSet::new();
+        loop {
+            match query.next() {
```
samscott89, line 115: Could write this as a combinator instead.

Maybe `query.try_fold(HashSet::new(), ...` ?

Would save you needing to deal with the Some/None case.

languages/rust/oso/tests/test_oso.polar
```diff
@@ -20,3 +20,9 @@ allow(actor: Actor, "frob", resource: Company) if
 
 allow(actor: Actor, "list", Company) if
    actor.name = "auditor";
+
```
samscott89, line 23: Can delete these if we have them inline in the tests.

languages/rust/oso/src/oso.rs
joshrotenberg: I'll give it a whirl.

samscott89: Hey @joshrotenberg! Are you going to make this change, or do you want me to do a final review as is? I don't think this is a must do. More of a personal preference

joshrotenberg: Sorry, I haven't had a chance to look at it, busy few days. If you're ok with the current implementation lets just go with it (and I can fix any minor issues if you have more) and then possibly revisit.

samscott89 (approved):
LGTM! Thanks for doing this.

PR #790, opened 2021-03-22
[Rust] Start work on test_oso.rs

joshrotenberg:
Start and stub out `test_oso.rs`. Following along with `test_oso.py` in terms of structure and tests though I imagine there will be some divergence as I dig further. Also, seems like there is an opportunity to incorporate a lot of what is currently in `lib_guide.rs` since that isn't run with tests but tests a lot of functionality. Merge the two?

### DIFF
--- a/languages/rust/oso/tests/test_oso.polar
+++ b/languages/rust/oso/tests/test_oso.polar
@@ -18,5 +18,10 @@ allow(actor: Actor, "create", resource: Company) if
 allow(actor: Actor, "frob", resource: Company) if
     resource in actor.companies();
 
-allow(actor: Actor, "list", Company) if
-    actor.name = "auditor";
+# XXX: test_oso::test_allow_models fails with this, but ...
+#allow(actor: Actor, "list", Company) if
+#   actor.name = "auditor";
+
+# XXX: ... works with this. why?
+allow(actor: Actor, "list", resource: Company) if
+   actor.name = "auditor";
--- a/languages/rust/oso/tests/test_oso.polar
+++ b/languages/rust/oso/tests/test_oso.polar
@@ -18,10 +18,5 @@ allow(actor: Actor, "create", resource: Company) if
 allow(actor: Actor, "frob", resource: Company) if
     resource in actor.companies();
 
-# XXX: test_oso::test_allow_models fails with this, but ...
-#allow(actor: Actor, "list", Company) if
-#   actor.name = "auditor";
-
-# XXX: ... works with this. why?
 allow(actor: Actor, "list", resource: Company) if
    actor.name = "auditor";
--- a/languages/rust/oso/tests/test_oso.polar
+++ b/languages/rust/oso/tests/test_oso.polar
@@ -1,0 +1,22 @@
+allow(actor, action, resource) if
+    allowRole(role, action, resource) and
+    actorInRole(actor, role, resource);
+
+allow(_: {sub: sub}, action, resource) if
+    allow(new Actor(name: sub), action, resource);
+
+allow("guest", action, resource) if
+    allow(new Actor(name: "guest"), action, resource);
+
+allow(_: {username: name}, action, resource) if
+    allow(new Actor(name: name), action, resource);
+
+allow(_actor: Actor, "get", _resource: Widget);
+allow(actor: Actor, "create", resource: Company) if
+    resource.role(actor) = "admin";
+
+allow(actor: Actor, "frob", resource: Company) if
+    resource in actor.companies();
+
+allow(actor: Actor, "list", Company) if
+    actor.name = "auditor";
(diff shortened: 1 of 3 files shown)


languages/rust/oso/tests/test_oso.polar
```diff
@@ -0,0 +1,27 @@
...
+    allow(new Actor(name: name), action, resource);
+
+allow(_actor: Actor, "get", _resource: Widget);
+allow(actor: Actor, "create", resource: Company) if
+    resource.role(actor) = "admin";
+
+allow(actor: Actor, "frob", resource: Company) if
+    resource in actor.companies();
+
+# XXX: test_oso::test_allow_models fails with this, but ...
```
joshrotenberg, line 21: The top rule, copied over from `test_oso.polar` in the Python tests, fails `test_allow_model` with:

```
Error: TypeError(TypeError { got: Some("oso::host::Class"), expected: "Company" })
thread 'test_allow_model' panicked at 'assertion failed: `(left == right)`
```

But the bottom succeeds. Not sure what I'm doing wrong.

samscott89: This is checking that the input is the _type_ `Company`, as opposed to an instance of a `Company`.

I'm not sure we have a suitable equivalent for this in Rust. It would either have to be by implementing `ToPolar` for `TypeId` so that you could do `TypeId::of::<Company>()` as the input value. Or using the built-in `Company::get_polar_class()`.

samscott89: On reflection, if the latter works then that is probably the best option for now.

samscott89:
This is looking great, thanks!

On:

> Also, seems like there is an opportunity to incorporate a lot of what is currently in lib_guide.rs since that isn't run with tests but tests a lot of functionality. Merge the two?

The lib guide is specifically (supposed to be) testing what's in the docs. We'll need to solve this separately though. The lib guide was broken because it fell out of sync with the docs.

languages/rust/oso/tests/test_oso.polar
joshrotenberg: I removed the comments and commented rule.

languages/rust/oso/tests/test_oso.rs
```diff
@@ -0,0 +1,161 @@
...
+        if actor.name == "president" {
+            "admin".to_string()
+        } else {
+            "guest".to_string()
+        }
+    }
+}
+
+fn test_oso() -> OsoTest {
+    let mut test = OsoTest::new();
```
samscott89, line 62: I would probably skip the use of `OsoTest` for this file, since this file is more about testing the publicly-facing part of Oso. The main purpose of `OsoTest` is to expose a bunch of useful helpers for unit testing functionality.

languages/rust/oso/tests/test_oso.polar
```diff
@@ -0,0 +1,22 @@
...
+    allow(new Actor(name: name), action, resource);
+
+allow(_actor: Actor, "get", _resource: Widget);
+allow(actor: Actor, "create", resource: Company) if
+    resource.role(actor) = "admin";
+
+allow(actor: Actor, "frob", resource: Company) if
+    resource in actor.companies();
+
+allow(actor: Actor, "list", resource: Company) if
```
samscott89, line 21: Let's keep this in line with the other tests, and mark the test `#[ignore]`
```suggestion
allow(actor: Actor, "list", Company) if
```

languages/rust/oso/tests/test_oso.rs
```diff
@@ -0,0 +1,161 @@
...
+    let actor = Actor::new(String::from("guest"));
+    let resource = Company::new(1);
+
+    assert!(oso.oso.is_allowed(actor, "frob", resource)?);
+
+    Ok(())
+}
+
+#[test]
+fn test_allow_model() -> oso::Result<()> {
```
samscott89, line 147: We can mark this `#[ignore]` and have the body as `todo!()` for now if we don't want to do it now (and update the policy as the other comment suggests).

samscott89 (changes requested):
This is great, thank you for doing this!

We should change the `allow_model` test, so the next person isn't too confused. And I think removing the use of `TestOso` for this would be important.

With those two changes done, I think this will be good to go!

languages/rust/oso/tests/test_oso.rs
samscott89: did this

samscott89: Currently fails because `PartialEq` is not configured for `Class` and neither is equality configured

samscott89 (approved):
I made the changes for you @joshrotenberg. Take a look and let me know if there are any questions! Otherwise we can merge this.

joshrotenberg (commented):
Looks good to me. Thanks for taking the time to clean it up.

PR #792, opened 2021-03-22
Fix for  695 better error handling and clearer message for inline queries that fail

### LINKED ISSUE #695: Error "inline query result was false", but which one?
When using `?= ...` inline queries to verify that a Polar policy behaves as expected, in the event that a query fails the error message output by Oso doesn't indicate the query that failed but only says _"inline query result was false"_. I encountered this with Oso 0.8.1 for Rust but looking at the [`code in the main`](https://github.com/osohq/oso/blob/main/languages/rust/oso/src/oso.rs#L75) branch it seems to be the same:

```rust
    fn check_inline_queries(&mut self) -> crate::Result<()> {
        while let Some(q) = self.inner.next_inline_query(false) {
            let query = Query::new(q, self.host.clone());
            match query.collect::<crate::Result<Vec<_>>>() {
                Ok(v) if !v.is_empty() => continue,
                Ok(_) => return lazy_error!("inline query result was false"),
                Err(e) => return lazy_error!("error in inline query: {}", e),
            }
        }
        check_messages!(self.inner);
        Ok(())
    }
```

Setting environment variable `POLAR_LOG=1` helps a bit as you can work out the last query that was being executed, but it's not that easy as you have to look at the log statement indentation to try and work out where the start of the query execution was and thus what the root query being executed was.

Anniepoo:
This PR closes #695 by creating a proper OsoError and including the source location in the location field.


Anniepoo:
obviously something bad in the last rebase - going to try again fresh

PR #794, opened 2021-03-22
Add context indicating which inline query failed

### LINKED ISSUE #695: Error "inline query result was false", but which one?

Anniepoo:
This PR fixes #695  by returning a new OsoError variant . The user now sees a line number and copy of the source line rather than the unhelpful 'inline query is false'.

PR checklist:

- [ X] Added changelog entry.

### DIFF
--- a/docs/changelogs/vNEXT.rst
+++ b/docs/changelogs/vNEXT.rst
@@ -41,6 +41,7 @@ Link to relevant documentation section
 Other bugs & improvements
 =========================
 
+- improved error messages for failing inline queries
 - bulleted list
 - improvements
 - of smaller
--- a/docs/changelogs/vNEXT.rst
+++ b/docs/changelogs/vNEXT.rst
@@ -41,7 +41,7 @@ Link to relevant documentation section
 Other bugs & improvements
 =========================
 
-- improved error messages for failing inline queries
+- Error messages for failing inline queries now contain location information indicating which inline query failed.
 - bulleted list
 - improvements
 - of smaller
--- a/languages/rust/oso/src/oso.rs
+++ b/languages/rust/oso/src/oso.rs
@@ -8,6 +8,7 @@ use std::sync::Arc;
 
 use crate::host::Host;
 use crate::query::Query;
+use crate::OsoError;
 use crate::{ToPolar, ToPolarList};
 
 /// Oso is the main struct you interact with. It is an instance of the Oso authorization library
@@ -69,10 +70,11 @@ impl Oso {
 
     fn check_inline_queries(&self) -> crate::Result<()> {
         while let Some(q) = self.inner.next_inline_query(false) {
+            let location = q.source_info();
             let query = Query::new(q, self.host.clone());
             match query.collect::<crate::Result<Vec<_>>>() {
                 Ok(v) if !v.is_empty() => continue,
-                Ok(_) => return lazy_error!("inline query result was false"),
+                Ok(_) => return Err(OsoError::InlineQueryFailedError { location }),
                 Err(e) => return lazy_error!("error in inline query: {}", e),
             }
         }
--- a/languages/rust/oso/tests/test_errors.rs
+++ b/languages/rust/oso/tests/test_errors.rs
@@ -89,9 +89,9 @@ fn test_unify_external_not_supported() -> oso::Result<()> {
     Ok(())
 }
 
-/// Test that failing directive sends a sane error message
+/// Test that failing inline query sends a sane error message
 #[test]
-fn test_failing_directive() {
+fn test_failing_inline_query() {
     common::setup();
 
     let oso = Oso::new();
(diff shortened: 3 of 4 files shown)


saolsen (approved):
Looks good!

languages/rust/oso/src/oso.rs
```diff
@@ -69,10 +70,11 @@ impl Oso {
 
     fn check_inline_queries(&self) -> crate::Result<()> {
         while let Some(q) = self.inner.next_inline_query(false) {
+            let location = q.source_info();
             let query = Query::new(q, self.host.clone());
             match query.collect::<crate::Result<Vec<_>>>() {
                 Ok(v) if !v.is_empty() => continue,
-                Ok(_) => return lazy_error!("inline query result was false"),
+                Ok(_) => return Err(OsoError::InlineQueryFailedError { location }),
                 Err(e) => return lazy_error!("error in inline query: {}", e),
```
dhatch, line 78: Should this error be updated too to the new type?

languages/rust/oso/tests/test_errors.rs
```diff
@@ -89,6 +89,26 @@ fn test_unify_external_not_supported() -> oso::Result<()> {
     Ok(())
 }
 
+/// Test that failing directive sends a sane error message
+#[test]
+fn test_failing_directive() {
```
dhatch, line 94: It would be good to rename this `test_failing_inline_query` since we don't use the term directive right now.

dhatch (changes requested):
Just a few small changes

languages/rust/oso/src/oso.rs
Anniepoo: darn git hell - yeah, somehow that goes backed out

Anniepoo: don't understand what you're asking for here.  this IS the new type?

dhatch: On L78 we still use `lazy_error`, but it's representing a failure in an inline query

languages/rust/oso/tests/test_errors.rs
Anniepoo: changed

languages/rust/oso/src/oso.rs
Anniepoo: that's actually a different error - one is failure, the other is that it threw
I think it's better to leave them separate - I suppose we could add a new variant for it.

docs/changelogs/vNEXT.rst
```diff
@@ -41,6 +41,7 @@ Link to relevant documentation section
 Other bugs & improvements
 =========================
 
+- improved error messages for failing inline queries
```
gj, line 44:
```suggestion
- Error messages for failing inline queries now contain location information indicating which inline query failed.
```

gj (approved):
Nice work Annie!

languages/rust/oso/src/oso.rs
gj: I agree that it's fine to leave it — the two errors are distinct. Whether we want to create a special new error type for errors (as distinct from failures) in inline queries feels outside the scope of this PR.

dhatch: Got it. I just looked into the other adapters.  When an error occurs in an inline query we just raise it directly without wrapping it in anything.

We also have a consistent way to attach source information to all errors in other adapters: https://github.com/osohq/oso/blob/83884a26ddb36d1c828187216d5aa713790ed469/languages/python/oso/polar/exceptions.py#L13

For now, I think we should also include the location information on L78 so that an error in an inline query can be debugged by our users (otherwise it will have no stack trace, like failed inline queries before this PR).

I'm going to write down another task in Notion to add stack traces to all errors in the Rust adapter (similar to Python).

dhatch: Er - never mind.  We don't need to do anything because the stack trace will be printed when the error is formatted if it has one.

https://github.com/osohq/oso/blob/83884a26ddb36d1c828187216d5aa713790ed469/polar-core/src/error.rs#L285

dhatch (approved):
lgtm!

PR #797, opened 2021-03-24
Fix typo in docs

AnIrishDuck:
I think this is just a typo that accidentally got missed. Ran into it [here](https://docs.osohq.com/rust/reference/polar/polar-syntax.html).


samscott89 (approved):
Good catch! Thanks!

PR #807, opened 2021-03-26
Annie/partialeval - Fix  f(x) if x.foo semantics

Anniepoo:
This PR alters the behavior of unbound variables encountered in callable positions to constrain them to true.
This fixes the issue "Querying for partial: f(x) if not x.boolean_attr and f(x) if x.boolean_attr"

Since this is a fairly core change and since we've gone a few rounds on what the correct behavior should be,
I've asked @samscott89  and @plotnick  to review. Want to get consensus before we accept.


polar-core/src/vm.rs
```diff
@@ -1406,9 +1406,34 @@ impl PolarVirtualMachine {
             Value::Expression(_) => {
                 return self.query_for_operation(&term);
             }
+            Value::Variable(_a_symbol) => {
+                let val = self.deref(term);
+
+                if val == *term {
+                    // variable was unbound
+                    // apply a constraint to variable that it must be truthy
+                    self.push_goal(Goal::Query {
```
samscott89, line 1415: You could do this more directly by pushing a `Goal::Unify` instead.

polar-core/src/vm.rs
```diff
@@ -1406,9 +1406,34 @@ impl PolarVirtualMachine {
...
+                    self.push_goal(Goal::Backtrack)?;
+                }
+
+                return Ok(QueryEvent::None);
+            }
             _ => {
-                let term = self.deref(term);
-                self.query_for_value(&term)?;
+                // everything else dies horribly and in pain
+                let term = self.deref(term); // shadows
```
samscott89, line 1432: Why do we need to deref here?

polar-core/src/vm.rs
```diff
@@ -1406,9 +1406,34 @@ impl PolarVirtualMachine {
...
+                let val = self.deref(term);
+
+                if val == *term {
+                    // variable was unbound
+                    // apply a constraint to variable that it must be truthy
+                    self.push_goal(Goal::Query {
+                        term: term.clone_with_value(value!(op!(Eq, term.clone(), term!(true)))),
+                    })?;
+                } else {
+                    self.push_goal(Goal::Query { term: val })?;
```
samscott89, line 1419: This is going to make the `x = f(1) and x` case "work", i.e. query `f(1)`.

We spoke about making predicates _not_ first class objects, so perhaps the place to handle this is instead the unification?

polar-core/src/vm.rs
Anniepoo: must attempt to deref to ensure it's unbound

polar-core/src/vm.rs
Anniepoo: When you have cycles you, I, and Alex should revisit all this.

dhatch: I don't necessarily need to be in a full discussion about this -- just adding an input. We generally try to have querying for unbound, constrained or bound values mean the same thing. Isa does not abide by this right now and it causes a bunch of issues.

As written, querying for an unbound results in `x == true`. Querying for a bound value may result in `x == true` if the value is bound to a boolean, or something else if the value is bound to another value.  

If we proceeded with this:

`x and x = f(x)`: What does this mean? `x` will add `x == true`. If the grounder works correctly with calls (it probably doesn't) this query would fail.
`x = f(x) and x` has a different result, we query for `f(x)`.

Having different results if the conjunction is written in a different order is not ideal.

polar-core/src/vm.rs
samscott89: We've already checked above for the `Value::Variable` case. There's no harm in doing it again, but it should be redundant.

Hmm... thinking about it, there's maybe a RestVariable case which this might do something with.

Anniepoo: Nope - sorry - line number shift - this deref is so the error message makes sense, but doh, that's only if it's a variable, and it's not. Getting there slowly

polar-core/src/vm.rs
Anniepoo: resolved by latest commit

samscott89 (commented):
Changes look fine to me.

My main question is whether we're planning to address in this PR whether or not to make `x = f(1)` break?
Today `x = f(1)` works and `x = f(1) and x` is an error
PR as it stands: `x = f(1) and x` will succeed/fail without error (which is better than prev)
Desired: `x = f(1)` errors

Are we making that last change in this PR or as a follow up?

Anniepoo:
re removing predicates as first class - let's defer this

samscott89 (approved):
This LGTM. Let's take removing predicates from being bound as a follow-up.

PR #809, opened 2021-03-28
go library on musl

saolsen:
Use a musl built static lib for golang library on linux.
Validates it works for glibc and musl based distros.

Tested here https://github.com/osohq/oso/runs/2212824195?check_suite_focus=true

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -483,7 +483,7 @@ jobs:
         working-directory: go
       - name: "test musl"
         if: runner.os == 'Linux'
-        run: docker run --rm -v `pwd`:/go golang:1.15-alpine /bin/sh -c 'go build && go test -v ./tests/'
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -481,6 +481,10 @@ jobs:
           go build
           go test -v ./tests/
         working-directory: go
```
gj, line 483: Should this step have a conditional `if: runner.os != 'Linux'`?

saolsen: I don't think so, why? We wanna run the tests on all platforms.

gj: Ahhhh, I was misreading what's going on here. So we've changed the _build_ step to use the musl lib instead of the glibc lib, and we're adding Alpine to the set of test environments to make sure that the new musl-powered lib works on macOS, Windows, Ubuntu, _and_ a musl-based distro.

gj (approved):

PR #810, opened 2021-03-29
Add parser benchmark.

samscott89:


polar-core/benches/bench.rs
```diff
@@ -191,6 +191,16 @@ pub fn many_rules(c: &mut Criterion) {
     });
 }
 
+fn parser(c: &mut Criterion) {
+    let policy = include_str!("roles_policy.polar");
+    c.bench_function("parse_policy", |b| {
+        b.iter(|| {
+            let polar = Polar::new();
+            polar.load_str(policy).unwrap()
```
saolsen, line 199: This is a lot more than just the parser. Would benching running `parse_lines` on the string be a better benchmark?

gj, line 199: Do you think this is better than moving `Polar` initialization outside of the loop and then inside the loop appending an incrementing byte to the policy (to avoid Polar rejecting it as a duplicate policy)? With the goal being benching parsing with as little overhead as possible.

saolsen (approved):

gj (approved):
One question out of curiosity, but happy to ship it regardless of response.

PR #811, opened 2021-03-29
More partial performance improvements

plotnick:
Queries that yield many constraints can take a long time to run. Most of that time is spent simplifying the resulting constraint expressions, and most of _that_ time is spent picking up and applying unifications one at a time. This work attempts to speed up such queries by doing more simplification work up front.

The current implementation is significantly faster (~2x), but drops some constraints (i.e., it is not correct). Draft status can be removed when it's both fast and correct.

PR checklist:

- [ ] Added changelog entry.


samscott89:
Continued in #831

PR #815, opened 2021-04-02
Draft Python add-to-app guide

litonico:
PR checklist:
- [ ] Added changelog entry.

(Shall I write a changelog entry for this?)

The beginning of this does overlap with the Quickstart, but I think that's okay! A little bit overlap in our documentation is good, especially if we expect someone to read the Quickstart and then come back later to read the Add-To-App article.

Here's what I tried to show:
- An easy example of authorization: "a user is allowed to view an expense if they submitted that expense"
- How to register classes with Oso
- How to call into Oso to find out if an action is allowed
- Where to put authorization calls
- How to write the policy we stated above in Polar
- Oso actually solves our problem!
- Polar is more flexible for authorization than Python code: we can easily add more policies.
- It's easy to add new information to a Polar file. (The multiple facts in `user_in_role` could have been an `or` statement, but I wanted to show one way in which Polar is different from other programming languages.)


litonico:
(I would love to sign the CLA, but I've also signed a more-broad contract with Oso that should take precedence! Let me know if I should go ahead and sign the CLA.)

gj:
> (Shall I write a changelog entry for this?)

We haven't been good about doing this for past documentation changes, but I think we should since they're impactful, user-facing changes.

The "next" changelog (which will be assigned a date once we know when the next release is going out) is located here: https://github.com/osohq/oso/blob/main/new-docs/content/any/project/changelogs/NEXT.md

The changelogs are typically organized by library, but I think we could make a new top-level section called "Documentation". What do you think?

> (I would love to sign the CLA, but I've also signed a more-broad contract with Oso that should take precedence! Let me know if I should go ahead and sign the CLA.)

That was my mistake — I should've added you to the allowlist previously. Done in #817.

new-docs/examples/python/getting-started/application/expenses-flask/app/authorization.py
```diff
@@ -21,22 +21,14 @@ def authorize(action, resource):
     else:
         raise Forbidden("Not Authorized!")
 
-
```
gj, line 24: We'll need to make these same changes to https://github.com/osohq/oso-flask-tutorial, which is what we direct readers to clone / follow along with.

Actually, might make the most sense to replace `new-docs/examples/python/getting-started/application/expenses-flask` with https://github.com/osohq/oso-flask-tutorial as a submodule. What do you think?

litonico: I agree! There's no sense in having to keep those manually in sync when we could use a submodule. I'll go ahead and write up a commit for that — would you like that on this PR or separately?

samscott89: This PR would be good.

new-docs/content/python/getting-started/application/index.md
```diff
@@ -54,267 +49,167 @@ The example comes with some example data, which you can load with:
...
-
 $ curl -H "user: alice@foo.com" localhost:5000/expenses/2
 Expense(amount=17743, description='Pug irony.', user_id=1, id=2)
 ```
 
-This pattern is pretty convenient. We can easily apply it elsewhere:
-
-{{< literalInclude path="examples/python/getting-started/application/expenses-flask/app/organization.py"
-                   lines="30-32" >}}
+It works!
```
gj, line 148: nit — imo reads a bit better to not repeat 'it works'

```suggestion
It does!
```

new-docs/content/python/getting-started/application/index.md
```diff
@@ -54,267 +49,167 @@ The example comes with some example data, which you can load with:
...
+Here, we'll add the concept of a *role*, like `accountant`.
+In this case, a user has the role of `accountant` if their job title is "Accountant".
 
-We could change the first highlighted line to:
+{{< literalInclude path="examples/python/getting-started/application/expenses-flask/app/authorization.polar" lines="4-5" >}}
 
-```python
-expense = authorize("create", Expense(**expense_data))
-```
+Here's one place Polar comes in handy: we can add extra information about roles ad-hoc.
```
gj, line 176: nit: from a quick search, I don't think we need the hyphen

```suggestion
Here's one place Polar comes in handy: we can add extra information about roles ad hoc.
```

gj (changes requested):
Have we done any brainstorming / outlining re: how this guide fits in with the Quickstart? It feels like there's a fair amount of overlap between the goals laid out in the OP under `Here's what I tried to show:` and the Quickstart's current goals.

I tried to mark all non-blocking feedback as nits. Overall, well written!

litonico:
I agree with @gj that there's too much overlap between the Quickstart and these add-to-app guide. That was the case even before I rewrote this!

I don't think we should remove the setup code from this guide. If we describe this guide as "how to add Oso to your app", we should walk people through every step of doing that.

One way to handle this would be to make this guide the Quickstart. Because this guide is a bit more complete, I think using this as the quickstart would work well! This guide is longer, but not by a substantial amount. (Our quickstart is 109 lines; this is 192 lines; and for reference, Stripe's quickstart is 151 lines)

We could also make the quickstart a conceptual, rather than in-practice, look at Oso — an "overview" rather than a "quickstart".

What do you think, @samscott89?

samscott89:
> What do you think, @samscott89?

Spoke about this with Lito live. Agree with the high-level suggestions from @gj, we'll do another iteration taking those into consideration.

litonico:
@gj @samscott89 Give this a read and let me know if the tone suits your goals better!

There are still a few things to do:
- The Python example on `main` has diverged — I'll need to read through that and find out how to integrate the changes
- We talked about moving the Flask example code into a 
- I haven't yet (but will) edit the examples to use `hlOpts`

litonico:
Hey @gj and @samscott89, did these updates address your concerns about the first draft? Thanks!

gj:
> Hey @gj and @samscott89, did these updates address your concerns about the first draft? Thanks!

I made some updates on the https://github.com/osohq/oso/tree/gj/lito-python-app-guide-edits branch and have it on my task list to get this PR over the line as part of the current sprint. Thanks for the ping!

samscott89:
Closing in favour of #912

PR #818, opened 2021-04-02
work around delve bug and remove debug printing

saolsen:
go library fixes

### DIFF
--- a/languages/go/internal/host/host.go
+++ b/languages/go/internal/host/host.go
@@ -331,7 +331,16 @@ func (h Host) ToPolar(v interface{}) (*Value, error) {
 		if err != nil {
 			return nil, err
 		}
-		repr := fmt.Sprintf("%v", v)
+		var repr string
+		value := reflect.ValueOf(v)
+		if value.Kind() == reflect.Ptr || value.Kind() == reflect.Interface {
+			value = value.Elem()
+		}
+		if value.IsZero() {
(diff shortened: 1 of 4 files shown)


languages/go/internal/host/host.go
```diff
@@ -331,7 +331,16 @@ func (h Host) ToPolar(v interface{}) (*Value, error) {
 		if err != nil {
 			return nil, err
 		}
-		repr := fmt.Sprintf("%v", v)
+		var repr string
+		value := reflect.ValueOf(v)
+		if value.Kind() == reflect.Ptr || value.Kind() == reflect.Interface {
```
samscott89, line 336: Isn't this the same thing we have on line 292? Are we missing the interface check there or something?

languages/go/internal/host/host.go
```diff
@@ -331,7 +331,16 @@ func (h Host) ToPolar(v interface{}) (*Value, error) {
...
+			value = value.Elem()
+		}
+		if value.IsZero() {
+			repr = fmt.Sprintf("%v", value.Type())
+		} else {
+			repr = fmt.Sprintf("%v", v)
```
samscott89, line 342: While we're making this change, it might be nice to format this as `fmt.Sprintf("%v{%v}", value.Type(), v)`. It can be confusing what type you have in the debugger/trace otherwise.

samscott89 (approved):
LGTM!

PR #819, opened 2021-04-03
Deduplicate spelling dictionary

litonico:
I noticed that I used a word in https://github.com/osohq/oso/pull/815 ("ad-hoc") that was in the old docs spelling dictionary, but not the new docs dictionary, so I merged the old into the new. This also adds adds a flag to `sort` to deduplicate our spelling dictionary.

### DIFF
--- a/new-docs/spelling/Makefile
+++ b/new-docs/spelling/Makefile
@@ -8,6 +8,6 @@ spelling-requirements: .make.spelling-requirements.installed
 	touch $@
 
 sort-allowed-words: allowed_words.txt
-	LANG=C sort -fo allowed_words.txt allowed_words.txt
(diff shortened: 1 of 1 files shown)


dhatch (approved):
lgtm

dhatch (changes requested):
Just discussed with @gj. We may want to just add the old words as needed, some of these are not words that should be in our dictionary (`rc`, `vec`, `usize`, `de`) that may have been needed due to errors / idiosyncrasies with the old docs spell checker.

The change to deduplicate the dictionary is a good idea, we should definitely keep that.

litonico:
Updated! This is now just the deduplication flag — I'll add whichever words I need on a per-PR basis.

dhatch (approved):
lgtm - thanks Lito!

Issue #816, opened 2021-04-02
Go - proposal for new (or alternative) class registration syntax


delicb:
Now, registering a class requires providing an instance of `reflect.Type`, like so:
```go
osoEngine.RegisterClass(reflect.TypeOf(User{}), nil)
```

Instead of that, I propose passing instances of a type directly, thus removing the need for users of the library to use `reflect` package. Previous call would look like this:
```go
osoEngine.RegisterClass(User{}, nil)
```

Of course, this would be applied to all `Register*` methods. 

I do remember seeing this kind of API in other libraries. Also, while not directly the same use case, some ORM libraries use similar approach as well. 

Regarding implementation, `RegisterClass` could accept `interface{}` and do `reflect.TypeOf` in implementation, so backward compatibility should be preserved regarding function signature. Additionally, it is possible to make a distinction between `reflect.Type` and anything else and adapt, so that existing code would work (of course, if someone needs to register type `reflect.Type` only option would be to pass `reflect.TypeOf(reflect.Type{})`, but I do not expect a lot of Polar policies to use `reflect.Type`).

delicb:
Since #820 is merged, (thanks @saolsen and @gj), I guess this can be closed. Or do you prefer to close tickets once the next version is shipped?

gj:
I'll close this after I make sure you're credited for your contribution in the upcoming release's changelog 🙂

PR #820, opened 2021-04-03
[Go] Use instance instead of register.Type for registration

delicb:
An alternative way of registering classes to be used in Polar.
Instead of passing reflect.Type during registration, an empty
instance of the type is passed. The old behavior is still supported,
so passing reflect.Type still works for backward compatibility.

see #816 

PR checklist:

- [x] Added changelog entry.


languages/go/polar.go
```diff
@@ -195,19 +195,28 @@ func (p Polar) registerClass(cls reflect.Type, ctor interface{}, name *string) e
...
 		className = *name
 	}
 
-	err := p.host.CacheClass(cls, className, constructor)
+	err := p.host.CacheClass(realType, className, constructor)
 	if err != nil {
 		return err
 	}
-	newVal := reflect.New(cls)
+	newVal := reflect.New(realType)
```
delicb, line 219: In case an instance (as opposed to `reflect.Type`) is provided, maybe it can be used here, instead of creating a new one?

saolsen: I'm not sure if that would work, I believe this has to be a zero value so safer to leave it like it is for now.

saolsen (commented):
Sorry it took me so long to get to this but it looks great!
I'm going to turn off the draft flag so we can see if anything breaks in the tests.

languages/go/polar.go
delicb: Sure, I do not have a deep understanding of what is happening with the instance value down the stack and how it is used. That is why I did not change it, just left the comment, so I am completely fine with not changing this at the moment. 

I will resolve the comment.

gj (approved):

PR #822, opened 2021-04-04
Added FromPolar and ToPolar implementations for various std::collection types

### LINKED ISSUE #821: rust: Implement ToPolar and FromPolar for HashSet<T>
I have noticed that the traits `ToPolar` and `FromPolar` are implemented for various basic types including `Vec<T>` and `HashMap<&str, T>` but not for `HashSet<T>`.

I think that the "ideal" way to handle this is to implement `ToPolar` for any type that implements `IntoIterator<T>` such that `T` implements `ToPolar`. A similar thing may be possible for  `FromPolar` for types that implement `Extend<T>` and `Default` but I am not so sure.

gjvnq:
This pull request closes #821

PR checklist:

- [ ] Added changelog entry.

### DIFF
--- a/languages/rust/oso/src/host/from_polar.rs
+++ b/languages/rust/oso/src/host/from_polar.rs
@@ -2,8 +2,9 @@
 //! Trait and implementations of `FromPolar` for converting from
 //! Polar types back to Rust types.
 
-use std::collections::hash_map::HashMap;
+use std::collections::{BTreeMap, BTreeSet, BinaryHeap, HashMap, HashSet, LinkedList, VecDeque};
 use std::convert::TryFrom;
+use std::hash::Hash;
 
 use impl_trait_for_tuples::*;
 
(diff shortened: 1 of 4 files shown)


gjvnq:

samscott89:
Hey @gjvnq! Thanks for opening the issue and the PR for it too :)

On first look, this all looks pretty reasonable. Will try to give it a more thorough review soon, however.

languages/rust/oso/src/host/from_polar.rs
```diff
@@ -135,6 +151,76 @@ impl<T: FromPolar> FromPolar for Vec<T> {
...
+                result.insert(T::from_polar(v)?);
+            }
+            Ok(result)
+        } else {
+            Err(TypeError::expected("List").user())
+        }
+    }
+}
+
+impl<T: Eq + Ord + FromPolar> FromPolar for BTreeSet<T> {
```
samscott89, line 196: What is the `Eq` bound needed for?

languages/rust/oso/src/host/from_polar.rs
```diff
@@ -135,6 +151,76 @@ impl<T: FromPolar> FromPolar for Vec<T> {
...
+                result.push_back(T::from_polar(v)?);
+            }
+            Ok(result)
+        } else {
+            Err(TypeError::expected("List").user())
+        }
+    }
+}
+
+impl<T: Eq + Hash + FromPolar> FromPolar for HashSet<T> {
```
samscott89, line 182: Are the `Eq + Hash` bound necessary for creating the set?

samscott89: Ignore this, I see the bounds now

samscott89 (changes requested):
This looks good to me! I think there is one small trait bound that isn't needed.

We should also document these changes in the reference doc: https://github.com/osohq/oso/blame/main/new-docs/content/rust/reference/polar/classes.md#L198-L199

and add a changelog entry. I'm happy to add those to this PR if you would prefer?

gjvnq:
> I'm happy to add those to this PR if you would prefer?

Please, go ahead!

languages/rust/oso/src/host/from_polar.rs
gjvnq: Because a set cannot have two equal elements like a Map can.

Also note how the [declaration for BTreeSet](https://doc.rust-lang.org/src/alloc/collections/btree/set.rs.html#62-66) imples the need for `Eq` and `Ord` while the [declaration for BTreeMap](https://doc.rust-lang.org/src/alloc/collections/btree/map.rs.html#132-136) does not.

samscott89: Ah it's derived on struct itself! That's why I couldn't see it in the docs. Thanks for clarifying!

samscott89 (approved):

samscott89:
Thank you @gjvnq <3

new-docs/content/rust/reference/polar/classes.md
```diff
@@ -101,7 +105,11 @@ matching](polar-syntax#patterns-and-matching).
 
 A Rust [`HashMap`][rust-hashmap] maps to a Polar
 [dictionary](polar-syntax#dictionaries) but requires that the `HashMap` key is
-a `String`:
+a `String`.
+
+Implementations also exist to convert `BTreeMap`s to and
+from Polar dictionaries, but dictionaries are treated as `HashMap` when calling methods.
```
gj, line 111:
```suggestion
from Polar dictionaries, but dictionaries are treated as `HashMap`s when calling methods.
```

PR #826, opened 2021-04-06
[Go] Remove logging call from query logic

stephen-opal:
Removes a superfluous logging call from the Golang query parsing logic. As is, this logging call floods stdout with noise.


stephen-opal:

samscott89 (approved):
Argh, good catch. Thank you!

PR #827, opened 2021-04-06
Algolia Index Generator

veverkap:
This PR adds in the Algolia search index format.  It will generate search.json files in the subfolders that have some duplicated records in them, unsuitable for indexing.  Therefore, in the /search folder there is a small Go program that ingests all of the files and dedupes the records and updates the Algolia index.


veverkap:

new-docs/search/main.go
```diff
@@ -0,0 +1,215 @@
...
+		// this is the main home page and we will include it
+		return 10 // this means it isn't weighted - default is 10
+	case "taxonomy":
+		// this is a taxonomy page, like https://docs.osohq.com/node/tags.html, so we ignore it
+		return 0
+	case "page":
+		return 10
+	case "section":
+		switch section {
+		case "getting-started":
```
dhatch, line 203: Curious why we you have this case for weighing getting started sections to 0

new-docs/Makefile
```diff
@@ -76,6 +80,14 @@ test: install
 clean:
 	ls -A public | grep -v 'favicon.ico' | xargs -I {} echo 'public/'{} | xargs rm -rf
 
+build-search:
+	cd search && go mod vendor && go build -o ../searcher
+
+require-algolia-vars: require-ALGOLIA_APPLICATION_ID require-ALGOLIA_ADMIN_API_KEY require-ALGOLIA_INDEX
+
+search: require-algolia-vars clean build-prod build-search
```
dhatch, line 88: Why is a clean required before building search?

dhatch (commented):
A few comments on the code, nothing significant.

I checked out the index on Algolia and had a few questions:

- do we need to index the page contents? Right now it looks like we only include the summary & description, so a search might not find a term mentioned the page itself.
- for pages in the `any` directory (like changelogs) we have multiple search entries (for example "Release 2021-02-03" matches multiple object ids one with every language). Is there a way to attach multiple language tags to a single search entry so that this result shows for any selected language, but we don't have duplicate results? This could be tricky because some pages in the any directory (like the quickstart) change based on language, and we probably want multiple results for each language so the user can choose the one they want (the any result would just take you to a language selector).  I guess one way to handle this would be a flag in the search index to manually indicate whether it's a page that dynamically changes based on language.

new-docs/Makefile
veverkap: Good question - it's possibly NOT required.  Let me double check if it overwrites the JSON

new-docs/search/main.go
veverkap: https://docs.osohq.com/rust/getting-started.html for example just holds links and a small amount of content for other pages.

veverkap:
>     * do we need to index the page contents? Right now it looks like we only include the summary & description, so a search might not find a term mentioned the page itself.
> 

I have gone and added the contents over and will include that in the index

>     * for pages in the `any` directory (like changelogs) we have multiple search entries (for example "Release 2021-02-03" matches multiple object ids one with every language). Is there a way to attach multiple language tags to a single search entry so that this result shows for any selected language, but we don't have duplicate results? This could be tricky because some pages in the any directory (like the quickstart) change based on language, and we probably want multiple results for each language so the user can choose the one they want (the any result would just take you to a language selector).  I guess one way to handle this would be a flag in the search index to manually indicate whether it's a page that dynamically changes based on language.

Changelogs are purposely downweighted to show up lower in the results, but we will also send along the language of the page that they are on as a facet to the search.

dhatch:
I'm not sure this weighting is working correctly (I may be testing incorrectly though, let me know). I made a test UI and performed a search, see the screenshot. Test ui [here](https://www.algolia.com/realtime-search-demo/test-ui-65fffd4c-365e-4977-a5f1-c38633c138b8)

Screen Shot 2021-04-09 at 5 10 19 PM

veverkap:
> I'm not sure this weighting is working correctly (I may be testing incorrectly though, let me know). I made a test UI and performed a search, see the screenshot. Test ui [here](https://www.algolia.com/realtime-search-demo/test-ui-65fffd4c-365e-4977-a5f1-c38633c138b8)
> 

That is an _okay_ way to test but it's more better to search directly at https://www.algolia.com/apps/KROZ8F05YT/explorer/browse/prod_OSODOCS?searchMode=search

That being said, your results were correct because I had the order of search sorting messed up.  First, I needed to add content:

image

Then I had to include sorting by the weight:

image

Which now yields:

image

dhatch (approved):
LGTM

veverkap:
@dhatch I don't have write access - can you merge this?

Thanks!

veverkap:
Also thanks @gj!

gj:
Yeah we'll handle getting it merged; prepping a new Oso version today so not exactly sure when it'll get in. Thanks @veverkap!

dhatch:
Rebased this and tested by running `make search` from the `docs/` directory (new-docs has been renamed). Looks good

PR #835, opened 2021-04-09
Remove first class references to predicates

Anniepoo:
This PR makes predicates 'Logical's instead of 'Value's and thus removes the ability to reference a predicate as a first class object.

PR checklist:

- [ X] Added changelog entry.


docs/changelogs/vNEXT.rst
```diff
@@ -19,14 +19,25 @@ Breaking changes
...
+Behavior of `a(x) if x` has changed
+===================================
+
    * Now equivalant to `a(x) if x == true`
    * Now works if x is unbound
 
+Polar predicates are no longer first class objects
+==================================================
+
+```
```
samscott89, line 34: We should add a little more context since this is a breaking change:

Previously it was possible to bind predicates to variables, e.g. `p = f(x)`. This supported an undocumented/not recommended feature for unifying predicates, e.g. `f(x) = f(1)` would unify `x = 1`.

This is no longer legal in this release. E.g. (your example).

A similar thing can be achieved by using lists. Instead of `f(x) = f(1)`, you can write `["f", x] = ["f", 1]` if necessary.

samscott89 (changes requested):
This is great. Love how small the change is. Just a few minor comments on the changelog + comment. But looks good otherwise.

polar-wasm-api/tests/polar.rs
```diff
@@ -90,15 +90,6 @@ fn new_query_from_str_succeeds() {
     assert!(is_done_event(event));
 }
 
-#[wasm_bindgen_test]
```
samscott89, line 93: We should keep the test but make it fail in a different way. E.g.

Makefile
```diff
@@ -81,6 +81,7 @@ fmt-js:
 
 fmt-python:
 	$(MAKE) -C languages/python fmt
+	black languages/python
```
dhatch, line 84: I don't know why we need this addition to the makefile, the sub makes should already run `black`.

@Anniepoo was seeing that test files were not properly reformatted

dhatch: However, black does run for test files.

```
(venv) ➜  oso git:(main) ✗ black -v .
... [snip] ...
/Users/dhatch/code/oso/languages/python/oso/tests/test_partial.py wasn't modified on disk since last run.
/Users/dhatch/code/oso/languages/python/oso/tests/test_polar.py wasn't modified on disk since last run.
```

samscott89: Yeah, was just checking the same thing

samscott89 (approved):

Makefile
gj: so do we need this or not?

new-docs/content/any/project/changelogs/NEXT.md
```diff
@@ -19,6 +19,26 @@ draft: true
...
+#### Breaking change 1
+
+Previously it was possible to bind predicates to variables, e.g. `p = f(x)`.
+This supported an undocumented feature for unifying predicates,
+e.g. `f(x) = f(1)` would unify `x = 1`.
+
+To avoid confusion with querying rules, binding predicates
+to variables is no longer legal in this release. For example:
+
+```
```
gj, line 31:
```suggestion
```polar
```

new-docs/content/any/project/changelogs/NEXT.md
```diff
@@ -19,6 +19,26 @@ draft: true
   before upgrading.
 {{% /callout %}}
 
+#### Breaking change 1
```
gj, line 22:
```suggestion
#### Unifying with a predicate now fails at parse time
```

new-docs/content/any/project/changelogs/NEXT.md
```diff
@@ -19,6 +19,26 @@ draft: true
...
+```
+f(x,y) if x < 3 and y = g(x);
+```
+
+will now fail at parse time.
+
+A similar thing can be achieved by using lists.
+Instead of `f(x) = f(1)`, you can write `["f", x] = ["f", 1]` if necessary.
+
+#### Breaking change 2
```
gj, line 40: Trying to come up with good title here but brain failing

polar-core/src/parser.rs
```diff
@@ -137,7 +133,7 @@ mod tests {
             "{}",
             exp2.to_polar()
         );
-
+        /*
```
gj, line 136: Kill this?

gj (changes requested):

polar-core/src/polar.lalrpop
```diff
@@ -248,6 +244,17 @@ LogExp: Term = {
     <ExpectLogical<Exp1<"Term">>>
 }
 
+// `ValueOrLogic` is used to do some simple parse-time 
+// checks for correctly formed expressions. For example,
+// For example, `x in y` is a logical expression, so it does
```
gj, line 249:
```suggestion
// `x in y` is a logical expression, so it does
```

polar-core/src/polar.lalrpop
```diff
@@ -248,6 +244,17 @@ LogExp: Term = {
...
+// not make sense to compare the result with a value,
```
gj, line 250:
```suggestion
// not make sense to compare the result with a value.
```

polar-core/src/polar.lalrpop
```diff
@@ -248,6 +244,17 @@ LogExp: Term = {
...
+// `x in y > 0` fails to parse because `A > B` expects
+// A and B to be values, but `x in y` is a logical term
```
gj, line 252:
```suggestion
// A and B to be values, but `x in y` is a logical term.
```

polar-wasm-api/tests/polar.rs
```diff
@@ -93,10 +93,10 @@ fn new_query_from_str_succeeds() {
 #[wasm_bindgen_test]
 fn new_query_from_str_errors() {
     let polar = polar_wasm_api::Polar::wasm_new();
-    let mut query = polar.wasm_new_query_from_str("x").unwrap();
+    let mut query = polar.wasm_new_query_from_str("[]").unwrap();
     let err: Error = query.wasm_next_event().unwrap_err().dyn_into().unwrap();
     assert_eq!(err.name(), "RuntimeError::TypeError");
-    assert_eq!(err.message(), "trace (most recent evaluation last):\n  in query at line 1, column 1\n    x\nType error: can't query for: x at line 1, column 1");
+    assert_eq!(err.message(), "trace (most recent evaluation last):\n  in query at line 1, column 1\n    []\nType error: [] isn\'t something that is true or false, so can\'t be a condition at line 1, column 1");
```
gj, line 99: doesn't have to be in this PR, but we should fix the comma splice in this error message

Makefile
samscott89: I honestly can't tell.

new-docs/content/any/project/changelogs/NEXT.md
samscott89: "Allow querying for unbound variables"?

gj: Did behavior *only* change when querying for an unbound?

samscott89: Well, technically the other PR #835 removed the other behaviour this created.

languages/go/internal/host/host.go
```diff
@@ -289,7 +289,7 @@ func (h Host) ToPolar(v interface{}) (*Value, error) {
 	// check composite types
 	rt := reflect.ValueOf(v)
 	// deref pointer
-	if rt.Kind() == reflect.Ptr || rt.Kind() == reflect.Interface{
+	if rt.Kind() == reflect.Ptr || rt.Kind() == reflect.Interface {
```
dhatch, line 292: did this somehow miss ci?

docs/content/any/project/changelogs/NEXT.md
```diff
@@ -19,9 +19,30 @@ draft: true
...
+
+A similar thing can be achieved by using lists.
+Instead of `f(x) = f(1)`, you can write `["f", x] = ["f", 1]` if necessary.
+
+#### Query for unbound variables
+
+Querying for an unbound variable `x` is inferred to mean `x = true`.
+
+Before: `f(x) if x;` would error if `x` was unbound.  
+Now: this is interpreted as `f(x) if x = true`;`
```
dhatch, line 45:
```suggestion
Now: this is interpreted as `f(x) if x = true`;
```

languages/go/internal/host/host.go
samscott89: We don't have a go check/lint/fmt step in CI by the looks of things.

docs/content/any/project/changelogs/NEXT.md
samscott89: Other way around technically

docs/content/any/project/changelogs/NEXT.md
```diff
@@ -19,9 +19,30 @@ draft: true
...
+
+A similar thing can be achieved by using lists.
+Instead of `f(x) = f(1)`, you can write `["f", x] = ["f", 1]` if necessary.
+
+#### Query for unbound variables
+
+Querying for an unbound variable `x` is inferred to mean `x = true`.
+
+Before: `f(x) if x;` would error if `x` was unbound.  
+Now: this is interpreted as `f(x) if x = true`;
```
samscott89, line 45:
```suggestion
Now: this is interpreted as `f(x) if x = true;`.
```

Makefile
Anniepoo: My PR was failing because one line in a python test was getting reformatted by black - strange part is, after I manually black'ed that directory, it's still failing, so as of this moment I'm baffled.

gj: Maybe you have a different version of `black` than the one used in CI?

gj: Either way, it seems like we should remove this line

dhatch: Going to drop this for now, we aren't entirely sure what caused black to not run on the test file on Annie's machine. If we see another formatting issue we can revisit and maybe just replace all the sub make calls with a top level `black languages/python`.

new-docs/content/any/project/changelogs/NEXT.md
gj: Answer after reading #807: Yes, behavior only changed when querying for an unbound, and your proposed changelog entry is 👍

gj (approved):

PR #837, opened 2021-04-12
Remove old docs

gj:
PR checklist:

- [ ] Update Go quickstart to conform to others & add it to test-quickstarts target.


.gitmodules
```diff
@@ -1,28 +1,22 @@
...
+	path = docs/examples/roles/sqlalchemy/basic
+	url = https://github.com/osohq/oso-sqlalchemy-roles-guide.git
+	branch = basic_roles_complete
+[submodule "docs/examples/quickstart/go"]
+	path = docs/examples/quickstart/go
+	url = https://github.com/osohq/oso-go-quickstart.git
+[submodule "docs/examples/quickstart/java"]
+	path = docs/examples/quickstart/java
 	url = https://github.com/osohq/oso-java-quickstart.git
-        branch = main
```
samscott89, line 4: This doesn't work for me without specifying all the branches as main.

gj: ah right; I think that's due to a difference in Git versions

gj: fixed in b83a9c9c

samscott89: Thank you kindly

samscott89 (approved):
I reviewed this as best as I could:

- ran the docs locally
- ran the docs tests
- looked at whatever changes seemed substantive. though there's a lot of noise with old docs files being "updated" with the move.

This looks good to me.

gj:
> Edit to add: I don't understand what
> 
> > ```
> > Update Go quickstart to conform to others & add it to test-quickstarts target.
> > ```
> 
> is referring to on the PR checklist. This is not introduced by this PR, right?

Correct, just a general observation that we weren't including the Go quickstart in the `test-quickstarts` make target that gates releases.

PR #838, opened 2021-04-12
Fix quickstart tests

gj:
Depends on:
- [ ] https://github.com/osohq/oso-go-quickstart/pull/4


.github/workflows/release.yml
```diff
@@ -687,7 +687,7 @@ jobs:
       - uses: actions/checkout@v2
       - name: Initialize submodules
         working-directory: docs/examples/quickstart
-        run: git submodule update --init -- oso-java-quickstart oso-nodejs-quickstart oso-python-quickstart oso-ruby-quickstart oso-rust-quickstart
+        run: git submodule update --init -- java nodejs python ruby rust
```
samscott89, line 690: How about using `make update-submodules` from `docs/examples` instead?

gj: good idea; updated

dhatch (approved):
lgtm

PR #845, opened 2021-04-14
Add go fmt check to CI.

samscott89:


samscott89:
I rebased so the history is hidden, but here's the successfully failing lint test: https://github.com/osohq/oso/actions/runs/746811036

gj (approved):

.github/workflows/test.yml
```diff
@@ -93,6 +93,11 @@ jobs:
           skipCommit: true
           args: "--dry-run --set-exit-if-changed"
 
+      ## Check Go
+      - name: Check formatting of Java library
```
gj, line 97:
```suggestion
      - name: Check formatting of Go library
```

PR #847, opened 2021-04-14
2021-04-14 changelog updates

### LINKED ISSUE #816: Go - proposal for new (or alternative) class registration syntax

gj:
closes #816

### DIFF
--- a/docs/content/any/project/changelogs/2021-03-25.md
+++ b/docs/content/any/project/changelogs/2021-03-25.md
@@ -35,7 +35,7 @@ in order to use across threads.
 
 ##### It's now possible to use Oso in Web Workers
 
-Big thanks to [@togmund](https://github.com/togmund) for submitting a patch
+Big thanks to [`@togmund`](https://github.com/togmund) for submitting a patch
 that enables Oso to run in [Web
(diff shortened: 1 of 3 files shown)


dhatch (approved):
thank you

PR #848, opened 2021-04-14
[Makefiles] Add clean targets for top level and sub projects

### LINKED ISSUE #836: clean target on main makefile
the main makefile has no clean target.  It would be nice to have one, so you could be assured you'd cleaned everything.

joshrotenberg:
Fixes #836.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -93,6 +93,11 @@ jobs:
           skipCommit: true
           args: "--dry-run --set-exit-if-changed"
 
+      ## Check Go
+      - name: Check formatting of Go library
+        working-directory: "languages/go"
+        run: make lint
+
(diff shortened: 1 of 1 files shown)


joshrotenberg:
The python and ruby Makefiles don't have a clean target. My experience with those ecosystems is out of date, so I'm going to naively assume that is because cleaning up a build for them isn't a thing. If that's not the case, though, I can add their targets and include them in the big clean.

gj (approved):

gj:
Pushed a small change to clean up the generated WASM files in the `languages/js/src` directory (`yarn clean` only cleans up the `languages/js/dist` dir).

LGTM; thanks for (yet another) great contribution @joshrotenberg!

PR #854, opened 2021-04-15
Correct follower backtracking bug.

dhatch:
Found this bug while investigating grounding & negation. The followers in the binding manager had an assumption that every binding or add constraint operation would always result in the same number of new bindings in the binding stack in the parent and followers.

This is not necessarily true (see the added regression test). If a variable is bound in a cycle in a parent binding manager, but not in a follower, adding a constraint on that variable will add multiple new bindings in the parent (one for each member of the cycle), but only 1 in the child. This results in failing to erase enough bindings in the follower on a backtrack. 

I picked this up from this panic:

```
panic!("Unexpected bound variable {} in constraint.", var);
```

To fix this, the `Bsp` records every `Bsp` in the followers. This could make things slightly slower because it needs to be cloned sometimes. We could try making `followers` in `Bsps` a `Vec` which may make cloning slightly faster.

PR checklist:

- [ ] Added changelog entry.

### DIFF
--- a/polar-core/src/bindings.rs
+++ b/polar-core/src/bindings.rs
@@ -216,9 +216,13 @@ impl BindingManager {
     }
 
     /// Reset the state of `BindingManager` to what it was at `to`.
-    pub fn backtrack(&mut self, mut to: Bsp) {
+    pub fn backtrack(&mut self, to: &Bsp) {
         self.do_followers(|follower_id, follower| {
-            follower.backtrack(to.followers.remove(&follower_id).unwrap_or_default());
+            if let Some(follower_to) = to.followers.get(&follower_id) {
+                follower.backtrack(follower_to);
+            } else {
+                follower.backtrack(&Bsp::default());
+            }
             Ok(())
         })
         .unwrap();
@@ -310,10 +314,10 @@ impl BindingManager {
     }
 
     pub fn variable_state(&self, variable: &Symbol) -> VariableState {
-        self.variable_state_at_point(variable, self.bsp())
+        self.variable_state_at_point(variable, &self.bsp())
     }
 
-    pub fn variable_state_at_point(&self, variable: &Symbol, bsp: Bsp) -> VariableState {
+    pub fn variable_state_at_point(&self, variable: &Symbol, bsp: &Bsp) -> VariableState {
         let bsp = bsp.bsp;
         let mut next = variable;
         while let Some(value) = self.value(next, bsp) {
@@ -356,16 +360,10 @@ impl BindingManager {
     }
 
     pub fn bindings(&self, include_temps: bool) -> Bindings {
-        self.bindings_after(
-            include_temps,
-            Bsps {
-                bsp: 0,
-                followers: HashMap::default(),
-            },
-        )
+        self.bindings_after(include_temps, &Bsp::default())
     }
 
-    pub fn bindings_after(&self, include_temps: bool, after: Bsp) -> Bindings {
+    pub fn bindings_after(&self, include_temps: bool, after: &Bsp) -> Bindings {
         let mut bindings = HashMap::new();
         for Binding(var, value) in &self.bindings[after.bsp..] {
             if !include_temps && var.is_temporary_var() {
@@ -528,11 +526,15 @@ impl BindingManager {
     }
 
     fn _variable_state(&self, variable: &Symbol) -> BindingManagerVariableState {
-        self._variable_state_at_point(variable, self.bsp())
+        self._variable_state_at_point(variable, &self.bsp())
     }
 
     /// Check the state of `variable` at `bsp`.
-    fn _variable_state_at_point(&self, variable: &Symbol, bsp: Bsp) -> BindingManagerVariableState {
+    fn _variable_state_at_point(
+        &self,
+        variable: &Symbol,
+        bsp: &Bsp,
+    ) -> BindingManagerVariableState {
         let bsp = bsp.bsp;
         let mut path = vec![variable];
         while let Some(value) = self.value(path.last().unwrap(), bsp) {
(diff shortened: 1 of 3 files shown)


polar-core/src/bindings.rs
```diff
@@ -16,9 +16,15 @@ pub struct Binding(pub Symbol, pub Term);
 pub type BindingStack = Vec<Binding>;
 pub type Bindings = HashMap<Symbol, Term>;
 
-pub type Bsp = usize;
+pub type Bsp = Bsps;
 pub type FollowerId = usize;
 
+#[derive(Clone, Debug, Default, PartialEq)]
+pub struct Bsps {
+    bsp: usize,
```
gj, line 24: What do you think about renaming this `pointer` or `ptr` or something to indicate that this is _actually_ the bsp? Or maybe renaming the struct to something like `BspWithFollowers` or `BspTree`?

Currently finding things like `self.bsp().bsp` and `let bsp = bsp.bsp;` a little hard to read.

polar-core/src/bindings.rs
```diff
@@ -505,11 +526,16 @@ impl BindingManager {
...
     }
 
     /// Check the state of `variable` at `bsp`.
-    fn _variable_state_at_point(&self, variable: &Symbol, bsp: Bsp) -> BindingManagerVariableState {
+    fn _variable_state_at_point(
+        &self,
+        variable: &Symbol,
+        bsp: &Bsp,
+    ) -> BindingManagerVariableState {
+        let bsp = bsp.bsp;
```
gj, line 538: If all we care about is the current `BindingManager`'s `bsp`, isn't calling `self.bsp()` a bit overkill in that it will also recursively calculate & collect bsps for followers and followers followers and so on?

polar-core/src/inverter.rs
```diff
@@ -208,8 +208,9 @@ impl Runnable for Inverter {
                         // out to the parent VM.
                         let constraints =
                             results_to_constraints(self.results.drain(..).collect::<Vec<_>>());
-                        let constraints =
-                            filter_inverted_constraints(constraints, &self.vm, self.bsp);
+                        let mut bsp = Bsp::default();
+                        std::mem::swap(&mut self.bsp, &mut bsp);
+                        let constraints = filter_inverted_constraints(constraints, &self.vm, bsp);
```
gj, line 213: this is neat, but I think a comment would be helpful

polar-core/src/bindings.rs
```diff
@@ -823,5 +849,28 @@ mod test {
...
+        let b2 = BindingManager::new();
+        let b2_id = b1.add_follower(b2);
+
+        b1.add_constraint(&term!(op!(Gt, term!(sym!("x")), term!(1))))
+            .unwrap();
+
+        let bsp = b1.bsp();
+
+        b1.bind(&sym!("a"), term!(sym!("x"))).unwrap();
+
```
gj, line 868: While the behavior is already tested in the `test_followers()` test, it would still be kind of nice to see an `assert!()` before the `backtrack()` to complement the one after.

gj (commented):
Some nits on naming & efficiency, but mostly LGTM

polar-core/src/bindings.rs
```diff
@@ -196,8 +202,12 @@ impl BindingManager {
...
                     op = e;
                 }
-                BindingManagerVariableState::Bound(_) => {
-                    panic!("Unexpected bound variable {} in constraint.", var);
+                BindingManagerVariableState::Bound(v) => {
+                    panic!(
+                        "Unexpected bound variable {} in constraint. {var} = {val}",
+                        var = var,
+                        val = v
+                    );
```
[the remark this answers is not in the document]
dhatch: Not sure how this compiled.

polar-core/src/bindings.rs
```diff
@@ -16,9 +16,15 @@ pub struct Binding(pub Symbol, pub Term);
 pub type BindingStack = Vec<Binding>;
 pub type Bindings = HashMap<Symbol, Term>;
 
-pub type Bsp = usize;
+pub type Bsp = Bsps;
 pub type FollowerId = usize;
 
+#[derive(Clone, Debug, Default, PartialEq)]
+pub struct Bsps {
```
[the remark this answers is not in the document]
dhatch: Having the indirection through `Bsp` makes it easier to change the type in the future. It made it pretty easy for me to change from `usize` to a struct rather than making updates everywhere.

polar-core/src/bindings.rs
dhatch: Renamed to `bindings_index`.

polar-core/src/bindings.rs
dhatch: Yeah, it is - but usually a binding manager will not have followers so I didn't think it's a big deal & didn't want to add some workaround to do it.

gj (approved):
one nit but LGTM

polar-core/src/bindings.rs
```diff
@@ -301,17 +317,18 @@ impl BindingManager {
     }
 
     pub fn variable_state(&self, variable: &Symbol) -> VariableState {
-        self.variable_state_at_point(variable, self.bsp())
+        self.variable_state_at_point(variable, &self.bsp())
     }
 
-    pub fn variable_state_at_point(&self, variable: &Symbol, bsp: Bsp) -> VariableState {
+    pub fn variable_state_at_point(&self, variable: &Symbol, bsp: &Bsp) -> VariableState {
+        let bsp = bsp.bindings_index;
```
gj, line 324: GitHub swallowed my first attempt to say this:

nit: I think we should rename the local variables here & on line 541 from `bsp` to `index` to make it clear that it's an index instead of a full `Bsp` struct.

dhatch: good call

PR #859, opened 2021-04-19
Derive `PolarClass` on Rust enums

seanchen1991:
- [ ] `PolarClass` may now be derived on Rust enums.  

Hello :wave:

I'm attempting to add support for Rust enums on the `#[derive(PolarClass)]` proc macro. This involves having the procedural macro generate code to store each enum variant on the `Class` that is being constructed. A loop in `register_class` then registers each constant. 

I sought to keep the API as simple as possible such that someone looking to add this functionality in their Rust project only needs to add `#[derive(PolarClass)]` onto their enums (thanks to @joshrotenberg for his input on this 🙂 ). However, this doesn't _quite_ work; I'm hoping some additional eyes on it will help. 

The core of the problem seems to be fact that the constants are initially stored on the class as trait objects. Those are then passed around a few times before they get handed to `register_constant`. The compiler complains about the size of the trait objects not being known. I believe there are ways of handling this in nightly Rust, but I'm not sure what the pattern for addressing this is on stable. 

There's also the distinct possibility that I'm missing something simple 🙂


seanchen1991:

dhatch:
Hi @seanchen1991! Thanks for this PR! I'll get someone on the team's eyes on it!

samscott89:
Hey @seanchen1991!

I like the proc macro part, that looks nicely done.

For the registration step, we currently require passing in an owned object to `register_constant`. That means it won't be possible to register the constants with this approach since it's wrapped inside Arc (and that's unavoidable since we need `Class` to be `Clone`).

We _could_ loosen this requirement... but that seems like a distraction here.

One way to avoid this would be to make the `add_constant` take in a closure... e.g.

```rust
class_builder.add_constant(|| Enum::Variant, "Enum::Variant")
```

where `add_constant` accepts a closure `F: Fn() -> T`?

The nice thing is this would allow somebody to initialise a constant at registration time too?

Alternatively, we could have a single `after_register_class` hook that takes in a closure to do any initialization logic.

seanchen1991:
A bunch of tests seem to be failing now as a result of the changes I introduced. It looks like deriving `PolarClass` on types introduced in test files isn't working; I'm getting a bunch of test errors saying `get_polar_class_builder` can't be found on types that derive them. 

I have a `main.rs` file that I added just to run `cargo expand` against to check the generated code from the proc macro and `get_polar_class_builder` is being generated for types that `#[derive(PolarClass)]`. Is there something that needs to be done in order to propagate the changes to `PolarClass` to the types in the tests?

seanchen1991:
@samscott89 Any suggestions on how this should be tested and where those tests should live? 🙂

languages/rust/oso-derive/src/lib.rs
```diff
@@ -94,21 +92,38 @@ pub fn derive_polar_class_impl(ts: TokenStream) -> TokenStream {
...
+        Data::Enum(DataEnum { variants, .. }) => {
+            for variant in variants {
+                let vident = variant.ident;
+                let vname = format!("{}::{}", class_name, vident);
+                constants.push(quote! {
+                    .add_constant(#type_name::#vident, #vname)
+                });
             }
         }
+        _ => unimplemented!(),
```
samscott89, line 126: Can we make this return a friendlier error?

seanchen1991:
I'm thinking of adding a few more assertions that ensure that enums with more complex shapes are registered correctly. Something like this:
```rust
    #[derive(Clone, Debug, PartialEq, PolarClass)]
    enum Event {
        PageLoad,
        PageUnload,
        KeyPress(char),
        Paste(String),
        Click { x: i64, y: i64 },
    }

    test.load_str(
        r#"
        is_pageload(Event::PageLoad); 
        is_pageunload(Event::PageUnload);
        is_keypress(Event::KeyPress);
        is_paste(Event::Paste);
        is_click(Event::Click);"#
    );
    
    test.qvar_one(r#"is_pageload(x)"#, "x", Event::PageLoad);
    test.qvar_one(r#"is_pageunload(x)"#, "x", Event::PageUnload);
    test.qvar_one(r#"is_keypress(x)"#, "x", Event::KeyPress);
    test.qvar_one(r#"is_paste(x)"#, "x", Event::Paste);
    test.qvar_one(r#"is_click(x)"#, "x", Event::Click);
```

The way the proc macro currently handles this is a variant such as `Event::Click { x: i64, y: i64 }` gets registered as just `Event::Click`. This won't compile though since `Event::Click` isn't a valid enum variant. Is there a way to check the stringified representation of the constant?

samscott89:
> 
> 
> I'm thinking of adding a few more assertions that ensure that enums with more complex shapes are registered correctly. Something like this:
> ...
> The way the proc macro currently handles this is a variant such as `Event::Click { x: i64, y: i64 }` gets registered as just `Event::Click`. This won't compile though since `Event::Click` isn't a valid enum variant. Is there a way to check the stringified representation of the constant?

Hmm, this is an interesting problem. Does it maybe suggest treating enum variants as classes instead of constants?

What would be awesome, is if you could write:

```
allow(user, "send", event: Event) if user.is_admin;

allow(user, "send", event: Event::Click { x: x, y: y}) if x > y;

allow(user, "send", Event::PageLoad);

```

Which would require:

- Register unit variants as constants still
- Making struct variants registered as classes
- Implement the `is_subclass` check for enums (we don't have anything like this right now, this could be a lot of work).

If this seems like too much to get it working, I think we should just throw an error instead for any enums with fields.

samscott89:
Also, I made a couple of small changes:

- https://github.com/osohq/oso/pull/859/commits/b7054e66d7818dfa09b69af4f1e6caa97d21ae7e
- https://github.com/osohq/oso/pull/859/commits/806b6b3b70541453082966ef137090c6e0caf65d

seanchen1991:
> Register unit variants as constants still
Making struct variants registered as classes
Implement the is_subclass check for enums (we don't have anything like this right now, this could be a lot of work).

I agree that this is an interesting idea. Perhaps it should be handled in a follow-up PR?

samscott89:
> 
> 
> > Register unit variants as constants still
> > Making struct variants registered as classes
> > Implement the is_subclass check for enums (we don't have anything like this right now, this could be a lot of work).
> 
> I agree that this is an interesting idea. Perhaps it should be handled in a follow-up PR?

Sounds good

languages/rust/oso-derive/src/lib.rs
```diff
@@ -94,21 +92,40 @@ pub fn derive_polar_class_impl(ts: TokenStream) -> TokenStream {
...
-                let attr = field.ident.unwrap();
-                let name = attr.to_string();
-                getters.push(quote! {
-                    .add_attribute_getter(#name, |recv: &#type_name| recv.#attr.clone())
-                })
+            Fields::Unnamed(_uf) => {}
+            Fields::Unit => {}
+        },
+        Data::Enum(DataEnum { variants, .. }) => {
+            for variant in variants {
```
samscott89, line 118: Return an error like below if the variant is not a unit variant

samscott89 (approved):
This looks great! Thanks for doing this! Just the one small place to do some error handling, and this looks good to go.

languages/rust/oso-derive/src/lib.rs
```diff
@@ -94,21 +92,47 @@ pub fn derive_polar_class_impl(ts: TokenStream) -> TokenStream {
...
+                match variant.fields {
+                    Fields::Unit => {
+                        let vident = variant.ident;
+                        let vname = format!("{}::{}", class_name, vident);
+                        constants.push(quote! {
+                            .add_constant(#type_name::#vident, #vname)
+                        });        
+                    } 
+                    _ => {
+                        return quote_spanned! { type_name.span() => compile_error!("#[derive(PolarClass)] is currently only supported on enums with unit variants."); }.into();
```
samscott89, line 128:
```suggestion
                        return quote_spanned! { vident.span() => compile_error!("#[derive(PolarClass)] is currently only supported on enums with unit variants."); }.into();
```

(I think?)

samscott89, line 128:
```suggestion
                        return quote_spanned! { variant.ident..span() => compile_error!("#[derive(PolarClass)] is currently only supported on enums with unit variants."); }.into();
```

(I think?)

samscott89:
Thanks @seanchen1991 !

PR #860, opened 2021-04-20
roles 2

saolsen:


languages/python/sqlalchemy-oso/tests/test_roles2.py
```diff
@@ -0,0 +1,146 @@
...
+    allow(actor, action, resource) if
+        Roles.role_allows(actor, action, resource);
+    """
+    oso.load_str(policy)
+
+    # engine = create_engine("sqlite:///roles.db")
+    engine = create_engine("sqlite://")
+    # @NOTE: Right now this has to happen after enabling oso roles to get the
+    #        tables.
+    Base.metadata.create_all(engine)
```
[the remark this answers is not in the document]
dhatch: Might be something to ask the SQLAlchemy guy, or look at other SQLAlchemy libs for.

languages/python/sqlalchemy-oso/tests/test_roles2.py
```diff
@@ -0,0 +1,146 @@
...
+    steve = User(name="steve")
+
+    objs = [leina, steve, osohq, oso_repo, bug]
+    for obj in objs:
+        session.add(obj)
+    session.commit()
+
+    # @TODO: How the heck should this work?
+    # not like this...
+    roles.session = session
```
dhatch, line 127: This session is used within the policy when `is_allowed` is called?

dhatch: If we were just writing this entire library for SQLAlchemy I'd say this should just be passed into `is_allowed` as you've done for `assign_role`. Obviously that would not keep the API consistent `Oso` when you're not using SQLAlchemy.

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles2.py
```diff
@@ -0,0 +1,477 @@
...
+        if not self.configured:
+            self._configure()
+        func(self, *args, **kwargs)
+
+    return wrapper
+
+
+class OsoRoles:
+    def __init__(self, oso, sqlalchemy_base, user_model):
+        # @Q: Is this where I should create the models?
```
dhatch, line 60: We also might want to think about how migrations would work. SQLAlchemy doesn't really have a standard migration tool, although I think a lot of people use Alembic.

Not for this first release... but we probably need to figure it out before the next one.

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles2.py
```diff
@@ -0,0 +1,477 @@
...
+        # Would this even work? Do these have to happen at a
+        # certain time or something to get created in the database?
+
+        self.session = None
+
+        user_pk_type = inspect(user_model).primary_key[0].type
```
dhatch, line 66: Ideally this would work on compound keys.

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles2.py
```diff
@@ -0,0 +1,477 @@
...
+            resource_id = Column(String)  # Most things can turn into a string lol.
+            role = Column(String)
+
+        class Permission(sqlalchemy_base):
+            __tablename__ = "permissions"
+            id = Column(Integer, primary_key=True)
+            resource_type = Column(String)
+            name = Column(String)
+
+        class Role(sqlalchemy_base):
```
dhatch, line 87: Unique constraint or compound PK on name & resource type?

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles2.py
```diff
@@ -0,0 +1,477 @@
...
+        # Would this even work? Do these have to happen at a
+        # certain time or something to get created in the database?
+
+        self.session = None
+
+        user_pk_type = inspect(user_model).primary_key[0].type
+        user_pk_name = inspect(user_model).primary_key[0].name
+        user_table_name = user_model.__tablename__
+
+        # Tables for the management api to save data.
```
dhatch, line 70: Need a way to define or provide the correct indexes for these tables.

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles2.py
```diff
@@ -0,0 +1,477 @@
...
+        class RolePermission(sqlalchemy_base):
+            __tablename__ = "role_permissions"
+            id = Column(Integer, primary_key=True)
+            role = Column(String, ForeignKey("roles.name"))
+            permission_id = Column(Integer, ForeignKey("permissions.id"))
+
+        class RoleImplication(sqlalchemy_base):
+            __tablename__ = "role_implications"
+            id = Column(Integer, primary_key=True)
+            from_role = Column(String, ForeignKey("roles.name"))
```
dhatch, line 101: This needs to reference name & resource type, right? (Makes me feel like we should have a compound PK).

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles2.py
```diff
@@ -0,0 +1,477 @@
...
+            user_id = Column(
+                user_pk_type, ForeignKey(f"{user_table_name}.{user_pk_name}")
+            )
+            resource_type = Column(String)
+            resource_id = Column(String)  # Most things can turn into a string lol.
+            role = Column(String)
+
+        class Permission(sqlalchemy_base):
+            __tablename__ = "permissions"
+            id = Column(Integer, primary_key=True)
```
dhatch, line 83: Any reason permission has an id col but not role?

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles2.py
```diff
@@ -0,0 +1,477 @@
...
+        # Tables for the management api to save data.
+        class UserRole(sqlalchemy_base):
+            __tablename__ = "user_roles"
+            id = Column(Integer, primary_key=True)
+            user_id = Column(
+                user_pk_type, ForeignKey(f"{user_table_name}.{user_pk_name}")
+            )
+            resource_type = Column(String)
+            resource_id = Column(String)  # Most things can turn into a string lol.
+            role = Column(String)
```
dhatch, line 79: Should the `role` column be a relationship?

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles2.py
```diff
@@ -0,0 +1,477 @@
...
+
+        self.configured = False
+
+    def _configure(self):
+        # @TODO: ALLLLL of the validation needed for the role model.
+
+        # @TODO: Figure out where this session should really come from.
+        assert self.session
+
+        self.resources = {}
```
dhatch, line 124: skipped reading most of this

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles2.py
```diff
@@ -0,0 +1,477 @@
...
+                    role = Role(
+                        name=name,
+                        python_class=python_class,
+                        permissions=role_permissions,
+                        implied_roles=implied_roles,
+                    )
+                    self.roles[role.name] = role
+
+        # Sync static data to the database.
+        self.session.execute("delete from role_permissions")
```
dhatch, line 273: This means any time the application starts (and calls `roles.enable()`) the entire role data is deleted and updated.

We probably need a workflow that performs this configuration & syncing of the policy with the database as a script, right?

Before release?

dhatch: Before the next release we are going to need some way to update this data from the policy without doing a complete replacement and a without eliminating any dynamic customizations (if we add that capability)

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles2.py
```diff
@@ -0,0 +1,477 @@
...
+        if role:
+            return True
+        else:
+            return False
+
+    def enable(self):
+        class Roles:
+            @staticmethod
+            def role_allows(user, action, resource):
+                if not self.configured:
```
dhatch, line 451: We probably shouldn't have this be lazily configured. (Maybe if we have a debug mode it's fine).

Otherwise the first request that performs authorization will create a bunch of data before executing

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles2.py
```diff
@@ -0,0 +1,477 @@
...
+        resource_pk_name = inspect(resource.__class__).primary_key[0].name
+        resource_id = str(getattr(resource, resource_pk_name))
+
+        user_role = self.UserRole(
+            user_id=user_id,
+            resource_type=resource_type,
+            resource_id=resource_id,
+            role=role_name,
+        )
+        session.add(user_role)
```
dhatch, line 477: How does conflict resolution work? (User already in role, user has another role on this resource)

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles2.py
dhatch: One note is we may want to add some validation of the data in the policy itself with helpful error messages to make up for the fact that we don't have a way to check that the configuration data is the right structure.

languages/python/sqlalchemy-oso/tests/test_roles2.py
```diff
@@ -0,0 +1,146 @@
...
+
+    # @NOTE: Need the users and resources in the db before assigning roles
```
dhatch, line 129: So I can't atomically create a user with some roles assigned to them? Or a resource?

I think that should be possible with SQLAlchemy. Is anything we did preventing that?

dhatch (commented):
I reviewed this mostly to get a sense of where it is at. Most of my comments are probably already stuff we intend to clean up, but it might be helpful to go through and figure out what we need to do before the first release & make sure we plan for the stuff we'll do later.

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles2.py
dhatch: Or we need the flag of whether it's configured to be persisted in the database so we can tell on startup.

dhatch:
Initial thoughts on what we should do when...

*before release*

- [ ] (M) More tests.
- [ ] (S) Provide [indexes for tables](https://github.com/osohq/oso/pull/860/files#r618414558)
- [ ] (S) Configuration nits ([permission](https://github.com/osohq/oso/pull/860/files#r618387891) and field names basically)
- [ ] (S) SQL data model nits ([unique constraint](https://github.com/osohq/oso/pull/860/files#r618414294), [roles compound key](https://github.com/osohq/oso/pull/860/files#r618415225), [etc.](https://github.com/osohq/oso/pull/860/files#r618416013))
- [ ] (S) [Indexes for role data model](https://github.com/osohq/oso/pull/860/files#r618414558)
- [ ] (M) Configure static role data [outside of application startup](https://github.com/osohq/oso/pull/860/files#r618418432). This should be a deployment step for most people. [Remove lazy config.](https://github.com/osohq/oso/pull/860/files#r618420768)
- [ ] (M) Figure out [session management](https://github.com/osohq/oso/pull/860/files#r618390555).

- [ ] (M) Some [validation of policy configuration data with error messages](https://github.com/osohq/oso/pull/860/files#r618422254)

*after release*

- [ ] Work on compound keys (for user and resource) https://github.com/osohq/oso/pull/860/files#r618393189
- [ ] (L) [Migrations of roles static data](https://github.com/osohq/oso/pull/860/files#r618419353)

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles2.py
saolsen: role just uses it's name as the id since it's globally unique, permission has no such builtin key

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles2.py
saolsen: name is unique which is why it's the primary key

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles2.py
saolsen: role names are globally unique

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles2.py
saolsen: yeah there's a bunch of validation stuff that needs to go in here

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles2.py
```diff
@@ -519,7 +523,7 @@ def _add_query_filter(oso, user, action, resource_model):
     # @OPT: It should be possible to pass the select sql as the in
     # parameter instead of doing two queries
     # but I'm not sure how you bind the variables.
-    results = oso.roles.session.execute(
+    results = session.execute(
```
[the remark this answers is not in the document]
saolsen: yeah it can't be an authorized session, has to be a normal session

languages/python/sqlalchemy-oso/tests/test_roles2.py
```diff
@@ -0,0 +1,146 @@
...
+
+    policy = """
+    resource(_type: Organization, "org", actions, roles) if
+        actions = [
+            "invite",
+            "create_repo"
+        ] and
+        roles = {
+            org_member: {
+                perms: ["create_repo"],
```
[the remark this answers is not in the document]
saolsen: I think of it as a permission is an action on a certain resource type. Def could be confusing though.

languages/python/sqlalchemy-oso/tests/test_roles2.py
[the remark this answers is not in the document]
saolsen: Yeah I was under the impression that they all had to use the same Base to be able to ensure we can join them up.

languages/python/sqlalchemy-oso/tests/test_roles2.py
saolsen: user_role references the user_id directly so the db verifies the user exists, I think if you use the same session maybe you can create them together?

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles2.py
saolsen: It could be, this model isn't exposed to the user though so it doesn't really matter.

languages/python/sqlalchemy-oso/sqlalchemy_oso/auth.py
```diff
@@ -79,13 +80,16 @@ def get_field_type(model, field):
...
+    has_role_allows = False
     for result in results:
         has_result = True
 
         resource_partial = result["bindings"]["resource"]
-        filter = partial_to_filter(
+        filter, role_allows = partial_to_filter(
             resource_partial, session, model, get_model=oso.get_class
         )
+        has_role_allows |= role_allows
```
dhatch, line 91: I don't think this logic is correct. If I write a policy like:

```
allow(actor, action, resource) if
    OsoRoles.role_allow(actor, action, resource) or
    resource.public = true;
```

The `role_allow` filter will be applied to both branches.

languages/python/sqlalchemy-oso/sqlalchemy_oso/partial.py
```diff
@@ -44,7 +45,32 @@ def flip_op(operator):
...
+
+
+def contains_role_allows(expression: Expression):
+    assert expression.operator == "And"
+    for expr in expression.args:
+        op = expr.operator
+        left = expr.args[0]
+        right = expr.args[1]
+        if (
+            op == "Unify"
```
dhatch, line 63: We have to check for this bidirectionally (we can't rely on `True` being on the left and `Dot` on the right).

languages/python/sqlalchemy-oso/sqlalchemy_oso/auth.py
dhatch: I think if we use a subquery we can apply the role filter only on the branch that is needed.  Generate something like:

```
SELECT * FROM (... roles query) WHERE ... normal filters ...;
```

or maybe

```
with roles_query AS roles_resources
SELECT * FROM organization INNER JOIN roles_resources USING (id) WHERE ... filters
```

dhatch: nvm won't quite work... anyway with more thought I think there's a way to write this that is correct

dhatch (commented):
Some comments on list filtering stuff

languages/python/sqlalchemy-oso/tests/test_roles2.py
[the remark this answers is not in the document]
dhatch: @leina05 did you look into this at all when you added this? https://github.com/osohq/oso/pull/860/commits/0e92f1e9f831283eaf59fd933dae7e4431816561

languages/python/sqlalchemy-oso/tests/test_roles2.py
```diff
@@ -0,0 +1,1465 @@
...
+
+    osohq = sample_data["osohq"]
+    oso_repo = sample_data["oso_repo"]
+    leina = sample_data["leina"]
+    steve = sample_data["steve"]
+
+    oso_roles.assign_role(leina, osohq, "org_member", session=session)
+
+    assert oso.is_allowed(leina, "invite", osohq)
+    assert oso.is_allowed(leina, "pull", oso_repo)
```
dhatch, line 584: Check that Leina cannot pull on a repository that is not a child of the org.

languages/python/sqlalchemy-oso/tests/test_roles2.py
```diff
@@ -0,0 +1,1465 @@
...
+
+    osohq = sample_data["osohq"]
+    oso_bug = sample_data["oso_bug"]
+    leina = sample_data["leina"]
+    steve = sample_data["steve"]
+
+    oso_roles.assign_role(leina, osohq, "org_member", session=session)
+
+    assert oso.is_allowed(leina, "invite", osohq)
+    assert oso.is_allowed(leina, "edit", oso_bug)
```
dhatch, line 658: Check for issue that is not a child.

languages/python/sqlalchemy-oso/tests/test_roles2.py
```diff
@@ -0,0 +1,1465 @@
...
+    osohq = sample_data["osohq"]
+    leina = sample_data["leina"]
+    steve = sample_data["steve"]
+
+    assert not oso.is_allowed(leina, "invite", osohq)
+
+    oso_roles.assign_role(leina, osohq, "org_member", session=session)
+    oso_roles.assign_role(steve, osohq, "org_owner", session=session)
+
+    assert oso.is_allowed(leina, "invite", osohq)
```
dhatch, line 727: negative case on other org

languages/python/sqlalchemy-oso/tests/test_roles2.py
```diff
@@ -0,0 +1,1465 @@
...
+    osohq = sample_data["osohq"]
+    oso_repo = sample_data["oso_repo"]
+    leina = sample_data["leina"]
+    steve = sample_data["steve"]
+
+    # org_member implies repo_read which has the "pull" permission
+    oso_roles.assign_role(leina, osohq, "org_member", session=session)
+
+    assert oso.is_allowed(leina, "invite", osohq)
+    assert oso.is_allowed(leina, "pull", oso_repo)
```
dhatch, line 805: non child

languages/python/sqlalchemy-oso/tests/test_roles2.py
```diff
@@ -0,0 +1,1465 @@
...
+
+    osohq = sample_data["osohq"]
+    oso_bug = sample_data["oso_bug"]
+    leina = sample_data["leina"]
+    steve = sample_data["steve"]
+
+    oso_roles.assign_role(leina, osohq, "org_member", session=session)
+
+    assert oso.is_allowed(leina, "invite", osohq)
+    assert oso.is_allowed(leina, "edit", oso_bug)
```
dhatch, line 885: non child

languages/python/sqlalchemy-oso/tests/test_roles2.py
```diff
@@ -0,0 +1,1465 @@
...
+    oso_roles.assign_role(steve, oso_repo, "repo_read", session=session)
+
+    # leina can invite to the org, pull from the repo, and edit the issue
+    assert oso.is_allowed(leina, "invite", osohq)
+    assert oso.is_allowed(steve, "pull", oso_repo)
+    assert oso.is_allowed(leina, "edit", oso_bug)
+
+    # steve can pull from the repo and edit the issue, but can NOT invite to the org
+    assert oso.is_allowed(steve, "pull", oso_repo)
+    assert oso.is_allowed(steve, "edit", oso_bug)
```
dhatch, line 992: non child

languages/python/sqlalchemy-oso/tests/test_roles2.py
```diff
@@ -0,0 +1,1465 @@
...
+
+    steve_roles = (
+        session.query(oso_roles.UserRole)
+        .filter(oso_roles.UserRole.user_id == steve.id)
+        .all()
+    )
+    assert len(steve_roles) == 1
+    assert steve_roles[0].role == "org_owner"
+
+    assert oso.is_allowed(leina, "invite", osohq)
```
dhatch, line 1187: I'm a little confused what this works without a commit.

polar-core/src/vm.rs
```diff
@@ -1926,6 +1931,74 @@ impl PolarVirtualMachine {
...
+        &mut self,
+        object: &Term,
+        field: &Term,
+        value: &Term,
+    ) -> PolarResult<bool> {
+        // If the lookup is a `Value::Call`, then we need to check for partial args
+        let mut has_partial_arg = false;
+        if let Value::Call(Call { name, args, kwargs }) = field.value() {
+            // get all arg values (args + kwargs)
+            let mut arg_values = args.clone();
```
dhatch, line 1947: Do we need all these clones?

polar-core/src/vm.rs
```diff
@@ -1926,6 +1931,74 @@ impl PolarVirtualMachine {
...
+                let mut v: Vec<Term> = kwarg_map.values().cloned().collect();
+                arg_values.append(&mut v)
+            }
+            for (i, arg) in arg_values.iter().enumerate() {
+                // Partial/unbound variables cannot be passed to external methods from Polar
+                if let Value::Variable(v) = arg.value() {
+                    match self.binding_manager.variable_state(v) {
+                        // bound variables are fine, continue
+                        VariableState::Bound(_) => continue,
+                        // partial variables are only fine if being passed into "role_allows"
```
dhatch, line 1958: Add a `TODO` here. This is a temporary hack.

polar-core/src/vm.rs
```diff
@@ -1926,6 +1931,74 @@ impl PolarVirtualMachine {
...
+                        VariableState::Partial => {
+                            if let Value::ExternalInstance(external) =
+                                self.deep_deref(&object).value()
+                            {
+                                if let Some(repr) = external.repr.clone() {
+                                    // only allow the third argument (resource) to be partially bound
+                                    if repr.contains("sqlalchemy_oso.roles2.OsoRoles")
+                                        && name.0 == "role_allows"
+                                        && i == 2
```
dhatch, line 1967: This `i == 2` check is not correct in the presence of keyword arguments.

polar-core/src/vm.rs
```diff
@@ -1926,6 +1931,74 @@ impl PolarVirtualMachine {
...
+                                            object.clone(),
+                                            field.clone_with_value(Value::Call(Call {
+                                                name: name.clone(),
+                                                args: vec![arg.clone()],
+                                                kwargs: None
+                                            }))
+                                        );
+                                        let term =
+                                            &op!(Unify, value.clone(), dot.into_term()).into_term();
+                                        self.add_constraint(term)?;
```
dhatch, line 1981: It's unexpected from the caller that this code adds a constraint. I'd either rename the function & update the doc string to make it more clear that a constraint will be added, or return an `Option<Term>` which contains the constraint to add, adding it in the calling code.

dhatch (commented):
Reviewed tests and rust code.

Still need to review `roles2.py` again.

saolsen:
@TODO: go through this again for todo items for followups but closing for now.

polar-core/src/vm.rs
leina05: I think so? It took me a while to make this happy, but I'd be glad to pair on cutting down the clones

polar-core/src/vm.rs
leina05: do we want it to accept kwargs?

leina05: Actually don't think this is true... why shouldn't it be correct if kwargs?

dhatch: The keyword arguments are unordered.  You mean to check that the parameter either in position 2 or with the name `resource` is partial. However, if someone uses all keyword arguments, there is not a guarantee that the `resource `argument is in position 2.

This function would be great to test.

polar-core/src/vm.rs
dhatch: @leina05 this one is the biggest deal of my comments tbh

polar-core/src/vm.rs
leina05: Ohhhh right sorry was misreading

languages/python/sqlalchemy-oso/tests/test_roles2.py
leina05: hmm, yeah

languages/python/sqlalchemy-oso/tests/test_roles2.py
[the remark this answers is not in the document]
dhatch: Opened as task

languages/python/sqlalchemy-oso/tests/test_roles2.py
saolsen: Hi it's steve from the future. Turns out none of these worked without a commit except on sqlite in memory. Makes sense.

dhatch: thanks future steve

PR #861, opened 2021-04-21
[sqlalchemy-oso] Ensure we use true filter where possible.

samscott89:
PR checklist:

- [x] Added changelog entry.


languages/python/sqlalchemy-oso/tests/test_sqlalchemy.py
```diff
@@ -290,3 +290,16 @@ def test_regular_session(engine, oso, fixture_data):
...
+def test_unconditional_policy_has_no_filter(engine, oso, fixture_data):
+    oso.load_str('allow("user", "read", post: Post) if post.id = 1; allow(_, _, _);')
+    session = AuthorizedSession(oso, user="user", action="read", bind=engine)
+
+    query = session.query(Post)
+
+    assert str(query) == (
+        "SELECT posts.id AS posts_id, posts.contents AS posts_contents, posts.title AS posts_title, "
+        + "posts.access_level AS posts_access_level, posts.created_by_id AS posts_created_by_id, "
+        + "posts.needs_moderation AS posts_needs_moderation \nFROM posts \nWHERE 1 = 1"
```
samscott89, line 304: This previously had a filter `post.id = ?` _instead_ of a true filter.

languages/python/sqlalchemy-oso/tests/test_post_relationship.py
```diff
@@ -644,9 +644,9 @@ def test_in_with_constraints_but_no_matching_objects(
         + " posts.access_level AS posts_access_level,"
         + " posts.created_by_id AS posts_created_by_id, posts.needs_moderation AS posts_needs_moderation"
         + " \nFROM posts"
-        + " \nWHERE (EXISTS (SELECT 1"
+        + " \nWHERE EXISTS (SELECT 1"
```
samscott89, line 647: No idea why. SQLalchemy version thingy?

languages/python/sqlalchemy-oso/sqlalchemy_oso/partial.py
```diff
@@ -41,6 +42,13 @@ def flip_op(operator):
     return flips[operator]
 
 
+def add_filter(current, new):
```
leina05, line 45:
```suggestion
def and_filter(current, new):
```

languages/python/sqlalchemy-oso/tests/test_post_relationship.py
leina05: Was this failing without your change? Or after the change?

languages/python/sqlalchemy-oso/tests/test_sqlalchemy.py
leina05: Looks good, we should probably add this test to the django adapter?

leina05 (approved):
Just documenting for myself/others: the bug here was that we were converting unconditional expressions to empty `AND` clauses, which when `OR`'d with other filters do nothing, when the desired behavior is to turn the filter into a `TRUE`/`WHERE 1 = 1` so that no conditions are applied.

languages/python/sqlalchemy-oso/tests/test_sqlalchemy.py
samscott89: Good idea.

languages/python/sqlalchemy-oso/tests/test_post_relationship.py
samscott89: I can check. I suspect there's an additional `and` that was stripped away from the query filter. e.g. `and(and(...))` vs `and(...)`

samscott89:
> 
> 
> Just documenting for myself/others: the bug here was that we were converting unconditional expressions to empty `AND` clauses, which when `OR`'d with other filters do nothing, when the desired behavior is to turn the filter into a `TRUE`/`WHERE 1 = 1` so that no conditions are applied.

Good summary! We had the same problem in the django adapter, which is why we have the `TRUE_FILTER` throughout.

languages/python/sqlalchemy-oso/tests/test_sqlalchemy.py
samscott89: Added in https://github.com/osohq/oso/pull/861/commits/03715a5e520289fdd0fc344286433a44f981f29a

languages/python/sqlalchemy-oso/sqlalchemy_oso/partial.py
samscott89: Added in https://github.com/osohq/oso/pull/861/commits/03715a5e520289fdd0fc344286433a44f981f29a

PR #865, opened 2021-04-22
Adds In Search UI

veverkap:
image

### DIFF
--- a/docs/themes/oso-tailwind/assets/css/site.css
+++ b/docs/themes/oso-tailwind/assets/css/site.css
@@ -12,9 +12,7 @@
 
 .search-button {
     display: flex;
-    -webkit-box-align: center;
     align-items: center;
-    -webkit-box-pack: justify;
     justify-content: space-between;
     width: 100%;
     border-radius: 0.25rem;
@@ -33,20 +31,12 @@
     }
 }
 
-.search-button:focus {
-    outline: transparent solid 2px;
-    outline-offset: 2px;
-}
-
-@media (min-width: 768px) {
-    .search-button:hover {
-        border-color: rgba(59, 130, 246, 0.5);
-    }
+.search-button:hover {
+    border-color: rgba(59, 130, 246, 0.5);
 }
 
 .search-inner-button {
     display: flex;
-    -webkit-box-align: center;
     align-items: center;
 }
 
@@ -84,11 +74,6 @@
     --tw-bg-opacity: 0.5;
     background-color: rgba(0, 0, 0, var(--tw-bg-opacity));
     z-index: 5050;
-}
-
-.hoJfGX:focus {
-    outline: transparent solid 2px;
-    outline-offset: 2px;
 }
 
 .searching {
@@ -141,9 +126,7 @@
 
 .result-link {
     display: flex;
-    -webkit-box-align: center;
     align-items: center;
-    -webkit-box-pack: justify;
     justify-content: space-between;
     padding-left: 0.75rem;
     padding-right: 0.75rem;
--- a/docs/themes/oso-tailwind/assets/css/site.css
+++ b/docs/themes/oso-tailwind/assets/css/site.css
@@ -1,5 +1,3 @@
-
-
 /* test {
     c100: #e1dbf6;
     c200: #afa0f1;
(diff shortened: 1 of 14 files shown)


docs/themes/oso-tailwind/assets/js/nav.js
```diff
@@ -46,6 +49,14 @@ window.onclick = function (event) {
       break;
     case langButton:
       break;
+    case searchInput:
+      break
+    case searchModal:
+      window.hideSearch(event);
```
veverkap, line 55: This makes it when you click off the modal, the search box goes away

dhatch:
Checking out!

docs/themes/oso-webpack/index.js
```diff
@@ -163,11 +206,136 @@ import('monaco-editor-core').then(monaco => {
...
+  const searchInput = document.getElementById('search-input');
+  if (searchInput) {
+    searchInput.addEventListener('input', (e) =>
+      window.searchInputKeyUp(e)
+    );
+  }
+
+  const searchResultsContainer = document.getElementById('search-results-container');
+  const algoliasearch = require("algoliasearch");
+  const client = algoliasearch("KROZ8F05YT", "13594a3b7da482e011ce0ab08fdb4c4d");
```
dhatch, line 260: Are these keys safe to include in code? I realize we may need to migrate the Algolia account.

dhatch (commented):
Hey @veverkap . I just tried this out locally. Is everything pushed up? I ran `make clean` and rebuilt the webpack modules a few times and tried a couple browsers but there seem to be some styling issues & a few other minor bugs. Let me know if you can't reproduce & I can dig a bit further. I used:

- Safari 14
- Firefox 85.0
- Chrome 90.0

OS: OS X 10.15

*Styling*

Screen Shot 2021-04-22 at 3 06 28 PM

Screen Shot 2021-04-22 at 3 06 17 PM


*Keyboard shortcuts*

Seem to be some issues with keyboard shortcuts working:

Screen Shot 2021-04-22 at 3 08 36 PM
Screen Shot 2021-04-22 at 3 07 14 PM

*Formatting of links*

links have `<em>` tags in them if a term in the link is matched (see the lower left hand corner).

Screen Shot 2021-04-22 at 3 06 44 PM

dhatch:
Can look a bit more later today and see if it's something with my setup

veverkap:
I’ll take a peek on another computer as well. Might be me

docs/themes/oso-webpack/index.js
veverkap: Yes these are the read only ones necessary for JS

veverkap:
> Can look a bit more later today and see if it's something with my setup

There was an issue with purgecss yanking out all of the styles because they were added in by a Handlebars template.

I fixed that by adding the styles to the safe list.

Can you give it another shot @dhatch

dhatch:
Took another look, it still seems like some CSS may be missing, tested in a few browsers and I'm getting similar results.

Screen Shot 2021-04-27 at 11 21 52 AM

I should have some time to debug more later today.

veverkap:
That's no good.  I've tried it on two other laptops and it works okay for me.  Check to see if the #search-result class is available in your CSS file

I was running it with `hugo server` in one window and `npm run serve` from the themes/oso-tailwind folder in another.  I think you have to get the CSS built?

dhatch:
I verified the issue is the CSS post processing by running it with the `HUGO_PURGECSS=off make run` command. The safelist in the config seems to be ignored for me. But I can add comments to control it as described here https://purgecss.com/safelisting.html#specific-selectors.

Probably something going on on my end, going to look into further and I'll let you know.

docs/themes/oso-tailwind/assets/css/site.css
```diff
@@ -10,4 +8,171 @@
     c700: #211164;
     c700: #170950;
     c900: #0e024e;
-} */
\ No newline at end of file
+} */
+
+.search-button {
+    display: flex;
+    -webkit-box-align: center;
```
dhatch, line 15: deprecated css

docs/themes/oso-tailwind/assets/css/site.css
```diff
@@ -10,4 +8,171 @@
...
+    border-radius: 0.25rem;
+    border-width: 1px;
+    border-color: #D6D8DC;
+    cursor: pointer;
+    padding: 0.5rem;
+    color: #B1B6BE;
+    text-align: left;
+}
+
+@media (min-width: 768px) {
```
dhatch, line 29: use tailwind for media queries if possible in case we reconfigure our size breaks

that way they will still match

docs/themes/oso-tailwind/assets/css/site.css
dhatch: use tailwind?

docs/themes/oso-tailwind/assets/css/site.css
```diff
@@ -10,4 +8,171 @@
...
+    right: 0px;
+    top: 0px;
+    bottom: 0px;
+    user-select: none;
+    --tw-bg-opacity: 0.5;
+    background-color: rgba(0, 0, 0, var(--tw-bg-opacity));
+    z-index: 5050;
+}
+
+.hoJfGX:focus {
```
dhatch, line 89: I don't see this class used anywhere.

docs/themes/oso-tailwind/layouts/_default/list.html
```diff
@@ -1,46 +1,47 @@
...
+<div class="w-3/5 text-black leading-normal pt-6 mx-auto static">
+  <h2 class="text-xl font-semibold text-primary-dark pb-4">
+    Choose your language:
+  </h2>
+  <div class="grid mx-6 grid-cols-3 gap-8 content-evenly">
+    {{ partial "language-chooser.html" . }}
+  </div>
+</div>
+{{ end }}
+{{- partial "search.html" . -}}
```
dhatch, line 46: This is the only actual change in this file, right? Can you update your formatting settings to match the existing format? I'm not sure if we are using a consistent formatter for docs... so perhaps we should set that up to make diffs easier to review in the future.

docs/themes/oso-tailwind/layouts/partials/footer.html
```diff
@@ -25,26 +24,23 @@
...
+                <button type="submit" data-wait="Churning..."
                     class="ml-2 px-3 py-2 rounded-md sm:text-sm text-gray-600 bg-gray-300 text-gray-700 hover:bg-gray-600 hover:text-gray-900 appearance-none focus:outline-none">
                     Subscribe
                 </button>
             </form>
         </div>
     </div>
 </footer>
-
+<script async="true" type="text/javascript" src="https://unpkg.com/tinykeys@1.1.2/dist/tinykeys.umd.js"></script>
```
dhatch, line 38: Can we include this in our built js? Is there a reason to do it this way?

docs/themes/oso-tailwind/layouts/partials/head.html
```diff
@@ -10,38 +10,51 @@
...
     <meta name="keywords" content="{{ delimit .Keywords " ," }}" />
     {{ end -}}
     {{- if .Params.Author }}
     <meta name="author" content="{{ .Params.Author }}" />
     {{ end -}}
+    <link rel="dns-prefetch" href="https://cdnjs.cloudflare.com/">
+    <link rel="dns-prefetch" href="https://cdn.jsdelivr.net/">
+    <meta id="facet-language" name="facet-language" content="{{ .Language.Lang }}" />
+    <script src="https://cdn.jsdelivr.net/npm/algoliasearch@4.5.1/dist/algoliasearch-lite.umd.js"></script>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/handlebars.js/4.7.7/handlebars.js"
```
dhatch, line 26: Include this in asset pipeline?

docs/themes/oso-tailwind/layouts/partials/footer.html
dhatch: It looks like it was included in webpack, so I don't think this is needed.

docs/themes/oso-tailwind/layouts/partials/head.html
dhatch: Also looks like this is in webpack, so I think we can drop this.

docs/themes/oso-tailwind/assets/css/site.css
```diff
@@ -10,4 +8,171 @@
...
+    }
+}
+
+.search-button:focus {
+    outline: transparent solid 2px;
+    outline-offset: 2px;
+}
+
+@media (min-width: 768px) {
+    .search-button:hover {
```
dhatch, line 42: Why is hover color disabled on smaller screens?

docs/themes/oso-webpack/index.js
```diff
@@ -163,11 +206,142 @@ import('monaco-editor-core').then(monaco => {
...
+      "snippetEllipsisText": "..."
+    }).then(({
+      hits
+    }) => {
+      window.processHits(hits);
+    });
+  }
+
+  // this search for a term WITH a facet
+  window.searchTermWithFacet = function(term, language) {
```
dhatch, line 323: Why do these functions need to be attached to the window? Are they used globally?

dhatch (commented):
Left some inline comments. I'm going to try to resolve most of these now if possible.

The CSS issue was caused by an incorrect purgecss configuration. We're using version 2, so the `safelist` configuration argument is actually `whitelist`. I found this from an old version of their documentation. I'm not sure why it was working locally for you, but once I swapped this the CSS issues were resolved. There was also a use of the CSS selector `inset` which does not seem to be supported in Safari. I pushed fixes for both of these to this branch.

I'm having issues with keyboard shortcuts in Safari, but they work fine in chrome & Firefox.

dhatch (approved):
looks good to me

dhatch:
I'm going to check out the algolia account and see if there's anything I need to do to transfer it before deploying this.

A next step might be adding search index generation to our CI process.

PR #870, opened 2021-04-27
core changes for roles2

saolsen:


dhatch (approved):
good for now @leina05 when you get a chance there are some comments on the original PR that would be nice to address

PR #881, opened 2021-05-04
Add CTA to look at roles library.

dhatch:
Rephrase some CTA to be focused on roles instead of SQLAlchemy.

PR checklist:

- [ ] Added changelog entry.

### DIFF
--- a/docs/content/any/getting-started/_index.md
+++ b/docs/content/any/getting-started/_index.md
@@ -9,7 +9,7 @@ any: false
 
 Learn the basics of using Oso and get Oso set up in your application.
 
-{{%ifLang "python" %}}
+{{% ifLang "python" %}}
 {{< callout "Early access to the next version of roles" "blue" >}}
 
 We just released early access to the next version of built-in support
--- a/docs/content/any/getting-started/_index.md
+++ b/docs/content/any/getting-started/_index.md
@@ -8,3 +8,16 @@ any: false
 # Jump in
 
 Learn the basics of using Oso and get Oso set up in your application.
+
+{{%ifLang "python" %}}
+{{< callout "Early access to the next version of roles" "blue" >}}
+
+We just released early access to the next version of built-in support
+for Roles. Initially supporting SQLAlchemy, this library, includes
+improved support for common authorization models, like role based access
+control.
+
+[Check it out here!](/new-roles)
+{{< /callout >}}
+
+{{% /ifLang %}}
--- a/docs/content/python/new-roles/_index.md
+++ b/docs/content/python/new-roles/_index.md
@@ -4,12 +4,14 @@ weight: 1
 no_nav: true
 ---
 
-SQLAlchemy Roles Early Access
-=============================
+Roles for SQLAlchemy Early Access
+=================================
 
-We're working on the next major version of the SQLAlchemy roles library.
-This library makes it easy to model role based access control in Polar
-and enforce it over your SQLAlchemy models.
+We're working on the next major version of support for roles in Oso. Our
+first release is in the SQLAlchemy library. This library makes it easy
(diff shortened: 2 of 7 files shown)


docs/content/any/getting-started/_index.md
```diff
@@ -8,3 +8,16 @@ any: false
 # Jump in
 
 Learn the basics of using Oso and get Oso set up in your application.
+
+{{%ifLang "python" %}}
```
dhatch, line 12:
```suggestion
{{% ifLang "python" %}}
```

docs/content/python/new-roles/_index.md
```diff
@@ -4,12 +4,14 @@ weight: 1
...
-We're working on the next major version of the SQLAlchemy roles library.
-This library makes it easy to model role based access control in Polar
-and enforce it over your SQLAlchemy models.
+Roles for SQLAlchemy Early Access
+=================================
+
+We're working on the next major version of support for roles in Oso. Our
+first release is in the SQLAlchemy library. This library makes it easy
+to model role based access control in Polar and enforce it over your
+SQLAlchemy models. If you're interested in using Roles with other
```
gj, line 13: There are a few places where you use lowercase `roles` and a few where you capitalize it as `Roles`. We should pick one.

docs/content/any/getting-started/_index.md
```diff
@@ -8,3 +8,16 @@ any: false
 # Jump in
 
 Learn the basics of using Oso and get Oso set up in your application.
+
+{{% ifLang "python" %}}
+{{< callout "Early access to the next version of roles" "blue" >}}
+
+We just released early access to the next version of built-in support
+for Roles. Initially supporting SQLAlchemy, this library, includes
```
gj, line 16:
```suggestion
for Roles. Initially supporting SQLAlchemy, this library includes
```

Also `Initially supporting SQLAlchemy, this library...` feels like an unnecessarily oblique way to indicate that support for the new roles feature is only in the `sqlalchemy-oso` library.

dhatch: Yeah, I actually agree. Originally these snippets said 

> "Using SQLAlchemy?"

> We just released early access to the next version of our SQLAlchemy library, includes improved support for common authorization models, like role based access control.

Which IMO is clearer. Maybe I should just update the title but keep the contents the same?

The goal behind these changes (from @gneray) was to make roles the headline instead of SQLAlchemy for the early access program so that people interested in other languages maybe look at it and ask for support.

docs/content/python/new-roles/_index.md
dhatch: You are right. Will update to lower case once I finish the wording.

docs/content/any/getting-started/_index.md
```diff
@@ -8,3 +8,16 @@ any: false
...
+improved support for common authorization models, like role based access
+control.
+
+[Check it out here!](/new-roles)
```
gj, line 20: @dhatch @gneray how about something like this

```suggestion
{{< callout "Built-in roles early access" "blue" >}}

We're hard at work on a major revision to Oso's built-in roles support.
We released [an early access version of our SQLAlchemy library](/new-roles)
containing the new features, which make it easy to model role-based access
control (RBAC) in Polar and enforce the policy over SQLAlchemy models.

[Check it out here!](/new-roles)
```

dhatch: We just discussed live a plan to:

1. Leave the SQLAlchemy callouts when the user has Python selected similar to as is (or maybe your suggestion)
2. Add callouts on roles pages for when users are not using Python that link to a higher level page describing features of the new roles work, with a link to try out in other languages that will email us.

dhatch:
replace with #889

PR #883, opened 2021-05-05
Check the variable first

veverkap:
This checks the variable is set before trying to modify it.

Also remove console.log


dhatch:
This fix doesn't quite make sense to me. If the variable is unset, won't hiding and showing search not work still?

Guessing a bit here, but I think the issue could be that the DOM isn't completely rendered when the JS file is being loaded, so the get element by id calls at the top level are returning undefined: https://github.com/osohq/oso/pull/883/files#diff-1b1024cc39deefd84f4cb43741f0294ab0e6af8f3807462519f4be35700b9c5dR37.

dhatch:
Fixed in #907

PR #885, opened 2021-05-06
Ctrl-K but with no formatting :)

veverkap:


dhatch:
TY!

dhatch (approved):
Thanks! Can do the formatter thing if you'd like (not going to have a chance today though)

veverkap:
> Thanks! Can do the formatter thing if you'd like (not going to have a chance today though)

We can look into that for sure - I just opened it in vim instead of VSCode :)

PR #887, opened 2021-05-06
`user_in_role` (now with list filtering)

saolsen:


languages/python/sqlalchemy-oso/sqlalchemy_oso/roles2.py
```diff
@@ -641,6 +723,44 @@ def _role_allows(self, user, action, resource):
...
+        if isinstance(resource, Variable):
+            # resource is a variable, so we are running as a partial. It may be allowed.
+            # Defer to later.
+            return True
+
+        session = self._get_session()
+
+        try:
+            user_pk_name = inspect(user.__class__).primary_key[0].name
+            user_id = getattr(user, user_pk_name)
```
leina05, line 736: Do we need to convert this to a `str` like we do with `resource_id`?

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles2.py
```diff
@@ -641,6 +723,44 @@ def _role_allows(self, user, action, resource):
         else:
             return False
 
+    def _user_in_role(self, user, role, resource):
```
leina05, line 726: Should we check that `role` is a `str`?

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles2.py
```diff
@@ -808,7 +928,7 @@ def _add_query_filter(oso, user, action, resource_model):
         return sql.false()
 
     try:
-        role_list_sql = oso.roles.list_filter_queries[resource_type]
+        role_list_sql = oso.roles.role_allow_list_filter_queries[resource_type]
```
leina05, line 931: Should we rename `_add_query_filter` to `_add_query_role_allow_filter`?

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles2.py
```diff
@@ -83,6 +83,67 @@ def user_in_role_query(
     return query
 
 
+def user_in_role_query(
```
leina05, line 86: So this gets both directly assigned and implied roles for the user?

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles2.py
saolsen: yeah

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles2.py
saolsen: yeah probably, gonna just note this for now because I have another branch where I'm restructuring a lot of this sql stuff, will do that there.

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles2.py
saolsen: Yep, it does basically the same thing as is_allowed but starts from a specific role instead of from any role that contains a specific permission.

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles2.py
```diff
@@ -584,15 +652,29 @@ def synchronize_data(self, session=None):
         resource_id_field = "cast(:resource_id as text)"
 
         has_relationships = len(self.relationships) > 0
-        self.sql_query = user_in_role_query(
+        self.role_allow_sql_query = role_allow_query(
```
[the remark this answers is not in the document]
saolsen: Yeah, a lot of this is changing soon so will document it better in the refactored version.

languages/python/sqlalchemy-oso/tests/test_roles2.py
```diff
@@ -1817,6 +1895,64 @@ def test_read_api(init_oso, sample_data, Repository, Organization):
...
+
+    oso_roles.assign_role(leina, osohq, "org_member")
+    oso_roles.assign_role(steve, oso_repo, "repo_read")
+
+    # Without data filtering
+    assert oso.is_allowed(leina, "read", oso_repo)
+    assert oso.is_allowed(steve, "read", oso_repo)
+    assert not oso.is_allowed(gabe, "read", oso_repo)
+
+    # With data filtering
```
leina05, line 1945: Need to add at least the same tests for `user_in_role` data filtering that we have for `role_allows`

leina05 (changes requested):
Should be good once we have more tests.

languages/python/sqlalchemy-oso/sqlalchemy_oso/partial.py
```diff
@@ -53,46 +53,48 @@ def and_filter(current, new):
...
-def contains_role_allows(expression: Expression):
-    def _is_role_allow(op, left, right):
-        op = expr.operator
-        left = expr.args[0]
-        right = expr.args[1]
-        is_role_allow = (
+def check_for_roles_method(expression: Expression):
+    def _is_roles_method(op, left, right):
+        is_roles_method = (
             left is True
```
gj, line 67: So the expression comes in from the core as `(true).user_in_role()`?

languages/python/sqlalchemy-oso/sqlalchemy_oso/partial.py
```diff
@@ -53,46 +53,48 @@ def and_filter(current, new):
...
+                right.args[1].name == "role_allows"
+                or right.args[1].name == "user_in_role"
+            )
         )
 
-        if is_role_allow:
+        method = None
+        if is_roles_method:
             if op != "Unify":
                 raise OsoError("Roles don't currently work with not.")
```
gj, line 79: Are the only two options for `op` `"Unify"` and `"Not"`?

languages/python/sqlalchemy-oso/sqlalchemy_oso/partial.py
```diff
@@ -53,46 +53,48 @@ def and_filter(current, new):
...
+        is_roles, method = _is_roles_method(expr.operator, expr.args[0], expr.args[1])
+        if is_roles:
+            expression.args.remove(expr)
+            return method
+        # Try with method call on left
+        is_roles, method = _is_roles_method(expr.operator, expr.args[1], expr.args[0])
+        if is_roles:
             expression.args.remove(expr)
-            return True
+            return method
```
gj, line 95: It seems like the current code won't support expressions containing multiple `role_allows()` and/or `user_in_role()` calls. Is that an oversight, a planned follow-up, or an intentional omission b/c we don't think it's a use case worth supporting?

languages/python/sqlalchemy-oso/tests/test_roles2.py
```diff
@@ -1567,6 +1569,52 @@ def test_reassign_user_role(init_oso, sample_data):
 # - [x] `role_allows` inside of a `not` (this probably won't work, so need error handling)
 
 
+@pytest.mark.xfail(
+    reason="Currently failing because we have a bug with filtering roles data--not returning relationships"
+)
```
gj, line 1574: TODO: verify whether #892 addresses this

languages/python/sqlalchemy-oso/tests/test_roles2.py
gj: Not sure exactly which tests this is referring to, but I'd love to see at least one test where we compose `Roles.user_in_role()` with additional constraints.

languages/python/sqlalchemy-oso/sqlalchemy_oso/partial.py
saolsen: I think this is called on each expression so would work fine for multiple calls.

gj: To my eyes it looks like it'll be called once per result but won't be called multiple times if a result contains multiple `Roles.xxx()` calls. Could we maybe add a test to verify?

languages/python/sqlalchemy-oso/sqlalchemy_oso/partial.py
leina05: It's actually `True = OsoRoles.user_in_role()`

leina05: The expression here is a `Unify` op

languages/python/sqlalchemy-oso/tests/test_roles2.py
gj: new tests look good

leina05 (approved):

gj (approved):

PR #891, opened 2021-05-10
Add vim support link to documentation.

dhatch:


docs/content/any/reference/tooling/ide.md
```diff
@@ -12,6 +12,7 @@ Oso’s IDE (Integrated Development Environment) integrations provide syntax hig
 ## Supported IDEs
 
 ### [Visual Studio Code](https://marketplace.visualstudio.com/items?itemName=osohq.oso)
+### [Vim (community maintained)](https://github.com/osohq/polar.vim)
```
gj, line 15: Is it really `community maintained` at this point? The repo was transferred to our org. The original author still has write access to it, but we discussed removing that since we bear responsibility for its contents now that it's in our org.

dhatch: Agreed. I just pushed a change removing this.

gj (approved):

PR #892, opened 2021-05-11
get_action -> get_checked_permissions

gj:
## TODO

- [x] Update doc strings. `new-roles` docs updates will be a separate PR once I update GitClub.
- [x] Changelog
- [x] Remove 8e9d158 after #900 goes in.


languages/python/sqlalchemy-oso/sqlalchemy_oso/session.py
```diff
@@ -51,62 +52,100 @@ def _before_compile(query):
...
+        # requiring authorization.
+        if alias in checked_permissions:
+            action = checked_permissions[alias]  # type: ignore
+        elif entity in checked_permissions:
+            action = checked_permissions[entity]  # type: ignore
+        else:
+            continue
+
+        session = query.session
+        assert isinstance(session, Session)
```
gj, line 103: This line is from trying to convince Pyright that `session` is indeed a `Session`. I can remove it if anyone minds.

languages/python/sqlalchemy-oso/sqlalchemy_oso/session.py
```diff
@@ -51,62 +52,100 @@ def _before_compile(query):
...
-    """Session factory for sessions with oso authorization applied.
-
-    :param get_oso: Callable that return oso instance to use for authorization.
-    :param get_user: Callable that returns user for an authorization request.
-    :param get_action: Callable that returns action for the authorization request.
+def authorized_sessionmaker(
+    get_oso: Callable[[], Oso],
+    get_user: Callable[[], Any],
+    get_checked_permissions: Callable[[], Permissions],
+    class_: Type[Session] = None,
```
gj, line 115: I added types to most of this file but haven't set up MyPy yet. Probably worth doing that & adding it to CI like some of the other Python libs.

languages/python/sqlalchemy-oso/sqlalchemy_oso/session.py
```diff
@@ -137,9 +188,12 @@ def scoped_session(get_oso, get_user, get_action, scopefunc=None, **kwargs):
     scopefunc = scopefunc or (lambda: None)
 
     def _scopefunc():
-        return (get_oso(), get_action(), get_user(), scopefunc())
+        checked_permissions = frozenset(get_checked_permissions().items())
```
gj, line 191: This is b/c the scope func's return value needs to be hashable.

languages/python/sqlalchemy-oso/tests/test_flask.py
```diff
@@ -29,18 +29,26 @@ def ctx(flask_app):
         yield ctx
 
 
+# Global because test_flask_model wants to enforce authorization on a model
+# that's created after initialization of the AuthorizedSQLAlchemy instance.
+checked_permissions = {}
```
gj, line 34: This is fairly self-explanatory but definitely a wowhack (lowercase) and worth calling out.

I know @leina05 mentioned wanting to remove the `flask_sqlalchemy` integration (`AuthorizedSQLAlchemy` in `languages/python/sqlalchemy-oso/sqlalchemy_oso/flask.py`) entirely, and I would be happy to do that as a follow-up to this PR.

languages/python/sqlalchemy-oso/tests/test_flask.py
```diff
@@ -56,14 +64,16 @@ def test_authorized_sqlalchemy(ctx, flask_app, oso, sqlalchemy, post_fixtures):
...
 
-    with flask_app.app_context():
-        assert sqlalchemy.session.query(Post).count() == 1
+    assert sqlalchemy.session.query(Post).count() == 1
 
 
-def test_flask_model(ctx, flask_app, oso, sqlalchemy):
+def test_flask_model(ctx, oso, sqlalchemy):
     class TestModel(sqlalchemy.Model):
         id = Column(Integer, primary_key=True)
```
gj, line 72: The wowhack is necessitated by the fact that we're retrieving the `AuthorizedSQLAlchemy` instance's base model (`Model` field), which is only available _after_ initializing the instance, in order to subclass this `TestModel` class that we want to perform authorization over. But we need to provide the `checked_permissions` map (enumerating every action-resource pair we want to authorize) at `AuthorizedSQLAlchemy` initialization time, so... conundrum.

gj:
Test failure is from the base branch: https://github.com/osohq/oso/runs/2549561228

languages/python/sqlalchemy-oso/sqlalchemy_oso/session.py
dhatch: Keep it it's a good idea

leina05 (changes requested):
Looks good, main thing is to update the functionality to be deny-by-default like we discussed,

languages/python/sqlalchemy-oso/sqlalchemy_oso/session.py
```diff
@@ -51,62 +52,100 @@ def _before_compile(query):
...
 
-    oso = session.oso_context["oso"]
+    oso: Oso = session.oso_context["oso"]
     user = session.oso_context["user"]
-    action = session.oso_context["action"]
+    checked_permissions: Permissions = session.oso_context["checked_permissions"]
+
+    # Early return if no authorization is to be applied.
+    if checked_permissions is None:
+        return None
```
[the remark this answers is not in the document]
gj: I think this pair of lines is actually fine as is — we discussed making it deny by default for any SQLAlchemy models _not_ explicitly enumerated in the `checked_permissions` map. I still want to provide a way for folks to use a single sessionmaker for queries with & without authorization applied, and passing `None` for `checked_permissions` seems like a reasonable signal value for "don't check any permissions for queries made via this session."

languages/python/sqlalchemy-oso/sqlalchemy_oso/flask.py
```diff
@@ -25,22 +25,22 @@ class AuthorizedSQLAlchemy(SQLAlchemy):
 
     :param get_oso: Callable that returns the :py:class:`oso.Oso` instance to use for authorization.
     :param get_user: Callable that returns the user to authorize for the current request.
-    :param get_action: Callable that returns the action to authorize for the current request.
+    :param get_checked_permissions: Callable that returns the permissions to authorize for the current request.
```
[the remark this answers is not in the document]
gj: Thoughts:
- I like how `checked_permissions=None` reads for the "I don't want to apply any authorization to this session" case. `allowed_permissions=None` kind of reads the opposite way to me.
- For the other cases where you _do_ want to apply authorization to a session, I agree that `checked_permissions` isn't the best way to say "these are the permissions that I'm going to check for this query, and attempts to load data corresponding to a model not present in this permissions map will be denied." 😅 

I would like something that reads well for both cases:
- `___=None` -> Don't check authorization for this session.
- `___={Foo: "read_foo", Bar: "write_bar"}` -> Check the specified actions for the specified models when querying data in this session. Anything not present in the provided map is denied by default.

languages/python/sqlalchemy-oso/sqlalchemy_oso/session.py
gj: Apparently it's already checked in CI and was just waiting for us to add annotations.

languages/python/sqlalchemy-oso/tests/test_sqlalchemy.py
```diff
@@ -303,3 +309,45 @@ def test_unconditional_policy_has_no_filter(engine, oso, fixture_data):
         + "posts.access_level AS posts_access_level, posts.created_by_id AS posts_created_by_id, "
         + "posts.needs_moderation AS posts_needs_moderation \nFROM posts \nWHERE 1 = 1"
     )
+
+
+def test_authorized_sessionmaker_multiple_actions(engine, oso):
```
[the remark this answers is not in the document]
gj: Added that functionality to the existing test for the relationships 'bug' (the one that was previously xfail-ed) and removed this test since it uses bespoke fixture data and is now duplicative of the other test.

languages/python/sqlalchemy-oso/sqlalchemy_oso/session.py
dhatch: Yep! It's not set to require annotations.

dhatch (commented):
Code & change looks good. I know there's some bike shedding on the name of the argument but I think we should just ship it.

Do we have tests for:

- [x] Deny by default
- [x] Deny all if `checked_permissions={}`
- [x] Deny if "Foo: None" in checked_permission map
- [x] Allow all if `None`

Is there any doc string we should update for this functionality (not sure if that's included in adding docs?)

gj:
> Do we have tests for:
> 
>     * [ ]  Deny by default
> 
>     * [x]  Deny all if `checked_permissions={}`
> 
>     * [x]  Deny if "Foo: None" in checked_permission map
> 
>     * [x]  Allow all if `None`

Not sure what you mean by "Deny by default", but I added coverage of the other three cases in e89590d.

> Is there any doc string we should update for this functionality (not sure if that's included in adding docs?)

Updated doc strings in 003255d.

dhatch (approved):
lgtm!

PR #903, opened 2021-05-17
Add webpack live reload instructions to readme.

dhatch:


gj (approved):
we should probably create a Make target but this is cool for now

PR #906, opened 2021-05-18
Move update subscription out of footer and into content body

samraper:
There's some code copied from the callout shortcode but as far as I can tell there isn't a good way to share code between shortcodes and partials


dhatch:
Screen Shot 2021-05-19 at 11 43 48 AM

Could you rearrange the "Read next" button and this form to look more like this? (Excuse my image editing....)

dhatch:
Nevermind, I didn't rebuild correctly locally. It looks good.

dhatch (approved):

PR #907, opened 2021-05-18
Get elements when needed instead of as constants

dhatch:
This fixes some race conditions occuring in production preventing search from working.
And, it makes the code a bit easier to read (no references across webpack and nav.js)


dhatch:
cc @veverkap this is the fix we were discussing

gj (approved):
I don't like attaching functions to the `window` object, but that's a bit outside the scope of this PR. This looks like it will fix the problem in question.

veverkap (approved):

PR #909, opened 2021-05-18
Use real type for resource id columns

saolsen:
Lets us type the resource id correctly (instead of casting it to a string).
Makes the sql more efficient and able to use indexes.

### DIFF
--- a/languages/python/sqlalchemy-oso/sqlalchemy_oso/oso.py
+++ b/languages/python/sqlalchemy-oso/sqlalchemy_oso/oso.py
@@ -30,7 +30,7 @@ class SQLAlchemyOso(Oso):
         self.base = sqlalchemy_base
         self._roles_enabled = False
 
-    def enable_roles(self, user_model, resource_id_column_type, session_maker):
+    def enable_roles(self, user_model, session_maker):
         """Enable the Oso role management API.
         Oso will create SQLAlchemy models to create and assign roles to users (stored in `user_model`).
         The roles API methods will be available on the `roles` property of `SQLAlchemyOso`.
@@ -39,7 +39,6 @@ class SQLAlchemyOso(Oso):
             oso=self,
             sqlalchemy_base=self.base,
             user_model=user_model,
-            resource_id_column_type=resource_id_column_type,
             session_maker=session_maker,
         )
         self._roles_enabled = True
--- a/languages/python/sqlalchemy-oso/sqlalchemy_oso/oso.py
+++ b/languages/python/sqlalchemy-oso/sqlalchemy_oso/oso.py
@@ -30,7 +30,7 @@ class SQLAlchemyOso(Oso):
         self.base = sqlalchemy_base
         self._roles_enabled = False
 
-    def enable_roles(self, user_model, session_maker):
+    def enable_roles(self, user_model, resource_id_column_type, session_maker):
         """Enable the Oso role management API.
         Oso will create SQLAlchemy models to create and assign roles to users (stored in `user_model`).
         The roles API methods will be available on the `roles` property of `SQLAlchemyOso`.
@@ -39,6 +39,7 @@ class SQLAlchemyOso(Oso):
             oso=self,
             sqlalchemy_base=self.base,
             user_model=user_model,
+            resource_id_column_type=resource_id_column_type,
             session_maker=session_maker,
         )
         self._roles_enabled = True
--- a/languages/python/sqlalchemy-oso/sqlalchemy_oso/roles2.py
+++ b/languages/python/sqlalchemy-oso/sqlalchemy_oso/roles2.py
@@ -501,10 +501,7 @@ def ensure_configured(func):
 
 
 class OsoRoles:
-    def __init__(
-        self, oso, sqlalchemy_base, user_model, resource_id_column_type, session_maker
-    ):
-        self.resource_id_column_type = resource_id_column_type
+    def __init__(self, oso, sqlalchemy_base, user_model, session_maker):
         self.session_maker = session_maker
 
         for cls in session_maker.class_.__mro__:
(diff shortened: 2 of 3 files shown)


gj (commented):
Would it be possible to build an internal map of resource_type -> pk_type during `SQLAlchemyOso` initialization (same place we `register_models()`) instead of (A) asking users to manually supply the `resource_id_column_type` and (B) forcing all resource ID columns to be the same type?

saolsen:
> Would it be possible to build an internal map of resource_type -> pk_type during `SQLAlchemyOso` initialization (same place we `register_models()`) instead of (A) asking users to manually supply the `resource_id_column_type` and (B) forcing all resource ID columns to be the same type?

Yeah it's possible. I'm up for getting the type in a more inferred way but it would have to be during the init which happens before register_models so I don't actually have the models at that point.

As far as not forcing all resource IDs to have the same type; Doing that would complicate the sql a lot and the way we store role assignments. Making them all the same type seems like something most people would be ok with and allows us to keep the model as it is.

dhatch:
> As far as not forcing all resource IDs to have the same type; Doing that would complicate the sql a lot and the way we store role assignments. Making them all the same type seems like something most people would be ok with and allows us to keep the model as it is.

We basically decided not to do this (https://www.notion.so/osohq/Table-per-role-resource-so-we-can-use-actual-id-columns-instead-of-strings-and-update-all-queries-to-dc8495cca8ba4c5da5a1a520ef5eb46a) so that we could spend less time on this.

@saolsen Did this change make it so that the single item queries are able to use index scans?

saolsen:
> @saolsen Did this change make it so that the single item queries are able to use index scans?

Not verified, though we also didn't verify that they couldn't before.

languages/python/sqlalchemy-oso/tests/test_roles2.py
```diff
@@ -2209,6 +2213,38 @@ def test_user_in_role(
         assert repo.org_id == "osohq"
 
 
+def test_mismatched_id_types_throws_error(engine, Base, User):
```
dhatch, line 2216: Should we add another test confirming that everything works with `Integer` pks? 

Maybe this one `test_homogeneous_role_perm` but with `Integer`?

gj:
> Yeah it's possible. I'm up for getting the type in a more inferred way but it would have to be during the init which happens before register_models so I don't actually have the models at that point.

doesn't `register_models` happen during `SQLAlchemyOso` init?

languages/python/sqlalchemy-oso/sqlalchemy_oso/oso.py
```diff
@@ -30,7 +30,7 @@ def __init__(self, sqlalchemy_base):
         self.base = sqlalchemy_base
         self._roles_enabled = False
 
-    def enable_roles(self, user_model, session_maker):
+    def enable_roles(self, user_model, resource_id_column_type, session_maker):
```
dhatch, line 33: Can we add documentation of this new parameter (and the others, I know they're missing) to this doc string?

languages/python/sqlalchemy-oso/tests/test_roles2.py
saolsen: Yeah that's a good idea.

saolsen:
> > Yeah it's possible. I'm up for getting the type in a more inferred way but **it would have to be during the init which happens before register_models** so I don't actually have the models at that point.
> 
> doesn't `register_models` happen during `SQLAlchemyOso` init?

Ah yeah `SQLAlchemyOso` init. Cool. I think I can manage to get this working.

dhatch:
> Not verified, though we also didn't verify that they couldn't before.

The reason I'm asking is after we do [this](https://www.notion.so/osohq/Rewrite-list-filtering-queries-to-not-table-scan-cea8aa3a882f42f9b05b53e0d7d0fad7) we shouldn't have any queries that have to do table scans, right?

I think the [`cast` in the join](https://github.com/osohq/oso/pull/909/files#diff-b0eb8cb3a6984c50da269d9c431c6649263f4b8782e73c8b2279d09ebc7d84a6L98) that you got rid have would have caused a table scan, but I'm not sure if there's anything else.

languages/python/sqlalchemy-oso/tests/test_roles2.py
```diff
@@ -2209,6 +2212,65 @@ def test_user_in_role(
...
+
+    Session = sessionmaker(bind=engine)
+
+    oso = SQLAlchemyOso(Base)
+
+    with pytest.raises(OsoError):
+        oso.enable_roles(User, Session)
+
+
+def test_integer_ids(engine, Base, User):
```
gj, line 2234: would be cool to parametrize this like

```suggestion
@pytest.mark.parametrize("sa_type,one_id", [(String, "1"), (Integer, 1)])
def test_integer_ids(engine, Base, User, sa_type, one_id):
```

and then update the body to use those two parametrized fixtures

gj (commented):
lookin' good; couple outstanding questions

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles2.py
```diff
@@ -514,6 +514,20 @@ def __init__(self, oso, sqlalchemy_base, user_model, session_maker):
...
+                continue
+            if model == user_model:
+                continue
+            _, id_type = get_pk(model)
+            if resource_id_column_type is None:
+                resource_id_column_type = id_type
+            elif resource_id_column_type.__class__ != id_type.__class__:
+                raise OsoError("All resource ids must match have the same id type.")
+
+        self.resource_id_column_type = resource_id_column_type
```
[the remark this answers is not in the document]
saolsen: Yeah, that would probably only happen if you didn't register your models on Base before enabling roles so will add a message saying that.

languages/python/sqlalchemy-oso/tests/test_roles2.py
saolsen: this is cool

gj (approved):

PR #915, opened 2021-05-20
Rewrite list filter queries to not table scan.

saolsen:
Creates a query that looks something like this

```
with allow_permission as (
    -- Find the permission
    select
            p.id
    from permissions p
    where p.resource_type = 'Repository' and p.name = 'pull' -- action and resource type
), permission_roles as (
    -- find roles with the permission
    select
            rp.role
    from role_permissions rp
    join allow_permission ap
    on rp.permission_id = ap.id
), relevant_roles as (
    -- recursively find all roles that have the permission or
    -- imply a role that has the permission
    with recursive relevant_roles (role) as (
            select
                    role
            from permission_roles
            union
            select
                    ri.from_role
            from role_implications ri
            join relevant_roles rr
            on ri.to_role = rr.role
    ) select * from relevant_roles
), user_relevant_roles as (
    select
        *
    from user_roles ur
    join relevant_roles rr
    on rr.role = ur.role
    where ur.user_id = 1 -- user_id
)
select
  R.id
from repositories R
join user_relevant_roles urr
on urr.resource_type = 'Repository' and urr.resource_id = R.id
union
select
  R.id
from repositories R
join organizations P
on R.org_id = P.id
join user_relevant_roles urr
on urr.resource_type = 'Organization' and urr.resource_id = P.id
```

### DIFF
--- a/languages/python/sqlalchemy-oso/sqlalchemy_oso/roles2.py
+++ b/languages/python/sqlalchemy-oso/sqlalchemy_oso/roles2.py
@@ -178,8 +178,6 @@ def list_filter_query(kind, resource, relationships, id_field):
                 found_parent = True
                 break
 
-    # user_id, resource_type, action or role
-
     sql = ""
     assert kind in ["role_allow", "user_in_role"]
     # Get the roles to start from.
@@ -254,7 +252,7 @@ def list_filter_query(kind, resource, relationships, id_field):
         role_joins += f"and {urr_table}.resource_id = r.{rel.parent_table}_id"
         role_filters += f" or {urr_table}.resource_id is not NULL"
 
-    # TODO: join to parents
+    # Join to parents
     sql += f"""
     resources_with_parents as (
         select
@@ -264,7 +262,7 @@ def list_filter_query(kind, resource, relationships, id_field):
     )
     """
 
-    # @TODO: join to roles and filter
+    # Join to roles and filter by having a role
     sql += f"""
     select
       r.id
@@ -272,8 +270,6 @@ def list_filter_query(kind, resource, relationships, id_field):
     {role_joins}
     {role_filters}
     """
-
-    # TODO: add the null checks
 
     return sql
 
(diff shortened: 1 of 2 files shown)


saolsen:
Still table scans, gonna do some tweaks though I have an idea.

saolsen:
ok updated and doesn't table scan with my examples

saolsen:
Not sure how long this link will work but here's a sample explain. https://explain.dalibo.com/plan/21b#plan

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles2.py
```diff
@@ -163,6 +162,113 @@ def user_in_role_query(
...
+        prev_joins.append(join)
+
+        sql += f"""
+        union
+        select
+        {tablename}.{id_field}
+        from {tablename}
+        """
+
+        for join in prev_joins:
```
dhatch, line 260: Are there tests that go across > 2 levels of relationships and use data filtering?

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles2.py
```diff
@@ -163,6 +162,113 @@ def user_in_role_query(
     return query
 
 
+def list_filter_query(kind, resource, relationships, id_field):
```
dhatch, line 165: I don't know the code enough to know if this would be helpful, so just a suggestion:

Is there a way to write the single item queries to just extend this? Would that make sense & make the code a bit easier to maintain? It would just involve adding the PK filters to the "select data" part of the sql generation.

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles2.py
saolsen: There's Issue list filtering tests that use 2 relationships

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles2.py
saolsen: yeah that would probably work, will try it as a followup

languages/python/sqlalchemy-oso/tests/test_roles2.py
```diff
@@ -2382,6 +2382,14 @@ def test_roles(init_oso, auth_sessionmaker, User, Organization, Repository, Issu
     assert demo_repo.id in result_ids
     assert ios.id not in result_ids
 
+    oso.actor = leina
+    oso.checked_permissions = {Issue: "edit"}
+    auth_session = auth_sessionmaker()
+
+    results = auth_session.query(Issue).all()
+    assert len(results) == 1
+    result_ids = [repo.id for repo in results]
```
dhatch, line 2391: The results are issues, not repos.
```suggestion
    result_ids = [issue.id for issue in results]
```

dhatch (approved):

PR #918, opened 2021-05-21
Add help.osohq.com messages.

dhatch:
These changes add links to a help endpoints that we can use to direct users to the appropriate resource for the type of error they are encountering.

### DIFF
--- a/languages/python/oso/oso/oso.py
+++ b/languages/python/oso/oso/oso.py
@@ -83,6 +83,6 @@ class Oso(Polar):
     def _print_polar_log_message(self):
         if os.environ.get("POLAR_LOG", None):
             print(
-                "It looks like you've enabled Polar tracing. Get help with "
+                "Polar tracing enabled. Get help with "
                 + "traces from our engineering team: https://help.osohq.com/trace"
             )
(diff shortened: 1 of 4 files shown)


languages/python/oso/oso/oso.py
```diff
@@ -77,3 +79,10 @@ def get_allowed_actions(self, actor, resource, allow_wildcard=False) -> list:
             actions.add(action)
 
         return list(actions)
+
+    def _print_polar_log_message(self):
+        if os.environ.get("POLAR_LOG", None):
+            print(
+                "It looks like you've enabled Polar tracing. Get help with "
```
saolsen, line 86:
```suggestion
                "Polar tracing enabled. Get help with "
```

saolsen (approved):

languages/python/oso/oso/oso.py
dhatch: good call

languages/python/oso/oso/oso.py
```diff
@@ -77,3 +79,10 @@ def get_allowed_actions(self, actor, resource, allow_wildcard=False) -> list:
             actions.add(action)
 
         return list(actions)
+
+    def _print_polar_log_message(self):
+        if os.environ.get("POLAR_LOG", None):
+            print(
+                "Polar tracing enabled. Get help with "
```
gj, line 86: Doesn't have to get figured out as part of this PR, but should this maybe be `OSO_LOG=1` and `Oso tracing enabled.` in light of our attempts to standardize on Oso > Polar in public?

gneray: yes!

PR #919, opened 2021-05-22
Adds Doc Search Generation To CI

veverkap:
This PR adds generating search index to CI.

Index preview_OSODOCS?searchMode=search is used on preview docs site.

### DIFF
--- a/.github/workflows/publish-docs.yml
+++ b/.github/workflows/publish-docs.yml
@@ -21,6 +21,7 @@ jobs:
       AWS_SECRET_ACCESS_KEY: ${{ secrets.DOCS_AWS_SECRET }}
       AWS_REGION: "us-east-1"
       PUBLISH_DOCS: 1
+      ALGOLIA_ADMIN_API_KEY: ${{ secrets.ALGOLIA_ADMIN_API_KEY }}
     steps:
       - uses: actions/checkout@v2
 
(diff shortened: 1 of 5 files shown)


docs/Makefile
```diff
@@ -85,6 +85,14 @@ build-search:
 
 require-algolia-vars: require-ALGOLIA_APPLICATION_ID require-ALGOLIA_ADMIN_API_KEY require-ALGOLIA_INDEX
 
+## This is the entry for all doc stuff for preview
+ci-preview-docs: require-ALGOLIA_ADMIN_API_KEY build build-search 
+	./searcher -index='preview_OSODOCS' -app_id='KROZ8F05YT'
+
+## This is the entry for all doc stuff for preview
```
samscott89, line 92:
```suggestion
## This is the entry for all doc stuff for production
```

.github/workflows/publish-docs.yml
```diff
@@ -90,17 +91,19 @@ jobs:
         uses: peaceiris/actions-hugo@v2
         with:
           hugo-version: '0.79.1'
-      - name: Build Hugo docs
+      - name: Build searcher
```
dhatch, line 94: Both steps build-search as a dependency, so this step should not be needed, right?

dhatch (approved):

PR #927, opened 2021-05-26
Fix cibuildwheel setup for sqlalchemy-oso

gj:


gj:
Successful test run: https://github.com/osohq/oso/runs/2677274457?check_suite_focus=true

languages/python/sqlalchemy-oso/tox.ini
```diff
@@ -1,5 +1,6 @@
 [tox]
-envlist = py3-sqlalchemy{13,14}-{earliest,latest}
+skip_missing_interpreters=true
+envlist = {py3,pypy3}-sqlalchemy{13,14}-{earliest,latest}
```
gj, line 3: Basically this is saying "if you find a `python3` executable, run the four tests (1.3.17, 1.3.latest, 1.4.0, 1.4.latest), and if you find a `pypy3` executable, do the same." It'll happily skip over either (or both) of those if they're missing without giving much in the way of visual confirmation that it skipped over one or both missing interpreters. The only visual confirmation is that the `OK` test runs will have completed _very_ quickly relative to runs that actually run the tests.

Implications:
- nothing changes for local development; everyone should have a `python3` executable and there's no need for everyone to install PyPy locally.
- For cibuildwheel, it spins up a bunch of Docker containers, some of which have `python3` available and some of which (1 of which?) have `pypy3` available. Tox will run the tests for whichever interpreter it finds and skip the tests for the missing interpreter.

dhatch (approved):

PR #929, opened 2021-05-26
Error when enable_roles() is called more than once

gj:
PR checklist:

- [x] Added changelog entry.


languages/python/sqlalchemy-oso/sqlalchemy_oso/oso.py
```diff
@@ -35,9 +35,10 @@ def enable_roles(self, user_model, session_maker):
         Oso will create SQLAlchemy models to create and assign roles to users (stored in `user_model`).
         The roles API methods will be available on the `roles` property of `SQLAlchemyOso`.
         """
+        if self._roles_enabled:
+            raise OsoError("Calling 'enable_roles()' twice has no effect.")
```
saolsen, line 39: Maybe "Roles already enabled". It seems that it does have the effect of throwing this error.

saolsen (approved):

PR #933, opened 2021-05-27
Fix bug with lookups in rule head

### LINKED ISSUE #913: Query evaluation doesn't fail on lookup in the head of a rule
Background: https://oso-oss.slack.com/archives/C017KSBTD4M/p1621457882015600

When lookup is done in the head of a rule, Polar doesn't backtrack on a failed unification.

E.g., the following code

```python
class User:
    def __init__(self, name):
        self.name = name


class Action(Enum):
    READ = "READ"
    WRITE = "WRITE"


oso = Oso()
oso.register_class(User)
oso.register_class(Action)
oso.load_str("allow(_user: User, Action.READ, _);")
alice = User("alice")
assert oso.is_allowed(alice, Action.READ, 1)
assert not oso.is_allowed(alice, Action.WRITE, 1)
```

results in: 

```
  File "test.py", line 21, in <module>
    assert not oso.is_allowed(alice, Action.WRITE, 1)
AssertionError
```

However, moving the lookup to the body of the rule fixes the problem, e.g.

```python
oso.load_str("allow(_user: User, action, _) if action = Action.READ;")
```

Some internal documentation here as well: https://www.notion.so/osohq/Failing-to-fail-on-lookup-in-the-head-of-a-rule-134725b87c97414b97709ca8e798951b

leina05:
Fixes #913

Issue was in rebinding variables to the results of external calls, even when the variable may already be bound.
This is an issue for lookups in rule heads, as the result variable for the lookup will already be bound to the argument passed into the rule.

E.g.,
The following code will fail.

```
oso.load_str("""allow("leina", Actions.READ, "doc");""")

assert not oso.is_allowed("leina", Actions.WRITE, "doc");

```

The Polar log explains a bit about what's happening:

```
RULE: allow("leina", Actions.READ, "doc") if ;
[debug]       QUERY: Actions.READ = _value_1_3, BINDINGS: {_value_1_3 = <Actions.WRITE: 2>, Actions = <enum 'Actions'>}
[debug]         QUERY: <enum 'Actions'>.READ = _value_1_3, BINDINGS: {_value_1_3 = <Actions.WRITE: 2>}
[debug]           LOOKUP: <enum 'Actions'>.READ()
[debug]           => <Actions.READ: 1>
[debug]           BACKTRACK
[debug]           HALT

```

What happens is that the argument passed in (`Actions.WRITE`) is bound to a temporary variable `_value_1_3`. We then query for `Actions.READ`.
The bug was caused by rebinding the result of the lookup to `_value_1_3` without checking if `_value_1_3` was already bound.
Rebinding the variable ignores the argument passed in, when what we should do is unify the result with `_value_1_3`, so that if the variable is bound we check that it matches the result. 

In order to switch from rebinding to unification in `external_call_result`, it was also necessary to change the way we initialize the result variables for external calls. Previously, we initialized the variable associated with an external `call_id` to `false`, which would always cause unification with a result to fail (unless the result was `false`). This was done with `new_call_var`, which takes an initial value.

To fix this, I stopped initializing to the `call_id` variable to `false` only for `LookupExternal` and `NextExternal` goals. For other goals `ExternalOp`, `ExternalIsa`, etc., where the result is always a boolean, we rely on the default value of the result variable being false, so I didn't change those. 

PR checklist:

- [x]  Added changelog entry.
- [x] Regression test in the core.

### DIFF
--- a/docs/content/any/project/changelogs/NEXT.md
+++ b/docs/content/any/project/changelogs/NEXT.md
@@ -8,17 +8,17 @@ description: >-
 draft: true
 ---
 
-## `RELEASED_PACKAGE_1` NEW_VERSION
+## `oso` v0.12.4
 
-### LANGUAGE (e.g., 'Core' or 'Python' or 'Node.js')
+### Core
 
 #### Breaking changes
 
 <!-- TODO: remove warning and replace with "None" if no breaking changes. -->
 
 {{% callout "Warning" "orange" %}}
-  This release contains breaking changes. Be sure to follow migration steps
-  before upgrading.
+This release contains breaking changes. Be sure to follow migration steps
+before upgrading.
 {{% /callout %}}
 
 ##### Breaking change 1
@@ -37,6 +37,4 @@ Link to [relevant documentation section]().
 
 #### Other bugs & improvements
 
-- Bulleted list
-- Of smaller improvements
-- Potentially with doc [links]().
+- Fixed bug with dot lookups in the head of a rule ([#933](https://github.com/osohq/oso/pull/933)).
--- a/languages/python/oso/tests/test_polar.py
+++ b/languages/python/oso/tests/test_polar.py
@@ -1,6 +1,7 @@
 from datetime import datetime
 from math import inf, isnan, nan
 from pathlib import Path
+from enum import Enum
 
 from polar import (
     polar_class,
@@ -12,7 +13,6 @@ from polar import (
     Pattern,
 )
 from polar.partial import TypeConstraint
-from polar.exceptions import InvalidCallError, UnexpectedPolarTypeError
 
 import pytest
 
@@ -123,7 +123,7 @@ def test_external(polar, qvar, qeval):
     polar.register_class(Foo)
     assert qvar("new Foo().a = x", "x", one=True) == "a"
     with pytest.raises(
-        InvalidCallError, match="tried to call 'a' but it is not callable"
+        exceptions.InvalidCallError, match="tried to call 'a' but it is not callable"
     ):
         assert not qeval("new Foo().a() = x")
     assert not qvar("new Foo().b = x", "x", one=True) == "b"
@@ -894,5 +894,52 @@ def test_unexpected_expression(polar):
     """Ensure expression type raises error from core."""
     polar.load_str("f(x) if x > 2;")
 
-    with pytest.raises(UnexpectedPolarTypeError):
+    with pytest.raises(exceptions.UnexpectedPolarTypeError):
         next(polar.query_rule("f", Variable("x")))
+
+
+def test_lookup_in_head(polar):
+    # Test with enums
+    class Actions(Enum):
+        READ = 1
+        WRITE = 2
+
+    polar.register_constant(Actions, "Actions")
+    p = """allow("leina", Actions.READ, "doc");"""
+    polar.load_str(p)
+
+    assert len(list(polar.query_rule("allow", "leina", Actions.WRITE, "doc"))) == 0
+    assert len(list(polar.query_rule("allow", "leina", "READ", "doc"))) == 0
+    assert len(list(polar.query_rule("allow", "leina", 1, "doc"))) == 0
+    assert len(list(polar.query_rule("allow", "leina", Actions, "doc"))) == 0
+    assert len(list(polar.query_rule("allow", "leina", Actions.READ, "doc"))) == 1
+
+    # Test lookup in specializer raises error
+    p = """allow("leina", action: Actions.READ, "doc");"""
+    with pytest.raises(exceptions.UnrecognizedToken):
+        polar.load_str(p)
+
+    # Test with normal class
+    class Resource:
+        def __init__(self, action):
+            self.action = action
+
+    polar.register_class(Resource, name="Resource")
+    p = """allow("leina", resource.action, resource: Resource);"""
+    polar.load_str(p)
+
+    r = Resource("read")
+
+    assert len(list(polar.query_rule("allow", "leina", "write", r))) == 0
+    assert len(list(polar.query_rule("allow", "leina", "read", r))) == 1
+
+
+# def test_specialize_on_constant(polar):
+#     const_dict = {"foo": "bar"}
+#     polar.register_constant(const_dict, "MY_DICT")
+
+#     p = """allow("leina", "read", resource: MY_DICT);"""
+#     polar.load_str(p)
+
+#     assert len(list(polar.query_rule("allow", "leina", "read", {"foo": "bar"}))) == 1
+#     assert len(list(polar.query_rule("allow", "leina", "read", {}))) == 0
--- a/languages/python/oso/tests/test_polar.py
+++ b/languages/python/oso/tests/test_polar.py
@@ -898,26 +898,24 @@ def test_unexpected_expression(polar):
         next(polar.query_rule("f", Variable("x")))
 
 
-def test_lookup_in_head(polar):
+def test_lookup_in_head(polar, is_allowed):
     # Test with enums
     class Actions(Enum):
         READ = 1
         WRITE = 2
 
     polar.register_constant(Actions, "Actions")
-    p = """allow("leina", Actions.READ, "doc");"""
-    polar.load_str(p)
+    polar.load_str('allow("leina", Actions.READ, "doc");')
 
-    assert len(list(polar.query_rule("allow", "leina", Actions.WRITE, "doc"))) == 0
-    assert len(list(polar.query_rule("allow", "leina", "READ", "doc"))) == 0
-    assert len(list(polar.query_rule("allow", "leina", 1, "doc"))) == 0
-    assert len(list(polar.query_rule("allow", "leina", Actions, "doc"))) == 0
-    assert len(list(polar.query_rule("allow", "leina", Actions.READ, "doc"))) == 1
+    assert not is_allowed("leina", Actions.WRITE, "doc")
+    assert not is_allowed("leina", "READ", "doc")
+    assert not is_allowed("leina", 1, "doc")
+    assert not is_allowed("leina", Actions, "doc")
+    assert is_allowed("leina", Actions.READ, "doc")
 
     # Test lookup in specializer raises error
-    p = """allow("leina", action: Actions.READ, "doc");"""
     with pytest.raises(exceptions.UnrecognizedToken):
-        polar.load_str(p)
+        polar.load_str('allow("leina", action: Actions.READ, "doc");')
 
     # Test with normal class
     class Resource:
@@ -925,10 +923,9 @@ def test_lookup_in_head(polar):
             self.action = action
 
     polar.register_class(Resource, name="Resource")
-    p = """allow("leina", resource.action, resource: Resource);"""
-    polar.load_str(p)
+    polar.load_str('allow("leina", resource.action, resource: Resource);')
 
     r = Resource("read")
 
-    assert len(list(polar.query_rule("allow", "leina", "write", r))) == 0
-    assert len(list(polar.query_rule("allow", "leina", "read", r))) == 1
+    assert not is_allowed("leina", "write", r)
+    assert is_allowed("leina", "read", r)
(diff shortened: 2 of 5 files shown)


languages/python/oso/tests/test_polar.py
```diff
@@ -894,5 +894,41 @@ def test_unexpected_expression(polar):
     """Ensure expression type raises error from core."""
     polar.load_str("f(x) if x > 2;")
 
-    with pytest.raises(UnexpectedPolarTypeError):
+    with pytest.raises(exceptions.UnexpectedPolarTypeError):
         next(polar.query_rule("f", Variable("x")))
+
+
+def test_lookup_in_head(polar):
```
leina05, line 901: Do we need the library tests if we have the regression test in the core?

dhatch:
@leina05 did you give any thought to changes we could make to prevent this bug in the future? The issue was lookup external relying on the result variable being unbound, but some callers were binding it (isa).

For example, one pretty easy thing we could do is assert that the call_id variable is unbound before returning the ExternalCall & ExternalNext query events.

languages/python/oso/tests/test_polar.py
gj: I don't mind keeping it, but we might consider moving it to the parity test suite. What do you think?

leina05:
> @leina05 did you give any thought to changes we could make to prevent this bug in the future? The issue was lookup external relying on the result variable being unbound, but some callers were binding it (isa).
> 
> For example, one pretty easy thing we could do is assert that the call_id variable is unbound before returning the ExternalCall & ExternalNext query events.

Yeah that's  a good idea.

polar-core/src/vm.rs
```diff
@@ -2932,14 +2939,17 @@ impl Runnable for PolarVirtualMachine {
...
 
-            self.rebind_external_answer(
-                &self
-                    .call_id_symbols
-                    .get(&call_id)
-                    .expect("unregistered external call ID")
-                    .clone(),
-                value,
-            );
+            // get the variable to bind result to
```
gj, line 2942:
```suggestion
            // Fetch variable to unify with call result.
```

gj (changes requested):
Agree w/ @dhatch about adding assertion. Otherwise LGTM

languages/python/oso/tests/test_polar.py
leina05: Why parity tests? Is this a feature we supported in the old Oso?

gj: Ah, there's an unfortunate nomenclature clash in our codebase.

You're talking about [the old Python parity tests](https://github.com/osohq/oso/tree/c453e478cf7dbde9ccb12d2ba73fca9915b8ca0d/languages/python/oso/tests/parity), which we've been meaning to remove from the codebase for a while. The only thing stopping us in the past was ensuring that we had captured *all* of the test cases in the old parity test suite in the new Python test suite. Those are effectively obsolete, and we should really clean them up one of these days.

I'm talking about [the language parity tests](https://github.com/osohq/oso/tree/c453e478cf7dbde9ccb12d2ba73fca9915b8ca0d/test), which exist to ensure parity of functionality across the various language integrations.

leina05:
> > @leina05 did you give any thought to changes we could make to prevent this bug in the future? The issue was lookup external relying on the result variable being unbound, but some callers were binding it (isa).
> > For example, one pretty easy thing we could do is assert that the call_id variable is unbound before returning the ExternalCall & ExternalNext query events.
> 
> Yeah that's a good idea.

After trying this, I realized this is actually not the correct behavior. The `call_id` variable will be bound in the case of lookups in the rule head, if the argument is a bound value. In that case, we will create a new `call_id` using the value passed in, so that the result of the lookup will unify with that value (https://github.com/osohq/oso/blob/c453e478cf7dbde9ccb12d2ba73fca9915b8ca0d/polar-core/src/vm.rs#L1887-L1903).

So the behavior we really want to check is: "make sure that we don't create artificially bound `call_id` variables ourselves, but if the variable is legitimately bound, then it's okay." That's a pretty hard thing to assert with the way the code is currently written.

leina05:
> > > @leina05 did you give any thought to changes we could make to prevent this bug in the future? The issue was lookup external relying on the result variable being unbound, but some callers were binding it (isa).
> > > For example, one pretty easy thing we could do is assert that the call_id variable is unbound before returning the ExternalCall & ExternalNext query events.
> > 
> > 
> > Yeah that's a good idea.
> 
> After trying this, I realized this is actually not the correct behavior. The `call_id` variable will be bound in the case of lookups in the rule head, if the argument is a bound value. In that case, we will create a new `call_id` using the value passed in, so that the result of the lookup will unify with that value (
> 
> https://github.com/osohq/oso/blob/c453e478cf7dbde9ccb12d2ba73fca9915b8ca0d/polar-core/src/vm.rs#L1887-L1903
> 
> ).
> So the behavior we really want to check is: "make sure that we don't create artificially bound `call_id` variables ourselves, but if the variable is legitimately bound, then it's okay." That's a pretty hard thing to assert with the way the code is currently written.

Fixed this by changing the way we create `LookupExternal` goals inside of `dot_op_helper`. Instead of using the value from the `Dot` operation as the `call_id` variable for the lookup, I create a fresh variable for the `call_id`, then push a `Unify` goal following the `LookupExternal` to unify this variable with the `Dot` value.

polar-core/src/vm.rs
```diff
@@ -419,6 +419,12 @@ impl PolarVirtualMachine {
         (call_id, Term::new_temporary(Value::Variable(sym)))
     }
 
+    fn get_call_sym(&self, call_id: &u64) -> &Symbol {
```
gj, line 422: nit: no need to take `call_id` as a ref; `u64` is `Copy` anyway

polar-core/src/vm.rs
```diff
@@ -530,6 +536,16 @@ impl PolarVirtualMachine {
             }
             .into());
         }
+        // For LookupExternal and NextExternal goals, make sure that the call id result variable is unbound
+        match goal {
+            Goal::LookupExternal { call_id, .. } | Goal::NextExternal { call_id, .. } => {
+                assert!(matches!(
+                    self.variable_state(self.get_call_sym(&call_id)),
+                    VariableState::Unbound
+                ));
```
gj, line 545: would be good to add an explanation as the second arg to this `assert!()` macro; once you've done that, the comment on 539 might be redundant

polar-core/src/vm.rs
```diff
@@ -2932,14 +2952,13 @@ impl Runnable for PolarVirtualMachine {
...
-                    .get(&call_id)
-                    .expect("unregistered external call ID")
-                    .clone(),
-                value,
-            );
+            // Fetch variable to unify with call result.
+            let sym = &self.get_call_sym(&call_id).to_owned();
+
+            self.push_goal(Goal::Unify {
+                left: Term::new_temporary(Value::Variable(sym.to_owned())),
```
gj, line 2959: if you remove the ref on 2956, you don't need the additional call to `.to_owned()` on 2959:
```suggestion
            let sym = self.get_call_sym(&call_id).to_owned();

            self.push_goal(Goal::Unify {
                left: Term::new_temporary(Value::Variable(sym)),
```

polar-core/src/vm.rs
```diff
@@ -2155,8 +2171,12 @@ impl PolarVirtualMachine {
...
             // Push an `ExternalLookup` goal for external instances
             (_, Value::ExternalInstance(_)) => {
-                // Generate symbol for next result and bind to `false` (default)
-                let (call_id, next_term) = self.new_call_var("next_value", Value::Boolean(false));
+                // Generate symbol for next result and leave the variable unbound, so that unification with the result does not fail
+                // Unification of the `next_term` variable with the result of `NextExternal` happens in `fn external_call_result()`
+                // `external_call_result` is the handler for results from both `LookupExternal` and `NextExternal`, so neither can bind the
+                // call ID variable to `false`.
+                let next_term = self.kb.read().unwrap().gensym("next_value");
+                let call_id = self.new_call_id(&next_term);
```
gj, line 2179: seems to be a symbol instead of a term
```suggestion
                // Generate symbol for next result and leave the variable unbound, so that unification with the result does not fail.
                // Unification of the `next_sym` variable with the result of `NextExternal` happens in `fn external_call_result()`
                // `external_call_result` is the handler for results from both `LookupExternal` and `NextExternal`, so neither can bind the
                // call ID variable to `false`.
                let next_sym = self.kb.read().unwrap().gensym("next_value");
                let call_id = self.new_call_id(&next_sym);
```

polar-core/src/vm.rs
```diff
@@ -2167,7 +2187,7 @@ impl PolarVirtualMachine {
                     },
                     Goal::Unify {
                         left: item.clone(),
-                        right: next_term,
+                        right: Term::new_temporary(Value::Variable(next_term)),
```
gj, line 2190:
```suggestion
                        right: Term::new_temporary(Value::Variable(next_sym)),
```

gj (changes requested):
mostly small stuff

languages/python/oso/tests/test_polar.py
leina05: Aha, yes makes sense. Okay I can move it to the language parity tests.

languages/js/test/parity.ts
```diff
@@ -185,5 +185,24 @@ oso.registerClass(E);
   if (((await result.next()).value as Map<string, any>).get('z') !== 3)
     throw new Error();
 
+  // Test lookup in rule head
+
+  // This query should succeed--is that what's happening????
```
gj, line 190:
```suggestion
```

languages/js/test/parity.ts
```diff
@@ -185,5 +185,24 @@ oso.registerClass(E);
...
+  if (
+    (await oso.queryRule('testLookupInRuleHead', 'attr', new A('attr')).next())
+      .done
+  )
+    throw new Error();
+
+  // This query should fail--is that what's happening????
```
gj, line 197:
```suggestion
```

polar-core/src/vm.rs
```diff
@@ -2155,8 +2170,12 @@ impl PolarVirtualMachine {
             }
             // Push an `ExternalLookup` goal for external instances
             (_, Value::ExternalInstance(_)) => {
-                // Generate symbol for next result and bind to `false` (default)
-                let (call_id, next_term) = self.new_call_var("next_value", Value::Boolean(false));
+                // Generate symbol for next result and leave the variable unbound, so that unification with the result does not fail
+                // Unification of the `next_term` variable with the result of `NextExternal` happens in `fn external_call_result()`
```
gj, line 2174:
```suggestion
                // Unification of the `next_sym` variable with the result of `NextExternal` happens in `fn external_call_result()`
```

polar-core/src/vm.rs
```diff
@@ -2932,14 +2951,13 @@ impl Runnable for PolarVirtualMachine {
...
-            self.rebind_external_answer(
-                &self
-                    .call_id_symbols
-                    .get(&call_id)
-                    .expect("unregistered external call ID")
-                    .clone(),
-                value,
-            );
+            // Fetch variable to unify with call result.
+            let sym = (&self).get_call_sym(call_id).to_owned();
```
gj, line 2955:
```suggestion
            let sym = self.get_call_sym(call_id).to_owned();
```

gj (changes requested):
Looks like we're missing parity tests for Rust & Go. Rust should go in `languages/rust/oso/tests/parity.rs`. Go might be a little trickier since it'll involve updating @samscott89's YAML biz, which currently runs as part of the Go test suite.

languages/js/test/parity.ts
leina05: So is this actually checking that the query succeeds?

gj: yep

docs/content/any/project/changelogs/NEXT.md
```diff
@@ -8,17 +8,17 @@ description: >-
 draft: true
 ---
 
-## `RELEASED_PACKAGE_1` NEW_VERSION
+## `oso` v0.12.4
```
gj, line 11:
```suggestion
## `oso` 0.12.4
```

gj (approved):
can also fix this in a follow-up PR if you just want to ship this

PR #934, opened 2021-05-28
Diff docs on PR run.

dhatch:
Add a job that looks for changes in the docs made by a diff by building the docs against the branch and base branch.

If there are changes, will comment on the PR, as in #937 and require a comment to acknowledge the changes (you must manually re-run the job). The intent of this is to catch PRs that inadvertently break the docs, like changes to code included in code snippets.

If you make a PR changing the docs intentionally, just comment "Docs changes ok" after uploading, and the job will pass on the first run.

If you don't make any changes to the docs (#936), the job also passes, with no comment required.

PR checklist:

- [ ] Added changelog entry.

### DIFF
--- a/.github/workflows/docs-diff.yml
+++ b/.github/workflows/docs-diff.yml
@@ -159,7 +159,7 @@ jobs:
           body="${body//$'\r'/'%0D'}"
           echo "::set-output name=body::$body"
       - name: Comment
-        uses: peter-evans/create_or_update_comment@v1
+        uses: peter-evans/create-or-update-comment@v1
         with:
           issue-number: ${{ github.event.number }}
           body: ${{ steps.get-comment-body.outputs.body }}
--- a/.github/workflows/docs-diff.yml
+++ b/.github/workflows/docs-diff.yml
@@ -2,11 +2,24 @@
 name: Docs Diff
 on:
   pull_request:
-    types: [opened, synchronize, reopened]
+    types: [opened, synchronize, reopened, labeled, unlabeled]
 jobs:
+  check-label:
+    name: Check for "Docs" label
+    runs-on: ubuntu-latest
+    outputs:
+      hadLabel: ${{ steps.label.outcome == "skipped" }}
+    steps:
+      - id: label
+        name: Check for "Docs" label
+        run: exit 1 # Fail the job!
+        if: "!contains(github.event.pull_request.labels.*.name, 'Docs')"
+        continue-on-error: true
   build-head:
     name: Build docs on HEAD.
     runs-on: ubuntu-latest
+    needs: check-label
+    if: ${{ !needs.check-id.outputs.hadLabel }}
     steps:
       - uses: actions/checkout@v2
       - uses: actions/cache@v2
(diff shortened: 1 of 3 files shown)


samscott89:
That was quick! What's the output of the run supposed to be? Fail if there's a change?

dhatch:
> That was quick! What's the output of the run supposed to be? Fail if there's a change?

Right now yeah, just testing it. I'm going to change it to just comment the diff summary.

dhatch:
Docs changes ok

docs/content/python/new-roles/getting-started.md
```diff
@@ -219,6 +219,8 @@ route handlers
     linenos=true
     >}}
 
+I changed this!
```
dhatch, line 222: Remove before merging.

.github/workflows/docs-diff.yml
```diff
@@ -0,0 +1,190 @@
...
+          body: ${{ steps.get-comment-body.outputs.body }}
+          edit-mode: replace
+        if: ${{ steps.diff.outcome == 'failure' }}
+      - name: Look for resolving comment.
+        id: comment-check
+        uses: peter-evans/find-comment@v1
+        with:
+          issue-number: ${{ github.event.pull_request.number }}
+          body-includes: "Docs changes ok"
+          comment-author: ${{ github.event.pull_request.user.login }}
```
samscott89, line 186: This might be annoying if someone wants to help someone else get something merged?

.github/workflows/docs-diff.yml
```diff
@@ -0,0 +1,190 @@
...
+      - name: Comment
+        uses: peter-evans/create-or-update-comment@v1
+        with:
+          comment-id: ${{ steps.comment-bot-check.outputs.comment-id }}
+          issue-number: ${{ github.event.pull_request.number }}
+          body: ${{ steps.get-comment-body.outputs.body }}
+          edit-mode: replace
+        if: ${{ steps.diff.outcome == 'failure' }}
+      - name: Look for resolving comment.
+        id: comment-check
```
samscott89, line 181: Could do it with this if we want to do labels instead:

```
if: contains(github.event.pull_request.labels.*.name, 'some-defined-label')
```

.github/workflows/docs-diff.yml
```diff
@@ -0,0 +1,190 @@
...
+        if: ${{ steps.diff.outcome == 'failure' }}
+      - name: Comment
+        uses: peter-evans/create-or-update-comment@v1
+        with:
+          comment-id: ${{ steps.comment-bot-check.outputs.comment-id }}
+          issue-number: ${{ github.event.pull_request.number }}
+          body: ${{ steps.get-comment-body.outputs.body }}
+          edit-mode: replace
+        if: ${{ steps.diff.outcome == 'failure' }}
+      - name: Look for resolving comment.
```
samscott89, line 180: Should this come up front to shortcut the whole job if it's already done?

.github/workflows/docs-diff.yml
dhatch: In those cases I think we could force the merge with administrator privileges. They don't happen too frequently. If the PR author intended to make docs changes they can just leave this comment right away.

.github/workflows/docs-diff.yml
dhatch: Cool I think that is better I'll update when I have a chance

dhatch:
Tried updating this to use labels but I can't seem to get it to run in CI thanks to GitHub Actions.

PR #938, opened 2021-05-28
Polish syntax for roles configuration 

leina05:
Syntax changes to make:
- [x] Rename "perms" to "permissions" in the `resource` predicate configuration
- [x] Add namespaces to role names

PR checklist:

- [x] Added changelog entry.
- [x] Updated GitClub + docs: https://github.com/osohq/gitclub-sqlalchemy-flask-react/pull/4

### DIFF
--- a/docs/content/any/project/changelogs/NEXT.md
+++ b/docs/content/any/project/changelogs/NEXT.md
@@ -19,7 +19,7 @@ This release contains breaking changes. Be sure to follow migration steps
 before upgrading.
 {{% /callout %}}
 
-We've made some updates to the syntax of the `resource` predicate, used to configure resources + roles for sqlalchemy-oso-preview. The goal of these changes is to improve the readability of the config and make the roles
+We've made some updates to the syntax of the `resource` predicate, used to configure resources + roles for sqlalchemy-oso-preview. The goal of these changes is to improve the readability of the configuration and make the roles
 features more intuitive to use.
 
 #### Rename "perms" -> "permissions" in resource roles configuration
--- a/docs/content/any/project/changelogs/NEXT.md
+++ b/docs/content/any/project/changelogs/NEXT.md
@@ -1,9 +1,9 @@
 ---
-title: Release YYYY-MM-DD
-menuTitle: YYYY-MM-DD
+title: Release 2021-05-16
+menuTitle: 2021-05-16
 any: true
 description: >-
-  Changelog for Release YYYY-MM-DD (RELEASED_VERSIONS) containing new features,
+  Changelog for Release 2021-05-16 (sqlalchemy-oso-preview) containing new features,
   bug fixes, and more.
 draft: true
 ---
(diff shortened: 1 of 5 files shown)


languages/python/sqlalchemy-oso/tests/test_roles2.py
```diff
@@ -222,7 +222,7 @@ def test_empty_role(init_oso):
     # defining role with no permissions/implications throws an error
     oso, session = init_oso
     policy = """
-    resource(_type: Organization, "org", actions, roles) if
+    resource(Organization, "org", actions, roles) if
```
leina05, line 225: Wondering if we should switch the order here so that the name comes first. What do others think?

saolsen:
Are classes always constants you can pass to things? I know they are in python but I'm not sure about go or rust. Worried this might not be implementable there.

languages/python/sqlalchemy-oso/tests/test_roles2.py
saolsen: Yeah that seems like a good idea.

saolsen (commented):
Looks good except for the question about passing the class. Seems safer to keep it as a specializer unless we know it'll work in rust / go.

saolsen (approved):

docs/content/any/project/changelogs/NEXT.md
```diff
@@ -15,13 +15,44 @@ draft: true
...
+
+#### Remove the `_type:` syntax from the `resource` predicate
+
+The `resource` predicate used to take the resource type parameter as a specializer. Now the predicate takes this parameter as a direct argument, removing the need for the `resource(_type: ResourceType, ...)` syntax. Now, the syntax is `resource(ResourceType, ...)`.
+
+Below is an example `resource` predicate that reflects the above changes. For a more in-depth example, see our [documentation](TODO!!!!) and [example repository](TODO!!!).
+
+```polar
+    # Removed `_type: ` from first argument
+    resource(Organization, "org", actions, roles) if
```
gj, line 42: Did you consider using the resource type as the namespace instead of requiring folks to pass in a string as the second arg? E.g., `"Organization:member"` instead of `"org:member"` and `resource(Organization, actions, roles) if` instead of `resource(Organization, "org", actions, roles) if`?

saolsen (changes requested):
Need to verify passing the class works in all our languages to make this change.

leina05:
> Looks good except for the question about passing the class. Seems safer to keep it as a specializer unless we know it'll work in rust / go.

Decided to just move this change into a separate PR: #943

docs/content/any/project/changelogs/NEXT.md
leina05: Yeah I actually did suggest this as another potential improvement: https://www.notion.so/osohq/Reorienting-the-quickstart-and-product-around-the-Oso-Config-f6ca52da55d54cb38c865bac219febc6#ff8f371a0b8e42a5889b8f6d1dab1e1a

We decided not to do it to reduce based on the benefit vs. time to implement. I'm starting to think it might make more sense though, but I think we should take it as a follow-up if we have time. Especially since we still need to implement the config across all languages, which is way higher priority so I'd rather get this done and help with that.

leina05:
> Need to verify passing the class works in all our languages to make this change.

Took out this change, so should be good now.

docs/content/any/project/changelogs/NEXT.md
gj: okay sounds good 👍

saolsen (approved):

PR #941, opened 2021-06-02
postgres tests

saolsen:

### DIFF
--- a/languages/python/sqlalchemy-oso/requirements-dev.txt
+++ b/languages/python/sqlalchemy-oso/requirements-dev.txt
@@ -3,3 +3,4 @@ flake8==3.9.2
 mypy==0.812
 sqlalchemy-stubs==0.4
 tox==3.23.1
+psycopg2
--- a/languages/python/sqlalchemy-oso/requirements-dev.txt
+++ b/languages/python/sqlalchemy-oso/requirements-dev.txt
@@ -3,4 +3,3 @@ flake8==3.9.2
 mypy==0.812
 sqlalchemy-stubs==0.4
 tox==3.23.1
-psycopg2
--- a/languages/python/sqlalchemy-oso/sqlalchemy_oso/roles2.py
+++ b/languages/python/sqlalchemy-oso/sqlalchemy_oso/roles2.py
@@ -34,7 +34,12 @@ def get_pk(model):
 
 
 def role_allow_query(
-    id_query, type_query, child_types, resource_id_field, has_relationships
+    id_query,
+    type_query,
+    child_types,
+    resource_id_field,
+    resource_id_type,
+    has_relationships,
 ):
     child_types = ",".join([f"'{ct}'" for ct in child_types])
 
@@ -53,7 +58,7 @@ def role_allow_query(
         with resources as (
             with recursive resources (id, type) as (
                 select
-                    {resource_id_field} as id,
+                    cast({resource_id_field} as {resource_id_type}) as id,
                     :resource_type as type
                 {recur}
             ) select * from resources
@@ -102,7 +107,12 @@ def role_allow_query(
 
 
 def user_in_role_query(
-    id_query, type_query, child_types, resource_id_field, has_relationships
+    id_query,
+    type_query,
+    child_types,
+    resource_id_field,
+    resource_id_type,
+    has_relationships,
 ):
     child_types = ",".join([f"'{ct}'" for ct in child_types])
 
@@ -121,7 +131,7 @@ def user_in_role_query(
         with resources as (
             with recursive resources (id, type) as (
                 select
-                    {resource_id_field} as id,
+                    cast({resource_id_field} as {resource_id_type}) as id,
                     :resource_type as type
                 {recur}
             ) select * from resources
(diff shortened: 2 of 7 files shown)


languages/python/sqlalchemy-oso/requirements-dev.txt
```diff
@@ -2,4 +2,4 @@ black==21.5b0
 flake8==3.9.2
 mypy==0.812
 sqlalchemy-stubs==0.4
-tox==3.23.1
+tox==3.23.1
```
gj, line 5: what in the heck changed here? I'm stumped

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles2.py
```diff
@@ -53,7 +58,7 @@ def role_allow_query(
         with resources as (
             with recursive resources (id, type) as (
                 select
-                    {resource_id_field} as id,
+                    cast({resource_id_field} as {resource_id_type}) as id,
```
gj, line 61: weren't we trying to avoid `cast()`s b/c they'd skip indices?

languages/python/sqlalchemy-oso/tests/test_roles2.py
```diff
@@ -150,7 +198,7 @@ def sample_data(init_oso, Organization, Repository, User, Issue):
 # - Passing a bad declarative_base to OsoRoles raises an exception
 
 
-def test_oso_roles_init(auth_sessionmaker, Base, User):
+def test_oso_roles_init(engine, auth_sessionmaker, Base, User):
```
gj, line 201: how was this passing before?

languages/python/sqlalchemy-oso/tests/test_roles2.py
```diff
@@ -299,6 +346,8 @@ def test_resource_with_roles_no_actions(init_oso, sample_data):
     oso.roles.assign_role(leina, osohq, "member", session)
     oso.roles.assign_role(steve, oso_repo, "repo_read", session)
 
+    session.commit()
```
gj, line 349: Why are all these `session.commit()` calls suddenly required?

languages/python/sqlalchemy-oso/tox.ini
```diff
@@ -3,6 +3,7 @@ skip_missing_interpreters=true
 envlist = {py3,pypy3}-sqlalchemy{13,14}-{earliest,latest}
 
 [testenv]
+passenv = *
```
gj, line 6: assuming we're doing this for the new postgres variables, how about this?

```suggestion
passenv = POSTGRES_*
```

https://tox.readthedocs.io/en/latest/config.html?highlight=passenv#conf-passenv

would feel better keeping the test envs as clean & isolated as possible

gj (changes requested):
few questions but mostly LGTM

languages/python/sqlalchemy-oso/requirements-dev.txt
saolsen: lol

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles2.py
saolsen: Casts were bad when we were casting the id column on a resource to be a string (if it was an int or something else it'd have to table scan instead of use the index).
In this case we're just casting the input argument to the query to match the id column type, so it's good.
This was failing in some cases when the arg came in as a `text` and the column was a `varchar`. Sqlite didn't care but postgres is stricter.

languages/python/sqlalchemy-oso/tests/test_roles2.py
saolsen: lol I don't really know

languages/python/sqlalchemy-oso/tests/test_roles2.py
saolsen: Because we change the roles with one session and then the is_allowed queries use new sessions. I think it makes sense that these have to be committed, I don't really know why it works on sqlite without committing them, maybe just some odd behavior of the in memory sqlite.

languages/python/sqlalchemy-oso/tox.ini
saolsen: oh cool

languages/python/sqlalchemy-oso/tests/test_roles2.py
saolsen: I think maybe because sqlalchemy will take a method that returns an engine as the engine so it was just using the fixture method directly to create an engine? Wild.

languages/python/sqlalchemy-oso/tests/test_roles2.py
saolsen: We could just not pass sessions to the `assign_role` calls and that would work too.

languages/python/sqlalchemy-oso/sqlalchemy_oso/roles2.py
gj: classic pg

languages/python/sqlalchemy-oso/tests/test_roles2.py
gj: > We could just not pass sessions to the `assign_role` calls and that would work too.

up 2 u

gj: > then the is_allowed queries use new sessions.

ah not that `is_allowed()` uses a session but that it's operating over uncommitted data, right? Wonder if `.flush()` would do the trick here, too.

> I don't really know why it works on sqlite without committing them, maybe just some odd behavior of the in memory sqlite.

yeah this seems real suspect to me — I'd bet it's some autocommit-on-access behavior, but that seems like it'd be a SQLAlchemy-level configuration instead of database specific.

gj (approved):
feel like the `session.commit()` calls are cluttering up the tests a bit and going with implicit sessions might be cleaner, but don't care enough to block shippin'

PR #948, opened 2021-06-04
clean up golang memory leaks

### LINKED ISSUE #944: Memory Leak in Golang
Hi OsoHQ,

Been enjoying the library, and I hope to use it for an authorization service, however I believe I found a memory leak in the Golang library.

## Setup

I was able to reproduce using the `oso-go-quickstart` example. I [forked that repo](https://github.com/bhb603/oso-go-quickstart) and the only changes I made were:
- upgraded to the latest oso  `v0.11.3 ->  v0.12.3` (https://github.com/bhb603/oso-go-quickstart/blob/main/go.mod#L5)
- added a policy:
    ```
    allow(actor: String, "GET", expense: Expense) if  
            expense.SubmittedBy = actor;
    ```
- [Dockerized](https://github.com/bhb603/oso-go-quickstart/blob/main/Dockerfile) the app and deployed it to a kubernetes cluster with [this simple deployment](https://github.com/bhb603/oso-go-quickstart/blob/main/k8s/deployment.yaml) (my ultimate use case would be to deploy in kubernetes, which is why I replicated that here)

## Testing

I have some load-testing tooling which can generate requests against the service. Since this quickstart example only has one endpoint, it just requests:
```
curl -H "user: alice@example.com" http://oso-go-quickstart.oso:5050/expenses/1
```
Which would check `oso.IsAllowed(User("alice@example.com"), "GET", expense)` and return:
```
HTTP/1.1 200 OK
Content-Length: 40
Content-Type: text/plain; charset=utf-8

Expense{{500 coffee alice@example.com}}
```

## Results

Memory Usage increases linearly when the service is under load, until it reaches the limit I set for the container (256M) at which point Kubernetes kills the container (OOMKIlled) and it restarts, producing the saw tooth graph below.

The rate of memory increase is proportional to the load (reqs per second), i.e. it increases ~2.5x faster when I was hitting it at 25rps vs 10rps, and flattens out with 0rps.

Memory usage under load:
Containers___Datadog

More Metrics:
image

saolsen:
Clean up golang memory leaks.
Wasn't calling query_free in all cases, wasn't calling string_free in some cases and wasn't freeing Cstrings allocated from go to pass through ffi calls.

fixes #944

### DIFF
--- a/languages/go/internal/ffi/ffi.go
+++ b/languages/go/internal/ffi/ffi.go
@@ -45,6 +45,7 @@ func (p *PolarFfi) delete() {
 func getError() error {
 	err := C.polar_get_error()
 	errStr := C.GoString(err)
+	C.string_free(err)
 	var polarError errors.FormattedPolarError
 	jsonErr := json.Unmarshal([]byte(errStr), &polarError)
 	if jsonErr != nil {
@@ -68,8 +69,10 @@ func processMessages(i ffiInterface) {
 			return
 		}
 		message := C.GoString(msgPtr)
+		C.string_free(msgPtr)
 		var messageStruct types.Message
 		err := json.Unmarshal([]byte(message), &messageStruct)
+
 		if err != nil {
 			panic(err)
 		}
@@ -96,9 +99,11 @@ func (p PolarFfi) NewId() (uint64, error) {
 
 func (p PolarFfi) Load(s string, filename *string) error {
 	cString := C.CString(s)
+	defer C.free(unsafe.Pointer(cString))
 	var cFilename *C.char
 	if filename != nil {
 		cFilename = C.CString(*filename)
+		defer C.free(unsafe.Pointer(cFilename))
 	}
 	result := C.polar_load(p.ptr, cString, cFilename)
 	processMessages(p)
@@ -119,6 +124,7 @@ func (p PolarFfi) ClearRules() error {
 
 func (p PolarFfi) NewQueryFromStr(queryStr string) (*QueryFfi, error) {
 	cs := C.CString(queryStr)
+	defer C.free(unsafe.Pointer(cs))
 	result := C.polar_new_query(p.ptr, cs, 0)
 	processMessages(p)
 	if result == nil {
@@ -131,6 +137,8 @@ func (p PolarFfi) NewQueryFromTerm(queryTerm types.Term) (*QueryFfi, error) {
 	json, err := ffiSerialize(queryTerm)
 	if err != nil {
 		return nil, err
+	} else {
+		defer C.free(unsafe.Pointer(json))
 	}
 	result := C.polar_new_query_from_term(p.ptr, json, 0)
 	processMessages(p)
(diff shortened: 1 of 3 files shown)


languages/go/internal/ffi/ffi.go
```diff
@@ -45,6 +45,7 @@ func (p *PolarFfi) delete() {
 func getError() error {
 	err := C.polar_get_error()
 	errStr := C.GoString(err)
```
samscott89, line 47: Minor: could move all go string construction into a method so we don't forget it in the future.

languages/go/internal/ffi/ffi.go
```diff
@@ -131,6 +137,8 @@ func (p PolarFfi) NewQueryFromTerm(queryTerm types.Term) (*QueryFfi, error) {
 	json, err := ffiSerialize(queryTerm)
 	if err != nil {
 		return nil, err
+	} else {
+		defer C.free(unsafe.Pointer(json))
```
samscott89, line 141: It's actually fine to call free on a nil pointer. So I would prefer we keep all the frees right next to where we get the string back. Will be easier to scan

samscott89: And avoid the else blocks that are a _little_ hard to follow.

languages/go/internal/ffi/ffi.go
```diff
@@ -152,10 +160,12 @@ func (p PolarFfi) NextInlineQuery() (*QueryFfi, error) {
 
 func (p PolarFfi) RegisterConstant(term types.Term, name string) error {
 	cName := C.CString(name)
+	defer C.free(unsafe.Pointer(cName))
 	cTerm, err := ffiSerialize(term)
 	if err != nil {
-		defer C.free(unsafe.Pointer(cName))
 		return err
+	} else {
```
samscott89, line 167: ditto

languages/go/internal/ffi/ffi.go
```diff
@@ -191,6 +201,8 @@ func (q QueryFfi) CallResult(callID uint64, term *types.Term) error {
 		s, err = ffiSerialize(term)
 		if err != nil {
 			return err
+		} else {
```
samscott89, line 204: ditto

languages/go/query.go
```diff
@@ -128,6 +132,7 @@ func (q *Query) Next() (*map[string]interface{}, error) {
 			return nil, fmt.Errorf("unexpected query event: %v", ev)
```
samscott89, line 132: Cleanup here too? Although memory leak wouldnt be the worst issue if this was hit a lot...

languages/go/oso.go
```diff
@@ -142,6 +142,7 @@ func (o Oso) IsAllowed(actor interface{}, action interface{}, resource interface
 	if err != nil {
 		return false, err
 	} else if results != nil {
+		defer query.Cleanup()
```
samscott89, line 145: It's very non-obvious why this is needed. Can we add a comment here and to the `Next` method stating that if you don't exhaust the iterator, you need to call `Cleanup` manually?

samscott89 (changes requested):
Looks good. I have a few suggestions to help avoid this in the future.

languages/go/query.go
saolsen: Yeah this is really like a panic case. Shouldn't ever happen.

samscott89 (approved):

PR #963, opened 2021-06-10
Api doc linking

samraper:
There are a couple of limitations here:

1) each language api reference has a different url scheme, so we will have to write one of these for each (not that big of a deal, just a little messy)
2) we sometimes use a shortcode to look up the name of a method for a specific language (basically just casing differences). You can't pass a shortcode as a parameter to another shortcode, but you can provide it as an inner shortcode. Just makes the  API a little messy

### DIFF
--- a/docs/layouts/shortcodes/apiDeepLink.html
+++ b/docs/layouts/shortcodes/apiDeepLink.html
@@ -1,19 +1,37 @@
-{{- /* apiDeepLink $class */ -}}
+{{- /* apiDeepLink $class $label $pythonFramework */ -}}
 {{- /* Makes a deep link to the currently selected language's documentation for a given class */ -}}
 {{- /* To link to a specific method of the class, supply it as an inner shortcode or string */ -}}
 {{- $class := (.Get 0) -}}
-{{- if eq $.Page.Language.Lang "node" -}}
-{{- $apiBase := printf "%s%s/%s" $.Site.BaseURL $.Page.Language "reference/api/classes" -}}
-{{- $classLink := printf "%[1]s/%[2]s.%[2]s-1.html" $apiBase (lower $class) -}}
-<code style="display: inline-block; line-height: initial;">
-  {{- if eq (len .Inner) 0 -}}
-  <a href="{{- $classLink -}}">
-    {{- $class -}}
-  </a>
-  {{- else -}}
-  <a href="{{- (printf "%s#%s" $classLink (lower .Inner)) | safeURL -}}">
-    {{- $class -}}.{{- .Inner -}}()
-  </a>
-  {{- end -}}
-</code>
+{{- $label := (.Get 1) -}}
+{{- $pythonFramework := (.Get 2) -}}
+{{- $href := "" -}}
+{{- $genText := "" -}}
+{{- $baseUrl := printf "%s%s" $.Site.BaseURL $.Page.Language -}}
+{{- if eq (len .Inner) 0 -}}
+  {{- $genText = $class -}}
+{{- else -}}
+  {{- $genText = printf "%s.%s()" $class .Inner -}}
 {{- end -}}
+{{- if eq $.Page.Language.Lang "node" -}}
+  {{- $apiBase := printf "%s/%s" $baseUrl "reference/api/classes" -}}
+  {{- $classLink := printf "%[1]s/%[2]s.%[2]s-1.html" $apiBase (lower $class) -}}
+  {{- if eq (len .Inner) 0 -}}
+    {{- $href = $classLink -}}
+  {{- else -}}
+    {{- $href = (printf "%s#%s" $classLink (lower .Inner)) -}}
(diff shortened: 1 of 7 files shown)


dhatch:
Overall seems like a pretty reasonable approach. Is there any input you need on this? How would this work for languages like Python where there are multiple index pages into the API reference?

samraper:
Updated to include the rest of the languages. Including deep links to all API documentation for python began feeling very very brittle and hacky as I needed to be able to handle top level classes, top level functions, methods on classes, modules (sqlalchemy_oso.roles has a different hash link than the other modules), all in the different frameworks and I was getting into if statement hell. I'm worried that the juice isn't worth the squeeze and it will just eventually break when something new is added. How do you feel about this approach where we just link to the top level documentation?

dhatch (commented):
I think this is fine for now. In the future it would be great to be able to link to methods, functions, etc. I'm not sure if this would help, but we could consider having several shortcakes for each type of link, somewhat like how [sphinx cross references data](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#python-roles)

Could you add a documentation string to `apiDeepLink`?

dhatch:
@gj I didn't get a chance to finish this review. Would you be able to help with it?

gj:
> @gj I didn't get a chance to finish this review. Would you be able to help with it?

yeah

docs/content/python/new-roles/examples.md
```diff
@@ -38,8 +38,9 @@ This defines two roles on repository, `"reader"` and `"writer"`.
 This allows us to assign users directly to repositories without
 giving them a role in the entire organization.
 
-Now, we can use `OsoRoles.assign_role` to assign role on
-repositories in addition to organizations.
+Now, we can use
+{{% apiDeepLink "OsoRoles" "OsoRoles.assign_role" "sqlalchemy" %}}assign_role{{% /apiDeepLink %}}
```
gj, line 42: This seems like the only example where we're passing more than one param to `apiDeepLink`, but I'm not totally sure why we need the middle param (`$label`) here. Won't `$genText` be equal to `OsoRoles.assign_role` as well?

Could you provide an example of where we would need to override `$genText` with `$label`?

docs/layouts/shortcodes/apiDeepLink.html
```diff
@@ -0,0 +1,37 @@
+{{- /* apiDeepLink $class $label $pythonFramework */ -}}
+{{- /* Makes a deep link to the currently selected language's documentation for a given class */ -}}
+{{- /* To link to a specific method of the class, supply it as an inner shortcode or string */ -}}
+{{- $class := (.Get 0) -}}
+{{- $label := (.Get 1) -}}
+{{- $pythonFramework := (.Get 2) -}}
```
gj, line 6: I think making these named params might make it a bit easier to figure out what we're linking to at the call site. Example from another shortcode:

https://github.com/osohq/oso/blob/6df569e81715627444fb260337162ea1eba733ae/docs/layouts/shortcodes/literalInclude.html#L50

docs/content/python/new-roles/examples.md
samraper: There are some cases in the sqlalchemy and django documentation where we need to differentiate which package something is coming from (eg `oso.Oso.foo` vs `sqlalchemy_oso.SqlalchemyOso.foo`) and I wanted to add a manual override for those cases

docs/layouts/shortcodes/apiDeepLink.html
samraper: agree! will do

docs/content/python/new-roles/examples.md
gj: Okay I think I follow, but would you mind doing one of those changes on this PR just so I can see what it looks like in the wild?

dhatch:
Screen Shot 2021-06-28 at 2 52 17 PM

Seems to render a bit off on http://localhost:1313/python/guides/data_access/sqlalchemy.html

dhatch:
> How do you feel about this approach where we just link to the top level documentation?

After trying this out I think we should try to get this to link to the right documentation string in Python. Most of our target users right now are in Python, and I find it pretty non-intuitive to not be directed to the right reference. I have some rough ideas, but maybe we could require the user to specify more (for example the fully qualified python path: `sqlalchemy_oso.roles.assign_role`) in order to generate the correct link.

dhatch (changes requested):
Requesting changes to include python method linking, see comment above.

samraper:
Ok, yeah it might take some more manual intervention, let me give it a try

samraper:
Can now handle any combination of module, class, and function. Hard coding the hash link would make it really clunky if we ever wanted to deep link to other languages in the future. So although this isn't very clean, I think it's worthwhile

dhatch (approved):
lgtm

PR #965, opened 2021-06-12
Make Query an enumerable to avoid running queries on fibers.

samscott89:
**For Ruby**

Minimal test case:

```ruby
require 'oso'

oso = Oso.new

class Foo
    def self.tv(name)
        Thread.current[name]
    end
end

oso.register_class(Foo)

Thread.current[:foo] = "bar"
oso.load_str('?= Foo.tv("foo") = "bar";')
```

Before this would fail, because the query runs on a fiber and thread variables are not accessible (without using `Thread.current.thread_variable_set` instead).

PR checklist:

- [x] Added changelog entry.

### DIFF
--- a/languages/ruby/lib/oso/polar/query.rb
+++ b/languages/ruby/lib/oso/polar/query.rb
@@ -6,8 +6,7 @@ module Oso
   module Polar
     # A single Polar query.
     class Query # rubocop:disable Metrics/ClassLength
-      # @return [Enumerator]
-      attr_reader :results
+      include Enumerable
 
       # @param ffi_query [FFI::Query]
       # @param host [Oso::Polar::Host]
@@ -15,7 +14,6 @@ module Oso
         @calls = {}
         @ffi_query = ffi_query
         @host = host
-        @results = start
       end
 
       private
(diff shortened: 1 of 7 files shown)


languages/ruby/spec/oso/polar/polar_spec.rb
```diff
@@ -476,13 +476,24 @@ def x
...
         subject.register_class(TestClass)
       end
 
       it 'accepts positional and keyword args' do
         expect(qvar(subject, 'x = (new TestClass()).my_method(1, 2, four: 4, three: 3)', 'x')).to eq([[1, 2, 3, 4]])
       end
+
+      it 'has access to the thread context' do
+        Thread.current["foo"] = "bar"
+        # you need to use load string here or it might not
```
samscott89, line 493: I don't fully get this, but if you use `expect(qvar(...` like the other test cases the bad behaviour doesn't get triggered.

languages/ruby/lib/oso/polar/query.rb
```diff
@@ -120,61 +118,59 @@ def handle_make_external(data) # rubocop:disable Metrics/AbcSize
...
-            else
-              raise "Unhandled event: #{JSON.dump(event.inspect)}"
+      def each # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/MethodLength, Metrics/PerceivedComplexity
+        loop do # rubocop:disable Metrics/BlockLength
+          event = ffi_query.next_event
+          case event.kind
+          when 'Done'
+            break
+          when 'Result'
+            yield event.data['bindings'].transform_values { |v| host.to_ruby(v) }
```
samscott89, line 128: I think?
```suggestion
            return event.data['bindings'].transform_values { |v| host.to_ruby(v) }
```

languages/ruby/spec/oso/polar/polar_spec.rb
samscott89: I'm guessing if you return the enumerator in the test, then Ruby figures out it doesn't need to create a new fiber to evaluate the query, so it all happens on the right thread.

gj (approved):
kinda want to wrap `.first.nil?` in a method, but to be fair 98% of the calls are in the parity tests, which weren't exactly a paragon of readability to begin with

gj (changes requested):
changelog first

good work tho 🦾

gj (approved):

PR #966, opened 2021-06-13
Java: update maven-javadoc-plugin to 3.3.0

joshrotenberg:
Intellij is showing that the `additionalOptions` configuration option for the `maven-javadoc-plugin` isn't supported:

Screen Shot 2021-06-12 at 17 49 20 PM

According to the [documentation](https://maven.apache.org/plugins/maven-javadoc-plugin/javadoc-mojo.html#additionalOptions), this option is supported in `3.0.0` and greater, but the `pom.xml` currently only has `2.9.1`. Update to the latest and greatest.

FWIW a diff of `target/site` before and after this change showed no changes.

PR checklist:

- [ ] Added changelog entry.

### DIFF
--- a/languages/java/oso/pom.xml
+++ b/languages/java/oso/pom.xml
@@ -120,7 +120,7 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-javadoc-plugin</artifactId>
-        <version>2.9.1</version>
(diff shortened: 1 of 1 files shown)


dhatch:
Thanks for this update @joshrotenberg. We are doing some other work on CI right now, and will test this alongside that before we merge.

PR #967, opened 2021-06-13
Java: Create OsoTest and add a test for isAllowed

joshrotenberg:
The top level `test/Test.java` does call `isAllowed` but this is the beginning of bringing the library tests in line with some of the other language implementations.

PR checklist:

- [ ] Added changelog entry.

### DIFF
--- a/languages/java/oso/src/test/java/com/osohq/oso/OsoTest.java
+++ b/languages/java/oso/src/test/java/com/osohq/oso/OsoTest.java
@@ -1,0 +1,110 @@
+package com.osohq.oso;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.net.URL;
+import java.util.List;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
+
+public class OsoTest {
+  protected Oso o;
+
+  public static class Actor {
+    public String name;
+
+    public Actor(String name) {
+      this.name = name;
+    }
+
+    public List<Company> companies() {
+      return List.of(new Company(1));
+    }
+  }
+
+  public static class Widget {
+    public int id;
+
+    public Widget(int id) {
+      this.id = id;
+    }
+  }
+
+  public static class Company {
+    public int id;
+
+    public Company(int id) {
+      this.id = id;
+    }
+
+    public String role(Actor a) {
+      if (a.name.equals("president")) {
+        return "admin";
+      }
+
+      return "guest";
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+      return obj instanceof OsoTest.Company && ((OsoTest.Company) obj).id == this.id;
+    }
+
+    @Override
+    public int hashCode() {
+      return this.id;
+    }
+  }
+
+  @BeforeEach
+  public void setUp() throws Exception {
+    try {
+      URL testOso = getClass().getClassLoader().getResource("test_oso.polar");
+
+      o = new Oso();
+      o.registerClass(Actor.class, "Actor");
+      o.registerClass(Widget.class, "Widget");
+      o.registerClass(Company.class, "Company");
+
+      o.loadFile(testOso.getPath());
+    } catch (Exception e) {
+      throw new Error(e);
+    }
+  }
+
+  @Test
+  public void testIsAllowed() throws Exception {
+    Actor guest = new Actor("guest");
+    Widget resource1 = new Widget(1);
+    assertTrue(o.isAllowed(guest, "get", resource1));
+
+    Actor president = new Actor("president");
+    Company company = new Company(1);
+    assertTrue(o.isAllowed(president, "create", company));
+  }
+
+  @Test
+  public void testFail() throws Exception {
+    Actor guest = new Actor("guest");
+    Widget widget = new Widget(1);
+    assertFalse(o.isAllowed(guest, "not_allowed", widget));
+  }
+
+  @Test
+  public void testInstanceFromExternalCall() throws Exception {
+    Company company = new Company(1);
+    Actor guest = new Actor("guest");
+    assertTrue(o.isAllowed(guest, "frob", company));
+  }
+
+  @Test
+  @Disabled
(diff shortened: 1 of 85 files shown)


dhatch:
Hey @joshrotenberg! Thanks for the PR, getting the tests more in line with other libraries is a great idea! Is there anything else you wanted to get to before we review (I noticed it's in draft)?

joshrotenberg:
> Hey @joshrotenberg! Thanks for the PR, getting the tests more in line with other libraries is a great idea! Is there anything else you wanted to get to before we review (I noticed it's in draft)?

I'm working on ~parity with the tests for `is_allowed` in the Rust tests (mainly because I know them a bit better than the others). I'll set it to ready for review when those are in, and then I have some other stuff I'd like to do in following PRs if that's cool.

joshrotenberg:
Also, that failure in the `check` part of the workflow is a weird one. Its failing on the `Test.java` which hasn't changed and I don't think any changes I've made should affect it. I'll dig into it and see.

languages/java/oso/src/test/java/com/osohq/oso/OsoTest.java
```diff
@@ -0,0 +1,110 @@
...
+  public void testFail() throws Exception {
+    Actor guest = new Actor("guest");
+    Widget widget = new Widget(1);
+    assertFalse(o.isAllowed(guest, "not_allowed", widget));
+  }
+
+  @Test
+  public void testInstanceFromExternalCall() throws Exception {
+    Company company = new Company(1);
+    Actor guest = new Actor("guest");
```
samscott89, line 98: I think this should just be the string `"guest"` so it hits the path for creating an instance from within Polar.

languages/java/oso/src/test/java/com/osohq/oso/OsoTest.java
```diff
@@ -0,0 +1,110 @@
...
+    assertTrue(o.isAllowed(guest, "frob", company));
+  }
+
+  @Test
+  @Disabled
+  public void testAllowModel() throws Exception {
+    Actor auditor = new Actor("auditor");
+
+    assertTrue(o.isAllowed(auditor, "list", new Company(1)));
+    assertFalse(o.isAllowed(auditor, "list", new Widget(1)));
```
samscott89, line 108:
```suggestion
    assertTrue(o.isAllowed(auditor, "list", Company.class));
    assertFalse(o.isAllowed(auditor, "list", Widget.class));
```

These should be testing whether you are allowed to "list companies". That would be why the test is not passing.

languages/java/oso/src/test/java/com/osohq/oso/OsoTest.java
```diff
@@ -0,0 +1,110 @@
...
+      if (a.name.equals("president")) {
+        return "admin";
+      }
+
+      return "guest";
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+      return obj instanceof OsoTest.Company && ((OsoTest.Company) obj).id == this.id;
```
samscott89, line 52: Does this pass if `equals` is only defined for `Company`? That would be how I think most people would want to write it.

samscott89 (changes requested):
Thank you for adding these! Much needed.

I have a couple of comments/suggestions, but on the whole looks fantastic.

languages/java/oso/src/test/java/com/osohq/oso/OsoTest.java
```diff
@@ -0,0 +1,110 @@
...
+
+  @Test
+  public void testInstanceFromExternalCall() throws Exception {
+    Company company = new Company(1);
+    Actor guest = new Actor("guest");
+    assertTrue(o.isAllowed(guest, "frob", company));
+  }
+
+  @Test
+  @Disabled
```
samscott89, line 103:
```suggestion
```

(Assuming my suggestion below fixes the test)

languages/java/oso/src/test/java/com/osohq/oso/OsoTest.java
joshrotenberg: `assertTrue(o.isAllowed("guest", "frob", company));`

gives me 

`[ERROR]   OsoTest.testInstanceFromExternalCall:99 » Instantiation constructor on class ...`

```java
com.osohq.oso.Exceptions$InstantiationError: constructor on class `Actor`
        at com.osohq.oso.Query.nextResult(Query.java:177)
        at com.osohq.oso.Query.<init>(Query.java:32)
        at com.osohq.oso.Polar.queryRule(Polar.java:111)
        at com.osohq.oso.Oso.isAllowed(Oso.java:27)
        at com.osohq.oso.OsoTest.testInstanceFromExternalCall(OsoTest.java:99)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:78)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.base/java.lang.reflect.Method.invoke(Method.java:567)
```

but then maybe the Rust test is wrong (which I modeled these on) since it also passes in an `Actor` struct and not just the bare string (unless I'm misunderstanding you):

https://github.com/osohq/oso/blob/d5ad7b57a3b1863361990c4ba0d2b29ad85ee0bc/languages/rust/oso/tests/test_oso.rs#L142

languages/java/oso/src/test/java/com/osohq/oso/OsoTest.java
joshrotenberg: Do you mean if its only defined for `Company` and not the other two classes? Seems to, yeah. 😸

languages/java/oso/src/test/java/com/osohq/oso/OsoTest.java
samscott89: Hmm, will take a look

samscott89: Okay, finally got around to this. The Rust test was also missing a check from the original Python test. Added it to both.

samscott89 (approved):
Okay, this looks good to go! Thank you.

PR #969, opened 2021-06-14
Python config validation code

saolsen:
We can't validate as much as we could before.

* Relationship stuff is much harder because we don't restrict the form of those rules the way we had to before.
* Anything checking valid sqlalchemy fields doesn't apply anymore.

This turns back on the tests we can validate.

### DIFF
--- a/languages/python/oso/oso/oso.py
+++ b/languages/python/oso/oso/oso.py
@@ -46,6 +46,17 @@ class Oso(Polar):
     def enable_roles(self):
         self.load_file(Path(__file__).parent / "roles.polar")
 
+    # @TODO: No good place to automatically call this as it has to be called
+    # after the policy is loaded which might happen before calling enable_roles
+    # and might happen after calling enable_roles. Seperate method for now.
+    def validate_config(self):
+        try:
+            self.load_file(Path(__file__).parent / "validate.polar")
+        except exceptions.InlineQueryFailedError as e:
+            # @TODO: No good way to turn an inline query error into what we want
+            # it to be. Need a better way to pass a failure string out of polar.
+            raise exceptions.OsoError(f"Config Error: {e.message}")
+
     def get_allowed_actions(self, actor, resource, allow_wildcard=False) -> list:
         """Determine the actions ``actor`` is allowed to take on ``resource``.
 
--- a/languages/python/oso/oso/oso.py
+++ b/languages/python/oso/oso/oso.py
@@ -5,6 +5,7 @@ __version__ = "0.12.4"
 import os
 from pathlib import Path
 from polar import Polar, Variable, exceptions
+from .config import validate_config
 
 
 class Oso(Polar):
@@ -45,17 +46,6 @@ class Oso(Polar):
 
     def enable_roles(self):
         self.load_file(Path(__file__).parent / "roles.polar")
-
-    # @TODO: No good place to automatically call this as it has to be called
-    # after the policy is loaded which might happen before calling enable_roles
-    # and might happen after calling enable_roles. Seperate method for now.
-    def validate_config(self):
-        try:
-            self.load_file(Path(__file__).parent / "validate.polar")
-        except exceptions.InlineQueryFailedError as e:
-            # @TODO: No good way to turn an inline query error into what we want
-            # it to be. Need a better way to pass a failure string out of polar.
-            raise exceptions.OsoError(f"Config Error: {e.message}")
 
     def get_allowed_actions(self, actor, resource, allow_wildcard=False) -> list:
         """Determine the actions ``actor`` is allowed to take on ``resource``.
@@ -101,3 +91,6 @@ class Oso(Polar):
                 "Polar tracing enabled. Get help with "
                 + "traces from our engineering team: https://help.osohq.com/trace"
             )
+
+    def validate_config(self):
+        validate_config(self)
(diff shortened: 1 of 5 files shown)


languages/python/oso/tests/test_polar_roles.py
```diff
@@ -338,10 +332,10 @@ def test_nested_dot_relationship(init_oso):
     oso.load_str(policy)
 
     with pytest.raises(OsoError):
-        oso.roles.synchronize_data()
+        oso.validate_config()
 
 
-@pytest.mark.skip("TODO: validation")
+@pytest.mark.skip("can't detect, not sqlalchemy")
```
gj, line 338: let's either kill this or update the message to mirror the reason we can't detect it
```suggestion
@pytest.mark.skip("Cannot generically walk fields of a Python class")
```

languages/python/oso/tests/test_polar_roles.py
```diff
@@ -367,10 +361,10 @@ def test_bad_relationship_lookup(init_oso):
     oso.load_str(policy)
 
     with pytest.raises(OsoError):
-        oso.roles.synchronize_data()
+        oso.validate_config()
 
 
-@pytest.mark.skip("TODO: validation")
+@pytest.mark.skip("Can't really test relationship types")
```
gj, line 367:
```suggestion
@pytest.mark.skip("Cannot generically walk fields of a Python class")
```

languages/python/oso/tests/test_polar_roles.py
```diff
@@ -560,10 +547,10 @@ def test_undeclared_role(init_oso):
     """
     oso.load_str(policy)
     with pytest.raises(OsoError):
-        oso.roles.synchronize_data()
+        oso.validate_config()
 
 
-@pytest.mark.skip("TODO: validation")
+@pytest.mark.skip("Can't really test relationship types")
```
gj, line 553: Without worrying about validating the dot lookup portion of the `parent/2` predicates, wouldn't it be possible for us to validate that there exist a chain of `parent/2` predicates that — if the dot lookups succeed — span from the implied namespace (`"repo"`) to the parent resource (`Org`)? That wouldn't be perfect, but it'd at least offer some level of assurance. Happy to book it as a follow-up if you don't think the payoff would be worth doing at the moment.

languages/python/oso/tests/test_polar_roles.py
```diff
@@ -591,10 +578,10 @@ def test_role_implication_without_relationship(init_oso):
...
```
gj, line 584: same question as above re: validating the presence of potentially valid `parent/2` predicates even if we can't validate the crucial dot lookup that validates the relationship

languages/python/oso/tests/test_polar_roles.py
```diff
@@ -1021,10 +998,9 @@ def test_grandparent_child_role_perm(init_oso, sample_data):
         role_allow(actor, action, resource);
     """
     oso.load_str(policy)
+    oso.validate_config()
     # TODO: big red button to reset roles policy?
     # oso.roles.config = None
```
gj, line 1003: now that we aren't persisting config on the Oso instance we can probably rm this pair of lines (and the other places it appears in this test suite)
```suggestion
```

ETA: will push a commit for this

languages/python/oso/oso/oso.py
```diff
@@ -90,3 +91,6 @@ def _print_polar_log_message(self):
                 "Polar tracing enabled. Get help with "
                 + "traces from our engineering team: https://help.osohq.com/trace"
             )
+
+    def validate_config(self):
```
gj, line 95: WDYT about calling this `validate_roles_config()` (or something similar) that indicates that it's only validating the 'Polar roles' predicates instead of all of the user's Polar code?

languages/python/oso/tests/test_polar_roles.py
saolsen: The way the constraints come out right now there's no way to really know the second arg has to be an "Org", the constraints are on the dot looked up var instead of the output var and we can't assume this rule even uses a dot lookup as they could call an external method or do anything else to get the org.

languages/python/oso/tests/test_polar_roles.py
saolsen: Same as above, without a specializer on the output var instead of an isa on the looked up var there's no real way to know in general what you'll get out. Maybe able to do after we change constraints output some.

languages/python/oso/oso/oso.py
saolsen: sure

gj:
For posterity: we can't really validate `actor_role/2` b/c we don't have a real actor to provide at validation time.

Would it be worth checking that `actor_role/2` has at least been defined?

leina05:
> For posterity: we can't really validate `actor_role/2` b/c we don't have a real actor to provide at validation time.
> 
> Would it be worth checking that `actor_role/2` has at least been defined?

I think so. Same goes for `parent_child` if we encounter anything in the config that requires a relationship. And ideally same for `role_allows` (trickier since it would be called within an allow rule)--although for that I think it would be best to issue a warning because it could be commented for testing reasons. All could probably be taken as follow-ups if we have time.

gj:
> I think so. Same goes for `parent_child` if we encounter anything in the config that requires a relationship. And ideally same for `role_allows` (trickier since it would be called within an allow rule)--although for that I think it would be best to issue a warning because it could be commented for testing reasons. All could probably be taken as follow-ups if we have time.

good idea — added to Notion task

gj:
Superseded by #976

PR #971, opened 2021-06-15
Include roles.polar file in built wheels

gj:
Tested by running `python setup.py bdist_wheel` before & after this
change


samscott89 (approved):
LGTM!

PR #972, opened 2021-06-15
Rename allow/3 -> role_allow/3 & create user_in_role/3

gj:
Changelog will be part of the large entry I need to write for the Polar roles stuff


saolsen (approved):
Looks good.
Namespacing our rules seems like a good idea. `__oso_internal__` is a pretty extreme one lol.

gj:
> Looks good.
> Namespacing our rules seems like a good idea. `__oso_internal__` is a pretty extreme one lol.

Yeah haha, I figured at least it would be ugly enough to ensure we don't forget about it and prod us to come up with a real namespacing solution

PR #973, opened 2021-06-15
Move Polar roles to core and add API for Node.js

gj:


languages/js/src/Polar.test.ts
```diff
@@ -802,3 +802,154 @@ describe('iterators', () => {
     ).toBe(6);
   });
 });
+
+// test_roles_integration
+describe('Polar roles', () => {
```
dhatch, line 807: I guess if this is easy enough to port seems reasonable (in reference to my comment in Notion)

languages/python/sqlalchemy-oso/sqlalchemy_oso/session.py
```diff
@@ -231,6 +232,11 @@ def __init__(self, oso: Oso, user, checked_permissions: Permissions, **options):
         responses from changing, ensuring that the session's identity map never
         contains unauthorized objects.
         """
+        if oso._polar_roles_enabled:
+            raise UnsupportedError(
+                "Data filtering not yet supported with Polar roles enabled."
```
samscott89, line 237: We should be directing people to using the sqlalchemy builtin roles

gj: good idea

languages/python/oso/polar/polar.py
```diff
@@ -86,6 +87,16 @@ def __del__(self):
...
 
+    def enable_roles(self):
+        class InternalRolesHelpers:
+            @staticmethod
+            def join(separator, left, right):
+                return separator.join([left, right])
+
+        self.register_constant(InternalRolesHelpers, "__oso_internal_roles_helpers__")
+        self.ffi_polar.enable_roles()
+        self._polar_roles_enabled = True
```
leina05, line 98: Should we check if this is already `True` to prevent them calling the method twice by accident?

languages/python/oso/polar/polar.py
```diff
@@ -120,6 +131,7 @@ def load_str(self, string, filename=None):
 
     def clear_rules(self):
         self.ffi_polar.clear_rules()
+        self._polar_roles_enabled = False
```
leina05, line 134: Hm, didn't think about this. Should we auto-reload the Polar roles policy if it was enabled with `clear_rules` was called? Or is it better to require the user to explicitly re-enable? I think from the user's perspective, they might not understand that "enabling roles" involves loading rules, so wouldn't realize that a side-effect of `clear_rules` is deactivating roles. But I don't really know when people use `clear_rules` anyway.

languages/python/oso/tests/test_polar_roles.py
```diff
@@ -2673,8 +2670,8 @@ def test_roles_integration(init_oso, sample_data):
     assert not oso.is_allowed(leina, "edit", ios_laggy)
     assert not oso.is_allowed(steve, "edit", ios_laggy)
 
-    oso.actor = leina
-    oso.checked_permissions = {Repo: "pull"}
+    # oso.actor = leina
+    # oso.checked_permissions = {Repo: "pull"}
     # auth_session = auth_sessionmaker()
```
leina05, line 2675: Can we delete all this commented code?

languages/python/oso/tests/test_polar_roles.py
```diff
@@ -56,9 +54,8 @@ def __repr__(self):
 class Issue(Base):  # type: ignore
     __tablename__ = "issues"
 
-    issue_id = Column(Integer, primary_key=True)
-    name = Column(String(256))
-    repo_id = Column(Integer, ForeignKey("repos.repo_id"))
+    name = Column(String(256), primary_key=True)
+    repo_name = Column(String(256), ForeignKey("repos.name"))
```
leina05, line 58: Let's pick a convention for defining these GitClub classes and stick with it (used in GitClub app + multiple tests). I'm assuming this is the naming used in the GitClub app, so we can just refer to that as the template from now on?

polar-core/src/polar.rs
```diff
@@ -296,6 +298,17 @@ impl Polar {
...
         self.messages.next()
     }
+
+    /// Load the Polar roles policy idempotently.
+    pub fn enable_roles(&self) -> PolarResult<()> {
+        match self.load(ROLES_POLICY, Some("Built-in Polar Roles Policy".to_owned())) {
+            Err(error::PolarError {
+                kind: error::ErrorKind::Runtime(error::RuntimeError::FileLoading { .. }),
+                ..
+            }) => Ok(()),
```
leina05, line 308: Why are we returning `Ok(())` on this error?

leina05: If it's to ignore duplicate file errors--is that the only case which could produce the `FileLoading` error, or would we potentially be ignoring any valid errors?

languages/js/src/Polar.test.ts
leina05: So is our plan to keep the tests in the host language libraries? Or are we moving testing into the core now that `enable_roles` is in the core?

leina05: Either way--I assume we'll have a follow-up PR for either finishing porting the node tests or porting all tests to the core, right?

leina05 (commented):
Left a few comments, mostly clarifying questions. Would also like to know what the plan is for testing, otherwise looks good.

languages/js/src/Polar.test.ts
gj: There are four pieces of testing split into two main categories: validation & functionality.

@saolsen is currently working on splitting the validation tests in twain, pulling the bits we can test in the core into the core and leaving another chunk in the host language test suite. The latter will have to be ported to each new host language as we enable the roles feature.

For functionality tests, we want a single, exhaustive test suite (which would ideally live in the core) and then one test per host language library that makes sure the library-specific bits of the API work. Now that the policy lives in the core and is shared across languages, we decided there wasn't really much difference between pulling the Python test suite into the core and leaving it in Python — and the former would be significantly more work.

So, to answer your questions explicitly: we're keeping some tests in the host language libraries and putting some in the core, and the only follow-up PR will be #969 for validation tests, which Steve's working on.

languages/python/oso/polar/polar.py
gj: That's a good question. Will think more about it.

languages/python/oso/tests/test_polar_roles.py
gj: Yeah I s'pose; was letting it hang around in the hopes we'd get to testing partial eval soon, but can always pull it out of the Git woodpile when we're ready to do that.

languages/python/oso/tests/test_polar_roles.py
gj: Not sure if this matches GitClub; the changes here are just simplifying the fixtures we copied from `sam/polar-roles` to remove columns that the tests weren't actually hitting.

polar-core/src/polar.rs
gj: > If it's to ignore duplicate file errors

yep

> is that the only case which could produce the FileLoading error

yep

> would we potentially be ignoring any valid errors?

nope

languages/python/oso/polar/polar.py
gj: Sure, everything's already idempotent, but doing the explicit check will make that more obvious

languages/python/oso/tests/test_polar_roles.py
gj: actually gonna avoid doing it for now b/c I think it'll cause @saolsen some merge conflict headaches on #969. will clean it up as part of that PR or as follow-up.

languages/python/sqlalchemy-oso/sqlalchemy_oso/session.py
gj: 8fb1530

languages/python/oso/polar/polar.py
gj: > Should we auto-reload the Polar roles policy if it was enabled with clear_rules was called? ... I think from the user's perspective, they might not understand that "enabling roles" involves loading rules, so wouldn't realize that a side-effect of clear_rules is deactivating roles.

Agree with this reasoning; implemented in 5ba7d14.

Only downside is that calling `clear_rules()` won't leave you with a completely empty KB, which might be unexpected, but as you said I'm not sure folks would realize `enable_roles()` loads rules unless they looked at the code or trace output.

languages/python/oso/polar/polar.py
gj: 5ba7d14df62b091354fdf8b7f30e02c8bd17595a

leina05 (approved):

PR #978, opened 2021-06-17
Support trailing commas in dictionaries and list.

samscott89:
Punted on doing calls/rules for now.

PR checklist:

- [x] Added changelog entry.

### DIFF
--- a/languages/python/sqlalchemy-oso/tests/test_roles.py
+++ b/languages/python/sqlalchemy-oso/tests/test_roles.py
@@ -486,7 +486,6 @@ def test_duplicate_role_name_same_resource(init_oso):
         """
     with pytest.raises(ParserError):
         oso.load_str(policy)
-    oso.roles.synchronize_data()
 
 
 def test_duplicate_role_name_different_resources(init_oso, sample_data):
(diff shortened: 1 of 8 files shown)


languages/python/sqlalchemy-oso/tests/test_roles.py
```diff
@@ -483,9 +484,9 @@ def test_duplicate_role_name_same_resource(init_oso):
             }
         };
         """
-    oso.load_str(policy)
-    with pytest.raises(OsoError):
-        oso.roles.synchronize_data()
+    with pytest.raises(ParserError):
+        oso.load_str(policy)
+    oso.roles.synchronize_data()
```
samscott89, line 489:
```suggestion
```

we wont even get here

languages/python/sqlalchemy-oso/tests/test_roles.py
```diff
@@ -483,9 +484,8 @@ def test_duplicate_role_name_same_resource(init_oso):
             }
         };
         """
-    oso.load_str(policy)
-    with pytest.raises(OsoError):
-        oso.roles.synchronize_data()
+    with pytest.raises(ParserError):
+        oso.load_str(policy)
```
gj, line 488: FYI @saolsen we don't need to check for duplicate keys in the same `resource/4` any more, but maybe we should still check for two `resource/4` predicates defined for the same type and the same role name?

gj (approved):

PR #981, opened 2021-06-18
validate node roles

saolsen:


leina05:
This isn't strictly related to Node validation, but I noticed when trying out Sam's Python [quickstart repo](https://github.com/osohq/oso-python-quickstart/pull/1) that the error message for calling `enable_roles()` before loading the policy isn't very helpful:

```
  File "app.py", line 11, in <module>
    oso.enable_roles()
  File "/Users/leinamcdermott/code/oso/languages/python/oso/polar/polar.py", line 124, in enable_roles
    self.ffi_polar.validate_roles_config(validation_query_results)
  File "/Users/leinamcdermott/code/oso/languages/python/oso/polar/ffi.py", line 30, in validate_roles_config
    check_result(result)
  File "/Users/leinamcdermott/code/oso/languages/python/oso/polar/ffi.py", line 165, in check_result
    raise Error().get()
polar.exceptions.RolesValidationError: Oso Roles Validation Error: Need to define resources to use oso roles.
        Get help with Oso from our engineers: https://help.osohq.com/error/RolesValidationError
```

Is there a way that we can catch this before running validation? E.g. with a flag when policies are loaded?

languages/js/src/Polar.ts
```diff
@@ -142,6 +188,11 @@ export class Polar {
       results.return();
       if (done) throw new InlineQueryFailedError(source);
     }
+
+    // TODO(gj): Make sure this is correct.
```
saolsen, line 192: This wouldn't do anything because the inline queries get added when you call `enable_roles`. It's technically ok and any config stuff you added would work but it wouldn't be validated.

languages/js/src/types.ts
```diff
@@ -470,3 +566,19 @@ export function isIterableIterator(x: any): boolean {
 export function isAsyncIterator(x: any): boolean {
   return Symbol.asyncIterator in Object(x);
 }
+
+/**
```
saolsen, line 570: big brain

languages/js/src/types.ts
```diff
@@ -470,3 +566,19 @@ export function isIterableIterator(x: any): boolean {
...
+ * JS objects, we lose the ability to distinguish between dictionaries and
+ * instances, since all JS instances are objects. By subclassing `Object`, we
+ * can use `instanceof` to determine if a JS value should be serialized as a
+ * Polar dictionary or external instance.
+ *
+ * @internal
+ */
+export class Dict extends Object {
+  [index: string]: any;
+}
```
saolsen, line 584: big brain

leina05:
> This isn't strictly related to Node validation, but I noticed when trying out Sam's Python [quickstart repo](https://github.com/osohq/oso-python-quickstart/pull/1) that the error message for calling `enable_roles()` before loading the policy isn't very helpful:
> 
> ```
> Traceback (most recent call last):
>   File "app.py", line 11, in <module>
>     oso.enable_roles()
>   File "/Users/leinamcdermott/code/oso/languages/python/oso/polar/polar.py", line 124, in enable_roles
>     self.ffi_polar.validate_roles_config(validation_query_results)
>   File "/Users/leinamcdermott/code/oso/languages/python/oso/polar/ffi.py", line 30, in validate_roles_config
>     check_result(result)
>   File "/Users/leinamcdermott/code/oso/languages/python/oso/polar/ffi.py", line 165, in check_result
>     raise Error().get()
> polar.exceptions.RolesValidationError: Oso Roles Validation Error: Need to define resources to use oso roles.
>         Get help with Oso from our engineers: https://help.osohq.com/error/RolesValidationError
> ```
> 
> Is there a way that we can catch this before running validation? E.g. with a flag when policies are loaded?

In addition to this, something that came up in checkin this morning (@samscott89 ) that we've talked about in the past was checking for things like having `actor_has_role_for_resource`, `role_allows`, and `parent_child` defined (rather than just query failing) if `enable_roles` is called. I know this adds a lot more to validation--happy to help out if that would be useful.

languages/js/src/Host.ts
```diff
@@ -308,31 +350,43 @@ export class Host {
     } else if (isPolarList(t)) {
       return await Promise.all(t.List.map(async el => await this.toJs(el)));
     } else if (isPolarDict(t)) {
+      const toJs = ([k, v]: [string, PolarTerm]) =>
```
gj, line 353:
```suggestion
      const valueToJs = ([k, v]: [string, PolarTerm]) =>
```

samscott89 (approved):
Talked through this live.

There is an unfortunate clash between how JS objects and Maps that don't cleanly map onto Polar dictionaries. This is a clash that has always existed, but is made evident when something makes a roundtrip JS<->Polar.

In this PR, this was a dict going Polar -> JS -> Polar. We address this by having a specific subclass of object to "track" that change.

There are also problems for objects going JS -> Polar -> JS. For example, if you have an objects `{a: 1}` created from JS, and then it goes to Polar, it _mostly_ behaves like a dictionary. But you cannot act on it like a native dict: iterators and unifying with dicts does not work.

This problem is not made any worse by this PR. 

The changes here for validation are good.

languages/js/src/Polar.ts
gj: Ahhh, okay... what do you think we should do if someone enables roles (running validation against all previously loaded policy) and then subsequently loads additional rules that might contain invalid roles predicates?

saolsen: Yeah all these different possible orders of doing things really sucks. If we're worried about that for now I think we should error if they load any polar after enabling roles.

samscott89: Shall we just take this out for now then?

samscott89: > 
> 
> Yeah all these different possible orders of doing things really sucks. If we're worried about that for now I think we should error if they load any polar after enabling roles.

That's not a bad idea. If you want dynamic policy loading, it's not yet supported with roles enabled. Seems like a reasonable trade-off.

languages/js/src/types.ts
gj: the galaxy brain meme but it goes Python -> Rust -> Idris -> JS

languages/js/src/Polar.ts
samscott89: > 
> 
> > Yeah all these different possible orders of doing things really sucks. If we're worried about that for now I think we should error if they load any polar after enabling roles.
> 
> That's not a bad idea. If you want dynamic policy loading, it's not yet supported with roles enabled. Seems like a reasonable trade-off.

(take as a follow up)

PR #982, opened 2021-06-21
Update JS lib dependencies

gj:
TypeDoc was broken again, so let's just bump everything while we're at it.

PR checklist:

- [ ] Added changelog entry.


.github/workflows/release.yml
```diff
@@ -659,7 +659,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        node-version: [10, 11, 12, 13, 14]
+        node-version: ['10.0.0', '12.0.0', '14.0.0', '16.0.0']
```
saolsen, line 662: Is there a reason we dropped 11?

saolsen (approved):
👍

.github/workflows/release.yml
gj: Figured testing [the LTS versions](https://nodejs.org/en/about/releases/) was good enough and wasn't sure if we were bumping up against some GitHub Actions limit on # of jobs for a single workflow haha. If you think it's cool, I'm happy to add 11 and 13 (and 15) back in.

PR #984, opened 2021-06-21
Add validation checks for calling enable_roles() after loading policy + actor_has_role_for_resource

leina05:
PR checklist:

- [ ] Update changelog entry/docs?

### DIFF
--- a/languages/python/oso/tests/test_polar_roles.py
+++ b/languages/python/oso/tests/test_polar_roles.py
@@ -139,11 +139,14 @@ def test_empty_role(init_oso):
         roles = {
             member: {}
         };
+    actor_has_role_for_resource(_, _, _);
     """
     oso.load_str(policy)
 
-    with pytest.raises(OsoError):
+    with pytest.raises(OsoError) as e:
         oso.enable_roles()
+
+    assert e.match("Must define actions or implications for a role.")
 
 
 @pytest.mark.skip(reason="TODO: More validation")
@@ -160,6 +163,7 @@ def test_bad_namespace_perm(init_oso):
                 permissions: ["repo:pull"]
             }
         };
+    actor_has_role_for_resource(_, _, _);
     """
     oso.load_str(policy)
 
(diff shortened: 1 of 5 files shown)


languages/python/oso/tests/test_polar_roles.py
```diff
@@ -139,6 +139,7 @@ def test_empty_role(init_oso):
         roles = {
             member: {}
         };
+    actor_has_role_for_resource(_, _, _) if False;
```
samscott89, line 142: Should be lower case `false`. `False` is unbound and evaluates to true...

leina05: ah I always mess this up

leina05: The return value doesn't actually matter--the rule just needs to be here so the validation error isn't thrown.

leina05: But yes I'll fix it so it's not confusing.

saolsen: > `False` is unbound and evaluates to true...

leina05: Actually it can't be `false` otherwise we won't recognize it when we conduct inline queries to check for it.

languages/python/oso/tests/test_polar_roles.py
```diff
@@ -335,11 +348,16 @@ def test_relationship_without_resources(init_oso):
...
+
+    actor_has_role_for_resource(_, _, _);
     """
     oso.load_str(policy)
 
-    with pytest.raises(OsoError):
+    with pytest.raises(OsoError) as e:
         oso.enable_roles()
+    assert e.match(
+        "Need to define at least one `resource/4` predicate to use oso roles"
```
saolsen, line 359: I know we use this syntax internally a lot and it's a common prolog thing `rulename/num_args` but I don't think we ever talk like this in the docs so could be confusing to people with no prior prolog knowledge.

saolsen (approved):

languages/python/oso/tests/test_polar_roles.py
saolsen: We also don't really say predicate in the docs, we call them rules.

leina05: Yeah ik, I could write out the arg names it's just kind of long. But prob a good idea.

languages/python/oso/tests/test_polar_roles.py
gj: @saolsen I laughed out loud when I saw that gif earlier today, then forgot about it, came back, and it got me again

PR #985, opened 2021-06-22
Fix Create Release workflow

gj:
docs for `yarn install --prod`:

> Yarn will not install any package listed in `devDependencies` if the
> `NODE_ENV` environment variable is set to `production`. Use this flag to
> instruct Yarn to ignore `NODE_ENV` and take its production-or-not status from
> this flag instead.

Seeing as how this step always installs all of our `devDependencies` anyway
(see: previous runs of the `Create Release` workflow), maybe it falls back to
installing `devDependencies` if there are no `dependendencies`? Either way, very
weird.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,417 +5,418 @@ on:
       - "v*" # Push events to matching v*, i.e. v1.0, v20.15.10
     branches:
       - main
+      - gj/fix-node-create-release-workflow
 
 jobs:
-  # TODO(gj): is it weird that the wheel jobs depend on `version` but the lib jobs don't?
-  version:
-    name: Compute and verify the version number
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v2
-      - uses: ./.github/actions/check_version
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -673,11 +674,8 @@ jobs:
         uses: actions/download-artifact@v1
         with:
           name: oso_js_node
-      - name: Run tests
-        # Timeout fix: https://stackoverflow.com/a/51508426
-        run: |
-          yarn install --prod  --network-timeout 100000
```
samscott89, line 679: I dont understand why we don't need to install dependencies here?

gj: There are no dependencies to install — Oso has zero runtime JS dependencies, and the built parity test in `oso_js_node` imports `Oso` & `Variable` from the built Oso in `oso_js_node` and `join` from Node's `path` module and then loads the `test.polar` file via a relative path.

samscott89: Ahhh, got it. Okay

samscott89 (approved):

PR #993, opened 2021-06-28
Fix roles validation bug

leina05:
PR checklist:

- [ ] Added changelog entry.


samscott89 (approved):

polar-core/src/roles_validation.rs
```diff
@@ -32,17 +34,47 @@ struct Resource {
 
 pub const VALIDATE_ROLES_CONFIG_RESOURCES: &str = "resource(resource, name, actions, roles)";
 
-pub fn validate_roles_config(roles_config: Vec<Vec<ResultEvent>>) -> PolarResult<()> {
+pub fn validate_actor_role(rules: &HashMap<Symbol, GenericRule>) -> PolarResult<()> {
+    if let Some(actor_role) = rules.get(&sym!("actor_has_role_for_resource")) {
```
gj, line 38: nit: variable naming could match rule name (`validate_actor_role` -> `validate_actor_has_role_for_resource`, etc.)

polar-core/src/roles_validation.rs
```diff
@@ -31,28 +33,39 @@ struct Resource {
...
     }
+    Ok(())
+}
 
-    let role_resources = roles_config.get(1).ok_or_else(|| {
+pub fn validate_roles_config(
+    polar: &Polar,
+    roles_config: Vec<Vec<ResultEvent>>,
+) -> PolarResult<()> {
+    let rules = &polar.kb.read().unwrap().rules;
```
gj, line 66: nit: since we pass a Polar in, immediately read its `rules`, and then don't use it for anything else, I'd probably just read the rules in the calling function and then pass them in here instead of a Polar.

gj (approved):
couple nits but seems fine

polar-core/src/roles_validation.rs
leina05: Fixed.

PR #994, opened 2021-06-28
Add integrations list page

samraper:
Need help on the content that we want to add.

### DIFF
--- a/docs/config.toml
+++ b/docs/config.toml
@@ -24,6 +24,9 @@ defaultContentLanguage = "any"
     weight = 1
     [languages.node.params]
       icon = "node-js"
+      frameworks = [
+        ["Express", "Express is a minimal and flexible Node.js web application framework that provides a robust set of features for web and mobile applications."]
+      ]
 
   [languages.python]
     languageName = "Python"
@@ -31,6 +34,9 @@ defaultContentLanguage = "any"
     weight = 2
     [languages.python.params]
       icon = "python"
+      frameworks = [
+        ["foo", "foo is a language"]
+      ]
 
   [languages.go]
     languageName = "Go"
--- a/docs/config.toml
+++ b/docs/config.toml
@@ -24,9 +24,6 @@ defaultContentLanguage = "any"
     weight = 1
     [languages.node.params]
       icon = "node-js"
-      frameworks = [
-        ["Express", "Express is a minimal and flexible Node.js web application framework that provides a robust set of features for web and mobile applications."]
-      ]
 
   [languages.python]
     languageName = "Python"
@@ -34,9 +31,6 @@ defaultContentLanguage = "any"
     weight = 2
     [languages.python.params]
       icon = "python"
-      frameworks = [
-        ["foo", "foo is a language"]
-      ]
 
   [languages.go]
     languageName = "Go"
--- a/docs/content/node/reference/frameworks/_index.md
+++ b/docs/content/node/reference/frameworks/_index.md
@@ -1,0 +1,11 @@
+---
+title: Frameworks
+weight: 3
+any: false
+frameworksTOC: true
+---
+
+# Framework Integrations
+
+Oso does not provide any Node framework integrations yet. Let us know which
+ones we should integrate with below.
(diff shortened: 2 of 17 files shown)


docs/config.toml
```diff
@@ -24,13 +24,19 @@ defaultContentLanguage = "any"
     weight = 1
     [languages.node.params]
       icon = "node-js"
+      frameworks = [
```
samscott89, line 27: I think it would be cleaner to keep this data in the frameworks folder. e.g. `lang/reference/frameworks/data/data.md`

samraper: How would you suggest having an arbitrary length data array in one of those files? I couldn't figure out a clean way, which I figure is just a side effect of using md as a data format. There are other data files that can be in toml, yaml, etc. but those are put at the hugo root, and I didn't know if this was worth adding another data storage.

samscott89: The way we do it with some of the other guides is we store it in the front matter: https://github.com/osohq/oso/blob/main/docs/content/python/getting-started/quickstart/data/data.md

Along with an index file in any that marks it as headless: https://github.com/osohq/oso/blob/main/docs/content/any/getting-started/quickstart/data/index.md

It's not idea, because then you're just writing yaml inside markdown.... But to me it would be more intuitive where the data lives relative to the content.

samraper:
I see - I didn't realize that the front matter supported all of yaml. I'll throw it in there

docs/config.toml
samraper: Still having trouble here - per [documentation](https://gohugo.io/content-management/page-bundles/) we can't a) have non-page resources nor b) have resources in a nested directory in a list page. Any suggestions?

dhatch:
Feedback on frontend:

- [ ] Buttons render a little off, that may just be a local CSS issue I have.
- [ ] Add ability to give example apps for integrations that we haven't yet built, or link to other content like a blog post. For languages besides Python, all integrations will fall into this category. 

For some languages we might have multiple example applications. Could we think about how to display multiple apps with maybe a description for each?

For the backend I think we should just do something similar to the error tracking API I built that uses the Notion API. That was pretty straightforward and would be an easy way to capture this data.

docs/content/node/reference/frameworks/_index.md
```diff
@@ -0,0 +1,11 @@
+---
+title: Frameworks
+weight: 3
+any: false
+frameworksTOC: true
```
dhatch, line 5: Specifying frameworks to request in the front matter here would be more clear than in config.toml.

docs/config.toml
```diff
@@ -24,13 +24,19 @@ defaultContentLanguage = "any"
...
+        ["Express", "Express is a minimal and flexible Node.js web application framework that provides a robust set of features for web and mobile applications."]
+      ]
 
   [languages.python]
     languageName = "Python"
     contentDir = "content/python"
     weight = 2
     [languages.python.params]
       icon = "python"
+      frameworks = [
```
dhatch, line 37: Could we rename this to `request_frameworks`?

docs/themes/oso-tailwind/layouts/partials/integrations.html
```diff
@@ -0,0 +1,61 @@
...
+      {{ $description := index . 1 }}
+      <div class="flex flex-col p-8 inline-block shadow-md hover:shadow-xl focus:shadow-xl transition-shadow">
+        <div class="text-lg">{{- $framework -}}</div>
+        <div class="flex-grow text-gray-500 my-2">{{- $description -}}</div>
+        <div class="grid grid-cols-2 gap-2">
+          <a
+              href=""
+              class="no-a inline-block px-4 py-2 bg-gray-200 rounded hover:bg-gray-300 transition-colors"
+          >
+            Request this feature
```
dhatch, line 55: We'll need to add an example button here so that we can link to example apps for integrations that already have them.

dhatch (commented):
Some feedback. I think we touched on these suggestions verbally but I'm not sure they were specified in the PR. Data for these pages coming soon.

docs/config.toml
dhatch: I also don't think descriptions are necessary for frameworks that we don't integrate with.

dhatch (approved):
lgtm

PR #995, opened 2021-06-29
Go: get allowed actions implementation

joshrotenberg:
My Go is out of date, so I'm sure there are issues with this, not the least of which being that it doesn't actually work.

Comments, suggestions, harsh criticism, banning from the project are all welcome.

See #746 

PR checklist:

- [ ] Added changelog entry.


dhatch:
🎉 🎉 🎉 Thanks for this @joshrotenberg! I'll make sure we get some eyes on this!

languages/go/oso.go
```diff
@@ -152,6 +157,42 @@ func (o Oso) IsAllowed(actor interface{}, action interface{}, resource interface
 	}
 }
 
+/*
+Return a set of actions allowed by the given (actor, resource) combination allowed
+by the policy.
+*/
+func (o Oso) GetAllowedActions(actor interface{}, resource interface{}, allowWildcard bool) (map[interface{}]interface{}, error) {
+	results := make(map[interface{}]interface{})
+	query, err := (*o.p).queryRule("allow", actor, resource)
```
samscott89, line 166: This needs to pass in `ValueVariable{"action"}` as the 3rd arg to queryRule.

languages/go/oso.go
```diff
@@ -152,6 +157,42 @@ func (o Oso) IsAllowed(actor interface{}, action interface{}, resource interface
...
+		return nil, err
+	}
+
+	for {
+		if v, err := query.Next(); err != nil {
+			return nil, err
+		} else if v == nil {
+			break
+		} else {
+			switch v := interface{}(v).(type) {
```
samscott89, line 177: The return type of `query.Next` is `*map[string]interface{}`

What you want to do is look for `v["action"]` (to match with the variable you pass in above).
And then on the result of that, you can apply this switch.

languages/go/oso.go
```diff
@@ -152,6 +157,42 @@ func (o Oso) IsAllowed(actor interface{}, action interface{}, resource interface
 	}
 }
 
+/*
+Return a set of actions allowed by the given (actor, resource) combination allowed
+by the policy.
+*/
+func (o Oso) GetAllowedActions(actor interface{}, resource interface{}, allowWildcard bool) (map[interface{}]interface{}, error) {
+	results := make(map[interface{}]interface{})
```
samscott89, line 165: Go doesn't have a set type apparently (just searched for this). We could maybe return a `map[interface{}]boolean` instead? Where the keys are any values returned by the query, and the value is just true for every present value?

samscott89: Apparently `map[interface{}]struct{}` is better: https://emersion.fr/blog/2017/sets-in-go/

languages/go/oso.go
joshrotenberg: Right, of course. Although that gave me a compiler error, I needed `types.ValueVariable("action")`.

languages/go/oso.go
joshrotenberg: Done ... I think. Compiles, so that's start.

samscott89 (approved):
I only just saw the updates now. This looks good!

PR #997, opened 2021-06-29
Cleaner imports

tomashozman:
I think it's much prettier and cleaner to do it like this, instead of having both objects from SQLAlchemy and SQLAlchemy itself in our namespace. Also better for consistency's sake.


tomashozman:

gj (approved):
Reasoning makes sense to me.

Thanks for the contribution!

PR #998, opened 2021-06-30
Improved formatting & imports for Python's oso and sqlalchemy-oso

tomashozman:
I've decided to try and improve the formatting and its consistency somewhat, as well as clean up the namespaces a bit. It's still not perfect, and I've probably missed a couple issues, but I think it's way better like this than it was before. I've also added __all__ variables to a couple __init__.py files.
The decision to not indent multi-line strings may be a bit controversial, as well as some other things. I can change those if necessary.
Please let me know what you think.


tomashozman:
Oh. I thought I checked it properly. Let's see.

PR #1001, opened 2021-07-06
Allow custom query timeouts

gkaemmer:
This change allows users to configure the query timeout using an environment variable `POLAR_TIMEOUT_MS`. If the timeout is set to `0`, the timeout mechanism is disabled entirely (which is useful for debugging)

### DIFF
--- a/polar-core/src/vm.rs
+++ b/polar-core/src/vm.rs
@@ -31,7 +31,7 @@ use crate::terms::*;
 use crate::traces::*;
 
 pub const MAX_STACK_SIZE: usize = 10_000;
-pub const QUERY_TIMEOUT_MILLIS: u64 = 30_000;
+pub const DEFAULT_TIMEOUT_MS: u64 = 30_000;
 
 #[derive(Debug, Clone)]
 #[must_use = "ignored goals are never accomplished"]
@@ -227,7 +227,7 @@ pub struct PolarVirtualMachine {
     query_start_time: Option<std::time::Instant>,
     #[cfg(target_arch = "wasm32")]
     query_start_time: Option<f64>,
-    query_timeout_millis: u64,
+    query_timeout_ms: u64,
 
     /// Maximum size of goal stack
     stack_limit: usize,
@@ -296,7 +296,9 @@ impl PolarVirtualMachine {
             goals: GoalStack::new_reversed(goals),
             binding_manager: BindingManager::new(),
             query_start_time: None,
-            query_timeout_millis: QUERY_TIMEOUT_MILLIS,
+            query_timeout_ms: std::env::var("POLAR_TIMEOUT_MS")
+                .map(|timeout_str| timeout_str.parse::<u64>().unwrap_or(DEFAULT_TIMEOUT_MS))
+                .unwrap_or(DEFAULT_TIMEOUT_MS),
             stack_limit: MAX_STACK_SIZE,
             csp: Bsp::default(),
             choices: vec![],
@@ -379,11 +381,6 @@ impl PolarVirtualMachine {
     #[cfg(test)]
     fn set_stack_limit(&mut self, limit: usize) {
         self.stack_limit = limit;
-    }
-
-    #[cfg(test)]
-    fn set_query_timeout(&mut self, timeout_s: u64) {
-        self.query_timeout_millis = timeout_s as u64 * 1_000;
     }
 
     pub fn new_id(&self) -> u64 {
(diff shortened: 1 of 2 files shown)


polar-core/src/vm.rs
```diff
@@ -389,7 +383,7 @@ impl PolarVirtualMachine {
 
     #[cfg(test)]
     fn set_query_timeout(&mut self, timeout_s: u64) {
-        self.query_timeout = std::time::Duration::from_secs(timeout_s);
+        self.query_timeout_millis = timeout_s as u64 * 1_000;
```
gj, line 386: looks like `timeout_s` is already a `u64`

gkaemmer:

polar-core/src/vm.rs
```diff
@@ -867,12 +861,12 @@ impl PolarVirtualMachine {
             .query_start_time
```
gj, line 861: Might consider breaking the current query duration calculation out into a separate function. Something like:

```rs
    #[cfg(not(target_arch = "wasm32"))]
    fn query_duration(&self) -> u64 {
        let now = std::time::Instant::now();
        let start = self.query_start_time.expect("Query start not recorded");
        (now - start).as_millis() as u64
    }

    #[cfg(target_arch = "wasm32")]
    fn query_duration(&self) -> u64 {
        let now: f64 = js_sys::Date::now();
        let start = self.query_start_time.expect("Query start not recorded");
        (now - start) as u64
    }

    // TODO (dhatch): How do we reliably not do this when debugging.
    fn check_timeout(&self) -> PolarResult<()> {
        let elapsed = self.query_duration();
        if elapsed > self.query_timeout_ms {
            return Err(error::RuntimeError::QueryTimeout {
                msg: format!(
                    "Query running for {}ms. Exceeded query timeout of {}ms",
                    elapsed, self.query_timeout_ms
                ),
            }
            .into());
        }
        Ok(())
    }
```

polar-core/src/vm.rs
```diff
@@ -296,7 +296,9 @@ impl PolarVirtualMachine {
             goals: GoalStack::new_reversed(goals),
             binding_manager: BindingManager::new(),
             query_start_time: None,
-            query_timeout_millis: QUERY_TIMEOUT_MILLIS,
+            query_timeout_ms: std::env::var("POLAR_TIMEOUT_MS")
+                .map(|timeout_str| timeout_str.parse::<u64>().unwrap_or(DEFAULT_TIMEOUT_MS))
+                .unwrap_or(DEFAULT_TIMEOUT_MS),
```
gj, line 301: nit: slight preference for 
```suggestion
            query_timeout_ms: std::env::var("POLAR_TIMEOUT_MS")
                .ok()
                .and_then(|ms| ms.parse::<u64>().ok())
                .unwrap_or(DEFAULT_TIMEOUT_MS),
```
to avoid the double `.unwrap_or(DEFAULT_TIMEOUT_MS)`

don't have to fix if you prefer it as-is

gj: also might want to move this logic up above the `let mut vm = Self {` to clean up the struct initialization a bit

(we should also probably do the same for `polar_log_stderr`, but don't need to do either as part of this PR)

gj (changes requested):
what do you think about setting `vm.query_timeout_ms = 0` when we hit a `debug()` so that we don't time folks out in the middle of a debug sesh?

polar-core/src/vm.rs
```diff
@@ -859,46 +850,39 @@ impl PolarVirtualMachine {
...
 
-        if now - start_time > self.query_timeout {
+        let elapsed = self.query_duration();
+        if elapsed > self.query_timeout_ms {
             return Err(error::RuntimeError::QueryTimeout {
                 msg: format!(
-                    "Query running for {}. Exceeded query timeout of {} seconds",
-                    (now - start_time) / 1_000.0,
-                    self.query_timeout / 1_000.0
+                    "Query running for {}ms. Exceeded query timeout of {}ms.",
```
gj, line 880:
```suggestion
                    "Query running for {}ms. Exceeded query timeout of {}ms. To disable the timeout, set the POLAR_TIMEOUT_MS environment variable to 0.",
```

gkaemmer:
> what do you think about setting `vm.query_timeout_ms = 0` when we hit a `debug()` so that we don't time folks out in the middle of a debug sesh?

This could be kinda weird if you hit a timeout, added a `debug` statement to your policy, and then stopped getting the timeout? Do many people use `debug` on the polar side?

gj:
> This could be kinda weird if you hit a timeout, added a `debug` statement to your policy, and then stopped getting the timeout? Do many people use `debug` on the polar side?

I'm not sure how prevalent `debug()` usage is, but this is a good point — let's punt for now.

gj (approved):
ship it

gj:
TODO changelog

PR #1008, opened 2021-07-12
Dynamically load constants in ruby library

gkaemmer:
This fixes issues caused by using Oso with code reloaders (i.e. rails and [zeitwerk](https://github.com/fxn/zeitwerk)), which effectively swap out the values of global constants when files are changed.

One user-space solution is to re-create an Oso instance (with `Oso.new`) every time a file changes, and call `register_class` with each new class constant, but that's kind of a hack. This change lets us take advantage of Ruby's global "constant table" to always use the most up-to-date version of a constant.

Note: it relies on this identity (which is always true AFAICT): `Object.const_get(klass.name) == klass`. That holds even for classes in a nested namespace, e.g. `Foo::Bar::Bat`.

PR checklist:

- [x] Added changelog entry.

### DIFF
--- a/languages/ruby/lib/oso/polar/host.rb
+++ b/languages/ruby/lib/oso/polar/host.rb
@@ -18,13 +18,18 @@ module Oso
     # look it up using const_get, which will always return the up-to-date
     # version of the class.
     class PolarClass
-      attr_reader :name
+      attr_reader :name, :anon_class
 
       def initialize(klass)
         @name = klass.name
+        # If the class doesn't have a name, it is anonymous, meaning we should
+        # actually store it directly
+        @anon_class = klass if klass.name.nil?
       end
 
       def get
+        return anon_class if anon_class
+
         Object.const_get(name)
       end
     end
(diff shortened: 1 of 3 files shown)


languages/ruby/lib/oso/polar/host.rb
```diff
@@ -35,7 +62,7 @@ def initialize_copy(other)
       def get_class(name)
         raise UnregisteredClassError, name unless classes.key? name
 
-        classes[name]
+        classes[name].get
```
saolsen, line 65: Could this whole thing just be done in get_class?
We could still check that the class is registered, but then return
```
Object.const_get(name)
```

Wrapping the class in another one adds a lot of this wrapping unwrapping code. Seems like just handling this here would be simpler.

saolsen: That would break `is_a?` though so I guess maybe it's not that simple and we'd have to do some class conversion stuff there too.

gkaemmer: Yeah the weird thing there is when you want to do a lookup/call on a class like `User.count` or `User.is_admin(user)`, which uses the value returned from `get_instance`.

So if we unwound this abstraction, we'd need to use `const_get` in two places (`get_class` and `get_instance`). I'm not opposed to that, but might also be a bit weird.

gkaemmer: Also, this reminded me about a particular edge case with anonymous classes (that I just added a test for and fixed), which now makes the abstraction a bit more reasonable imo

saolsen (approved):
Cool, I'm into it.

Issue #1010, opened 2021-07-12
Enable partial evaluation from the java API


MFAshby:
what:
[data filtering](https://docs.osohq.com/python/project/internals.html#data-filtering) is a very powerful capability of oso. It allows me to apply attribute based access control to both read and search operations using write the same, simple policy for both. I'd like to make that capability available on Java also.

why:
I haven't yet found another authorization library with this capability in the java ecosystem!

how:
Implement the necessary types and conversion routines to be able to evaluate policies with Variables in them and return Expressions.

PR #1011, opened 2021-07-12
java: Parse Expressions and Operators returned from the polar core.

MFAshby:
Remove test for throwing assertion when Expression is returned.

Ports tests for partial evaluation from the python code to java.

https://github.com/osohq/oso/issues/1010

PR checklist:

- [ ] Added changelog entry.
-> I'm not sure of the format yet, so I haven't done this.


MFAshby:

MFAshby:
I've ported over the relevant tests from the python module, I welcome suggestions for more testing though.

gj (commented):
This is awesome! To my currently very tired eyes, this looks like a great start. I'm out tomorrow, but I'll make sure that this either (A) gets looked at more closely by someone else tomorrow or (B) is first thing on my plate Thursday morning.

dhatch:
@MFAshby added an `acceptExpression` flag on the Host class. We use this to protect users that are not expecting to receive a partial result from receiving expressions in the query output. All the python adapters set this flag to true when running queries, but by default it is False. The equivalent functionality in Python is here: https://github.com/osohq/oso/blob/main/languages/python/oso/polar/host.py#L237

languages/java/oso/src/main/java/com/osohq/oso/Exceptions.java
```diff
@@ -5,6 +5,13 @@
 
 @SuppressWarnings("serial")
 public class Exceptions {
+  public static final String UNEXPECTED_EXPRESSION_MESSAGE =
+      "Recieved Expression from Polar VM. The Expression type is only supported when\n"
+          + "using django-oso or sqlalchemy-oso's data filtering features. Did you perform an "
+          + "operation over an unbound variable in your policy?\n\n"
+          + "To silence this error and receive an Expression result, pass\n"
+          + "accept_expression=True to Oso.queryRule.";
```
samscott89, line 13: It's not a kwarg, right?

languages/java/oso/src/main/java/com/osohq/oso/Polar.java
```diff
@@ -88,15 +92,65 @@ public void loadStr(String str) throws Exceptions.OsoException {
...
+  /**
+   * Query for a predicate, applying bindings and optionally accepting the expression type as a
+   * result.
+   *
+   * @param acceptExpression Set to true to accept an Expression as a result from the VM.
+   */
+  public Query query(Predicate query, Map<String, Object> bindings, boolean acceptExpression)
+      throws Exceptions.OsoException {
+    Host new_host = host.clone();
+    new_host.setAcceptExpression(acceptExpression);
```
samscott89, line 141: Why don't we make `setAcceptExpression` a method on the Query object that calls this? And then we don't need the multiple APIs for it?

languages/java/oso/src/test/java/com/osohq/oso/PolarTest.java
```diff
@@ -591,11 +601,102 @@ public void testIterators() throws Exception {
...
+  public void testPartialConstraint() {
+    p.registerClass(User.class, "User");
+    p.registerClass(Post.class, "Post");
+    p.loadStr("f(x: User) if x.user = 1;");
+    p.loadStr("f(x: Post) if x.post = 1;");
+    Variable x = new Variable("x");
+
+    Predicate rule = new Predicate("f", List.of(x));
+    List<HashMap<String, Object>> results =
+        p.query(rule, Map.of("x", new TypeConstraint(x, "User")), true).results();
```
samscott89, line 690: How is this working if ToPolar isn't implemented for it?

samscott89 (commented):
A few comments/suggestions.

languages/java/oso/src/test/java/com/osohq/oso/PolarTest.java
```diff
@@ -591,11 +601,102 @@ public void testIterators() throws Exception {
...
+    assertEquals(2, results.size());
```
samscott89, line 691: This should only have one result from the rule matching `User`, not the rule matching `Post`.

languages/java/oso/src/test/java/com/osohq/oso/PolarTest.java
MFAshby: It's possible that it's _not_ working, this might be the reason for 2 results instead of 1

MFAshby (commented):
Thanks @dhatch ! I didn't know the motivation behind that argument, but it makes sense now.

dhatch:
> It's possible that it's not working, this might be the reason for 2 results instead of 1

The same test in Python is also returning two results, but I would expect it to return one. I'm going to do a little investigation on that. I'll update you tomorrow @MFAshby!

languages/java/oso/src/main/java/com/osohq/oso/Exceptions.java
```diff
@@ -5,6 +5,13 @@
 
 @SuppressWarnings("serial")
 public class Exceptions {
+  public static final String UNEXPECTED_EXPRESSION_MESSAGE =
+      "Recieved Expression from Polar VM. The Expression type is only supported when\n"
```
gj, line 9:
```suggestion
      "Received Expression from Polar VM. The Expression type is only supported when\n"
```

languages/java/oso/src/main/java/com/osohq/oso/Exceptions.java
```diff
@@ -5,6 +5,13 @@
...
+          + "using django-oso or sqlalchemy-oso's data filtering features. Did you perform an "
```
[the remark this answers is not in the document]
gj: How about something like "The Expression type is not yet generally supported in the Java library. Did you perform an operation..."

languages/java/oso/src/test/java/com/osohq/oso/PolarTest.java
dhatch: This is working now, I added the implementation to the PR. I noticed this test was different though as ported. The python test [expects 1 result](https://github.com/osohq/oso/blob/574c80b357ca2e2487c8d6442f86308ffd3e684f/languages/python/oso/tests/test_polar.py#L840).

Are there any other tests you changed while porting?

languages/java/oso/src/main/java/com/osohq/oso/Polar.java
dhatch: Unfortunately Query gets a result when it is constructed from the core, so this parameter needs to be set on Host before query construction.

This also (roughly) matches the API in other languages, but I agree it's not ideal.

languages/java/oso/src/test/java/com/osohq/oso/OsoTest.java
```diff
@@ -167,4 +170,12 @@ public void testGetAllowedActionsWildcard() throws Exception {
     assertEquals(Set.of("*"), o.getAllowedActions(actor, widget, true));
     assertThrows(Exceptions.OsoException.class, () -> o.getAllowedActions(actor, widget, false));
   }
+
+  @Test
+  public void testNotEqalOperator() {
```
dhatch, line 175:
```suggestion
  public void testNotEqualOperator() {
```

languages/java/oso/src/test/java/com/osohq/oso/PolarTest.java
MFAshby: Not to my knowledge. I erroneously thought this should return 2 results for some reason.

languages/java/oso/src/main/java/com/osohq/oso/Host.java
```diff
@@ -191,6 +226,29 @@ public JSONObject toPolarTerm(Object value) throws Exceptions.OsoException {
           "Call", new JSONObject(Map.of("name", pred.name, "args", javaListToPolar(pred.args))));
     } else if (value != null && value instanceof Variable) {
       jVal.put("Variable", value);
+    } else if (value != null && value instanceof Expression) {
+      Expression expression = (Expression) value;
```
MFAshby, line 230: Thanks for this. I completely missed transforming in this direction.

languages/java/oso/src/test/java/com/osohq/oso/PolarTest.java
dhatch: Cool, I think it's just because the python test looks for two arguments in the And expression. I got a bit confused myself reading it.

I checked the other tests & they look good to me!

samscott89 (approved):
LGTM

MFAshby:
Does the changelog need updating here or in another PR?

PR #1015, opened 2021-07-15
Singleton variables are errors

gw000:
What it does:

- singleton variables are now errors!
- fixes a bug where SingletonVisitor wasn't walking dicts contained in
  patterns
- updates the singleton variable test, and removes singleton variables
  from other tests

What it doesn't do but maybe should:

- let users choose the old behavior (emit a warning)

PR checklist:

- [x] Added changelog entry.

### DIFF
--- a/docs/content/any/project/changelogs/NEXT.md
+++ b/docs/content/any/project/changelogs/NEXT.md
@@ -44,18 +44,29 @@ Link to [relevant documentation section]().
 
 #### Breaking changes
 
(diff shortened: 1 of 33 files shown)


gw000:

saolsen (approved):
Looks Great!

saolsen:
Changelog and msg look good, you can merge it when the tests pass!

gw000:
It's listed in the changelog as a breaking change, and the error message is updated to include the fix, which is very easy. So we're not going to add a compatibility option.

PR #1019, opened 2021-07-19
Add roles to Rust library

gw000:
PR checklist:

- [x] Added changelog entry.

### DIFF
--- a/languages/rust/oso/src/oso.rs
+++ b/languages/rust/oso/src/oso.rs
@@ -1,7 +1,7 @@
 //! Communicate with the Polar virtual machine: load rules, make queries, etc/
 
-use polar_core::terms::{Call, Symbol, Term, Value};
 use polar_core::roles_validation::ResultEvent;
+use polar_core::terms::{Call, Symbol, Term, Value};
 
 use std::collections::HashSet;
 use std::fs::File;
@@ -55,7 +55,11 @@ impl Oso {
         let inner = Arc::new(polar_core::polar::Polar::new());
         let host = Host::new(inner.clone());
 
-        let mut oso = Self { inner, host, polar_roles_enabled: false };
+        let mut oso = Self {
+            inner,
+            host,
+            polar_roles_enabled: false,
+        };
 
         for class in crate::builtins::classes() {
             oso.register_class(class)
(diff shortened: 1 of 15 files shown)


languages/rust/oso/src/oso.rs
```diff
@@ -131,9 +138,11 @@ impl Oso {
     }
 
     /// Clear out all files and rules that have been loaded.
-    pub fn clear_rules(&self) {
+    pub fn clear_rules(&mut self) -> crate::Result<()> {
```
dhatch, line 141: I'm pretty sure these methods take a `&self` so that the `Oso` object can be shared across threads. Why did we need to change them?

languages/rust/oso/src/oso.rs
```diff
@@ -166,14 +175,24 @@ impl Oso {
...
+        self.enable_roles()
+    }
+
     /// Load a string of polar source directly.
     /// # Examples
     /// ```ignore
     /// oso.load_str("allow(a, b, c) if true;");
     /// ```
-    pub fn load_str(&self, s: &str) -> crate::Result<()> {
+    pub fn load_str(&mut self, s: &str) -> crate::Result<()> {
```
dhatch, line 192: In other languages `load_str` calls `load_file`, so the roles library is reinitialized for both.

In Rust, both are separately implemented. We'll need to either call `load_str` from `load_file` or add `reinitialize_roles` to `load_file`.

languages/rust/oso/src/oso.rs
dhatch: They are initialization methods, and other initialization methods like `register_class` take `&mut self`, so this might be okay if we can't avoid it.

gw000: `enable_roles` sets a flag on the `Oso` object so any method that might call it needs to have a mutable reference.

languages/rust/oso/src/oso.rs
gw000: good catch, thanks!

languages/rust/oso/src/oso.rs
```diff
@@ -241,6 +261,46 @@ impl Oso {
...
+
+    pub fn enable_roles(&mut self) -> crate::Result<()> {
+        if self.polar_roles_enabled {
+            return Ok(());
+        }
+
+        self.inner.enable_roles()?;
+
+        if !self.host.has_class(OSO_INTERNAL_ROLES_HELPER) {
+            self.register_class(
```
gw000, line 273: It would be simpler to just include this with the built-in classes and register it when the host is initialized, before `enable_roles`. Is there any reason why we might not want to do that?

gw000: I went ahead & did it but let me know if there's a problem!

dhatch: That seems fine to me.

saolsen (approved):
This looks great.

PR #1020, opened 2021-07-20
Change literalInclude to enable highlighting by content

samraper:


dhatch (approved):
I think this contains the search changes again, but besides that looks good to me.

PR #1022, opened 2021-07-21
set context on singleton variable errors

gw000:


dhatch (approved):
LGTM if you confirmed that the error messages from Python now contain the location information.

Issue #795, opened 2021-03-23
[Go] Clean up ExternalOp comparisons


gj:
When handling an `ExternalOp` event, the Go adapter currently relies on operands implementing the `Comparer` interface defined here:

https://github.com/osohq/oso/blob/992b72881dfe528a116630573941eaa816e5900a/languages/go/interfaces/interfaces.go#L6-L11

If one of the operands doesn't implement `Comparer` — for example, if it's a type that doesn't have an `Equal()` method, such as a new type arising from a type alias — then an error is thrown.

As a quick fix, we should probably add the same `reflect.DeepEqual()` fallback that we have in the unification code path…

https://github.com/osohq/oso/blob/992b72881dfe528a116630573941eaa816e5900a/languages/go/internal/host/host.go#L192-L197

…to the equality checks we perform in the ExternalOp code path:

https://github.com/osohq/oso/blob/992b72881dfe528a116630573941eaa816e5900a/languages/go/query.go#L248-L263

However that's just for the `Equal()` piece of the `Comparer` interface — I'm not sure what to do for a generic `Lt()` fallback, or if one even makes sense to begin with.

### Alternatives

- Catch the runtime error and throw our own (more understandable/instructive) error telling users to implement `Equal()` / `Lt()` for whatever type is missing the method(s)
- Be explicit in the (forthcoming) Go docs that users must ensure their application types implement the `Comparer` interface.

samscott89:
Adding on to this, the Comparer interface itself is a bit rough:

```go
type Foo string

func (left Foo) Equal(right Foo) bool {
    // ...
}
```

isn't sufficient to compare `Foo`s with a string because: 

- The input type needs to match exactly (i.e. be other Comparer)
- We don't currently support comparing instances with not instances. So if you do foo = "abc" where foo is an instance of `Foo`, it's not going to work

First it needs to be:

```go
type Foo string
func (left Foo) Equal(right Comparer) bool {
    if other, ok := right.(string); ok    {
		return left == other
	}
}
```

Secondly we need to fix (I think this is true across all languages) comparing native + host types.

gj:
@gw000 or @samscott89 — was this fixed by #1023?

samscott89:
Yes!

PR #1023, opened 2021-07-21
Unify across host/native types

gw000:
This allows unification between host and native data of "equivalent" types, so queries like

```polar
new Integer(1) = 1
```
can succeed.

The specific example above will only work in Python, since different host languages have different rules about applying using the "new" operator built-in types, for example Ruby and Java can't create integers this way. But where creating a new instance of a type is supported by the host language, they should now be transparently unifiable with native data.

This PR also (at least partially) implements comparison operations for all host languages, and improves Python error handling when creating external instances.

PR checklist:

- [x] Added changelog entry.
- [x] Add appropriate tests for languages other than Python.

### DIFF
--- a/languages/go/query.go
+++ b/languages/go/query.go
@@ -122,8 +122,6 @@ func (q *Query) Next() (*map[string]interface{}, error) {
 			err = q.handleExternalIsSubSpecializer(ev)
 		case QueryEventExternalIsSubclass:
 			err = q.handleExternalIsSubclass(ev)
-		case QueryEventExternalUnify:
-			err = q.handleExternalUnify(ev)
 		case QueryEventExternalOp:
 			err = q.handleExternalOp(ev)
 		case QueryEventNextExternal:
@@ -223,14 +221,6 @@ func (q Query) handleExternalIsSubSpecializer(event types.QueryEventExternalIsSu
 
 func (q Query) handleExternalIsSubclass(event types.QueryEventExternalIsSubclass) error {
 	res, err := q.host.IsSubclass(string(event.LeftClassTag), string(event.RightClassTag))
-	if err != nil {
-		return err
-	}
-	return q.ffiQuery.QuestionResult(event.CallId, res)
-}
-
-func (q Query) handleExternalUnify(event types.QueryEventExternalUnify) error {
-	res, err := q.host.Unify(event.LeftInstanceId, event.RightInstanceId)
 	if err != nil {
 		return err
 	}
--- a/languages/go/query.go
+++ b/languages/go/query.go
@@ -245,32 +245,32 @@ func (q Query) handleExternalOp(event types.QueryEventExternalOp) error {
 	leftCmp, leftOk := left.(interfaces.Comparer)
 	rightCmp, rightOk := right.(interfaces.Comparer)
 
-  if !leftOk || !rightOk {
-    switch event.Operator.OperatorVariant.(type) {
-    case OperatorEq:
-      answer = reflect.DeepEqual(left, right)
-    default:
-      return fmt.Errorf("Unsupported operation: %v", event.Operator.OperatorVariant)
-    }
-  } else {
-    // @TODO: Where are the implementations for these for builtin stuff (numbers mainly)
-    switch event.Operator.OperatorVariant.(type) {
-    case OperatorLt:
-      answer = leftCmp.Lt(rightCmp)
-    case OperatorLeq:
-      answer = leftCmp.Lt(rightCmp) || leftCmp.Equal(rightCmp)
-    case OperatorGt:
-      answer = rightCmp.Lt(leftCmp)
-    case OperatorGeq:
-      answer = !leftCmp.Lt(rightCmp)
-    case OperatorEq:
-      answer = leftCmp.Equal(rightCmp)
-    case OperatorNeq:
-      answer = !leftCmp.Equal(rightCmp)
-    default:
-      return fmt.Errorf("Unsupported operation: %v", event.Operator.OperatorVariant)
-    }
-  }
+	if !leftOk || !rightOk {
+		switch event.Operator.OperatorVariant.(type) {
+		case OperatorEq:
+			answer = reflect.DeepEqual(left, right)
+		default:
+			return fmt.Errorf("Unsupported operation: %v", event.Operator.OperatorVariant)
+		}
+	} else {
+		// @TODO: Where are the implementations for these for builtin stuff (numbers mainly)
+		switch event.Operator.OperatorVariant.(type) {
+		case OperatorLt:
+			answer = leftCmp.Lt(rightCmp)
+		case OperatorLeq:
+			answer = leftCmp.Lt(rightCmp) || leftCmp.Equal(rightCmp)
+		case OperatorGt:
+			answer = rightCmp.Lt(leftCmp)
+		case OperatorGeq:
+			answer = !leftCmp.Lt(rightCmp)
+		case OperatorEq:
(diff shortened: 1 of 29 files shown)


languages/go/query.go
```diff
@@ -249,25 +239,37 @@ func (q Query) handleExternalOp(event types.QueryEventExternalOp) error {
...
-	case OperatorGeq:
-		answer = !leftCmp.Lt(rightCmp)
-	case OperatorEq:
-		answer = leftCmp.Equal(rightCmp)
-	case OperatorNeq:
-		answer = !leftCmp.Equal(rightCmp)
-	default:
-		return fmt.Errorf("Unsupported operation: %v", event.Operator.OperatorVariant)
+
+	leftCmp, leftOk := left.(interfaces.Comparer)
```
samscott89, line 245: I think we want to make this so that if _either_ the left or the right implements `Comparer`, it can use that implementation of `Comparer`. And update the `Comparer` interface itself so that it takes in `interface{}` as the argument.

That way you can implement comparer for typedefs in Go in a sensible way. There's a little more information in https://github.com/osohq/oso/issues/795#issue-838921051 but the short of it is, IMO the ideal Go interface would be:

```go
type Comparer interface {
	// Should return `true` when values are equal; `false` when not equal.
	Equal(other interface{}) bool
	// Should return `true` when the comparer is less than `other`; `false` when not less than.
	Lt(other interface{}) bool
}
```

User implements:
```go
type MyString string

func (left MyString ) Equal(right interface{}) bool {
    if other, ok := right.(string); ok    {
		return left == other
	}
}
```

It needs to check that at least one of the two operands implementing Comparer, because you might end up with: `new MyString("abc") == "abc"` or you might have `"abc" == new MyString("abc")` and both should work without needing to implement `Comparer` for string

gw000: 👍 done, check it out; it makes things a little more complicated because you can't just negate opposite order relations when x and y might have totally different types.

languages/java/oso/src/main/java/com/osohq/oso/Host.java
```diff
@@ -167,15 +167,10 @@ public boolean isSubclass(String leftTag, String rightTag) {
...
-  public boolean unify(long leftId, long rightId) throws Exceptions.UnregisteredInstanceError {
-    Object left = getInstance(leftId);
-    Object right = getInstance(rightId);
-    if (left == null) {
-      return right == null;
-    } else {
-      return left.equals(right);
-    }
+  public boolean operator(String op, List<Object> initargs) throws Exceptions.OsoException {
+    Object l = initargs.get(0), r = initargs.get(1);
```
dhatch, line 171: `left` and `right` would be clearer names. We try to avoid abbreviations.

languages/java/oso/src/main/java/com/osohq/oso/Host.java
```diff
@@ -167,15 +167,10 @@ public boolean isSubclass(String leftTag, String rightTag) {
...
+    if (op.equals("Eq")) return l == null ? l == r : l.equals(r);
```
dhatch, line 172: This would be easier to read if it was less compressed, like the previous implementation of `unify`

languages/python/oso/polar/host.py
```diff
@@ -76,7 +76,7 @@ def make_instance(self, name, args, kwargs, id):
         cls = self.get_class(name)
         try:
             instance = cls(*args, **kwargs)
-        except TypeError as e:
+        except Exception as e:
```
dhatch, line 79: Why did this exception type change?

languages/ruby/lib/oso/polar/host.rb
```diff
@@ -149,6 +149,33 @@ def make_instance(cls_name, args:, kwargs:, id:)
...
+      #
+      # @param op [String] operation to perform.
+      # @param args [Array<Object>] left and right args to operation.
+      # @raise [PolarRuntimeError] if operation fails or is unsupported.
+      # @return [Boolean]
+      def operator(operation, args)
+        op = OPS[operation]
+        raise PolarRuntimeError, "Unsupported external operation '#{l.class} #{operation} #{r.class}'" if op.nil?
+
+        l, r = args
```
dhatch, line 171: Same comment about 1 letter variable names as in Java.

polar-core/src/vm.rs
```diff
@@ -2318,44 +2309,16 @@ impl PolarVirtualMachine {
...
-                if left_instance != right_instance {
-                    self.push_goal(Goal::UnifyExternal {
-                        left_instance_id: *left_instance,
-                        right_instance_id: *right_instance,
-                    })?;
-                }
+            // If either operand is an external instance, let the host
+            // compare them for equality. This handles unification between
+            // "equivalent" host and native types transparently.
+            (Value::ExternalInstance(_), _) | (_, Value::ExternalInstance(_)) => {
```
dhatch, line 2315: We should keep the previous case that compared both instances by ID to avoid an external event if the operators are both external instances.

languages/java/oso/src/main/java/com/osohq/oso/Host.java
```diff
@@ -167,15 +167,10 @@ public boolean isSubclass(String leftTag, String rightTag) {
...
-  /** Check if two instances unify. */
-  public boolean unify(long leftId, long rightId) throws Exceptions.UnregisteredInstanceError {
-    Object left = getInstance(leftId);
-    Object right = getInstance(rightId);
-    if (left == null) {
-      return right == null;
-    } else {
-      return left.equals(right);
-    }
+  public boolean operator(String op, List<Object> initargs) throws Exceptions.OsoException {
```
dhatch, line 170: Why is this called `initargs` rather than just `args` like other languages?

dhatch (changes requested):
Some minor comments but overall looks good.

I didn't check the Go logic very closely, but I can if that's helpful.

languages/python/oso/polar/host.py
gw000: Because if python throws another error type, like a `ValueError` for example, it won't be caught and polar will crash. Try `x = new Integer("")`. As an aside, `x = new List(x)` will cause a segfault even with this change, but I haven't investigated why.

polar-core/src/vm.rs
gw000: 👍 it's back in

languages/java/oso/src/main/java/com/osohq/oso/Host.java
gw000: I was trying to be consistent with `makeExternal` but in hindsight it's obviously short for "initializer args" which makes no sense here. Changed it :)

languages/python/oso/polar/host.py
gw000: This btw is the "improves Python error handling" change mentioned up top. Fwiw I think it makes the python library's behavior consistent with ruby.

dhatch: > As an aside, x = new List(x) will cause a segfault even with this change, but I haven't investigated why.

Ooof could you open a bug for this?

languages/java/oso/src/main/java/com/osohq/oso/Host.java
dhatch: Ohh I didn't see `makeExternal`.

dhatch (approved):
lgtm!

languages/python/oso/polar/host.py
gw000: Done

PR #1025, opened 2021-07-22
Debugger: show correct value for non-temporary names

gw000:
Automatically map user variables to temporary names.

Screen Shot 2021-07-23 at 6 22 21 PM

PR checklist:

- [x] Added changelog entry.

### DIFF
--- a/polar-core/src/debugger.rs
+++ b/polar-core/src/debugger.rs
@@ -293,18 +293,23 @@ impl Debugger {
                 if parts.len() > 1 {
                     let vars: Vec<Binding> = parts[1..]
                         .iter()
-                        .map(|nom| {
-                            let var = Symbol::new(nom);
+                        .map(|name| {
+                            // *** variable name mapping ***
+                            // if the requested variable is bound, then we return that binding.
+                            // otherwise, we look for the matching bound temp variable with the
+                            // highest numeric component in its name, and return that binding
+                            // if we find it. otherwise, show that the variable is unbound.
+                            let var = Symbol::new(name);
                             let bindings = vm.bindings(true);
                             bindings.get(&var).cloned().map_or_else(|| {
-                                let pref = KnowledgeBase::temp_prefix(nom);
+                                let prefix = KnowledgeBase::temp_prefix(name);
                                 bindings.keys()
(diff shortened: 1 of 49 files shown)


polar-core/src/debugger.rs
```diff
@@ -292,12 +293,20 @@ impl Debugger {
                 if parts.len() > 1 {
                     let vars: Vec<Binding> = parts[1..]
                         .iter()
-                        .map(|var| {
-                            let var = Symbol::new(var);
-                            let value = vm.bindings(true).get(&var).cloned().unwrap_or_else(|| {
-                                Term::new_temporary(Value::Variable(Symbol::new("<unbound>")))
-                            });
-                            Binding(var, value)
+                        .map(|nom| {
```
saolsen, line 296: What is `nom` here?

polar-core/src/kb.rs
```diff
@@ -45,16 +45,18 @@ impl KnowledgeBase {
         self.id_counter.clone()
     }
 
+    pub fn temp_prefix(nom: &str) -> String {
```
saolsen, line 48: Could we maybe just add a comment explaining what this is for?

polar-core/src/debugger.rs
```diff
@@ -292,12 +293,20 @@ impl Debugger {
...
+                            let var = Symbol::new(nom);
+                            let bindings = vm.bindings(true);
+                            bindings.get(&var).cloned().map_or_else(|| {
+                                let pref = KnowledgeBase::temp_prefix(nom);
```
saolsen, line 300: Lengthening this to `prefix` would help me too.

saolsen (approved):
Looks good, I'm excited to use this it would have come in handy last week for me.
I had a little trouble following how it worked, I think maybe a few comments and more descriptive variable names might have helped. Made a few suggestions.

polar-core/src/kb.rs
gw000: 👍

polar-core/src/debugger.rs
gw000: 👍

polar-core/src/debugger.rs
gw000: Just a synonym for name, changed it.

PR #1027, opened 2021-07-23
Debugger: break on error

gw000:
Just what it says! Continuing past an error isn't possible, but the stack/bindings at the point where the error is thrown can be examined.

Screen Shot 2021-07-23 at 6 15 47 PM

PR checklist:

- [x] Added changelog entry.


saolsen (approved):
This is awesome. Was just playing around with it and it's super useful.

My only nitpick would be the term "CAUGHT ERROR". I think maybe "ENCOUNTERED ERROR" or even just "ERROR" would be more appropriate.

As a followup to this and the temp vars PR we need to update the debugger docs page to add this new command and update the section on `var`.

PR #1029, opened 2021-07-26
Add Rust roles getting started guide

gw000:
Also takes out the "load policies before enabling roles" callouts from the python/ruby docs, since we actually do revalidate the roles config.

### DIFF
--- a/docs/content/rust/guides/roles/getting-started.md
+++ b/docs/content/rust/guides/roles/getting-started.md
@@ -1,0 +1,281 @@
+---
+title: Getting started
+weight: 2
+description: >
+    Get started with Oso Roles for Rust
+---
+
+[rust-polar-classes]: reference/polar/classes.html
+
+# Getting started
+
+When managing access to resources within an application, it can be useful to
+group permissions into **roles** and assign these roles to users. This is
+known as **Role-Based Access Control (RBAC).** The Oso library
+comes with built-in configuration for role-based access control.
+
+In this guide, we'll walk through the basics of starting to use the
+roles feature.
+
+## Setting up the Oso instance
+
+First, we'll cover some of the basics of integrating Oso into your
+application.
+
+The `oso::Oso` struct is the entrypoint to using Oso in our application.
+We usually will have a global instance that is created
+during application initialization and shared across requests.
+
+```rust
+use oso::Oso;
+
+fn init_oso() -> Oso {
+  let oso = Oso::new();
+```
+
+### Using Rust types with Oso
+
+Next, register any Rust types you'll be using in your policies.
+
+{{% callout "Rust type configuration" "blue" %}}
+
+Rust structs and enums will need
+[some extra configuration][rust-polar-classes] to work with Oso.
+
+{{% /callout %}}
+
+
+```rust
+  // ... register classes used by the policy (see below)
+  oso.register_class(User::get_polar_class());
+  oso.register_class(Org::get_polar_class());
+  oso.register_class(OrgRole::get_polar_class());
+```
+
+### Loading our policy
+
+Oso uses the [Polar language](/reference/polar/polar-syntax) to define authorization
+policies. An authorization policy specifies what requests are allowed and what data a
+user can access. The policy is stored in a Polar file, along with your code.
+
+Load the policy at application start using the `Oso::load_file()` function:
+
+```rust
+  // load the policy from a file
+  oso.load_file('app/policy/authorization.polar');
+```
+
+### Enable Oso Roles
+
+In order to enable the built-in roles features, we call the
+`Oso::enable_roles` function:
+
+{{% callout "Load policies before enabling roles" "blue" %}}
+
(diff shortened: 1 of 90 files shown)


docs/content/rust/guides/roles/getting-started.md
```diff
@@ -0,0 +1,274 @@
...
+
+Oso uses the [Polar language](/reference/polar/polar-syntax) to define authorization
+policies. An authorization policy specifies what requests are allowed and what data a
+user can access. The policy is stored in a Polar file, along with your code.
+
+Load the policy at application start using the `Oso::load_file()` function:
+
+```rust
+  // load the policy from a file
+  oso.load_file('app/policy/authorization.polar');
```
gkaemmer, line 65: I think we want double-quotes here. Also (minor nit) but I chose this `app/policy/authorization.polar` path because it's where the file would be in a ruby setup, but probably for the rust example we can just use:
```suggestion
  oso.load_file("authorization.polar");
```

docs/content/rust/guides/roles/getting-started.md
```diff
@@ -0,0 +1,274 @@
...
+
+### Enable Oso Roles
+
+In order to enable the built-in roles features, we call the
+`Oso::enable_roles` function:
+
+```rust
+  oso.enable_roles();
+  oso
+}
```
gkaemmer, line 76: This is consistent with our other guides as it is now so this is probably fine, but what do you think about including the whole `init_oso` method in one snippet to make it easier to copy/paste?

I think it's particularly noticeable in rust because the snippets are each a "part" of the function and include unmatched curly braces.

docs/content/rust/guides/roles/getting-started.md
```diff
@@ -0,0 +1,274 @@
...
+Let's walk through the resource definition for `Org`.
+
+```polar
+resource(_type: Org, "org", actions, roles) if
+```
+
+The rule head has 4 parameters:
+
+- `_type` is the Rust class the resource definition is associated with.
+  **NOTE**: you must have registered this class with `OSO.register_class` before
```
gkaemmer, line 105:
```suggestion
  **NOTE**: you must have registered this class with `Oso::register_class` before
```

docs/content/rust/guides/roles/getting-started.md
```diff
@@ -0,0 +1,274 @@
...
+    // ...
+  }
+}
+```
+
+And the `actor_has_role_for_resource` would be implemented as:
+
+```polar
+actor_has_role_for_resource(actor, role_name, resource) if
+    role in actor.roles and
```
gkaemmer, line 216: Does this need parentheses? Like `role in actor.roles()`? It's possible it doesn't, but just want to make sure

docs/content/rust/guides/roles/getting-started.md
```diff
@@ -0,0 +1,274 @@
...
+```
+
+The `"owner"` role now implies the `"member"` role. Any user with the
+`"owner"` role will be granted all permissions associated with both
+roles.
+
+Here's the full `Org` resource definition from the GitClub example app:
+
+{{< literalInclude
+    path="examples/gitclub-sqlalchemy-flask-react/backend/app/authorization.polar"
```
gkaemmer, line 259: Hmm (not related to this PR) I wonder if these `literalIncludes` will break with the repo name change to `gitclub`. I'll file a pull request to fix all those.

docs/content/rust/guides/roles/getting-started.md
```diff
@@ -0,0 +1,274 @@
...
+    from="docs: begin-org-resource"
+    to="docs: end-org-resource"
+    gitHub="https://github.com/osohq/gitclub-sqlalchemy-flask-react"
+    linenos=true
+>}}
+
+Notice the `"repo:reader"` and `"repo:admin"` implications. These are
+roles defined on another resource, `Repo`. In the next guide, we'll
```
gkaemmer, line 267: We had a question yesterday in slack about this sentence in the ruby docs: https://oso-oss.slack.com/archives/C017KSBTD4M/p1627313868367300

Should we remove this paragraph for now and maybe replace with a link to the example app code?

docs/content/rust/guides/roles/getting-started.md
gw000: oops thank you!

docs/content/rust/guides/roles/getting-started.md
gw000: 👍

docs/content/rust/guides/roles/getting-started.md
gw000: Good question, I noticed that yesterday too. Maybe we should just write the next guide? Anyone else have any thoughts?

docs/content/rust/guides/roles/getting-started.md
gw000: 👍

saolsen (approved):

PR #1030, opened 2021-07-27
Change feedback to be thumbs up/down

samraper:
Not tested E2E, blocked on getting a new database in notion

### DIFF
--- a/docs/themes/oso-webpack/index.js
+++ b/docs/themes/oso-webpack/index.js
@@ -1,4 +1,4 @@
-import {postIntegrationRequest} from './backend';
+import {postIntegrationRequest, postFeedback} from './backend';
 import {get, set} from './localStorage';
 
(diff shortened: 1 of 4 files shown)


docs/themes/oso-webpack/index.js
```diff
@@ -356,3 +356,23 @@ function makePromptsUnselectable() {
     });
   })
 };
+
+window.recordFeedback = (isUp) => {
+  recordFeedback(isUp).then(() => {
```
dhatch, line 361: Could we disable the input before the request is complete? The notion recorder can be a little slow so this would give more immediate feedback to the user. Ideally we'd have a submitting state but I don't think that's really necessary right now.

dhatch (approved):
Small comment but overall LGTM assuming it works with the database. Sent you credentials for that in quill.

PR #1034, opened 2021-07-27
Improve circular data handling

gw000:
Right now trying to create circular data in Polar will cause a stack overflow due to unbounded recursion:

Screen Shot 2021-07-27 at 6 48 57 PM

This branch changes the dereferencing code to handle circular data a little more gracefully

Screen Shot 2021-07-27 at 6 55 33 PM

which still isn't super useful but probably better than crashing.

To manage this it does at least two slightly iffy things:

- It completely replaces `BindingManager::deref` with a slightly refactored `BindingManager::deep_deref`;
- To detect reference cycles it uses the same hash-set-of-hash-values-so-we-don't-have-to-clone strategy as the simplifier.

Local benchmarks don't seem to show much performance impact from deep-derefing everything, but someone else should test that also. A bigger concern would be if `deref` instead of `deep_deref` is actually necessary for some reason that's not covered by the tests.

### DIFF
--- a/polar-core/src/bindings.rs
+++ b/polar-core/src/bindings.rs
@@ -4,7 +4,7 @@
 use std::collections::{HashMap, HashSet};
 
 use crate::error::{PolarResult, RuntimeError};
-use crate::folder::{fold_term, Folder};
+use crate::folder::{fold_list, fold_term, Folder};
 use crate::formatting::ToPolarString;
 use crate::terms::{has_rest_var, Operation, Operator, Symbol, Term, Value};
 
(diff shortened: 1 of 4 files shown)


polar-core/src/bindings.rs
```diff
@@ -787,37 +780,6 @@ mod test {
         }
     }
 
-    #[test]
-    fn deref() {
```
dhatch, line 791: This test also covers some other logic in the binding manager, can we keep it but just replace `deref` with `deep_deref`?

polar-core/src/vm.rs
```diff
@@ -719,7 +719,7 @@ impl PolarVirtualMachine {
     /// The exception is for lists, so that we can correctly handle rest variables.
     /// We also support cycle detection, in which case we return the original term.
     fn deref(&self, term: &Term) -> Term {
```
dhatch, line 721: I don't think we need a `deref` function and a `deep_deref` function if they both do the same thing. What do you think?

dhatch (approved):
Small comments, lgtm. I don't need to review again after you address unless you have a question!

polar-core/src/bindings.rs
gw000: 👍

polar-core/src/vm.rs
gw000: No, of course I agree. I'll do what you suggested and try to replace `deref` with `variable_state` where possible though.

gw000: (there turned out to be only one such case)

PR #1028, opened 2021-07-26
Add language server implementation to VSCode integration

samscott89:
Thanks to @feliksleus for helping with the implementation!

PR checklist:

- [ ] Add tests for polar analyzer
- [ ] Find and address any blocking bugs
- [ ] Add changelog entry.

PR #1035, opened 2021-07-28
Make it possible to remove files from knowledge base.

samscott89:
Core changes needed to support #1028 .

- Make it possible to _remove_ a file (we could consider exposing this functionality, but not in this PR).
- Add sources for rules.

Theoretically, there's a user-facing change which is that previously if a policy failed to load, then all rules loaded up until that point would have been stored. So if you caught that error and tried again you might end up with duplicate rules?!?

### DIFF
--- a/polar-core/src/kb.rs
+++ b/polar-core/src/kb.rs
@@ -97,7 +97,7 @@ impl KnowledgeBase {
         Ok(src_id)
     }
 
-    pub fn clear_sources(&mut self) {
+    pub fn clear_rules(&mut self) {
         self.rules.clear();
         self.sources = Sources::default();
         self.inline_queries.clear();
--- a/polar-core/src/kb.rs
+++ b/polar-core/src/kb.rs
@@ -105,14 +105,23 @@ impl KnowledgeBase {
         self.loaded_files.clear();
     }
 
+    /// Removes a file from the knowledge base by finding the associated
+    /// `Source` and removing all rules for that source, and
+    /// removes the file from loaded files.
+    ///
+    /// Optionally return the source for the file, returning `None`
+    /// if the file was not in the loaded files.
     pub fn remove_file(&mut self, filename: &str) -> Option<String> {
         self.loaded_files
             .get(filename)
             .cloned()
-            .map(|src_id| self.remove_source(Some(filename.to_string()), src_id))
+            .map(|src_id| self.remove_source(src_id))
     }
 
-    pub fn remove_source(&mut self, filename: Option<String>, source_id: u64) -> String {
+    /// Removes a source from the knowledge base by finding the associated
+    /// `Source` and removing all rules for that source. Will
+    /// also remove the loaded files if the source was loaded from a file.
+    pub fn remove_source(&mut self, source_id: u64) -> String {
         // remove from rules
         self.rules.retain(|_, gr| {
             let to_remove: Vec<u64> = gr.rules.iter().filter_map(|(idx, rule)| {
@@ -134,8 +143,7 @@ impl KnowledgeBase {
             .sources
             .remove_source(source_id)
             .expect("source doesn't exist in KB");
-
-        assert_eq!(source.filename, filename);
+        let filename = source.filename;
 
         // remove queries
         self.inline_queries
--- a/polar-core/src/kb.rs
+++ b/polar-core/src/kb.rs
@@ -88,7 +88,7 @@ impl KnowledgeBase {
     pub fn add_source(&mut self, source: Source) -> PolarResult<u64> {
         let src_id = self.new_id();
         if let Some(ref filename) = source.filename {
-            self.check_file(&source.src, &filename)?;
+            self.check_file(&source.src, filename)?;
             self.loaded_content
                 .insert(source.src.clone(), filename.to_string());
             self.loaded_files.insert(filename.to_string(), src_id);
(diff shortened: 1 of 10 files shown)


polar-core/src/kb.rs
```diff
@@ -75,4 +85,102 @@ impl KnowledgeBase {
...
+    }
+
+    pub fn remove_file(&mut self, filename: &str) -> Option<String> {
+        self.loaded_files
+            .get(filename)
+            .cloned()
+            .map(|src_id| self.remove_source(Some(filename.to_string()), src_id))
+    }
+
+    pub fn remove_source(&mut self, filename: Option<String>, source_id: u64) -> String {
```
dhatch, line 116: Why is filename optional?

dhatch: Does this function need to be public? Can we just call `remove_file` from `polar.rs` instead?

polar-core/src/rules.rs
```diff
@@ -16,24 +18,73 @@ impl Parameter {
...
+        }
+    }
 }
 
 pub type Rules = Vec<Arc<Rule>>;
 
 type RuleSet = BTreeSet<u64>;
 
-#[derive(Clone, Default, Debug)]
+#[derive(Clone, Default, Debug, Deserialize, Serialize)]
```
dhatch, line 87: Why does this need to be serializable? It's an internal data structure to optimize looking up rules. Are we transferring the index and using it somewhere else?

polar-core/src/rules.rs
```diff
@@ -94,10 +152,10 @@ impl RuleIndex {
     }
 }
 
-#[derive(Clone)]
+#[derive(Clone, Deserialize, Serialize)]
```
dhatch, line 155: Same question about serialization here.

dhatch (changes requested):
I don't think we should make it possible for users to remove files. They should just make a new instance if they want to change what policy is loaded. (Obviously we have clear rules but even that I don't like. I think we just added it to make testing easier in instances where the Oso instance is global)

polar-core/src/polar.rs
```diff
@@ -160,109 +155,80 @@ impl Polar {
...
-                        .rules
-                        .entry(name.clone())
-                        .or_insert_with(|| GenericRule::new(name, vec![]));
-                    generic_rule.add_rule(Arc::new(rule));
-                }
-                parser::Line::Query(term) => {
-                    kb.inline_queries.push(term);
+        let source_id = kb.add_source(source.clone())?;
+
+        // extract this into a seperate function to we can catch all errors
```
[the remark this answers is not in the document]
samscott89: Nope. that's a description of what this is doing.

polar-core/src/polar.rs
```diff
@@ -160,109 +155,80 @@ impl Polar {
...
+        match load_source(source_id, &source, &mut kb) {
+            Ok(warnings) => {
+                self.messages.extend(warnings.iter().map(|m| Message {
+                    kind: MessageKind::Warning,
+                    msg: m.to_owned(),
+                }));
+                Ok(())
+            }
+            Err(e) => {
+                kb.remove_source(source.filename, source_id);
```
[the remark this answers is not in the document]
samscott89: Hmm, yeah that's a lot simpler. You can still do the rewrites as you go, but just defer inserting them into the KB until the very end.

polar-core/src/rules.rs
samscott89: Thanks, good catch on these. Not needed.

polar-core/src/kb.rs
samscott89: Yes. And  we can merge these together.

polar-core/src/polar.rs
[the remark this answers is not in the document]
samscott89: So we actually depend on the sources being inserted before parsing to make good error messages. So this seems like the only way to achieve this without rewriting a ton of things.

polar-core/src/kb.rs
samscott89: Okay, we still need this. The problem is if you load a raw string in that has errors, then you need the source in the KB so you can print nice errors, but then you want to remove the source. But there's no filename.

samscott89:
> I don't think we should make it possible for users to remove files. They should just make a new instance if they want to change what policy is loaded. (Obviously we have clear rules but even that I don't like. I think we just added it to make testing easier in instances where the Oso instance is global)

Agree we shouldn't expose this to users if we don't yet have a use case for it. This PR only makes this possible for anyone using the polar-core crate itself.

polar-core/src/polar.rs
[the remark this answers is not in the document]
samscott89: Updated the comment to be clearer

polar-core/src/kb.rs
dhatch: Okay can we make this function just take `source_id` as an argument then so that it's clear when to use `remove_source` and when to use `remove_file`?

We can look at the values in `loaded_files` to clean up the filenames without needing the filename to be passed in.

dhatch (approved):
One more suggestion about `remove_source`. Would prefer to include that before merging but if you don't agree or I'm missing something merge w/o.

polar-core/src/kb.rs
samscott89: Yeah, that seems reasonable. Let me look.

samscott89: Good suggestion. Removed a redundant assert too.

PR #1037, opened 2021-07-29
Lazily print reprs in log, error, and debugging output

gkaemmer:
Only python for now. Once I get a review, I can port to ruby and potentially node. Other languages (I think) are unlikely to have slow reprs, so it's less important to do this.

PR checklist:

- [x] Added changelog entry.

### DIFF
--- a/languages/python/oso/polar/ffi.py
+++ b/languages/python/oso/polar/ffi.py
@@ -50,23 +50,19 @@ class Polar:
         self.process_messages()
         self.check_result(result)
 
-    def new_query_from_str(self, query_str, enrich_message):
+    def new_query_from_str(self, query_str):
         new_q_ptr = lib.polar_new_query(self.ptr, to_c_str(query_str), 0)
         self.process_messages()
         query = self.check_result(new_q_ptr)
-        query = Query(query)
-        query.set_message_enricher(enrich_message)
-        return query
+        return Query(query)
 
-    def new_query_from_term(self, query_term, enrich_message):
+    def new_query_from_term(self, query_term):
         new_q_ptr = lib.polar_new_query_from_term(
             self.ptr, ffi_serialize(query_term), 0
         )
         self.process_messages()
         query = self.check_result(new_q_ptr)
-        query = Query(query)
-        query.set_message_enricher(enrich_message)
-        return query
+        return Query(query)
 
     def next_inline_query(self):
         q = lib.polar_next_inline_query(self.ptr, 0)
--- a/languages/python/oso/polar/ffi.py
+++ b/languages/python/oso/polar/ffi.py
@@ -157,7 +157,7 @@ class Query:
         self.enrich_message = enrich_message
 
     def check_result(self, result):
-        return check_result(result, lambda msg: self.enrich_message(msg))
+        return check_result(result, self.enrich_message)
 
     def process_messages(self):
         assert self.enrich_message, (
--- a/languages/python/oso/polar/ffi.py
+++ b/languages/python/oso/polar/ffi.py
@@ -88,16 +88,23 @@ class Query:
             value = ffi.NULL
         else:
             value = ffi_serialize(value)
-        check_result(lib.polar_call_result(self.ptr, call_id, value), self._temp_process_message)
+        check_result(
+            lib.polar_call_result(self.ptr, call_id, value), self._temp_process_message
+        )
 
     def question_result(self, call_id, answer):
         answer = 1 if answer else 0
-        check_result(lib.polar_question_result(self.ptr, call_id, answer), self._temp_process_message)
+        check_result(
+            lib.polar_question_result(self.ptr, call_id, answer),
+            self._temp_process_message,
+        )
 
     def application_error(self, message):
         """Pass an error back to polar to get stack trace and other info."""
         message = to_c_str(message)
-        check_result(lib.polar_application_error(self.ptr, message), self._temp_process_message)
+        check_result(
+            lib.polar_application_error(self.ptr, message), self._temp_process_message
+        )
 
     def next_event(self):
         event = lib.polar_next_query_event(self.ptr)
--- a/languages/python/oso/polar/ffi.py
+++ b/languages/python/oso/polar/ffi.py
@@ -88,7 +88,9 @@ class Polar:
         return check_result(result, self.enrich_message)
 
     def process_messages(self):
-        assert self.enrich_message, "No message enricher on this instance of FfiPolar. You must call set_message_enricher before using process_messages."
+        assert (
+            self.enrich_message
+        ), "No message enricher on this instance of FfiPolar. You must call set_message_enricher before using process_messages."
         for msg in process_messages(self.next_message):
             print(self.enrich_message(msg))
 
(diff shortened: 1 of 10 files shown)


languages/python/oso/polar/ffi.py
```diff
@@ -14,70 +17,78 @@ def __del__(self):
...
     def clear_rules(self):
         """Clear all rules from the Polar KB"""
         result = lib.polar_clear_rules(self.ptr)
-        process_messages(self.next_message)
-        check_result(result)
+        self.process_messages()
+        self.check_result(result)
 
-    def new_query_from_str(self, query_str):
+    def new_query_from_str(self, query_str, host):
```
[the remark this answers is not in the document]
dhatch: Nvm I see. Would be good to add a comment about this though.

languages/python/oso/polar/ffi.py
```diff
@@ -14,70 +17,78 @@ def __del__(self):
...
             self.ptr, ffi_serialize(query_term), 0
         )
-        process_messages(self.next_message)
-        query = check_result(new_q_ptr)
-        return Query(query)
+        self.process_messages()
+        query = self.check_result(new_q_ptr)
+        return Query(query, host)
 
     def next_inline_query(self):
```
dhatch, line 63: This needs to use the same host class as the host for the inline query.

languages/python/oso/polar/polar.py
```diff
@@ -64,6 +64,7 @@ class Polar:
     def __init__(self, classes=CLASSES):
         self.ffi_polar = FfiPolar()
         self.host = Host(self.ffi_polar)
+        self.ffi_polar.host = self.host
```
dhatch, line 67: This could be something like `self.ffi_polar.set_message_enricher(callable)` if you want to take my suggestion from above.

languages/python/oso/polar/ffi.py
```diff
@@ -88,43 +99,50 @@ def call_result(self, call_id, value):
...
         # TODO(gj): Do we need to process_messages here?
-        process_messages(self.next_message)
-        check_result(result)
+        self.process_messages()
+        self.check_result(result)
+
+    def check_result(self, result):
+        return check_result(result, lambda msg: self.host.enrich_message(msg))
+
+    def process_messages(self):
```
dhatch, line 143: Let's add an assert here that `self.host is not None` to catch forgetting to set the `self.host` property after the constructor. This will help us catch that even if we don't write a test that emits messages.

languages/python/oso/tests/test_polar_roles.py
```diff
@@ -26,6 +26,10 @@ class Org(Base):  # type: ignore
     name = Column(String(), primary_key=True)
 
     def __repr__(self):
+        # TODO: remove this slowdown?
+        import time
```
dhatch, line 30: Remove before merging?

languages/python/oso/polar/host.py
```diff
@@ -190,10 +210,21 @@ def to_polar(self, v):
                     }
                 }
         else:
+            # BEGIN HACK:
```
dhatch, line 213: Thanks for documenting this.

languages/python/oso/polar/ffi.py
```diff
@@ -190,7 +208,10 @@ def process_messages(next_message_method):
         kind = message["kind"]
         msg = message["msg"]
 
+        if postprocess:
```
dhatch, line 211: You never used `postprocess`, right?

dhatch (changes requested):
A few small changes.

I left a longer inline comment about this, overall I'm a bit hesitant about introducing this circular dependency. I have some suggestions if you want to take them that could limit the damage a bit but I can't think of another approach that's much better.

languages/python/oso/polar/ffi.py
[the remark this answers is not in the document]
gkaemmer: I think this is clearer now that we always use `set_message_enricher`

languages/python/oso/polar/ffi.py
```diff
@@ -1,11 +1,14 @@
 import json
 
 from _polar_lib import ffi, lib
+from polar.host import Host
 
 from .errors import get_python_error
 
 
 class Polar:
+    host: Host
```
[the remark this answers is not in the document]
gkaemmer: Yeah I think that makes a lot of sense. I removed `host` as an attribute from the `ffi` classes, instead giving them both a `set_message_enricher` method that has to be called before `process_messages` can succeed.

languages/python/oso/polar/ffi.py
```diff
@@ -14,70 +17,78 @@ def __del__(self):
...
-        process_messages(self.next_message)
-        check_result(result)
+        self.process_messages()
+        self.check_result(result)
 
     def next_message(self):
         return lib.polar_next_polar_message(self.ptr)
 
+    def check_result(self, result):
+        return check_result(result, lambda msg: self.host.enrich_message(msg))
```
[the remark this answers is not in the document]
gkaemmer: TIL -- in node the method would end up unbound to the original `self.host`, but in python that works!

languages/python/oso/polar/polar.py
```diff
@@ -109,6 +109,7 @@ def join(separator, left, right):
                 query = self.ffi_polar.next_inline_query()
                 if query is None:  # Load is done
                     break
+                query.set_message_enricher(self.host.enrich_message)
```
dhatch, line 112: You'll need to use the copied host made on L114 here. If the inline query creates external instances they are attached to that one instead of `self.host`.

gkaemmer:
@dhatch this is ready for another look. Do you think I should make the ruby/node/etc changes in this PR or in follow-ons? Happy to do either.

languages/python/oso/polar/polar.py
```diff
@@ -151,6 +152,7 @@ def load_str(self, string, filename=None):
             if query is None:  # Load is done
                 break
             else:
+                query.set_message_enricher(self.host.enrich_message)
```
dhatch, line 155: Same here, use the copied host.

languages/python/oso/polar/polar.py
```diff
@@ -214,9 +218,10 @@ def repl(self, files=[]):
 
             host = self.host.copy()
             host.set_accept_expression(True)
+            ffi_query.set_message_enricher(host.enrich_message)
             result = False
             try:
-                query = Query(ffi_query, host=host).run()
+                query = Query(ffi_query).run()
```
dhatch, line 224: Why did you remove this `host=`?

languages/python/oso/polar/polar.py
dhatch: This is really error prone & these would be very hard errors to detect through testing. I think `query.Query` (not `ffi.Query`) can call `ffi.Query.set_message_enricher` in its initializer to ensure that the host classes on the two objects match.

dhatch (changes requested):
Few more things about host

languages/python/oso/polar/polar.py
gkaemmer: Oh wow good catch -- scary that this doesn't cause any CI failures. I think I got my wires crossed when removing the `host` argument from the `ffi.Query` constructor. Will revert this

languages/python/oso/polar/polar.py
gkaemmer: Ah yeah that makes a lot more sense. I just fixed that and its way clearer I think

languages/python/oso/polar/polar.py
dhatch: Yeah unfortunately the `host` handling in here is very error prone. (Before your PR it was already a huge rough spot). If you have any thoughts about cleaning it up now that you've looked at it a bit it might be worth us spending time on.

dhatch (approved):
This looks good to me, can you add a changelog entry before you merge?

Thanks for going through those cleanups! It's fine to do other languages as follow ups in a separate PR (& probably better that way - it's easier to review)

PR #1039, opened 2021-07-30
Fix grounding bugs

gw000:
Well, fundamentally I think there was only ever one grounding bug, but hopefully this should mostly fix it.

The problem in a nutshell was that when Polar tried to constrain an unbound variable, it would place that variable in what we call a "partial" state, which basically means it was bound to a list of constraints that would *only ever be changed by adding to it*, irrespective of whether the additional constraints were consistent or whether the variable should later have been bound to a value. So for example:

```polar
foo(a, b) if b > 0 and a = b;
bar(a, b) if a = b and b > 0;
# but ...
?= foo(1, b) and b = 2; # this succeeds, which is wrong!
?= bar(1, b) and b = 2; # this fails as expected
```

The bug arises when `foo` is called with a free second argument: the constraint `b > 0` then prevents `b` from being correctly unified with `a`.

The fix is basically to allow rebinding of partially bound variables to ground values, in which case the VM reevaluates the prior constraint to ensure that everything is still consistent.

PR checklist:

- [x] Added changelog entry.


polar-core/src/bindings.rs
```diff
@@ -153,6 +153,21 @@ impl BindingManager {
         Self::default()
     }
 
+    fn partial_bind(&mut self, partial: Operation, var: &Symbol, val: Term) -> PolarResult<Goal> {
+        match partial.ground(var, val.clone()) {
+            None => Err(RuntimeError::IncompatibleBindings {
```
samscott89, line 158: I'm not sure we should be using a runtime error here to mean "the grounding failed". Runtime errors are supposed to capture the VM entering invalid states, or similar. Would it be possible to instead return a backtrack goal?

polar-core/src/bindings.rs
```diff
@@ -179,35 +194,36 @@ impl BindingManager {
...
+                    let p = p.clone();
+                    let val = val.clone();
+                    goal = Some(self.partial_bind(p, var, val)?)
                 }
-                .into());
-            }
 
-            self.add_binding(var, val.clone());
+                BindingManagerVariableState::Bound(_) => {
+                    return Err(RuntimeError::IncompatibleBindings {
```
samscott89, line 210: Ditto here. I don't think this is a runtime error. This is the VM correctly determining that the current query path is false.

polar-core/src/bindings.rs
```diff
@@ -179,35 +194,36 @@ impl BindingManager {
...
+                        msg: format!("Cannot rebind {:?}", var),
```
samscott89, line 211: This is a good example of where runtime error is the correct behaviour -- the VM shouldn't be reaching this state where it attempts to rebind a value.

polar-core/src/bindings.rs
```diff
@@ -243,28 +259,35 @@ impl BindingManager {
...
+
+        let vars = op.variables();
+        let mut varset = vars.iter().collect::<HashSet<_>>();
+        for var in vars.iter() {
+            if let BindingManagerVariableState::Bound(val) = self._variable_state(var) {
+                varset.remove(var);
+                match op.ground(var, val) {
+                    Some(o) => op = o,
+                    None => {
+                        return Err(RuntimeError::IncompatibleBindings {
```
samscott89, line 278: Same thing here: not an error?

polar-core/src/bindings.rs
```diff
@@ -243,28 +259,35 @@ impl BindingManager {
...
-                        "Unexpected bound variable {var} in constraint. {var} = {val}",
-                        var = var,
-                        val = v
-                    );
+                BindingManagerVariableState::Partial(e) => op = e.clone().merge_constraints(op),
+                _ => {}
+            }
+        }
+
+        let vars = op.variables();
```
samscott89, line 270: Why is this a separate loop from the one above?

polar-core/src/bindings.rs
gw000: You're totally right, this is just pre-existing logic I didn't want to mess with. You can see in `vm.rs` most callers to `bind` check for an error and backtrack if one happens.

polar-core/src/bindings.rs
gw000: Because the first loop may have added new variables to the expression.

languages/python/sqlalchemy-oso/sqlalchemy_oso/auth.py
```diff
@@ -79,17 +79,22 @@ def get_field_type(model, field):
 
     combined_filter = None
     has_result = False
+    pkey = roles.get_pk(model)[0]
```
dhatch, line 82: We should remove this & use inspect.

languages/python/sqlalchemy-oso/sqlalchemy_oso/auth.py
```diff
@@ -79,17 +79,22 @@ def get_field_type(model, field):
...
     for result in results:
         has_result = True
 
         resource_partial = result["bindings"]["resource"]
-        filter, role_method = partial_to_filter(
-            resource_partial, session, model, get_model=oso.get_class
-        )
+        if isinstance(resource_partial, model):
+            filter = getattr(model, pkey) == getattr(resource_partial, pkey)
```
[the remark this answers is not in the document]
dhatch: You could also just call `partial_to_filter(Expression("Eq", [Variable("_this"), model]), ...)`

polar-core/benches/benchmarks/partial.rs
```diff
@@ -61,7 +61,7 @@ pub fn partial_and(c: &mut Criterion) {
 /// a partial that calls 5 rules and does 5 field comparisons.
 pub fn partial_rule_depth(c: &mut Criterion) {
     let mut group = c.benchmark_group("partial_rule_depth");
-    for n in &[1, 5, 10, 20, 40, 80, 100] {
+    for n in &[1, 5, 10, 20, 40] {
```
dhatch, line 64: Let's discuss this.

polar-core/src/bindings.rs
```diff
@@ -153,6 +153,21 @@ impl BindingManager {
...
+    fn partial_bind(&mut self, partial: Operation, var: &Symbol, val: Term) -> PolarResult<Goal> {
+        match partial.ground(var, val.clone()) {
+            None => Err(RuntimeError::IncompatibleBindings {
+                msg: "Grounding failed".into(),
+            }
+            .into()),
+            Some(grounded) => {
+                self.add_binding(var, val);
+                Ok(Goal::Query {
+                    term: grounded.into_term(),
```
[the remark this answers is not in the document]
dhatch: Nvm I see you implemented `merge_constraints` to avoid duplicates.

polar-core/src/partial/partial.rs
```diff
@@ -76,13 +76,16 @@ impl Operation {
     pub fn variables(&self) -> Vec<Symbol> {
         struct VariableVisitor {
             seen: HashSet<Symbol>,
-            vars: Vec<Symbol>,
+            vars: Vec<Symbol>, // FIXME gw
```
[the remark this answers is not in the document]
dhatch: 😭 😭 😭

dhatch (changes requested):
I left some comments. Most are fairly minor. I haven't gotten too much of a chance to think deeply on this besides the code. Some other things I'd like to do before merging:

- [ ] talk about perf impact
- [ ] think a bit more about edge cases... one that comes to mind is performing a grounding over an expression with an OR. Something like `not (x > 1 and x < 3) and x = 2`. Would both sides of the `or` run as choice points in the VM, and is this correct behavior? I think we would only need to find one compatible branch to be correct.
- [ ] think more about negation, although I think this is probably fine and we should just address negation issues as separate work

polar-core/src/vm.rs
```diff
@@ -2159,15 +2160,36 @@ impl PolarVirtualMachine {
...
-                        "cannot unify expressions directly `{}` = `{}`",
-                        left.to_polar(),
-                        right.to_polar()
-                    ),
-                ));
+            (Value::Expression(op), other) | (other, Value::Expression(op)) => {
+                match op {
+                    // this branch handles dot ops that were rewritten for inclusion
+                    // in a partial, but then queried again after the partial was bound;
+                    // cf. Vm::dot_op_helper(), Vm::bind()
```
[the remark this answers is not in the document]
gw000: https://en.wikipedia.org/wiki/Cf.

polar-core/src/vm.rs
```diff
@@ -2159,15 +2160,36 @@ impl PolarVirtualMachine {
...
-                return Err(self.type_error(
-                    left,
-                    format!(
-                        "cannot unify expressions directly `{}` = `{}`",
-                        left.to_polar(),
-                        right.to_polar()
-                    ),
-                ));
+            (Value::Expression(op), other) | (other, Value::Expression(op)) => {
+                match op {
```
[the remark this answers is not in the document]
gw000: That would exacerbate the performance issue even more though.

polar-core/src/bindings.rs
samscott89: Oh yeah, I was thinking of `OperationalError` which we have to represent these internal errors.

polar-core/src/bindings.rs
```diff
@@ -179,35 +194,36 @@ impl BindingManager {
     ///
     /// If a binding between two variables is made, and one is bound and the other unbound, the
     /// unbound variable will take the value of the bound one.
-    pub fn bind(&mut self, var: &Symbol, val: Term) -> PolarResult<()> {
+    pub fn bind(&mut self, var: &Symbol, val: Term) -> PolarResult<Option<Goal>> {
```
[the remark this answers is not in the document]
dhatch: The piece that's still missing is the `Option<Goal>`.

If `Some(Goal)` is returned, the calling VM must evaluate the goal to determine whether the binding is compatible.

polar-core/src/vm.rs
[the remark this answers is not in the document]
dhatch: Ok, it would be clearer to replace this with something like


> this branch handles dot ops that were rewritten for inclusion in a partial in Vm::dot_op_helper() but then queried again after the partial was bound by a call to Vm::bind()

For the same reasons that we try to avoid abbreviations. It makes reading the code harder if you don't know what `cf.` is.

polar-core/src/vm.rs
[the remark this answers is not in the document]
dhatch: Fair point on the rewriting all being in the VM. Let's leave it. FWIW this comment is coming from wanting to make the partial representation internal & not something the VM needs to worry about so that we are free to change to a more efficient representation than a vec of constraints. But, there are a bunch of other things that also have to happen to get there.

I'm not sure I follow why though on the performance issue. Because we'd iterate over the entire partial to produce the query goal?

dhatch (commented):
Resolved some comments.

dhatch:
Just doing some more testing locally... it seems there are still some cases that don't ground out properly when more than one variable is involved:

```
$ python -m polar
query> x = y and y > 3 and x = 2
y = Expression(And, [Expression(Gt, [2, 3])])
x = 2
# Expected result is false because 2 > 3 is false

query> x = y and not (x < 1 and x > 3) and x = 2
x = 2
y = Expression(And, [Expression(Unify, [2, Variable('_this')]), Expression(Or, [Expression(And, []), Expression(And, [])])])

# Expected result is x = 2, y =  2
```

gw000:
@dhatch those cases come out as expected for me. did you run `make python-build`?

Screen Shot 2021-08-03 at 1 18 42 PM

languages/python/sqlalchemy-oso/sqlalchemy_oso/auth.py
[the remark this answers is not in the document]
gw000: Changed, lmk if the new way seems right

dhatch (approved):
LGTM but could you consider below before merging:

I'd prefer if we just used macros like:

`term!(op!(...))`
`term!(str!(...))`

I think having two macros that are similar is confusing.

Another option would be something like...

`op_t!(...)`
`string_t!(...)`
`sym_t!(...)`

if you really want to keep them. This way there's a consistent naming & difference between the Value versions & the term versions.

PR #1042, opened 2021-08-02
Fix in operator & and/or precedence

gw000:
Sorry to drop these so close to release time but ...

- Fix the `in` operator to handle rest variables
- Make `or` have lower precedence than `and`, to match every other language

PR checklist:

- [x] Added changelog entry.


saolsen:
Making `and` have the lowest precedence was an intentional move so that rule bodies would always be an `and` and to match the way it works in prolog. I'm not really sure if that matters though or if we depend on it anywhere. We did that back when we had prolog syntax and didn't call them `and` and `or`. This has tricked users before so I'm into changing it, just a little worried that we depend on it in some subtle ways. I think it's probably fine if the tests pass. This is a big one to flag in the changelog too.

gw000:
Interesting. I'm happy to take that change out if we want to stick with that design choice, but it does seem pretty confusing. It may be internally convenient for rule bodies to always be an `and`, but that shouldn't be allowed to dictate the user experience IMO. It also breaks the analogy `&& : || :: * : +`.

polar-core/tests/integration_tests.rs
```diff
@@ -1568,8 +1568,18 @@ fn test_rest_vars() -> TestResult {
...
     qeval(&mut p, "not append([1,2,3], [4], [1,2,3])");
 
+    qeval(
+        &mut p,
+        "a = [1, *b] and b = [2, *c] and c=[3] and 1 in a and 2 in a and 3 in a",
+    );
+
     let a = &var(&mut p, "[*_] in [*a] and [*b] in [*_] and b = 1", "a")[0];
-    assert!(matches!(a, Value::List(b) if matches!(b[0].value(), Value::RestVariable(_))));
+    assert!(matches!(a,
```
gw000, line 1577: So I updated this test to expect a new expression, and the expression seems to be correct, but actually I'm a little confused about what we're trying to test here. The constraint described by the expression is basically "a is a list containing a list".

saolsen:
> Interesting. I'm happy to take that change out if we want to stick with that design choice, but it does seem pretty confusing. It may be internally convenient for rule bodies to always be an `and`, but that shouldn't be allowed to dictate the user experience IMO. It also breaks the analogy `&& : || :: * : +`.

Yeah I agree. I also don't think the rule body thing is even a problem after thinking about it, the whole body is wrapped in an `and` anyway so it'd be an `and` with an `or` in it and work just fine.

saolsen (changes requested):
There's a table in `formatting.rs` that needs to be updated that records all the precedences and uses them to output terms in polar syntax.

dhatch:
What's this part of the change?

> Fix the in operator to handle rest variables

Does it need to be in the changelog?

dhatch:
> What's this part of the change?
> 
> > Fix the in operator to handle rest variables
> 
> Does it need to be in the changelog?

Nvm.

dhatch:
Can you add a new test for the `in` behavior though? I don't see it covered in any here.

polar-core/src/polar.lalrpop
```diff
@@ -478,15 +478,15 @@ OrExp<T>: Value = {
     },
 }
 
-Exp2<T>: ValueOrLogical = {
+Exp1<T>: ValueOrLogical = {
```
dhatch, line 481: Can we swap this & `OrExp` w/ `Exp1` & `AndExp` so this file remains in precedence order after the change?

gw000:
precedence change isn't going out this release; other fix is in https://github.com/osohq/oso/pull/1047

PR #1044, opened 2021-08-02
CI Optimization: always upload build cache

gkaemmer:
By appending the git SHA to the cache key, we tell github to always upload a new version of the cache after the build step. Previously, we'd only upload new cache directories when Cargo.lock changed, so the test jobs weren't using the artifacts built by the build job.

It also has the added benefit that the build job runs with a more up-to-date version of the cache, so it finishes a bit faster too.

This speeds up tests by about ~1:30 when it needs to rebuild rust, but when it doesn't need to rebuild rust, the speedup is more like 3-4 mins:
Screen Shot 2021-08-02 at 3 40 48 PM


.github/workflows/test.yml
```diff
@@ -121,10 +121,52 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/cache@v2
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            target
+          key: ${{ runner.os }}-cargo-test-${{ hashFiles('**/Cargo.lock') }}
```
samscott89, line 134: I _think_ you want to differentiate the cache keys so both get cached separately.

saolsen:
Does this really speed it up? Both jobs still compile the core which is probably the majority of the time.

gkaemmer:
> Does this really speed it up? Both jobs still compile the core which is probably the majority of the time.

Yeah, seems to shave off about a minute from that first step -- indeed both jobs need to compile the core but it still is a bit faster to do them both at the same time.

Old:
Screen Shot 2021-08-02 at 1 54 32 PM

New:
Screen Shot 2021-08-02 at 1 54 24 PM

saolsen:
> > Does this really speed it up? Both jobs still compile the core which is probably the majority of the time.
> 
> Yeah, seems to shave off about a minute from that first step -- indeed both jobs need to compile the core but it still is a bit faster to do them both at the same time.
> 
> Old:
> Screen Shot 2021-08-02 at 1 54 32 PM
> 
> New:
> Screen Shot 2021-08-02 at 1 54 24 PM

dope

gkaemmer:
@saolsen Actually I'm settling on a different change after fiddling a bit. By appending the git SHA to the cache key, we tell github to always upload a new version of the cache after the `build` step. Previously, we'd only upload new cache directories when `Cargo.lock` changed, so the `test` jobs weren't using the artifacts built by the `build` job.

It also has the added benefit that the `build` job runs with a more up-to-date version of the cache, so it finishes a bit faster too.

This speeds up tests by about ~1:30 when it needs to rebuild rust, but when it _doesn't_ need to rebuild rust, the speedup is more like 3-4 mins.

Screen Shot 2021-08-02 at 3 40 48 PM

saolsen:
I don't really get this. Wouldn't appending the sha mean that we never hit the cache and it always rebuilds since each push is a new commit?

saolsen:
> I don't really get this. Wouldn't appending the sha mean that we never hit the cache and it always rebuilds since each push is a new commit?

ohhh n/m. I just looked up how `restore-keys` works. This is great!

saolsen (approved):

PR #1047, opened 2021-08-03
Fix in operator for rest variables

gw000:
(replaces https://github.com/osohq/oso/pull/1042)

before change:
Screen Shot 2021-08-03 at 11 24 42 AM

notice `d` receives two bindings, the second being `b`

after change:
Screen Shot 2021-08-03 at 11 25 34 AM

`d` now receives the expected 3 bindings.

PR checklist:

- [x] Added changelog entry.


polar-core/src/vm.rs
```diff
@@ -2042,82 +2042,71 @@ impl PolarVirtualMachine {
         assert_eq!(args.len(), 2);
         let item = &args[0];
         let iterable = &args[1];
+        let item_is_ground = item.is_ground();
 
-        match (item.value(), iterable.value()) {
-            (_, Value::List(list)) if list.is_empty() => {
-                // Nothing is in an empty list.
-                self.backtrack()?;
```
samscott89, line 2049: If anyone else is curious why this is no longer needed:

We used to iterate through the list, and push an alternative for every element of the list. Which meant that "no alternatives" just evaluated to true.

At some point we changed this to an iterator and `self.choose` which already handles the "no alternatives" case as meaning backtrack.

Cool that we can delete this code!

polar-core/tests/integration_tests.rs
```diff
@@ -1569,7 +1569,12 @@ fn test_rest_vars() -> TestResult {
     qeval(&mut p, "not append([1,2,3], [4], [1,2,3])");
 
     let a = &var(&mut p, "[*_] in [*a] and [*b] in [*_] and b = 1", "a")[0];
-    assert!(matches!(a, Value::List(b) if matches!(b[0].value(), Value::RestVariable(_))));
+    assert!(matches!(a,
```
gw000, line 1572: like I mentioned in the other PR, I kept this test but I'm not sure what it's supposed to be testing.

samscott89: https://github.com/osohq/oso/pull/644

samscott89 (approved):
LGTM! Good catch

polar-core/tests/integration_tests.rs
samscott89: Feel free to replace with a clearer test. Agree that this test isn't very easy to follow!

samscott89: List in list doesn't make much sense here.

Maybe changing them to unifies? `[*_] = [*a] and [*b] = [*_] and b = [1]` ?

gw000: can I replace it with a test that just ensures `a` and `b` get different bindings? the specific outcome we were/are testing for doesn't seem "correct" as much as just "not specifically incorrect".

PR #1056, opened 2021-08-05
Redirect use cases page to .com.

dhatch:


README.md
```diff
@@ -14,17 +14,31 @@
...
+customize as you see fit.
+
+Developers can typically write a working Oso policy in <5 minutes, add
+Oso to an app in <30 minutes, and use Oso to solve real authorization
+problems within a few hours. To get started, you add the library to your
+application, create a new Oso instance and load an Oso policy. You can
+mix and match any of Oso’s authorization APIs to implement features like
+roles with custom policies that you write to suit your application.
+
+Check out [Use Cases][use-cases] to learn more about how teams are using
```
dhatch, line 35: Only change was here.

dhatch: Rest is formatting. @gneray good with this change? Now redirects to `Use Cases` on .com. I changed the wording a bit because the past description sounded like you would find out about how to use Oso for things other than application authorization from following this link.

gneray (approved):
More stunning engineering right here

PR #1059, opened 2021-08-05
Diff docs

dhatch:


dhatch:
@samscott89 this is ready for review. I closed the other branch because GitHub Actions wasn't working. Once we merge we should mark the check as required.

.github/workflows/docs-diff.yml
```diff
@@ -0,0 +1,240 @@
...
+    if: ${{ needs.check-label.outputs.hadLabel == 'false' }}
+    steps:
+      - uses: actions/cache@v2
+        id: cache
+        with:
+          path: |
+            docs/public
+          key: docs-${{ github.sha }}
+      - uses: actions/checkout@v2
+        if: steps.cache.outputs.cache-hit != 'true'
```
samscott89, line 31: What are these?

dhatch: This caches the doc build so that we don't have to rebuild on every run. For example if you add the label after the first run of this, the docs will already be cached.

samscott89: Isn't the `actions/cache` steps doing the caching?

I'm specifically wondering about the `if: steps.cache.outputs.cache-hit != 'true'` multiple times. 

Is the idea that there will only be a cache hit if the job made it all the way to the end. Therefore a cache hit implies that the CI run was previously successful for the same files?
And we need to do the skipping because even with the docs build cached it still takes a little while to run?

dhatch: It still takes around 4 minutes to run even if the docs build is cached unless we skip.

samscott89: Okay. A little ugly, but seems like the best way to do that currently.

We're also going to need to keep this in sync with any other docs CI changes.

samscott89 (approved):
This seems worth trying out!

PR #1062, opened 2021-08-09
Combine timeout tests to reduce flakiness.

dhatch:
Rust runs all tests in parallel in threads, but environment variables are set at the process level. The `test_timeout` and `test_timeout_configuration` tests could step on each other if they overlap, causing somewhat frequent failures in CI.

### DIFF
--- a/polar-core/src/vm.rs
+++ b/polar-core/src/vm.rs
@@ -3718,18 +3718,15 @@ mod tests {
     }
 
     #[test]
-    fn test_timeout_configuration() {
+    fn test_timeout() {
         let vm = PolarVirtualMachine::default();
         assert!(vm.query_timeout_ms == DEFAULT_TIMEOUT_MS);
 
         std::env::set_var("POLAR_TIMEOUT_MS", "0");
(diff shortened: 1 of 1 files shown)


polar-core/src/vm.rs
```diff
@@ -3718,18 +3718,15 @@ mod tests {
...
 
     #[test]
-    fn test_timeout_configuration() {
+    fn test_timeout() {
         let vm = PolarVirtualMachine::default();
         assert!(vm.query_timeout_ms == DEFAULT_TIMEOUT_MS);
 
         std::env::set_var("POLAR_TIMEOUT_MS", "0");
         let vm = PolarVirtualMachine::default();
         std::env::remove_var("POLAR_TIMEOUT_MS");
```
gkaemmer, line 3727: Maybe we don't need this line anymore? Totally optional if you want to keep it so that the tests are still logically separated.

gkaemmer (approved):
Thank you for fixing this -- I actually didn't know these got parallelized in CI!

PR #1063, opened 2021-08-10
Rule prototypes

leina05:
Rule prototypes provide a way to constrain how certain rules must be defined. 

A rule prototype looks like this:
```python
type f(_a: Foo, _b: Bar);
```
It consists of the keyword `type` followed by what looks like a normal Polar rule head.
With the above rule prototype defined, any rule named `f` would have to match the prototype in order to be valid.

A rule matches a prototype if it has the same **arity** (number of parameters) as the prototype, and if each parameter matches the corresponding prototype parameter. This logic is defined in `validate_rules` in `kb.rs`.

A rule parameter matches a prototype parameter according to `isa` or `matches` logic. This logic is defined in `rule_params_match` in `kb.rs`.

Multiple prototypes can be defined for a given rule name.

When a rule fails to match any defined prototype, the user receives an error message that looks like:

```
polar.exceptions.ValidationError: Invalid rule: f(_x: Bad{}); Must match one of the following rule prototypes:
E           
E           f(_x: Integer{});
E               Failed to match because: Rule specializer Bad on parameter 1 must be a subclass of prototype specializer Integer
E           
E           f(_x: Foo{});
E               Failed to match because: Rule specializer Bad on parameter 1 must be a subclass of prototype specializer Foo
E           
E           f(_x: Foo{}, _y: Bar{});
E               Failed to match because: Different number of parameters. Rule has 1 parameter(s) but prototype has 2.
E            at line 1, column 11
E               Get help with Oso from our engineers: https://help.osohq.com/error/ValidationError
```

PR checklist:

- [ ] Added changelog entry.
- [x] Standardize naming of rule prototypes
- [x] Test `validate_rules`

### DIFF
--- a/languages/python/oso/polar/host.py
+++ b/languages/python/oso/polar/host.py
@@ -24,6 +24,7 @@ class Host:
         self,
         polar,
         classes=None,
+        class_ids=None,
         cls_names=None,
         instances=None,
         get_field=None,
@@ -34,6 +35,9 @@ class Host:
         self.ffi_polar = polar  # a "weak" handle, which we do not free
         self.classes = (classes or {}).copy()
         self.cls_names = (cls_names or {}).copy()
+        self.class_ids = (
+            class_ids or {}
+        ).copy()  # Map from class name (Python) => instance ID used every time the class is converted to polar
         self.instances = (instances or {}).copy()
         self.types = (types or {}).copy()
         self.fetchers = (fetchers or {}).copy()
@@ -66,6 +70,7 @@ class Host:
             self.ffi_polar,
             classes=self.classes,
             cls_names=self.cls_names,
+            class_ids=self.class_ids,
             instances=self.instances,
             get_field=self.get_field,
             types=self.types,
@@ -86,6 +91,7 @@ class Host:
             raise DuplicateClassAliasError(name, self.get_class(name), cls)
 
         self.classes[name] = cls
+        self.class_ids[cls] = self.cache_instance(cls)
         return name
 
     def get_instance(self, id):
@@ -241,25 +247,27 @@ class Host:
                     }
                 }
         else:
-            # BEGIN HACK:
-            # The polar core uses the .repr property to determine whether or not
-            # to allow Roles.role_allows to be called with unbound variables as
-            # arguments (only for sqlalchemy_oso)
-            # Because of this, we need to continue to send the repr for
-            # sqlalchemy_oso.roles.OsoRoles.Roles ONLY
+            instance_id = None
             repr_str = None
             import inspect
 
-            if (
-                inspect.isclass(v)
-                and "OsoRoles" in v.__qualname__
-                and v.__module__ == "sqlalchemy_oso.roles"
-            ):
-                repr_str = repr(v)
-            # END HACK
+            if inspect.isclass(v):
+                instance_id = self.class_ids.get(v)
+                # BEGIN HACK:
+                # The polar core uses the .repr property to determine whether or not
+                # to allow Roles.role_allows to be called with unbound variables as
+                # arguments (only for sqlalchemy_oso)
+                # Because of this, we need to continue to send the repr for
+                # sqlalchemy_oso.roles.OsoRoles.Roles ONLY
+                if (
+                    "OsoRoles" in v.__qualname__
+                    and v.__module__ == "sqlalchemy_oso.roles"
+                ):
+                    repr_str = repr(v)
+                # END HACK
             val = {
                 "ExternalInstance": {
-                    "instance_id": self.cache_instance(v),
+                    "instance_id": self.cache_instance(v, instance_id),
                     "repr": repr_str,
                 }
             }
--- a/languages/python/oso/polar/polar.py
+++ b/languages/python/oso/polar/polar.py
@@ -1,6 +1,7 @@
 """Communicate with the Polar virtual machine: load rules, make queries, etc."""
 
 from datetime import datetime, timedelta
+import inspect
 import os
 from pathlib import Path
 import sys
@@ -146,6 +147,18 @@ class Polar:
 
     def load_str(self, string, filename=None):
         """Load a Polar string, checking that all inline queries succeed."""
+        # Get MRO of all registered classes
+        # NOTE: not ideal that the MRO gets updated each time load_str is
+        # called, but since we are planning to move to only calling load once
+        # with the include feature, I think it's okay for now.
+        for (cls_name, cls) in self.host.classes.items():
+            mro = [
+                self.host.class_ids.get(c)
+                for c in inspect.getmro(cls)
+                if c in self.host.class_ids
+            ]
+            self.ffi_polar.register_mro(cls_name, mro)
+
         self.ffi_polar.load(string, filename)
 
         # check inline queries
@@ -257,6 +270,9 @@ class Polar:
             self.host.types[cls_name] = types
         if fetcher:
             self.host.fetchers[cls_name] = fetcher
+        self.ffi_polar.register_mro(
+            cls_name, self.host.to_polar(list(inspect.getmro(cls)))
+        )
 
     def register_constant(self, value, name):
         """Register `value` as a Polar constant variable called `name`."""
--- a/languages/python/oso/polar/polar.py
+++ b/languages/python/oso/polar/polar.py
@@ -270,9 +270,6 @@ class Polar:
             self.host.types[cls_name] = types
         if fetcher:
             self.host.fetchers[cls_name] = fetcher
-        self.ffi_polar.register_mro(
-            cls_name, self.host.to_polar(list(inspect.getmro(cls)))
-        )
 
     def register_constant(self, value, name):
         """Register `value` as a Polar constant variable called `name`."""
--- a/languages/python/oso/tests/test_polar.py
+++ b/languages/python/oso/tests/test_polar.py
@@ -13,6 +13,7 @@ from polar import (
     Pattern,
 )
 from polar.partial import TypeConstraint
+from polar.errors import ValidationError
 
 import pytest
 
@@ -948,3 +949,59 @@ def test_lookup_in_head(polar, is_allowed):
 
     assert not is_allowed("leina", "write", r)
     assert is_allowed("leina", "read", r)
+
+
+def test_rule_prototypes_with_subclass_check(polar):
+    class Foo:
+        pass
+
+    class Bar(Foo):
+        pass
+
+    class Baz(Bar):
+        pass
+
+    class Bad:
+        pass
+
+    # NOTE: keep this order of registering classes--confirms that MROs are added at the correct time
+    polar.register_class(Baz)
+    polar.register_class(Bar)
+    polar.register_class(Foo)
+    polar.register_class(Bad)
+
+    p = """
+    type f(_x: Integer);
+    f(1);
+    """
+    polar.load_str(p)
+
+    p = """
+    type f(_x: Foo);
+    type f(_x: Foo, _y: Bar);
+    f(_x: Bar);
+    f(_x: Baz);
+    """
+    polar.load_str(p)
+
+    with pytest.raises(ValidationError) as e:
+        polar.load_str("f(_x: Bad);")
+
+    polar.clear_rules()
+
+    # Test with fields
+    p = """
+    type f(_x: Foo{id: 1});
+    f(_x: Bar{id: 1});
+    f(_x: Baz{id: 1});
+    """
+    polar.load_str(p)
+    with pytest.raises(ValidationError) as e:
+        polar.load_str("f(_x: Baz);")
+
+    # Test invalid rule prototype
+    p = """
+    type f(x: Foo, x.baz);
+    """
+    with pytest.raises(ValidationError) as e:
+        polar.load_str(p)
--- a/languages/python/oso/tests/test_polar.py
+++ b/languages/python/oso/tests/test_polar.py
@@ -984,7 +984,7 @@ def test_rule_prototypes_with_subclass_check(polar):
     """
     polar.load_str(p)
 
-    with pytest.raises(ValidationError) as e:
+    with pytest.raises(ValidationError):
         polar.load_str("f(_x: Bad);")
 
     polar.clear_rules()
@@ -996,12 +996,12 @@ def test_rule_prototypes_with_subclass_check(polar):
     f(_x: Baz{id: 1});
     """
     polar.load_str(p)
-    with pytest.raises(ValidationError) as e:
+    with pytest.raises(ValidationError):
         polar.load_str("f(_x: Baz);")
 
     # Test invalid rule prototype
     p = """
     type f(x: Foo, x.baz);
     """
-    with pytest.raises(ValidationError) as e:
+    with pytest.raises(ValidationError):
         polar.load_str(p)
--- a/polar-core/src/kb.rs
+++ b/polar-core/src/kb.rs
@@ -105,7 +105,7 @@ impl KnowledgeBase {
     /// Validate that all rules loaded into the knowledge base are valid based on rule prototypes.
     pub fn validate_rules(&self) -> PolarResult<()> {
         for (rule_name, generic_rule) in &self.rules {
-            if let Some(prototypes) = self.rule_prototypes.get(&rule_name) {
+            if let Some(prototypes) = self.rule_prototypes.get(rule_name) {
                 // If a prototype with the same name exists, then the parameters must match for each rule
                 for rule in generic_rule.rules.values() {
                     let mut msg = "Must match one of the following rule prototypes:\n".to_owned();
--- a/polar-core/src/kb.rs
+++ b/polar-core/src/kb.rs
@@ -443,10 +443,7 @@ impl KnowledgeBase {
     pub fn add_rule_prototype(&mut self, prototype: Rule) {
         let name = prototype.name.clone();
         // get rule prototypes
-        let prototypes = self
-            .rule_prototypes
-            .entry(name.clone())
-            .or_insert_with(|| vec![]);
+        let prototypes = self.rule_prototypes.entry(name).or_insert_with(Vec::new);
         prototypes.push(prototype);
     }
 
--- a/polar-core/src/kb.rs
+++ b/polar-core/src/kb.rs
@@ -324,7 +324,7 @@ impl KnowledgeBase {
                 )
                 | (Value::Variable(_), Some(Value::Pattern(prototype_spec)), rule_value, None) => {
                     match prototype_spec {
-                        // Template specializer is an instance pattern
+                        // Prototype specializer is an instance pattern
                         Pattern::Instance(InstanceLiteral {
                             tag,
                             fields: prototype_fields,
@@ -356,7 +356,7 @@ impl KnowledgeBase {
                                 ))
                             }
                         }
-                        // Template specializer is a dictionary pattern
+                        // Prototype specializer is a dictionary pattern
                         Pattern::Dictionary(prototype_fields) => {
                             if let Value::Dictionary(rule_fields) = rule_value {
                                 if self.param_fields_match(prototype_fields, rule_fields) {
@@ -971,7 +971,7 @@ mod tests {
         // Orange is a subclass of Citrus
         kb.add_mro(sym!("Orange"), vec![3, 2, 1]).unwrap();
 
-        // Template applies if it has the same name as a rule
+        // Prototype applies if it has the same name as a rule
         kb.add_rule_prototype(rule!("f", ["x"; instance!(sym!("Orange"))]));
         kb.add_rule(rule!("f", ["x"; instance!(sym!("Orange"))]));
         kb.add_rule(rule!("f", ["x"; instance!(sym!("Fruit"))]));
@@ -984,7 +984,7 @@ mod tests {
             }
         ));
 
-        // Template does not apply if it doesn't have the same name as a rule
+        // Prototype does not apply if it doesn't have the same name as a rule
         kb.clear_rules();
         kb.add_rule_prototype(rule!("f", ["x"; instance!(sym!("Orange"))]));
         kb.add_rule(rule!("f", ["x"; instance!(sym!("Orange"))]));
@@ -992,7 +992,7 @@ mod tests {
 
         kb.validate_rules().unwrap();
 
-        // Template does apply if it has the same name as a rule even if different arity
+        // Prototype does apply if it has the same name as a rule even if different arity
         kb.clear_rules();
         kb.add_rule_prototype(rule!("f", ["x"; instance!(sym!("Orange")), value!(1)]));
         kb.add_rule(rule!("f", ["x"; instance!(sym!("Orange"))]));
--- a/polar-core/src/kb.rs
+++ b/polar-core/src/kb.rs
@@ -1,26 +1,42 @@
 use std::collections::HashMap;
 
-use crate::error::PolarResult;
+use crate::error::ParameterError;
+use crate::error::{PolarError, PolarResult};
 
 pub use super::bindings::Bindings;
 use super::counter::Counter;
 use super::rules::*;
 use super::sources::*;
 use super::terms::*;
+use std::sync::Arc;
 
-/// A map of bindings: variable name → value. The VM uses a stack internally,
-/// but can translate to and from this type.
+enum RuleParamMatch {
+    True,
+    False(String),
+}
+
+impl RuleParamMatch {
+    #[cfg(test)]
+    fn is_true(&self) -> bool {
+        matches!(self, RuleParamMatch::True)
+    }
+}
 
 #[derive(Default)]
 pub struct KnowledgeBase {
+    /// A map of bindings: variable name → value. The VM uses a stack internally,
+    /// but can translate to and from this type.
     pub constants: Bindings,
+    /// Map of class name -> MRO list where the MRO list is a list of class instance IDs
+    mro: HashMap<Symbol, Vec<u64>>,
 
     /// Map from loaded files to the source ID
     pub loaded_files: HashMap<String, u64>,
     /// Map from source code loaded to the filename it was loaded as
     pub loaded_content: HashMap<String, String>,
 
-    pub rules: HashMap<Symbol, GenericRule>,
+    rules: HashMap<Symbol, GenericRule>,
+    rule_prototypes: HashMap<Symbol, Vec<Rule>>,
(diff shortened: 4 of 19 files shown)


polar-core/src/polar.lalrpop
```diff
@@ -62,6 +62,7 @@ extern {
         "or" => lexer::Token::Or,           // or
         "not" => lexer::Token::Not,         // not
         "matches" => lexer::Token::Matches, // matches
+        "type" => lexer::Token::Type        // type
```
gkaemmer, line 65: Should we call this "template" instead of "type" if we're referring to them as templates?

polar-core/src/polar.rs
```diff
@@ -183,19 +182,19 @@ impl Polar {
...
                         warnings.append(&mut rule_warnings);
                         let rule = rewrite_rule(rule, kb);
-
-                        let name = rule.name.clone();
-                        let generic_rule = kb
-                            .rules
-                            .entry(name.clone())
-                            .or_insert_with(|| GenericRule::new(name, vec![]));
-                        generic_rule.add_rule(Arc::new(rule));
+                        kb.add_rule(rule);
```
gkaemmer, line 185: Love this refactor

gkaemmer (commented):
Some comments!

languages/python/oso/polar/host.py
```diff
@@ -55,6 +61,7 @@ def cache_class(self, cls, name=None):
             raise DuplicateClassAliasError(name, self.get_class(name), cls)
 
         self.classes[name] = cls
+        self.class_ids[cls.__name__] = self.cache_instance(cls)
```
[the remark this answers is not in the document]
gkaemmer: Actually as I glance at this with fresh eyes, it seems like you could use `name` instead of `cls.__name__`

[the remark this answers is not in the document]
leina05: Ah, good call. Mapping from `cls` should be good. `name` doesn't work because it can't easily be retrieved from the raw Python class.

polar-core/src/polar.lalrpop
leina05: I think I'm going to standardize on "rule prototype." "Template" is left over from a hackathon project.

polar-core/src/kb.rs
```diff
@@ -1,26 +1,30 @@
...
+    mro: HashMap<Symbol, TermList>,
 
     /// Map from loaded files to the source ID
     pub loaded_files: HashMap<String, u64>,
     /// Map from source code loaded to the filename it was loaded as
     pub loaded_content: HashMap<String, String>,
 
-    pub rules: HashMap<Symbol, GenericRule>,
+    rules: HashMap<Symbol, GenericRule>,
+    rule_templates: HashMap<Symbol, GenericRule>,
```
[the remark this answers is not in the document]
leina05: The main reason for this is to reuse other code for parsing in rewriting rules. Since rule prototypes have the same structure as rules I think it makes sense.

polar-core/src/kb.rs
```diff
@@ -75,11 +81,253 @@ impl KnowledgeBase {
...
+                        let mut found_match = false;
+                        for t in templates {
+                            if self.rule_params_match(r.as_ref(), t.as_ref()) {
+                                found_match = true;
+                                break;
+                            }
+                        }
+                        if !found_match {
+                            let err: PolarError = error::RuntimeError::FileLoading {
+                                msg: format!("Rule {}({:#?}) not allowed because parameters did not match rule template.", r.name, r.params),
```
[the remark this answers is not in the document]
leina05: Definitely. Working on this next!

polar-core/src/kb.rs
```diff
@@ -75,11 +81,253 @@ impl KnowledgeBase {
...
+        for (symbol, generic_rule) in self.rules.clone() {
+            let generic_template = self.rule_templates.get(&symbol);
+            if let Some(generic_template) = generic_template {
+                // If a template with the same name exists, then the parameters must match for each rule
+                for (_, r) in generic_rule.rules {
+                    // get templates with correct arity
+                    let templates = generic_template
+                        .rules
+                        .values()
+                        .map(|t| {
```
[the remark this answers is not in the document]
leina05: This is selecting the templates that have the same arity as the rule we are currently checking. If we don't find any, then the current rule doesn't have a corresponding template with the same arity, so we error on line 131.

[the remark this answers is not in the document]
gkaemmer: I _think_ `templates` may end up as `None` here if any of `generic_template.rules` have different arities. I.e. if I have templates with arity 2 and 3, and my rule has arity 2. Haven't tested though

The FromIterator implementation for `Option` is kinda weird -- behaves like an "all" check to make sure there are no `None`s in the iterator: https://doc.rust-lang.org/std/iter/trait.FromIterator.html#method.from_iter-16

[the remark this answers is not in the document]
leina05: Oh yeah you're totally right. Good catch, thanks!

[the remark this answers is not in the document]
leina05: Updated to user `filter`.

polar-core/src/kb.rs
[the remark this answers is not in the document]
dhatch: Why do rule prototypes need rewriting?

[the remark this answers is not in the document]
dhatch: It doesn't really look like you're using `GenericRule` at all. You  might want to just store this as `HashMap<Symbol, Vec<Rule>>` Or `HashMap<(Symbol, u64), Vec<Rule>>` if you want to include arity in the key.

polar-core/src/kb.rs
```diff
@@ -75,11 +81,247 @@ impl KnowledgeBase {
...
+            .or_insert_with(|| GenericRule::new(name, vec![]));
+        generic_rule.add_rule(Arc::new(rule));
+    }
+
+    pub fn validate_rules(&self) -> PolarResult<()> {
+        for (symbol, generic_rule) in self.rules.clone() {
+            let generic_template = self.rule_templates.get(&symbol);
+            if let Some(generic_template) = generic_template {
+                // If a template with the same name exists, then the parameters must match for each rule
+                for (_, r) in generic_rule.rules {
```
dhatch, line 98: `r` => `rule`

polar-core/src/kb.rs
```diff
@@ -75,11 +81,247 @@ impl KnowledgeBase {
...
+        let name = rule.name.clone();
+        let generic_rule = self
+            .rules
+            .entry(name.clone())
+            .or_insert_with(|| GenericRule::new(name, vec![]));
+        generic_rule.add_rule(Arc::new(rule));
+    }
+
+    pub fn validate_rules(&self) -> PolarResult<()> {
+        for (symbol, generic_rule) in self.rules.clone() {
```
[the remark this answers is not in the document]
dhatch: The clone on this line seems unnecessary.

polar-core/src/kb.rs
```diff
@@ -75,11 +81,247 @@ impl KnowledgeBase {
...
+                    .fields
+                    .get(k)
+                    .map(|rule_value| rule_value == template_value)
+                    .unwrap_or_else(|| false)
+            })
+            .all(|v| v);
+    }
+
+    fn rule_params_match(&self, rule: &Rule, template: &Rule) -> bool {
+        for (rule_param, template_param) in rule.params.iter().zip(template.params.iter()) {
```
dhatch, line 151: You can use https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.all here to clean up this code a bit. Breaking it into several functions may also help.

polar-core/src/kb.rs
```diff
@@ -75,11 +81,247 @@ impl KnowledgeBase {
...
+            let res = match (
+                template_param.parameter.value(),
+                template_param.specializer.as_ref().map(Term::value),
+                rule_param.parameter.value(),
+                rule_param.specializer.as_ref().map(Term::value),
+            ) {
+                (
+                    Value::Variable(_),
+                    Some(Value::Pattern(template_spec)),
+                    Value::Variable(_),
```
dhatch, line 161: Variable names don't have to match?

polar-core/src/kb.rs
```diff
@@ -75,11 +81,247 @@ impl KnowledgeBase {
...
+                            return Err(err.set_context(None, Some(&r.body)));
+                        }
+                    }
+                }
+            }
+        }
+        Ok(())
+    }
+
+    fn param_fields_match(&self, template_fields: &Dictionary, rule_fields: &Dictionary) -> bool {
```
dhatch, line 136: Add doc string about behavior of this. Rule fields must be a superset of template fields & values must match.

polar-core/src/kb.rs
```diff
@@ -75,11 +81,247 @@ impl KnowledgeBase {
...
+        return template_fields
+            .fields
+            .iter()
+            .map(|(k, template_value)| {
+                rule_fields
+                    .fields
+                    .get(k)
+                    .map(|rule_value| rule_value == template_value)
```
dhatch, line 144: Does this need to be recursive? What about something like:

```
type foo(a: Foo{a: Bar{b: 1}});

foo(a: Foo{a: Bar{b: 1, c: 2}, d: 2})
```

polar-core/src/kb.rs
```diff
@@ -75,11 +81,247 @@ impl KnowledgeBase {
...
+                            if template_instance.tag == rule_instance.tag {
+                                self.param_fields_match(
+                                    &template_instance.fields,
+                                    &rule_instance.fields,
+                                )
+                            // If tags don't match, then rule specializer must be a subclass of template specializer
+                            } else if let Some(template_class) =
+                                self.constants.get(&template_instance.tag)
+                            {
+                                if let Some(rule_mro) = self.mro.get(&rule_instance.tag) {
```
dhatch, line 181: Would it be easier for mro to just contain class names?

polar-core/src/kb.rs
```diff
@@ -75,11 +81,247 @@ impl KnowledgeBase {
...
+                            }),
+                        ) => self.param_fields_match(template_fields, rule_fields),
+                        (
+                            Pattern::Instance(InstanceLiteral {
+                                tag,
+                                fields: template_fields,
+                            }),
+                            Pattern::Dictionary(rule_fields),
+                        ) if tag == &sym!("Dictionary") => {
+                            self.param_fields_match(template_fields, rule_fields)
```
dhatch, line 208: Dictionary specializer doesn't actually work, right?

polar-core/src/kb.rs
```diff
@@ -75,11 +81,247 @@ impl KnowledgeBase {
...
+                // template has value, rule has value
+                (
+                    Value::Variable(_),
+                    Some(template_value),
+                    Value::Variable(_),
+                    Some(rule_value),
+                )
+                | (Value::Variable(_), Some(template_value), rule_value, None)
+                | (template_value, None, rule_value, None) => match (template_value, rule_value) {
+                    (Value::List(template_list), Value::List(rule_list)) => template_list
```
dhatch, line 267: What's the logic here? Why are we not checking that the lists are equal?

polar-core/src/kb.rs
```diff
@@ -191,3 +433,210 @@ impl KnowledgeBase {
...
+        assert!(kb.rule_params_match(
+            &rule!("f", ["x"; instance!(sym!("Foo"))]),
+            &rule!("f", ["x"; instance!(sym!("Foo"))])
+        ));
+        // rule: f(x: Foo), template: f(x: Bar) => FAIL if Foo is not subclass of Bar
+        assert!(!kb.rule_params_match(
+            &rule!("f", ["x"; instance!(sym!("Foo"))]),
+            &rule!("f", ["x"; instance!(sym!("Bar"))])
+        ));
+        // TODO: rule: f(x: Foo), template: f(x: Bar) => PASS if Foo is subclass of Bar
```
dhatch, line 455: Check more than 1 level of subclass

languages/python/oso/polar/polar.py
```diff
@@ -235,6 +236,9 @@ def register_class(self, cls, *, name=None):
         """Register `cls` as a class accessible by Polar."""
         cls_name = self.host.cache_class(cls, name)
         self.register_constant(cls, cls_name)
+        self.ffi_polar.register_mro(
+            cls_name, self.host.to_polar(list(inspect.getmro(cls)))
```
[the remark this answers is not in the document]
leina05: Does that matter thought? All classes have to be registered by this point in order to use them as specializers.

polar-core/src/kb.rs
```diff
@@ -1,26 +1,30 @@
...
 use super::terms::*;
+use std::sync::Arc;
 
 /// A map of bindings: variable name → value. The VM uses a stack internally,
 /// but can translate to and from this type.
 
 #[derive(Default)]
 pub struct KnowledgeBase {
     pub constants: Bindings,
+    mro: HashMap<Symbol, TermList>,
```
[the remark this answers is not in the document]
leina05: Yeah that's probably better.

languages/python/oso/polar/polar.py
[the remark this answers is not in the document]
leina05: Yeah this is a problem that I didn't think of. I guess we need to add all the MROs after every class has been registered, which is unfortunate.

polar-core/src/polar.rs
```diff
@@ -183,19 +182,19 @@ impl Polar {
...
                         kb.inline_queries.push(term);
                     }
+                    parser::Line::RuleTemplate(template) => {
+                        let template = rewrite_rule(template, kb);
+                        kb.add_rule_template(template);
+                    }
                 }
             }
+            // check rules are valid against rule templates
+            kb.validate_rules()?;
```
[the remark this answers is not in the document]
leina05: Yeah, I'm working on error handling now.

polar-core/src/kb.rs
```diff
@@ -75,11 +81,247 @@ impl KnowledgeBase {
...
+    }
+
+    pub fn add_rule_template(&mut self, template: Rule) {
+        let name = template.name.clone();
+        // get rule templates
+        let generic_template = self
+            .rule_templates
+            .entry(name.clone())
+            .or_insert_with(|| GenericRule::new(name, vec![]));
+        generic_template.add_rule(Arc::new(template));
```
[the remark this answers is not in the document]
leina05: Yes this is allowed.

polar-core/src/kb.rs
leina05: This is related to the bug that prevents pattern specializers from working on fields. Since that doesn't work right now, I didn't account for it here.

polar-core/src/kb.rs
leina05: No, not sure that would make any sense.

polar-core/src/kb.rs
```diff
@@ -75,11 +81,247 @@ impl KnowledgeBase {
...
+                            Pattern::Instance(template_instance),
+                            Pattern::Instance(rule_instance),
+                        ) => {
+                            // if tags match, all template fields must match those in rule fields, otherwise false
+                            if template_instance.tag == rule_instance.tag {
+                                self.param_fields_match(
+                                    &template_instance.fields,
+                                    &rule_instance.fields,
+                                )
+                            // If tags don't match, then rule specializer must be a subclass of template specializer
```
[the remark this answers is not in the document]
leina05: Yep that's missing good catch.

polar-core/src/kb.rs
leina05: I like your suggestion of the MRO containing instance IDs.

polar-core/src/kb.rs
leina05: No because of a bug.

polar-core/src/kb.rs
leina05: This is the `isa` logic that the list needs to contain at least all the elements of the other list in order to match, but could be a superset.

languages/python/oso/polar/polar.py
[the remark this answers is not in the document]
leina05: Moved registering MROs with Oso from `register_class()` to the `load()` method. They will be re-registered every time `load()` is called, which I think is necessary in case classes are registered between loads. Not great but I don't think it's a big issue.

polar-core/src/kb.rs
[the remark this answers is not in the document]
leina05: Added fields check + test in Python

polar-core/src/kb.rs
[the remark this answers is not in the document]
leina05: Made this change.

polar-core/src/kb.rs
leina05: Done.

polar-core/src/kb.rs
```diff
@@ -75,11 +81,247 @@ impl KnowledgeBase {
...
+                    if templates.is_empty() {
+                        let err: PolarError = error::RuntimeError::FileLoading {
+                            msg: format!("Rule {}({:#?}) not allowed because arity did not match rule template.", r.name, r.params),
+                        }
+                        .into();
+                        return Err(err.set_context(None, Some(&r.body)));
+                    } else {
+                        // these templates should all have matching arity
+                        // TODO: check params of rule against rule templates
+                        let mut found_match = false;
```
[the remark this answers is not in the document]
leina05: Done.

languages/python/oso/polar/polar.py
```diff
@@ -144,6 +144,18 @@ def load_file(self, policy_file):
...
         """Load a Polar string, checking that all inline queries succeed."""
+        # Get MRO of all registered classes
+        # NOTE: not ideal that the MRO gets updated each time load_str is
+        # called, but since we are planning to move to only calling load once
+        # with the include feature, I think it's okay for now.
+        for (cls_name, cls) in self.host.classes.items():
+            mro = [
+                self.host.class_ids.get(c)
+                for c in inspect.getmro(cls)
+                if self.host.class_ids.get(c)
```
dhatch, line 155:
```suggestion
                if c in self.host.class_ids
```

Would be the typical way I'd write this

languages/python/oso/tests/test_polar.py
```diff
@@ -971,9 +977,23 @@ class Baz:
...
     type f(_x: Foo);
+    type f(_x: Foo, _y: Bar);
     f(_x: Bar);
+    f(_x: Baz);
     """
     polar.load_str(p)
 
     with pytest.raises(Exception) as e:
-        polar.load_str("f(_x: Baz")
+        polar.load_str("f(_x: Bad);")
```
dhatch, line 987: It would be good to at least check that this is the right type of exception

polar-core/src/kb.rs
```diff
@@ -90,196 +102,338 @@ impl KnowledgeBase {
...
+        &self,
+        index: usize,
+        rule_value: &Value,
+        prototype_value: &Value,
+    ) -> PolarResult<RuleParamMatch> {
+        Ok(match (prototype_value, rule_value) {
+            (Value::List(prototype_list), Value::List(rule_list)) => {
+                if prototype_list
+                    .iter()
+                    .map(|t| rule_list.contains(t))
```
dhatch, line 262: I think you can make this one line: `all(|t| rule_list_contains(t))`

polar-core/src/kb.rs
```diff
@@ -90,196 +102,338 @@ impl KnowledgeBase {
...
+                    if let Some(rule_mro) = self.mro.get(&rule_instance.tag) {
+                        if rule_mro.contains(instance_id)
+                            && self.param_fields_match(
+                                &prototype_instance.fields,
+                                &rule_instance.fields,
+                            )
+                        {
+                            RuleParamMatch::True
+                        } else {
+                            RuleParamMatch::False(format!("Rule specializer {} on parameter {} must be a subclass of prototype specializer {}", rule_instance.tag,index, prototype_instance.tag))
```
dhatch, line 206: What if it's a subclass, but the parameters don't match?

languages/python/oso/polar/polar.py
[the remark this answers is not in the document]
dhatch: This would be an issue though if you register a class after load, right?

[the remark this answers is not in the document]
dhatch: I don't think we should fix it now but another point in favor of improving init process for Oso instance.

polar-core/src/kb.rs
```diff
@@ -1,26 +1,42 @@
...
+    mro: HashMap<Symbol, Vec<u64>>,
 
     /// Map from loaded files to the source ID
     pub loaded_files: HashMap<String, u64>,
     /// Map from source code loaded to the filename it was loaded as
     pub loaded_content: HashMap<String, String>,
 
-    pub rules: HashMap<Symbol, GenericRule>,
+    rules: HashMap<Symbol, GenericRule>,
+    rule_prototypes: HashMap<Symbol, GenericRule>,
```
dhatch, line 39: Left a comment about this before, but I can't find the resolution. Why can't this just be `HashMap<Symbol, Vec<Rule>>`?

polar-core/src/polar.rs
```diff
@@ -183,19 +182,19 @@ impl Polar {
...
-                            .entry(name.clone())
-                            .or_insert_with(|| GenericRule::new(name, vec![]));
-                        generic_rule.add_rule(Arc::new(rule));
+                        kb.add_rule(rule);
                     }
                     parser::Line::Query(term) => {
                         kb.inline_queries.push(term);
                     }
+                    parser::Line::RuleTemplate(prototype) => {
+                        let prototype = rewrite_rule(prototype, kb);
```
dhatch, line 191: Why do we need to rewrite prototypes?

polar-core/src/kb.rs
dhatch: Ok, maybe leave a todo here?

dhatch (commented):
Few small comments, looks pretty good.

languages/python/oso/polar/polar.py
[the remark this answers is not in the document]
leina05: As long as `load()` is called again after registering more classes, it would be fine. It's already an issue to register classes after `load()` because you'll get unrecognized specializer errors. But yeah the init needs to be generally improved.

polar-core/src/kb.rs
leina05: There's not a real reason. Can change it if you want.

polar-core/src/polar.rs
leina05: I don't know, there's no point since we never check the body but I wasn't sure if there's a case where a head-only rule would get rewritten.

dhatch: Rewrites always go into the body, so we shouldn't need them. They're used for operations that return a value, like `dot`, `add`, `sub`, etc. Imo none of these should be allowed in rule prototypes, but I don't think we'd actually catch them.

I guess we could run the rewrites and then error if the rule has a body afterwards? Which would mean it's an invalid prototype also?

polar-core/src/kb.rs
dhatch: The rationale for changing is generic rule has a rule index for efficiently looking up rules in the VM. We don't use that index here so we aren't really getting any benefit from it.

You can decide if you want to change it or not.

polar-core/src/kb.rs
leina05: good catch

polar-core/src/polar.rs
leina05: Yeah we could do that

polar-core/src/kb.rs
```diff
@@ -1,26 +1,42 @@
...
 use super::sources::*;
 use super::terms::*;
+use std::sync::Arc;
 
-/// A map of bindings: variable name → value. The VM uses a stack internally,
-/// but can translate to and from this type.
+enum RuleParamMatch {
+    True,
+    False(String),
+}
```
gj, line 16: Feel like this would be better as

```suggestion
type RuleParamMismatch = Option<String>;
```

so we can leverage all of the Rust combinators over `Option`, but don't feel too strongly about it

polar-core/src/kb.rs
```diff
@@ -75,11 +93,379 @@ impl KnowledgeBase {
...
+                // Rule has value or value specializer, prototype has pattern specializer
+                (
+                    Value::Variable(_),
+                    Some(Value::Pattern(prototype_spec)),
+                    Value::Variable(_),
+                    Some(rule_value),
+                )
+                | (Value::Variable(_), Some(Value::Pattern(prototype_spec)), rule_value, None) => {
+                    match prototype_spec {
+                        // Template specializer is an instance pattern
```
gj, line 327:
```suggestion
                        // Prototype specializer is an instance pattern
```

gj: Few other occurrences of "template" -- probably worth a global find-and-replace

polar-core/src/kb.rs
```diff
@@ -75,11 +93,379 @@ impl KnowledgeBase {
...
+
+    /// Check that a rule parameter that is a value matches a prototype parameter that is a value
+    fn check_value_param(
+        &self,
+        index: usize,
+        rule_value: &Value,
+        prototype_value: &Value,
+    ) -> PolarResult<RuleParamMatch> {
+        Ok(match (prototype_value, rule_value) {
+            (Value::List(prototype_list), Value::List(rule_list)) => {
```
gj, line 259: Might want to reuse the list unification logic from the VM to account for *rest vars.

Also not totally sure w/o putting more thought into it what the semantics should be for the interactions between prototypes & *rest vars.

languages/python/oso/tests/test_polar.py
```diff
@@ -948,3 +949,59 @@ def __init__(self, action):
...
+
+    p = """
+    type f(_x: Integer);
+    f(1);
+    """
+    polar.load_str(p)
+
+    p = """
+    type f(_x: Foo);
+    type f(_x: Foo, _y: Bar);
```
gj, line 981: For the future, it'd be cool to disable Singleton Var checks for rule prototypes so we can write this w/o the underscores

gj (commented):
Not sure how blocking the *rest var question should be

languages/python/oso/tests/test_polar.py
dhatch: Oh yeah that would definitely be good

dhatch (approved):
lgtm

polar-core/src/kb.rs
leina05: Don't really want to do this refactor at this point but could do it as a follow up

PR #1064, opened 2021-08-10
data filtering poc cont'd

gw000:
Includes grounding fixes


saolsen:
uh @dhatch, why does it say absolutely every docs file changed?

dhatch:
I think because we added a changelog on main. not certain

polar-core/src/debugger.rs
```diff
@@ -186,7 +184,13 @@ impl Debugger {
             }
         }
         let parts: Vec<&str> = command.split_whitespace().collect();
-        match *parts.get(0).unwrap_or(&"help") {
+        let dflt = match self.last.take() {
```
dhatch, line 187: What do these variable names mean? We shouldn't use abbreviations.

polar-core/src/debugger.rs
```diff
@@ -80,6 +83,7 @@ pub struct Debugger {
     /// - `Some(step)`: View the stopping logic in
     ///   [`maybe_break`](struct.Debugger.html#method.maybe_break).
     step: Option<Step>,
+    last: Option<String>,
```
dhatch, line 86: What's `last`? Could you add a doc string?

polar-core/src/data_filtering.rs
```diff
@@ -0,0 +1,799 @@
...
+    partial_results: PartialResults,
+    variable: &str,
+    class_tag: &str,
+) -> PolarResult<FilterPlan> {
+    let mut filter_plan = FilterPlan {
+        result_sets: vec![],
+    };
+    // @NOTE(steve): Just reading an env var here sucks (see all the stuff we had to do
+    // to get POLAR_LOG to work in all libs, wasm etc...) but that's what I'm doing today.
+    // At some point surface this info better.
```
dhatch, line 719: +1. It's probably best for us to pass in all these options & read the env vars from the host libraries. If you want you could open a task for this & mark as good for onboarding & cleanup.

dhatch: Other aspect of this is having various logging we can turn on and off. See the DEBUG constant at the top of simplifier for another crappy example of this that I added.

polar-core/src/data_filtering.rs
```diff
@@ -0,0 +1,799 @@
...
+        assert_eq!(exp.operator, Operator::And);
+
+        if explain {
+            eprintln!("  {}: {}", i, term.to_polar());
+        }
+
+        let var_info = process_result(exp);
+        let vars = collapse_vars(var_info);
+
+        if explain {
```
dhatch, line 739: Breaking this explain into a separate function would make the main control flow easier to follow.

polar-core/src/data_filtering.rs
```diff
@@ -0,0 +1,799 @@
...
+            }
+        }
+    }
+}
+
+pub type Types = HashMap<String, HashMap<String, Type>>;
+pub type PartialResults = Vec<ResultEvent>;
+
+#[derive(Debug)]
+struct VarInfo {
```
dhatch, line 114: Some comments on these fields & their meaning would be helpful.

polar-core/src/data_filtering.rs
```diff
@@ -0,0 +1,799 @@
...
+    eq_values: Vec<(Symbol, Term)>,                     // x = 1;
+    contained_values: Vec<(Term, Symbol)>,              // 1 in x
+    field_relationships: Vec<(Symbol, String, Symbol)>, // x.a = y
+    in_relationships: Vec<(Symbol, Symbol)>,            // x in y
+}
+
+// @TODO(steve): Better way to handle these checks than just unwraps and asserts.
+
+fn process_result(exp: &Operation) -> VarInfo {
+    let mut var_info = VarInfo {
```
dhatch, line 126: `impl Default`

`let mut var_info = VarInfo::default();`

polar-core/src/data_filtering.rs
```diff
@@ -0,0 +1,799 @@
...
+        Operator::Mod => unimplemented!("`mod` is not supported for data filtering."),
+        Operator::Rem => unimplemented!("`rem` is not supported for data filtering."),
+        Operator::Add => unimplemented!("addition is not supported for data filtering."),
+        Operator::Sub => unimplemented!("subtraction is not supported for data filtering."),
+        Operator::Geq => unimplemented!("`>=` is not supported for data filtering."),
+        Operator::Leq => unimplemented!("`<=` is not supported for data filtering."),
+        Operator::Neq => unimplemented!("`!=` is not supported for data filtering."),
+        Operator::Gt => unimplemented!("`>` is not supported for data filtering."),
+        Operator::Lt => unimplemented!("`<` is not supported for data filtering."),
+        // @TODO(steve): Expand or expressions to multiple bindings in the simplifier.
```
dhatch, line 299: I think a better way to accomplish this would be in the inverter. Then we can drop the Or operator completely from partials.

polar-core/src/data_filtering.rs
```diff
@@ -0,0 +1,799 @@
...
+            } else {
+                unimplemented!("Non pattern specializers are not yet supported for data filtering.")
+            }
+        }
+        Operator::Unify | Operator::Eq | Operator::Assign => {
+            assert_eq!(exp.args.len(), 2);
+
+            let mut lhs = exp.args[0].clone();
+            if let Value::Expression(op) = lhs.value() {
+                lhs = process_exp(var_info, op).unwrap();
```
dhatch, line 215: This means we expect lhs & rhs to be Dot, right?

Could we break `process_exp` into `process_exp_dot`, etc. and only have `process_exp_dot` return a `Term`?

polar-core/src/data_filtering.rs
```diff
@@ -0,0 +1,799 @@
...
+        Operator::Dot => {
+            // Dot operations return a var that can be unified with.
+            // We create a new var to represent the result of the operation.
+            assert_eq!(exp.args.len(), 2);
+            let mut var = exp.args[0].clone();
+            if let Ok(inner_exp) = var.value().as_expression() {
+                if inner_exp.operator != Operator::Dot {
+                    unimplemented!("Operations other than dot nested within a dot are not yet supported for data filtering.")
+                }
+                var = process_exp(var_info, inner_exp).unwrap();
```
dhatch, line 181: Same here w/ `process_exp_dot`

polar-core/src/data_filtering.rs
```diff
@@ -0,0 +1,799 @@
...
+        args,
+    }) = var.value().as_expression()
+    {
+        var = Term::new_temporary(Value::Variable(dot_var(
+            var_info,
+            args[0].clone(),
+            &args[1],
+        )))
+    }
+    // TODO(steve): There's a potential name clash here which would be bad. Works for now.
```
dhatch, line 151: If I have `x.foo > 2 and x.foo < 1` that would be a name clash, right?

polar-core/src/data_filtering.rs
```diff
@@ -0,0 +1,799 @@
...
+    // but should probably generate this var better.
+    let sym = var.value().as_symbol().unwrap();
+    let field_str = field.value().as_string().unwrap();
+    let new_var = Symbol::new(&format!("{}_dot_{}", sym.0, field_str));
+
+    // Record the relationship between the vars.
+    var_info
```
dhatch, line 158: It seems like this function just produces the variable name but it actually also updates `var_info`. Any way to keep this in `process_exp`?

polar-core/src/data_filtering.rs
```diff
@@ -0,0 +1,799 @@
...
+                rhs = process_exp(var_info, op).unwrap();
+            };
+
+            match (lhs.value(), rhs.value()) {
+                // l in r
+                (Value::Variable(l), Value::Variable(r)) => {
+                    var_info.in_relationships.push((l.clone(), r.clone()));
+                }
+                // var in [1, 2, 3]
+                (Value::Variable(_var), _val) => {
```
dhatch, line 262: I don't think this every happens because VM would produce choice points & multiple bindings.

polar-core/src/data_filtering.rs
```diff
@@ -0,0 +1,799 @@
...
+        }
+        // Create new cycle if we couldn't
+        let mut new_cycle = HashSet::new();
+        new_cycle.insert(l);
+        new_cycle.insert(r);
+        joined_cycles.push(new_cycle);
+    }
+
+    let mut next_id = 0;
+    let mut get_id = move || {
```
dhatch, line 339: did not know you could do this

polar-core/src/data_filtering.rs
```diff
@@ -0,0 +1,799 @@
...
+        let id = next_id;
+        next_id += 1;
+        format!("{}", id)
```
dhatch, line 342: why is id not an int?

polar-core/src/data_filtering.rs
```diff
@@ -0,0 +1,799 @@
...
+    };
+
+    // Give each cycle an id
+    let mut variables: HashMap<String, HashSet<Symbol>> = HashMap::new();
+    for joined_cycle in joined_cycles {
+        let id = get_id();
+        variables.insert(id, joined_cycle);
+    }
+
+    // Substitute in relationships
```
dhatch, line 352: @dhatch stopped reading here

dhatch (commented):
Ok I started reviewing but I ran out of steam for today. Here's my comments so far which are not too substantial & some may be clean ups you don't want to take.

I haven't gotten through the meat of this which is reviewing `data_filtering.rs` beyond `process_exp`.

languages/python/oso/polar/polar.py
```diff
@@ -247,6 +255,45 @@ def get_class(self, name):
...
+        """
+        # Data filtering.
+        resource = Variable("resource")
+        # Get registered class name somehow
+        class_name = self.host.cls_names[cls]
+        constraint = Expression(
+            "And", [Expression("Isa", [resource, Pattern(class_name, {})])]
+        )
+        results = list(
+            self.query_rule(
```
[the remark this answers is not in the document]
saolsen: Yeah one day, not for this first version. You run into the same problem that the validation queries had where you need to run the query from the language side so that external checks all work. We can do this later.

languages/python/oso/polar/host.py
```diff
@@ -20,25 +20,56 @@
...
+            return self.types_get_field(obj, field)
 
         self.get_field = get_field or default_get_field
 
+    # @Q: I'm not really sure what I'm returning here.
+    def types_get_field(self, obj, field):
+        obj_type_name = self.cls_names[obj]
+        if obj_type_name in self.types:
+            obj_type_info = self.types[obj_type_name]
+            if field in obj_type_info:
```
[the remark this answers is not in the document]
saolsen: It checks the registered type info to see what the type of the field lookup is. I'll add a comment.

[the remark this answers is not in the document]
saolsen: This also will probably go away later. Eventually the type info will be passed into the core and we wont need this anymore it can all be resolved there (not for this release though).

languages/python/oso/polar/query.py
```diff
@@ -94,7 +95,34 @@ def handle_external_call(self, data):
 
         # Lookup the attribute on the instance.
         try:
-            attr = getattr(instance, attribute)
+            # Check if it's a relationship
+            attr = None
```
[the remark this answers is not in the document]
saolsen: yep and yep

polar-core/src/events.rs
```diff
@@ -96,3 +96,16 @@ pub enum QueryEvent {
         iterable: Term,
     },
 }
+
+// A struct for just Result Events. Used to pass data back into
+// the core for validation and data filtering.
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct ResultEvent {
```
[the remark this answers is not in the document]
saolsen: Agree.

polar-core/src/data_filtering.rs
saolsen: yeah, that is a good onboarding task

polar-core/src/data_filtering.rs
saolsen: Yeah, even better. Will make a task, that's a followup thing.

polar-core/src/data_filtering.rs
saolsen: Maybe, I'm not sure yet if anything else will have to pass out a term. Will put on my list of cleanups for later if not.

polar-core/src/data_filtering.rs
saolsen: No that would be fine. Those are both the same var. The name clash this refers to is if you had `x.foo = 1 and x_dot_foo = 2`. It would be a really weird policy lol but those are actually different vars but I'd end up treating them the same.

polar-core/src/data_filtering.rs
saolsen: Right

polar-core/src/data_filtering.rs
dhatch: Oh I got it

polar-core/src/data_filtering.rs
dhatch: Sg

polar-core/src/data_filtering.rs
```diff
@@ -0,0 +1,799 @@
...
+    process_exp(&mut var_info, exp);
+    var_info
+}
+
+fn dot_var(var_info: &mut VarInfo, var: Term, field: &Term) -> Symbol {
+    let mut var = var;
+    while let Ok(Operation {
+        operator: Operator::Dot,
+        args,
+    }) = var.value().as_expression()
```
dhatch, line 143: Do we need this loop here & also the recursive call in `process_exp`? I didn't think about this too long but it seems like this might run several times for every part of the expression.

polar-core/src/data_filtering.rs
```diff
@@ -0,0 +1,799 @@
...
+    // Give each cycle an id
+    let mut variables: HashMap<String, HashSet<Symbol>> = HashMap::new();
+    for joined_cycle in joined_cycles {
+        let id = get_id();
+        variables.insert(id, joined_cycle);
+    }
+
+    // Substitute in relationships
+    let mut parent_ids = vec![];
+    'parent_relationships: for (parent, _, _) in &var_info.field_relationships {
```
dhatch, line 354: Note to self: `field_relationships` is (var, field, other_var) (var.field = other_var)

polar-core/src/data_filtering.rs
```diff
@@ -0,0 +1,799 @@
...
+        let mut new_set = HashSet::new();
+        new_set.insert(child.clone());
+        variables.insert(new_id.clone(), new_set);
+        child_ids.push(new_id);
+    }
+
+    // If a.b = c and a.b = d, that means c = d.
+    // @Sorry(steve): Wow, what a loop. Maybe just loop over indexes instead.
+    let mut new_unifies: Vec<(String, String)> = vec![];
+    for (i, ((parent_id1, child_id1), (_, field1, _))) in parent_ids
```
dhatch, line 387: Might call `i` `child_id`

polar-core/src/data_filtering.rs
```diff
@@ -0,0 +1,799 @@
...
+        .iter()
+        .zip(child_ids.iter())
+        .zip(var_info.field_relationships.iter())
+        .enumerate()
+    {
+        for (j, ((parent_id2, child_id2), (_, field2, _))) in parent_ids
```
dhatch, line 393: Are these things actually ids? Aren't they the variable names of the parent & the children still?

polar-core/src/data_filtering.rs
```diff
@@ -0,0 +1,799 @@
...
+                new_unifies.push((child_id1.clone(), child_id2.clone()));
+            }
+        }
+    }
+
+    // @TODO(steve): There are absolutely bugs in here.
+    // If we're turning 0 into 1 and then 0 into 2 it'll just blow up
+    // not correctly turn 0 and 1 into 2. Needs some tests.
+    for (x, y) in &new_unifies {
+        if x != y {
```
dhatch, line 410: Why is this logic different than the original logic you used when building `variables`?

polar-core/src/data_filtering.rs
```diff
@@ -0,0 +1,799 @@
...
+        {
+            if i != j && parent_id1 == parent_id2 && field1 == field2 && child_id1 != child_id2 {
+                // Unify children
+                new_unifies.push((child_id1.clone(), child_id2.clone()));
+            }
+        }
+    }
+
+    // @TODO(steve): There are absolutely bugs in here.
+    // If we're turning 0 into 1 and then 0 into 2 it'll just blow up
```
dhatch, line 407: This comment means if we have `0 = 1` & `0 = 2` which would stem from something like `x.bar = y and x.bar = z and x.bar = a` (although not sure how this case would get out of the simplifier)?

polar-core/src/data_filtering.rs
```diff
@@ -0,0 +1,799 @@
...
+                other_field,
+                ..
+            } = typ
+            {
+                constrain_var(result_set, types, vars, child, other_class_tag);
+
+                // If the constrained child var doesn't have any constraints on it, we don't need to
+                // constrain this var. Otherwise we're just saying field foo in all Foos which
+                // would fetch all Foos and not be good.
+                if let Some(child_result) = result_set.requests.remove(child) {
```
dhatch, line 595: Can we just do this later before we insert the request into the result set on L663? Only insert if there are constraints? Then here we just look up the result set & if it exists we add the In & ref constraint.

polar-core/src/data_filtering.rs
```diff
@@ -0,0 +1,799 @@
...
+                });
+                contributed_constraints = true;
+            }
+            assert!(contributed_constraints);
+        }
+    }
+
+    // Constrain any vars that are `in` this var.
+    // Add their constraints to this one.
+    // @NOTE(steve): I think this is right, but I'm not totally sure.
```
dhatch, line 652: So this is for something like:

`y in _this.bar and y.foo = 2`

polar-core/src/data_filtering.rs
```diff
@@ -0,0 +1,799 @@
...
+    // This might assume that the current var is a relationship of kind "children".
+    for (lhs, rhs) in &vars.in_relationships {
+        if rhs == var_id {
+            constrain_var(result_set, types, vars, lhs, var_type);
+            let in_result_set = result_set.requests.remove(lhs).unwrap();
```
dhatch, line 657: Could you just return a `FetchRequest` from `constrain_var`? Then you don't need this logic here

polar-core/src/data_filtering.rs
```diff
@@ -0,0 +1,799 @@
...
+        let new_id = get_id();
+        let mut new_set = HashSet::new();
+        new_set.insert(var.clone());
+        variables.insert(new_id.clone(), new_set);
+        types.insert(new_id, typ);
+    }
+
+    let mut this_id = String::new();
+    for (id, set) in &variables {
+        if set.contains(&Symbol::new("_this")) {
```
dhatch, line 527: Is it a problem if this isn't found?

polar-core/src/data_filtering.rs
```diff
@@ -0,0 +1,799 @@
...
+        Operator::Or => unimplemented!("`or` is not supported for data filtering."),
+        Operator::ForAll => unimplemented!("`forall` is not supported for data filtering."),
+    }
+    None
+}
+
+#[derive(Debug)]
+struct Vars {
+    variables: HashMap<String, HashSet<Symbol>>,
+    field_relationships: HashSet<(String, String, String)>,
```
dhatch, line 309: If we define `Id` as a type, this would be `HashSet<(Id, String, Id)>` which would make it easier to tell what's going on.

polar-core/src/data_filtering.rs
```diff
@@ -0,0 +1,799 @@
...
+                .filter(|r| r.0 == *parent && r.1 != *field && r.2 == *child)
+            {
+                request.constraints.push(Constraint {
+                    kind: ConstraintKind::Eq,
+                    field: field.clone(),
+                    value: ConstraintValue::Field(eqf.1.clone()),
+                });
+                contributed_constraints = true;
+            }
+            assert!(contributed_constraints);
```
dhatch, line 646: So we'd hit this assert if the relationship wasn't defined rn?

polar-core/src/data_filtering.rs
```diff
@@ -0,0 +1,799 @@
...
+) {
+    // @TODO(steve): Probably should check the type against the var types. I think???
+    let type_def = types
+        .iter()
+        .find(|r| r.0 == var_type)
+        .map(|r| r.1.clone())
+        .unwrap_or_else(HashMap::new);
+
+    let mut request = if result_set.requests.contains_key(var_id) {
+        result_set.requests.remove(var_id).unwrap()
```
dhatch, line 568: wouldn't this branch result in pushing the `var_id` to the resolve order twice?

polar-core/src/data_filtering.rs
```diff
@@ -0,0 +1,799 @@
...
+
+    result_set.requests.insert(var_id.to_string(), request);
+    result_set.resolve_order.push(var_id.to_string());
+}
+
+pub fn opt_pass(filter_plan: &mut FilterPlan, explain: bool) -> bool {
+    let mut optimized = false;
+
+    // Remove duplicate result set in a union.
+    let mut drop_plan = None;
```
dhatch, line 671: How does this happen? If two rules are identical?

polar-core/src/data_filtering.rs
```diff
@@ -0,0 +1,799 @@
...
+}
+
+fn constrain_vars(types: &Types, vars: &Vars, this_type: &str) -> ResultSet {
+    let mut result_set = ResultSet {
+        requests: HashMap::new(),
+        resolve_order: vec![],
+        result_id: vars.this_id.clone(),
+    };
+    constrain_var(&mut result_set, types, vars, &vars.this_id, this_type);
+    result_set
```
dhatch, line 550: How are we actually checking that everything in Vars is converted into a constraint?

For example if I wrote something like `x.bar = x.y.foo`? My undersatnding is that isn't supported right now, but I think it would just silently not be checked. Is that right?

polar-core/src/partial/simplify.rs
```diff
@@ -597,44 +601,55 @@ impl Simplifier {
...
-            term.value(),
-            Value::Dictionary(_) | Value::Call(_) | Value::List(_) | Value::Expression(_)
-        ) {
-            let value = term.mut_value();
-            match value {
-                Value::Dictionary(dict) => {
-                    for (_, v) in dict.fields.iter_mut() {
-                        self.simplify_term(v, simplify_operation.clone());
-                    }
+        if self.seen.contains(term) {
```
dhatch, line 604: I don't have the context to review this, what's this doing?

languages/python/oso/polar/data_filtering.py
```diff
@@ -0,0 +1,148 @@
...
+
+    return Constraint(kind=kind, field=field, value=value)
+
+
+def ground_constraints(polar, results, filter_plan, constraints):
+    for constraint in constraints:
+        if isinstance(constraint.value, Ref):
+            ref = constraint.value
+            constraint.value = results[ref.result_id]
+            if ref.field is not None:
```
dhatch, line 108: Is it valid for `field` to be None?

dhatch (commented):
Okay, I've reviewed everything besides simplifier changes & isa changes so far.

polar-core/src/partial/isa_constraint_check.rs
```diff
@@ -65,75 +74,98 @@ impl IsaConstraintCheck {
...
         }
 
-        // a.b.c vs. d
-        if constraint_path
+        let just_vars = constraint_path.len() == 1
+            && proposed_path.len() == 1
+            && matches!(&constraint.args[0].value().as_symbol(), Ok(Symbol(_)))
+            && matches!(&self.proposed.args[0].value().as_symbol(), Ok(Symbol(_)));
+
+        // FIXME(gw): this logic is hard to follow!
```
dhatch, line 85: Can you explain what this logic is doing?

polar-core/src/data_filtering.rs
saolsen: Hmm yeah, gotta fix resolve order in general, It's not right. Will make that a followup task.

polar-core/src/data_filtering.rs
saolsen: Oh there were a few reasons. One was so I could re-use a Vec<(String, String)> for variables and ids but I didn't end up actually doing that. The other is that keys in a `json` map have to be strings so I'd have to convert later if I didn't just do it here.

polar-core/src/partial/isa_constraint_check.rs
gw000: first exclude a bunch of cases where we don't care about this constraint: it's not an `Isa`, it's not checking variables, it's checking two unrelated variables, or the paths don't match. then check for the cases we do care about: the sides are the same (is subclass check), or we already have a constraint on a prefix of an access path (isa with path check).

the only real difference from before is (supposed to be) that now we check if the "root" variable of an existing constraint is an alias for the root of the proposed constraint, instead of if it's identical.

polar-core/src/data_filtering.rs
saolsen: They are ids. I'm for sure rewriting these loops.

polar-core/src/data_filtering.rs
saolsen: Because we're dealing with ids now instead of variables. We're merging two different variable sets.

polar-core/src/partial/simplify.rs
gw000: see [here](https://www.notion.so/osohq/Simplifier-infinite-recursion-ad0479a286854af687b40f7e00db0876). it's kind of a simple-minded fix, I know there are still expressions that the simplifier will choke on, although I'm not sure how to produce them now. I actually think this might not be needed anymore now that we supply aliases to the simplifier, i'll do some testing

polar-core/src/data_filtering.rs
saolsen: Yeah.
I haven't actually seen it so maybe it is true that the simplifier takes care of that. Should write a test to try and hit this.

polar-core/src/data_filtering.rs
saolsen: Yeah I think that works.

polar-core/src/data_filtering.rs
saolsen: Yeah

polar-core/src/data_filtering.rs
saolsen: Yeah maybe.

polar-core/src/data_filtering.rs
saolsen: Not sure, that's a good test to see what happens.

polar-core/src/data_filtering.rs
saolsen: We get this from the builtin roles policy. Not sure why, maybe because there's like a parent check and an ancestor check?

languages/python/oso/polar/data_filtering.py
saolsen: Yeah

polar-core/src/partial/isa_constraint_check.rs
saolsen: lol gabe

dhatch (approved):

PR #1065, opened 2021-08-10
Improve grounding performance for dot operations

dhatch:
Makes `partial_and` and `partial_rule_filter` benchmarks faster than previously. This fix works by rewriting dot operations (only) as:

```
_value_3 = 2 and _value_1.a = _value_2 and _value_2.b = _value_3
```

compared to the previous rewriting of

```
_value_1.a = _value_2 and _value_2.b = _value_3 and _value_3 = 2
```

This improves performance for partials because grounding is no longer required for a dot operation like `x.foo = 2`. Instead, the constraint added is `x.foo = 2` (compared with adding `x.foo = _value_1` and then grounding later when `_value_1 = 2` is processed). This is a special case, but seems worth it in my opinion because these types of operations are probably the most common.

- [ ] Rewrite code is pretty challenging to work with, I think it may be worth a bit of a cleanup while I'm looking at it.
- [ ] Also needs a test

### DIFF
--- a/languages/python/oso/polar/exceptions.py
+++ b/languages/python/oso/polar/exceptions.py
@@ -16,7 +16,7 @@ class OsoError(Exception):
     @classmethod
     def add_get_help(cls, message):
         return (
-            message
+            str(message)
             + f"\n\tGet help with Oso from our engineers: https://help.osohq.com/error/{cls.__name__}"
         )
 
--- a/languages/python/oso/setup.py
+++ b/languages/python/oso/setup.py
@@ -39,7 +39,7 @@ def get_version(rel_path):
 setup(
     name="oso",
     version=get_version("oso/oso.py"),
-    description="oso is an open source policy engine for authorization that’s embedded in your application",
+    description="oso is an open source policy engine for authorization that's embedded in your application",
     long_description=long_description,
     long_description_content_type="text/markdown",
     author="Oso Security, Inc.",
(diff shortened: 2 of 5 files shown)


samscott89 (commented):
Might it make sense to look for any expressions `x.y = z` or `z = x.y` and rewrite them into the 3-op form `.(x, y, z)` straight away?

samscott89:
BTW awesome work for spotting this. How did you know this would be where the perf hit is?

did a quick test locally, this seems to also have a perf improvement over main:

```
diff --git a/polar-core/src/rewrites.rs b/polar-core/src/rewrites.rs
index ebb1caf4..0085a5ea 100644
--- a/polar-core/src/rewrites.rs
+++ b/polar-core/src/rewrites.rs
@@ -132,6 +132,25 @@ impl<'kb> Folder for Rewriter<'kb> {
                 }
                 new
             }
+            Value::Expression(o) if o.operator == Operator::Unify => {
+                let v = match (o.args[0].value(), o.args[1].value()) {
+                    (Value::Expression(op), _) if op.operator == Operator::Dot => {
+                        let mut dot_op = op.clone();
+                        dot_op.args.push(o.args[1].clone());
+                        dot_op
+                    }
+                    (_, Value::Expression(op)) if op.operator == Operator::Dot => {
+                        let mut dot_op = op.clone();
+                        dot_op.args.push(o.args[0].clone());
+                        dot_op
+                    }
+                    _ => {
+                        return fold_term(t, self);
+                    }
+                };
+
+                fold_term(t.clone_with_value(Value::Expression(v)), self)
+            }
             Value::Expression(o) if self.needs_rewrite(o) => {
                 // Rewrite sub-expressions, then push a temp onto the args.
                 let mut new = fold_operation(o.clone(), self);
```

passes all python tests, breaks rewriting + partial tests due to the names of things changing

PR #1068, opened 2021-08-12
Enforcement API

gkaemmer:
Adds an `Enforcer` type to the python library, which wraps a policy instance (aka an `Oso` instance) and exposes the `authorize` and `authorize_request` methods. The `Enforcer` class has an API to customize the errors raised by those methods.

PR checklist:

- [ ] Added changelog entry.


samscott89 (commented):
This is a little different to what I had thought we had discussed. My impression was that we would make the existing `Oso` object be the `Policy` object (still subclass of Polar, still has the same methods on it). For backwards compatibility we could alias `Oso = Policy` for the next release, and keep `is_allowed` on the `Oso` object.

And then introduce a _new_ object `Enforcer`, which would have the new APIs on it.

languages/python/oso/polar/polar.py
```diff
@@ -196,6 +196,20 @@ def query_rule(self, name, *args, **kwargs):
         """
         return self.query(Predicate(name=name, args=args), **kwargs)
 
+    def query_rule_once(self, name, *args, **kwargs):
```
gkaemmer, line 199: Don't love this name -- very open to alternatives. It seems like a useful method (I believe I've seen folks ask about something similar in slack)

languages/python/oso/oso/enforcer.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def authorize_request(self, actor, request):
+        """Ensure that ``actor`` is allowed to send ``request`` to the server.
+
+        If there is a matching ``allow_request(user, request)`` rule in the
+        policy, then this method returns ``None``. Otherwise, this method raises
+        a ``ForbiddenError``.
+
+        :param actor: The actor performing the request.
+        :param request: The request that was sent by the user. Can be a string
```
samscott89, line 78: Let's revisit this when we do the docs. But I think we should suggest this is always the request object.

languages/python/oso/oso/exceptions.py
```diff
@@ -0,0 +1,13 @@
+class AuthorizationError(Exception):
+    def __init__(self, actor, action, resource):
+        self.actor = actor
```
samscott89, line 3: I'm not seeing the benefit of providing actor/action/resource here. Maybe keep out for now and we can add back later?

languages/python/oso/oso/enforcer.py
```diff
@@ -0,0 +1,122 @@
...
+        self.policy = policy
+        if get_error is None:
+            self._get_error = self._default_get_error
+        else:
+            self._get_error = get_error
+        self.read_action = read_action
+
+        super().__init__()
+
+    def _default_get_error(self, is_not_found, actor, action, resource):
```
samscott89, line 33: Same comment as lower down: maybe skip needing `actor, action, resource` args for now, and keep this simpler?

samscott89 (approved):
Cool, this looks good!

I have a few small comments on docs, and the construction of the error. But this is clean.

PR #1070, opened 2021-08-13
data filtering fallback mode

gw000:
If the data filtering code path fails for some reason, use the fetcher function to enumerate all the options and ask polar which ones satisfy the partials. This might be slower, depending on what the fetcher does, but it should transparently handle most of the cases we don't support yet under the filter plan model. Also useful for testing & validation!


languages/python/oso/polar/data_filtering.py
```diff
@@ -83,30 +84,41 @@ def parse_constraint(polar, constraint):
...
-    for constraint in constraints:
-        if isinstance(constraint.value, Ref):
-            ref = constraint.value
-            constraint.value = results[ref.result_id]
-            if ref.field is not None:
-                constraint.value = [getattr(v, ref.field) for v in constraint.value]
+    refs, rest = partition(constraints, lambda c: isinstance(c.value, Ref) and c.value.result_id is not None)
+    for rid, cons in group_refs(refs).items():
+        fields = [[getattr(r, c.value.field) for c in cons] for r in results[rid]]
+        rest.append(Constraint(value=fields, kind='Ins', field=[c.field for c in cons]))
```
gw000, line 120: this is pretty appalling ... we need to merge all the ref constraints on a given result set in order to get the intersection (ie. records satisfying all constraints) instead of the union (records satisfying any constraint)

PR #1071, opened 2021-08-13
add docs for debugger break-on-rule

gw000:
(broken out of data filtering pr)


polar-core/src/partial/isa_constraint_check.rs
```diff
@@ -70,6 +70,7 @@ impl IsaConstraintCheck {
 
         let constraint_path = path(&constraint.args[0]);
         let proposed_path = path(&self.proposed.args[0]);
+        //                println!("EXISC {:?} :: {:?} {:?}", self.proposed.operator, proposed_path, constraint_path);
```
saolsen, line 73: 🙃

saolsen (approved):

PR #1073, opened 2021-08-13
rerelease

gw000:
for breaking changes

PR checklist:

- [x] Added changelog entry.


dhatch (changes requested):
You'll also need to bump `flask-oso` so there's a version of it that is compatible with `0.15.0`

saolsen (approved):

dhatch (approved):

PR #1075, opened 2021-08-13
Remove commenting behavior for docs diff.

dhatch:


.github/workflows/docs-diff.yml
```diff
@@ -200,9 +200,9 @@ jobs:
         id: diff
         run: set -o pipefail && diff -q --recursive base head | tee diff.txt
         continue-on-error: true
-      - id: get-comment-body
+      - id: Print docs changes
```
gj, line 203: I think?
```suggestion
      - name: Print docs changes
```

samscott89 (approved):

PR #1079, opened 2021-08-16
SQLAlchemy enforcer

gkaemmer:
PR checklist:

- [ ] Added changelog entry.

### DIFF
--- a/languages/python/sqlalchemy-oso/sqlalchemy_oso/enforcer.py
+++ b/languages/python/sqlalchemy-oso/sqlalchemy_oso/enforcer.py
@@ -1,0 +1,48 @@
+from oso import Enforcer
+from .oso import SQLAlchemyPolicy
+from .auth import authorize_model
+
+class SQLAlchemyEnforcer(Enforcer):
+    """
+    NOTE: This is a preview feature.
+
+    Custom Oso enforcer for SQLAlchemy.
+    """
+    def __init__(self, policy: SQLAlchemyPolicy, session_maker, *args, **kwargs):
+        """Construct a new SQLAlchemyEnforcer
+
+        >>> policy = SQLAlchemyPolicy(Base)
+        >>> oso = SQLAlchemyEnforcer(policy, Session)
+        >>> # ...
+        >>> oso.authorize_query(user, Article).all()
+
+        :param policy: An instance of ``SQLAlchemyPolicy``
+        :param session_maker: A SQLAlchemy session maker instance"""
+        self.session_maker = session_maker
+        super().__init__(policy, *args, **kwargs)
+
+    def authorize_query(self, actor, model, action=None):
+        """Authorize a model query, returning a SQLAlchemy ``Query`` instance.
+
+        Uses the ``allow`` rule to determine which constraints to apply. The
+        query instance returned contains only results that the actor can
+        ``"read"``.
+
+        :param actor: The current actor
+        :param model: A SQLAlchemy model class
+        :param action: Optionally override the action used to filter results.
+        Defaults to the ``read_action`` of this enforcer, which is normally the
+        string ``"read"``.
+        """
+        if action is None:
+            action = self.read_action
+
+        session = self.session_maker()
+        filter = authorize_model(
+            oso=self.policy,
+            actor=actor,
+            action=self.read_action,
+            session=session,
+            model=model,
+        )
+        return session.query(model).filter(filter)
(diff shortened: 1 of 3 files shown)


languages/python/sqlalchemy-oso/sqlalchemy_oso/enforcer.py
```diff
@@ -0,0 +1,50 @@
+from oso import Enforcer
+from .oso import SQLAlchemyPolicy
+from .auth import authorize_model
+
+
+class SQLAlchemyEnforcer(Enforcer):
```
samscott89, line 6: What do you think about removing the `SQLAlchemy` prefix? I feel like the pro of removing, is that it will remove some redundancy (`from sqlalchemy_oso import SQLAlchemyEnforcer` etc). Cons is that it might be harder to know from context in a docs guide which is imported.

gkaemmer: It's a good question. My gut is to keep it, and encourage the use of longer-named enforcers. The cost isn't huge (they're typically imported and used once), and like you said, it makes the docs a bit clearer what to use when.

It does make me wonder -- if I wanted to use a `FlaskEnforcer` AND a `SQLAlchemyEnforcer` as a single enforcer, I wonder how I might allow that. For now I don't think it's a huge deal, but it might become something we want to support. But I think that type of thing would be also be easier to reason about if enforcers had specific names.

languages/python/sqlalchemy-oso/sqlalchemy_oso/enforcer.py
```diff
@@ -0,0 +1,50 @@
...
+        > oso = SQLAlchemyEnforcer(policy, Session)
+        > # ...
+        > oso.authorize_query(user, Article).all()
+
+        :param policy: An instance of ``SQLAlchemyPolicy``
+        :param session_maker: A SQLAlchemy session maker instance"""
+        self.session_maker = session_maker
+        super().__init__(policy, *args, **kwargs)
+
+    def authorize_query(self, actor, model, action=None):
```
dhatch, line 26: We intentionally don't expose a query authorization API for SQLAlchemy currently. `authorize_model` is internal and only used on the session API.

Authorization is associated with the session because SQLAlchemy may issue queries under the hood, and because sessions cache objects that have already been retrieved from the database.

For example, consider this policy:

```
allow(user: User, "read", post: Post) if
   post.owner = user;

allow(user: User, "read", _post_analytics: PostAnalytics) if
    user.is_admin;
```

Assume the current user is not an admin:

```
enforcer = SQLAlchemyEnforcer(session=session)
posts = enforcer.authorize_query(user, Post).all()

analytics = [p.analytics for p in posts]
return json.dumps(analytics)
```

This would return analytics data to a user, because accessing the `p.analytics` property issues a query to the database under the hood. This can unintentionally expose data that should be authorized.

With the current API, authorization is attached to a session.

```
session = AuthorizedSession(oso, user, {Post: "read", PostAnalytics: "read"})
posts = session.query(Post).all()

analytics = [p.analytics for p in posts]
# analytics will be an array of [None, None, None, ... ] since the user isn't authorized to read this data
```

I don't think the current API is ideal either, we've gotten some feedback that it's non-intuitive. But, it does prevent unintentionally fetching objects without applying authorization.

dhatch (commented):
I'm not sure about this approach. We intentionally made session the entrypoint for authorization in SQLAlchemy to prevent exposing unauthorized models from queries.

I left some details in the comment. LMK what you think, definitely open to talking this through further but I wanted to make sure you had this context.

languages/python/sqlalchemy-oso/sqlalchemy_oso/enforcer.py
gkaemmer: Mostly resolved this offline, but I am curious whether you think it makes more sense to directly return the result of `authorize_model` instead of attaching it to a session. That's a bit of a sqlalchemy style question, so I might defer to you.

The current API I have is:
```
query = oso.authorize_query(user, Expense)
query.all()
```

We might also consider something like this (which might be more sqlalchemy-y):
```
query = oso.authorize_query(user, Expense)
db.session.query(Expense).filter(query).all()
```

Do you have a preference?

dhatch: Not really. Although one advantage of the latter is we wouldn't need the Enforcer to know how to construct a session.

We could also achieve that by passing the session in to `authorize_query`, or passing the query object itself into `authorize_query` (which would already have the session associated with it).

dhatch (approved):

languages/python/sqlalchemy-oso/sqlalchemy_oso/enforcer.py
gkaemmer: I tried out this other API last night, but given the way that `authorize_model` works, we need the session anyway. I'm going to merge this as-is, but we should definitely revisit this once we're using the new data filtering methods.

PR #1083, opened 2021-08-16
Add field-level methods to Enforcer API

gkaemmer:
PR checklist:

- [ ] Added changelog entry.

### DIFF
--- a/languages/python/oso/oso/enforcer.py
+++ b/languages/python/oso/oso/enforcer.py
@@ -122,3 +122,20 @@ class Enforcer:
             actions.add(action)
 
         return list(actions)
+
+    def authorize_field(self, actor, action, resource, field):
+        """Ensure that ``actor`` is allowed to perform ``action`` on a given
+        ``resource``'s ``field``.
+
+        If the action is permitted with an ``allow_field`` rule in the policy,
+        then this method returns ``None``. If the action is not permitted by the
+        policy, this method will raise a ``ForbiddenError`.
+
+        :param actor: The actor performing the request.
+        :param action: The action the actor is attempting to perform on the
+        field.
+        :param resource: The resource being accessed.
+        :param field: The name of the field being accessed.
+        """
+        if not self.policy.query_rule_once("allow_field", actor, action, resource, field):
+            raise self._get_error(False)
--- a/languages/python/oso/oso/enforcer.py
+++ b/languages/python/oso/oso/enforcer.py
@@ -1,4 +1,4 @@
-from typing import List
+from typing import List, Any
 from polar import Variable, exceptions
 from .exceptions import NotFoundError, ForbiddenError
 from .oso import Policy
@@ -84,7 +84,7 @@ class Enforcer:
         if not self.policy.query_rule_once("allow_request", actor, request):
             raise self._get_error(False)
 
-    def authorized_actions(self, actor, resource, allow_wildcard=False) -> List[any]:
+    def authorized_actions(self, actor, resource, allow_wildcard=False) -> List[Any]:
         """Determine the actions ``actor`` is allowed to take on ``resource``.
 
         Collects all actions allowed by allow rules in the Polar policy for the
@@ -145,7 +145,7 @@ class Enforcer:
 
     def authorized_fields(
         self, actor, action, resource, allow_wildcard=False
-    ) -> List[any]:
+    ) -> List[Any]:
         """Determine the fields of ``resource`` on which ``actor`` is allowed to
         perform  ``action``.
 
--- a/languages/python/oso/oso/enforcer.py
+++ b/languages/python/oso/oso/enforcer.py
@@ -1,3 +1,4 @@
+from typing import List
 from polar import Variable, exceptions
 from .exceptions import NotFoundError, ForbiddenError
 from .oso import Policy
@@ -83,7 +84,7 @@ class Enforcer:
         if not self.policy.query_rule_once("allow_request", actor, request):
             raise self._get_error(False)
 
-    def authorized_actions(self, actor, resource, allow_wildcard=False) -> list:
+    def authorized_actions(self, actor, resource, allow_wildcard=False) -> List[any]:
         """Determine the actions ``actor`` is allowed to take on ``resource``.
 
         Collects all actions allowed by allow rules in the Polar policy for the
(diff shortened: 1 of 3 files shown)


languages/python/oso/oso/enforcer.py
```diff
@@ -122,3 +122,62 @@ def authorized_actions(self, actor, resource, allow_wildcard=False) -> list:
...
+        field.
+        :param resource: The resource being accessed.
+        :param field: The name of the field being accessed.
+        """
+        if not self.policy.query_rule_once(
+            "allow_field", actor, action, resource, field
+        ):
+            raise self._get_error(False)
+
+    def authorized_fields(self, actor, action, resource, allow_wildcard=False) -> list:
```
samscott89, line 145: Would it make sense to combine this with an `authorize` check? (ditto for `authorize_field`).

I think the two would almost always be used together. First you check the action on the resource, then you check the fields.

gkaemmer: Hmm, I'm thinking about this. Not sure it'd be ergonomic. It makes some sense for `read`, but I feel like as a user I'd expect to explicitly perform both resource-level and field-level authorization on my own terms. I.e. they might happen in slightly different places, and the errors might be handled differently.

For example, I've been thinking about `authorize_field` method primarily being useful in a graphql setting, where you would have already resolved (and authorized) the parent resource.

And the `authorized_fields` would be useful for filtering the fields of an object, either coming in (POST parameters) or going out (JSON response). But again I'd expect the resource-level authorization to happen in a different place.

Also I think of field level authorization as `can this user perform this action on this field`, which doesn't _necessarily_ mean that the answer to `can this user perform this action on this resource` is true. In other words the "action space" of fields might be different than the "action space" of resources.

Does that make sense?

samscott89: > Hmm, I'm thinking about this. Not sure it'd be ergonomic. It makes some sense for `read`, but I feel like as a user I'd expect to explicitly perform both resource-level and field-level authorization on my own terms. I.e. they might happen in slightly different places, and the errors might be handled differently.
> 
> For example, I've been thinking about `authorize_field` method primarily being useful in a graphql setting, where you would have already resolved (and authorized) the parent resource.
> 
> And the `authorized_fields` would be useful for filtering the fields of an object, either coming in (POST parameters) or going out (JSON response). But again I'd expect the resource-level authorization to happen in a different place.

Yeah, I can definitely see the argument that it should be possible to split these two. 

> Also I think of field level authorization as `can this user perform this action on this field`, which doesn't _necessarily_ mean that the answer to `can this user perform this action on this resource` is true. In other words the "action space" of fields might be different than the "action space" of resources.

I don't agree with this. In my mind, having access to the resource should be a necessary prerequisite to having access to a field. You could suggest people do this in the policy:
```
allow_field(user, action, resource, field) if allow(user, action, resource) and ...
```

or through the enforcement API. Or by asking people to do the two authorize calls.

Depends on what we think the 80% case is... Should we optimise for the two being coupled? For example, an option would be making it `oso.authorize(user, "read", repo, fields=["name", "email"])` with fields as an optional param that does the two checks we spoke about. And `fields=None` is the default.

Pros/cons of that specific example: makes the API more complex, but it folds the best practice into one check.


Where I'm at is that we don't yet have enough knowledge on what the 80% case is here. So better to keep the two separate methods for now, and revisit later?

gkaemmer: Yeah good points -- I'd be curious to see more of how people end up using field-level enforcement. It's possible that `"read"` is in fact the 80% use-case and we should just optimize the methods around read-only access (making use of `enforcer.read_action`).

But for now, I think I share your hesitance to combine them into one method. How does this look otherwise?

languages/python/oso/oso/enforcer.py
```diff
@@ -122,3 +122,62 @@ def authorized_actions(self, actor, resource, allow_wildcard=False) -> list:
...
+
+        Uses ``allow_field`` rules in the policy to find all allowed fields.
+
+        :param actor: The actor for whom to collect allowed fields.
+        :param action: The action being taken on the field.
+        :param resource: The resource being accessed.
+
+        :param allow_wildcard: Flag to determine behavior if the policy \
+        includes a wildcard field. E.g., a rule allowing any field: \
+        ``allow(_actor, _field, _resource)``. If ``True``, the method will \
```
samscott89, line 157: This should be `allow_field(_actor, _action, _resource, _field)`

languages/python/oso/oso/enforcer.py
```diff
@@ -122,3 +122,62 @@ def authorized_actions(self, actor, resource, allow_wildcard=False) -> list:
             actions.add(action)
 
         return list(actions)
+
+    def authorize_field(self, actor, action, resource, field):
+        """Ensure that ``actor`` is allowed to perform ``action`` on a given
+        ``resource``'s ``field``.
+
+        If the action is permitted with an ``allow_field`` rule in the policy,
```
samscott89, line 130:
```suggestion
        If the action is permitted by an ``allow_field`` rule in the policy,
```

samscott89 (approved):
A couple of small docs nits and this looks good to me!

languages/python/oso/oso/enforcer.py
```diff
@@ -122,3 +122,62 @@ def authorized_actions(self, actor, resource, allow_wildcard=False) -> list:
...
+        :param resource: The resource being accessed.
+        :param field: The name of the field being accessed.
+        """
+        if not self.policy.query_rule_once(
+            "allow_field", actor, action, resource, field
+        ):
+            raise self._get_error(False)
+
+    def authorized_fields(self, actor, action, resource, allow_wildcard=False) -> list:
+        """Determine the fields of ``resource`` on which ``actor`` is allowed to
```
dhatch, line 146: Pretty sure this type hint should be `List[str]` since we support Python < 3.9. https://mypy.readthedocs.io/en/stable/builtin_types.html?highlight=list#generic-types

If CI isn't catching this we should look into why.

samscott89: It can't be `List[str]` though because you might get an enum type back as an action. `List[any]`?

languages/python/oso/oso/enforcer.py
```diff
@@ -122,3 +123,65 @@ def authorized_actions(self, actor, resource, allow_wildcard=False) -> list:
...
+
+        Uses ``allow_field`` rules in the policy to find all allowed fields.
+
+        :param actor: The actor for whom to collect allowed fields.
+        :param action: The action being taken on the field.
+        :param resource: The resource being accessed.
+
+        :param allow_wildcard: Flag to determine behavior if the policy \
+        includes a wildcard field. E.g., a rule allowing any field: \
+        ``allow_field(_actor, _field, _resource, _field)``. If ``True``, the \
```
samscott89, line 160:
```suggestion
        ``allow_field(_actor, _action, _resource, _field)``. If ``True``, the \
```

gkaemmer: Good catch, my bad!

PR #1084, opened 2021-08-17
Register classes as authorization entity types

leina05:
This feature adds the `entity_type` argument to `register_class`. 

Valid options for `entity_type` are currently `OsoActor, OsoResource, and OsoGroup`. Passing one of these to `register_class` will register that class as the provided entity type.

When a class is registered as an entity type, instances of the class will match the entity type specializer in Polar rules at evaluation time, e.g.

```python
allow(actor: OsoActor, "read", resource: OsoResource) if...
```
An allow query with `actor=User()` and `resource=Repository()` will succeed if `User` is registered as `OsoActor` and `Repository` is registered as `OsoResource`.

Rules written with the class as a specializer will also match rule prototypes written with entity type specializers, e.g.

```python
type has_role(_actor: OsoActor, _role: String, _resource: OsoResource);

# this rule will match the above prototype if User is registered as an OsoActor and Repository is registered as an OsoResource
has_role(actor: User, role: String, resource: Repository) if...
```

PR checklist:

- [ ] Added changelog entry.


languages/python/oso/polar/host.py
```diff
@@ -84,14 +143,16 @@ def get_class(self, name):
         except KeyError:
             raise UnregisteredClassError(name)
 
-    def cache_class(self, cls, name=None):
+    def cache_class(self, cls, entity_type, name=None):
```
samscott89, line 146:
```suggestion
    def cache_class(self, cls, entity_type=None, name=None):
```

languages/python/oso/polar/host.py
```diff
@@ -17,6 +17,62 @@
...
+class OsoActor:
+    pass
+
+
+class OsoGroup:
+    pass
+
+
+class EntityMap:
+    ENTITY_TYPES = {
```
samscott89, line 36: I don't quite see why these need to be stringified, and then the API takes in a string. Why not have the user import these as classes or constants?

samscott89 (commented):
@gj in #1066 we spoke about groups, and we ended up using them implicitly (i.e. "teams" were referenced in the data lookup, but we never wrote any rules _for_ the teams).

Do we need to have a group type? If we got rid of that, might it make sense to instead have `register_actor` and `register_resource` as the two methods? (Rationale being: for three, maybe it makes sense to make it a param, but for two maybe two methods is clearer).

languages/python/oso/polar/host.py
```diff
@@ -17,6 +17,62 @@
 from .expression import Expression, Pattern
 
 
+# Entity classes
+
+
+class OsoResource:
```
samscott89, line 23: How about just `Resource` and `Actor` as the names?

samscott89: I suppose that's more likely to collide with an actual class... maybe? Or we namespace them in polar as `oso::Actor`. But that's basically the same thing.

languages/python/oso/polar/polar.py
```diff
@@ -84,6 +84,9 @@ def __init__(self, classes=CLASSES):
         self.register_class(str, name="String")
         self.register_class(datetime, name="Datetime")
         self.register_class(timedelta, name="Timedelta")
+        self.register_class(OsoActor, name="OsoActor", entity_type="OsoActor")
```
samscott89, line 87: This probably isn't a big deal, but it's a little weird that a user could start doing things like `actor = new OsoActor()` and it's going to create a random instance. Not sure there's any need to fix this, but there might be similar implications?

PR #1085, opened 2021-08-17
data filtering cleanup

gw000:
Addresses issues that came up during the review of the data filtering POC branch.

- Code reorganized for clarity
- Nested loops refactored
- Less string-typed things
- Fix variable set merging
- Use unique temp var names
- Clean up Python type info
- Handle non-partial results from allow query

*most* of the functions in `data_filtering.rs` haven't been changed that much, just moved into struct/interface implementations & maybe tweaked a lil bit to take out the explicit loops. `collapse_vars` is entirely rewritten as a `From<VarInfo>` implementation for `Vars`.


polar-core/src/data_filtering.rs
```diff
@@ -63,737 +67,635 @@ pub struct FetchRequest {
...
+            // Give each cycle an id
+            .into_iter()
+            .map(|c| (counter.next(), c))
+            .collect::<HashMap<_, _>>();
+
+        let mut assign_id = |item| get_var_id(&mut variables, item, &counter);
+
+        // now convert the remaining VarInfo fields into equivalent Vars fields.
+
+        let in_relationships = info
```
saolsen, line 178: The reason this was down lower before is because we don't know if `p1 = p2` yet. They could be `foo` and `bar` and maybe somewhere else there's a `foo = bar` but if we do this check before knowing that we'll miss it.

saolsen: We should probably add a test for that case since it's only in my mind.

gw000: are you thinking of `field_relationships`?

languages/python/oso/polar/data_filtering.py
```diff
@@ -60,21 +61,23 @@ class Constraint:
     field: str
     value: Any
 
-    def to_predicate(self):
-        def known_value(x):
-            return self.value
+    def __post_init__(self):
```
saolsen, line 64: I didn't know about `__post_init__`, kewl.

languages/python/oso/polar/host.py
```diff
@@ -23,75 +34,69 @@ class Host:
...
         self._accept_expression = False  # default, see set_accept_expression
 
-        # Check the types.
-        def default_get_field(obj, field):
-            return self.types_get_field(obj, field)
-
-        self.get_field = get_field or default_get_field
+        self.get_field = get_field or self.types_get_field
 
     # @Q: I'm not really sure what I'm returning here.
```
saolsen, line 49: Might as well drop this commentnow.

languages/python/oso/polar/polar.py
```diff
@@ -312,14 +306,19 @@ def get_allowed_resources(self, actor, action, cls) -> list:
...
-        plan = self.ffi_polar.build_filter_plan(types, results, "resource", class_name)
-        return filter_data(self, plan)
+                if isinstance(v, Expression):
+                    partial.append({"bindings": {k: self.host.to_polar(v)}})
+                else:
+                    complete.append(v)
+
+        types = serialize_types(self.host.distinct_user_types(), self.host.types)
+        plan = self.ffi_polar.build_filter_plan(types, partial, "resource", class_name)
+        complete += filter_data(self, plan)
```
saolsen, line 320: 🥇

polar-core/src/data_filtering.rs
```diff
@@ -63,737 +67,632 @@ pub struct FetchRequest {
...
     eq_values: Vec<(Symbol, Term)>,                     // x = 1;
     contained_values: Vec<(Term, Symbol)>,              // 1 in x
     field_relationships: Vec<(Symbol, String, Symbol)>, // x.a = y
     in_relationships: Vec<(Symbol, Symbol)>,            // x in y
+    counter: Counter,
+}
+
+#[derive(Debug)]
+struct Vars {
+    variables: HashMap<Id, HashSet<Symbol>>,
```
saolsen, line 96: Dave requested comments, similar to the ones on VarInfo, that explain what each of these are.

saolsen (approved):
Beautiful

PR #1087, opened 2021-08-18
Add target to run docs without api docs.

dhatch:


gkaemmer (changes requested):
Let's get rid of the package-lock, unless you added it for a reason. Other than that looks good.

Maybe also add a `dev` task? 😃
```
dev: install
  HUGO_PURGECSS=false hugo server
```

docs/Makefile
```diff
@@ -1,6 +1,9 @@
 run: api_docs install
 	hugo server
```
gj, line 2: nit

```suggestion
run: api_docs
	$(MAKE) run_without_api_docs
```

gj (approved):
Don't particularly care whether or not these lockfiles remain version-controlled, but we should probably standardize on Yarn or NPM project-wide. I think we use Yarn in most places.

dhatch:
> should probably standardize on Yarn or NPM project-wide. I think we use Yarn in most places.

@gj Is using Yarn vs. NPM the reason the lock file version changed?

gkaemmer (approved):

gj:
> > should probably standardize on Yarn or NPM project-wide. I think we use Yarn in most places.
> 
> @gj Is using Yarn vs. NPM the reason the lock file version changed?

nope; using Yarn would've generated a (new) `yarn.lock` file in addition to the existing `package-lock.json` file generated by NPM

Issue #1089, opened 2021-08-18
Required file not being autoloaded in gem


davidpelaez:
Given a Gemfile declaring the Oso gem as a dependency with `gem 'oso-oso', '~> 0.15.0'` when the gem is loaded it fails with:

```ruby
Traceback (most recent call last):
	3: from server.rb:9:in `<main>'
	2: from server.rb:9:in `new'
	1: from /Users/david/Ephemeral/oso-auth/oso-ruby-quickstart/vendor/oso/languages/ruby/lib/oso/oso.rb:9:in `initialize'
/Users/david/Ephemeral/oso-auth/oso-ruby-quickstart/vendor/oso/languages/ruby/lib/oso/polar/polar.rb:38:in `initialize': uninitialized constant Oso::Polar::Polar::FFI (NameError)
```

Ignore the `vendor/...`, that's showing up because of my setup for development purposes but it matches exactly the current head of the Oso main repo.

This happens because the gem by default loads https://github.com/osohq/oso/blob/main/languages/ruby/lib/oso/oso.rb instead of https://github.com/osohq/oso/blob/main/languages/ruby/lib/oso.rb (which is one level up in the file tree). As a result the following requires are not executing:

https://github.com/osohq/oso/blob/main/languages/ruby/lib/oso.rb#L3-L5

The root issue is that since the gem is called `oso-oso` requiring the gem will load `lib/oso-oso.rb` but not `lib/oso.rb`. 

I came across this issue running the Ruby quickstart which doesn't work even if the version of the gem is updated to the latest. It can be fixed with a manual require but ideally it should work out of the box simply by putting it in the Gemfile of my project.

PR #1090, opened 2021-08-18
Fix file loading issue with Ruby gem - Fixes osohq/oso#1089

davidpelaez:
Here's a small fix to the Ruby gem for Oso that wasn't loading as expected out of the box. It fixes https://github.com/osohq/oso/issues/1089

I wasn't able to find an applicable changelog file to register this. Nor did I know if I was supposed to bump the version in the code. Please let me know if this solution works.

Since it's my first time contributing please give me instructions on accepting the contributors agreement.

Thank you for Oso and the help on Slack!


davidpelaez:
Just noticed an issue with the tests, updating this moment

davidpelaez:

saolsen:
Thanks for the pr! Bumping the version of the gem happens when we cut a new release. Can you please reset that back to what it was (otherwise the tests will fail). It will probably be 0.15.1 but not until the next release which should be next week.

davidpelaez:
@saolsen just undo the change to the version. I also updated my email because the CLA bot wasn't recognizing me. Hope it works now.

davidpelaez:
This failed because the file isn't called `oso_oso.rb` however it's called that way because it's the name of the gem. What should I do in this case?

davidpelaez:
I made some research and the entry file of the gem has to be called like the gem itself. Therefore I added an exception on the Rubocop file to cover this and solve that error. The other error seems to be related to the require calling `oso` but the gem is actually `oso-oso` so I suppose that error comes from some earlier time before the gem was published and you thought the name `oso` was available. Also the Gemfile defined the dependency but without requiring Bundler it wasn't being loaded. I simply added the required setup for the Gemfile to be considered and updated the requires to match the actual name of the gem as a regular user would fix it. This should (hopefully) now make all the tests pass. Please let me know if there any other required changes that you'd like to be made. Thank you 🤓👍

davidpelaez:
One more error but I don't understand what the error "Not okay because no label, and diff failed." means in the PR Docs changes check. Could you please provide instructions on how to fix? The other errors related to the Ruby code seem to be solved now.

gj:
Hey @davidpelaez, thank you so much for the contribution! We're looking into the implications right now and will deal with it ASAP.

At the moment, I'm having trouble reproducing the issue with https://github.com/osohq/oso-ruby-quickstart. Would you be able to provide details about your environment (OS, Ruby version, etc.) and the commands you're running that result in the error? Thanks!

languages/ruby/.rubocop.yml
```diff
@@ -3,5 +3,6 @@ AllCops:
   Exclude:
     - '**/*~'
     - 'bin/oso'
+    - 'lib/oso-oso.rb'
```
gkaemmer, line 6: Instead of excluding this file from `AllCops`, let's just exclude it from `Naming/FileName`:

```yaml
Naming/FileName:
  Exclude:
    - 'lib/oso-oso.rb'
```

gkaemmer (dismissed):
Hey -- thanks for this PR! This is definitely a good thing to fix, especially because the error is hard to pin down at a glance.

Backwards-compatibility might be an issue here -- I think this will break things for people who are only running `require "oso"` which some of our docs suggest is possible.

Instead of renaming `oso.rb` to `oso-oso.rb`, could we instead keep `oso.rb` as-is and add a shim in `oso-oso.rb` that looks like this?

```ruby
require_relative "./oso"
```

That way, both `require "oso"` AND `require "oso-oso"` would work, and we're not under pressure to immediately update the docs or notify folks of a breaking change.

PR #1091, opened 2021-08-18
Ruby data filtering

gw000:
Pretty complete. Also

- implements `ExternalIsSubclass` and `ExternalIsaWithPath` for the Ruby library (needed for data filtering)
- lets data filtering results be compared directly, instead of just their fields (requires a lil bit of host support which is added to Python)
- fixes a couple of infinite loops in the VM (now we can answer `x = [x] and x in x`)

PR checklist:

- [ ] Added changelog entry.


languages/ruby/lib/oso/polar/host.rb
```diff
@@ -190,6 +210,10 @@ def subspecializer?(instance_id, left_tag:, right_tag:)
         left_index && right_index && left_index < right_index
       end
 
+      def subclass?(left_tag:, right_tag:)
+        get_class(left_tag) <= get_class(right_tag)
```
gkaemmer, line 214: Lol this is the most cursed less-than-or-equal-to operator ever built

languages/ruby/lib/oso/polar/polar.rb
```diff
@@ -88,6 +88,49 @@ def self.join(separator, left, right)
...
+        if host.types.key? klass
+          class_name = host.types[klass].name
+        elsif host.types.key? klass.name
+          class_name = host.types[klass.name].name
+        else
+          raise "Unknown class `#{klass}`"
+        end
+
+        constraint = Expression.new(
+          'And',
```
gkaemmer, line 104: Why does this need to be an `And` with a single argument?

languages/ruby/lib/oso/polar/polar.rb
```diff
@@ -88,6 +88,49 @@ def self.join(separator, left, right)
...
+        resource = Variable.new 'resource'
+
+        # FIXME(gw) kind of a hack because of class autoreloading, maybe not needed?
+        if host.types.key? klass
+          class_name = host.types[klass].name
+        elsif host.types.key? klass.name
+          class_name = host.types[klass.name].name
+        else
+          raise "Unknown class `#{klass}`"
+        end
```
gkaemmer, line 101: Hmm this is kind of weird -- seems like we kind of want to store a mapping from user-supplied type name AND from class name so that you can look up the `UserType` based on a class even if the `name` they provided doesn't match.

```ruby
oso.register_class(Expense, "MyApp::Expense")
# ... later
oso.get_allowed_resources(actor, "read", Expense)
# OR
oso.get_allowed_resources(actor, "read", "Expense")
# OR
oso.get_allowed_resources(actor, "read", "MyApp::Expense")
```

languages/ruby/lib/oso/polar/polar.rb
```diff
@@ -88,6 +88,49 @@ def self.join(separator, left, right)
...
+          action,
+          resource,
+          bindings: { 'resource' => constraint },
+          accept_expression: true
+        )
+
+        complete = []
+        partial = []
+
+        results.to_a.each do |result|
```
gkaemmer, line 120: Nit alert: do you need `to_a` here?

languages/ruby/lib/oso/polar/polar.rb
```diff
@@ -88,6 +88,49 @@ def self.join(separator, left, right)
...
+          result.to_a.each do |key, val|
+            if val.is_a? Expression
+              partial.push({ 'bindings' => { key => host.to_polar(val) } })
+            else
+              complete.push val
+            end
+          end
+        end
+        types = host.serialize_types
```
gkaemmer, line 129: Hmm ... should these be stored on the VM? Perhaps out of scope here, but a thought for the future perhaps

languages/ruby/lib/oso/polar/polar.rb
```diff
@@ -181,8 +236,8 @@ def query_rule(name, *args)
       # under a previously-registered name.
       # @raise [FFI::Error] if the FFI call returns an error.
       # @return [self] for chaining.
-      def register_class(cls, name: nil)
-        name = host.cache_class(cls, name: name || cls.name)
+      def register_class(cls, name: nil, fields: {}, fetcher: nil)
```
gkaemmer, line 239: This is going to have to change a bit with the enforcer stuff, but I suppose that's a job for next cycle

languages/ruby/lib/oso/polar/query.rb
```diff
@@ -71,6 +72,9 @@ def application_error(message)
       # @raise [Error] if the FFI call raises one.
       def handle_call(attribute, call_id:, instance:, args:, kwargs:) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
         instance = host.to_ruby(instance)
+        rel = get_relationship(instance.class, attribute)
+        return handle_relationship(call_id, instance, rel) unless rel.nil?
```
gkaemmer, line 76: I think I'm confused -- why does this need to be handled here? Shouldn't the VM be the one handling relationships?

Does the host library handle relationships on bound variables while the VM handles relationships on unbound variables?

```
allow(user, "read", document) if
  folder in user.folders and
#               ^ here, user is bound
  document in folder.documents;
#                   ^ here, folder is not bound
```

languages/ruby/lib/oso/polar/host.rb
saolsen, line 214: oh wow, ruby is cool sometimes

languages/ruby/lib/oso/polar/polar.rb
```diff
@@ -150,17 +193,27 @@ def load_str(str, filename: nil) # rubocop:disable Metrics/AbcSize, Metrics/Meth
...
+      end
+
+      # Query the knowledge base to determine whether an actor is allowed to
+      # perform an action upon a resource.
+      #
+      # @param actor [Object] Subject.
+      # @param action [Object] Verb.
+      # @param resource [Object] Object.
+      # @return [Boolean] An access control decision.
+      def allowed?(actor:, action:, resource:)
```
saolsen, line 215: Why did this get moved to polar.rb?

gkaemmer (commented):
Some comments!

polar-core/src/vm.rs
```diff
@@ -1393,6 +1392,13 @@ impl PolarVirtualMachine {
     /// consists of unifying the rule head with the arguments, then
     /// querying for each body clause.
     fn query(&mut self, term: &Term) -> PolarResult<QueryEvent> {
+        /*
+        if self.queries.iter().find(|q| *q == term).is_some() {
+            self.push_goal(Goal::Backtrack)?;
+            return Ok(QueryEvent::None);
+        }
+        */
+
```
saolsen, line 1401: commented out code

saolsen (commented):
Looks really good!

languages/ruby/lib/oso/polar/polar.rb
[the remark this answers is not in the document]
gw000: to make it accessible in the tests

languages/ruby/lib/oso/polar/host.rb
gw000: 🤣

languages/ruby/lib/oso/polar/polar.rb
gw000: that's just kind of the format the vm uses internally for partials right now

languages/ruby/lib/oso/polar/polar.rb
gw000: yes there does, otherwise it complains about calling a private method, go fig

languages/ruby/lib/oso/polar/polar.rb
gw000: I think the config work is moving in that direction so we'll probably and up sharing that functionality eventually

languages/ruby/lib/oso/polar/query.rb
gw000: no, it's not about bound vs. unbound. if we end up storing type information in the VM then we could do some of the work there. but here a relationship could be handled by an ORM or something, so it has to be handled in the host.

languages/ruby/lib/oso/polar/polar.rb
gw000: to make it accessible from tests without changing them to use an `Oso` instead of a `Polar`

polar-core/src/vm.rs
gw000: ✂️

languages/ruby/lib/oso/polar/ffi/polar.rb
```diff
@@ -51,6 +57,15 @@ def validate_roles_config(config)
           handle_error if result.zero?
         end
 
+        def build_filter_plan(types, partials, variable, class_tag)
+          types = JSON.dump(types)
+          partials = JSON.dump(partials)
+          plan = Rust.build_filter_plan(self, types, partials, variable, class_tag)
```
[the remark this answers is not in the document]
gw000: no reason except that's not how we do it in python. maybe something to do next cycle

languages/ruby/lib/oso/polar/polar.rb
gw000: that's a good idea but what about name collisions?

gkaemmer: Probably going with the name of the registered class would be a safe bet? Not sure though. I think the most important test case to get working here would be:
```ruby
oso.register_class(Expense, "MyApp::Expense")

oso.get_allowed_resources(actor, "read", Expense)
```
Which should be able to find the `MyApp::Expense` name to send with the query. I think right now that would throw an `Unknown class "Expense"` error.

languages/ruby/lib/oso/polar/query.rb
gkaemmer: Ohh I think I maybe understand this now. Question: if we were to get rid of this, data filtering would still work, right? It just would potentially eagerly load one "layer" of relationships?

```
allow(user, "read", document) if
  folder in user.folders and
  document in folder.documents;
```
^ For example, here `user.folders` would end up bound to an array of folders, instead of being bound to a variable?

languages/ruby/lib/oso/polar/ffi/polar.rb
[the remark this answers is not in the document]
gw000: Ok having thought about it a little, I think an FFI class would be overkill since the filter plan is just data with no Rust code it needs to call. But I did add some classes to separate the json wrangling from the actual filter logic, lmk if it's better

saolsen (approved):

languages/ruby/lib/oso/polar/polar.rb
gw000: Added a test. It works since we use the class object as a key as well as the name

languages/ruby/lib/oso/polar/query.rb
gw000: no, it would break. a relationship is kind of like a "virtual" field on a type, it doesn't necessarily exist on the external instance, but the host library handles it so it looks as if it did.

gkaemmer (approved):
LGTM

PR #1092, opened 2021-08-18
Enforcement docs

gkaemmer:
Note that `query.md` is not really in scope here -- it's going to need to wait for the next version of data filtering.


docs/content/any/getting-started/enforcement/_index.md
```diff
@@ -0,0 +1,22 @@
...
+any: true
+# showContentForAnyLanguage: true
+# draft: True
+---
+
+# Use Oso Policies in an App
+
+To use an Oso policy in your app, you'll need to "enforce" it. A policy is
+useless without an app that consults the policy on user actions. For most apps,
+policies can be enforced on four "levels":
```
samscott89, line 13: This says four, but query is commented out

docs/content/python/guides/enforcement/_index.md
```diff
@@ -0,0 +1,32 @@
+---
+title: Enforcing an Oso Policy (Oso 0.20.0 Alpha)
+weight: 10
+any: true
+# showContentForAnyLanguage: true
+# draft: True
+---
+
+{{% callout "Note: 0.20.0 Alpha Feature" %}}
```
dhatch, line 9: Might be good to make this a shortcake as we'll have it on several pages now.

docs/content/python/guides/enforcement/_index.md
```diff
@@ -0,0 +1,32 @@
+---
+title: Enforcing an Oso Policy (Oso 0.20.0 Alpha)
+weight: 10
```
dhatch, line 3: Can you weight so it shows up above `More` for Python

docs/content/python/guides/enforcement/enforcer.md
```diff
@@ -0,0 +1,58 @@
...
+  # Make the oso enforcer accessible throughout the app
+  app.oso = oso
+```
+
+An enforcer exposes a number of useful methods that your app can use. Each one
+queries your policy for `allow`, `allow_request`, or `allow_field` rules. To
+learn more about writing those rules, read about [writing a
+policy](../../getting-started/policies).
+
+- `authorize(user, action, resource)`: Ensure that a user can perform an action
```
dhatch, line 48: Would be great to link to API documentation here for each method. You can use the apiDeepLink shortcode for that.

docs/content/python/guides/enforcement/resource.md
```diff
@@ -0,0 +1,115 @@
...
+Is the current user allowed to perform a certain action on a certain resource?
+This is the central question of **"resource-level" enforcement.**
+
+Resource-level enforcement is the bread-and-butter of application authorization.
+If you only perform one type of authorization in your app, it should be
+this. **Just about every endpoint in your application should perform some kind
+of resource-level enforcement.**
+
+The method you use for resource-level authorization is called `authorize`, and
+is exposed by an Oso enforcer. The `authorize` method takes three arguments,
```
dhatch, line 32: Link to API docs

docs/content/python/guides/enforcement/resource.md
```diff
@@ -0,0 +1,115 @@
...
+oso.authorize(user, "approve", expense)
+```
+
+The `authorize` method checks all of the `allow` rules in your policy, and
+ensures that there is an `allow` rule for that applies to the given user,
+action, and resource:
+
+```polar
+allow(user: User, "approve", expense: Expense) if
+    expense.org = org and
```
dhatch, line 50: Is there a reason you wrote this `expense.org = org` instead of `org = expense.org`? I usually write the variable in an assignment on the LHS so that it's more intuitive for users used to `=` meaning assignment.

docs/content/python/guides/enforcement/resource.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Authorization Failure
+
+What happens when the authorization fails? That is, what if there is not an
+`allow` rule that gives the user permission to perform the action on the
+resource? In that case, the `authorize` method will raise an
+`AuthorizationError`. There are actually two types of authorization errors,
+depending on the situation.
+
+- `NotFound` errors are for situations where the user should not even know
```
dhatch, line 77: These are called `ForbiddenError` & `NotFoundError` in the API. Use those names instead here?

Also link to API docs.

docs/content/python/guides/enforcement/resource.md
```diff
@@ -0,0 +1,115 @@
...
+the resource, the server should act as though it doesn't exist.
+{{% /minicallout %}}
+
+You could handle these errors at each place you call `authorize`, but that would
+mean a lot of error handling. We recommend handling `NotFound` and `Forbidden`
+errors globally in your application, using middleware or something similar.
+Ideally, you can perform resource-level authorization by adding a single line of
+code to each endpoint.
+
+As an example, here's what a global error handler looks like in a flask app:
```
dhatch, line 98:
```suggestion
As an example, here's what a global error handler looks like in a Flask app:
```

docs/content/python/guides/enforcement/resource.md
```diff
@@ -0,0 +1,115 @@
...
+
+Is the current user allowed to perform a certain action on a certain resource?
+This is the central question of **"resource-level" enforcement.**
+
+Resource-level enforcement is the bread-and-butter of application authorization.
+If you only perform one type of authorization in your app, it should be
+this. **Just about every endpoint in your application should perform some kind
+of resource-level enforcement.**
+
+The method you use for resource-level authorization is called `authorize`, and
```
dhatch, line 31: Maybe an h2 here: "Using `authorize`"? So that it shows up in page TOC on RHS

docs/content/python/guides/enforcement/field.md
```diff
@@ -0,0 +1,80 @@
...
+{{% /callout %}}
+
+<div class="pb-10"></div>
+
+# Field-level Enforcement
+
+Perhaps you're creating a user profile page, and in your app, only a user's
+friends can access their last check-in location. To enforce this type of
+authorization, it often makes sense to use **"field-level" enforcement**, by
+explicitly allowing access to certain fields on your domain objects.
```
dhatch, line 26: How do I know whether to use `field-level enforcement` or `resource-level enforcement`?

In this example, I could write a rule over the Location itself and only allow users to access the location if it's their own, or they're in the friends list of the user for the location.

docs/content/python/guides/enforcement/field.md
```diff
@@ -0,0 +1,80 @@
...
+
+Field-level authorization gives you fine-grained control over who can access
+exactly what bits of information. In polar, you can write field-level rules like
```
dhatch, line 29:
```suggestion
exactly what bits of information. In Polar, you can write field-level rules like
```

docs/content/python/guides/enforcement/field.md
```diff
@@ -0,0 +1,80 @@
...
+this:
+
+```polar
+allow_field(user, "read", profile, "last_check_in_location") if
```
dhatch, line 33: I think we should include specializers on most rules in our documentation examples since it's best practice.

docs/content/python/guides/enforcement/field.md
```diff
@@ -0,0 +1,80 @@
...
+    profile.user in user.friends;
+```
+
+Notice that an `allow_field` rule is just like an `allow` rule, except that it
+takes an additional argument: the field name.
+
+To enforce field-level authorization in your app, you use the `authorize_field`
+method provided by Oso enforcers.
```
dhatch, line 41:
```suggestion
method provided by an Oso enforcer.
```

this is the wording you've used previously

docs/content/python/guides/enforcement/field.md
```diff
@@ -0,0 +1,80 @@
...
+
+{{% minicallout %}}
+**Note**: You'll need an Oso enforcer to use the `authorize_field` method. Check out
+[how to build one](enforcer.html).
+{{% /minicallout %}}
+
+```python
+def get_last_check_in_location(profile, current_user):
+    oso.authorize_field(current_user, "read", profile, "last_check_in_location");
```
dhatch, line 50:
```suggestion
    oso.authorize_field(current_user, "read", profile, "last_check_in_location")
```

docs/content/python/guides/enforcement/field.md
```diff
@@ -0,0 +1,80 @@
...
+    return profile.last_check_in.location
+```
+
+Like `authorize`, `authorize_field` will raise an `AuthorizationError` when the
+user is not allowed to perform the given action. This is an error that you
+should handle globally in your app. You can read more details about this on the
```
dhatch, line 56:
```suggestion
should handle globally in your app. You can read more details about this in the
```

docs/content/python/guides/enforcement/field.md
```diff
@@ -0,0 +1,80 @@
...
+[Resource-level Enforcement Guide](resource.html#authorization-failure).
+
+Sometimes it is helpful to get
```
dhatch, line 59: Maybe a header for this?

docs/content/python/guides/enforcement/field.md
```diff
@@ -0,0 +1,80 @@
...
+_all_ fields that a user can access, and for this there is a separate method
+called `authorized_fields`:
+
+```python
+# Serialize only the fields of profile that the current user is allowed to read
+def serialize_profile(profile, current_user):
+    fields = oso.authorized_fields(current_user, "read", profile)
+    return { field: profile[field] for key in fields }
```
dhatch, line 67:
```suggestion
    return { key: profile[key] for key in fields }
```

And same in other example

dhatch: Also I believe it's best style in python to write as `{field: ... for key in fields}` (no leading or trailing spaces)

docs/content/python/guides/enforcement/field.md
```diff
@@ -0,0 +1,80 @@
...
+authorization, it often makes sense to use **"field-level" enforcement**, by
+explicitly allowing access to certain fields on your domain objects.
+
+Field-level authorization gives you fine-grained control over who can access
+exactly what bits of information. In polar, you can write field-level rules like
+this:
+
+```polar
+allow_field(user, "read", profile, "last_check_in_location") if
+    profile.user in user.friends;
```
dhatch, line 34: Is this backwards?

```suggestion
    user in profile.friends;
```

docs/content/python/guides/enforcement/field.md
```diff
@@ -0,0 +1,80 @@
...
+is explicitly allowed to read, or can similarly be used to filter _incoming_
+parameters from a user for a call to, say, an `update` method. In that case, you
+might use an `"update"` action in the call to `authorized_fields`:
+
+```python
+# Filter update_params by the fields on profile that the user can update
+def filter_update_params(profile, raw_update_params, current_user):
+    fields = oso.authorized_fields(current_user, "update", profile)
+    return { field: raw_update_params[field] for key in fields }
+```
```
dhatch, line 80: It would be good to include in this guide some advice about handling large number of fields, allowing users to read all fields, etc.

Field level enforcement is a bit unfortunate b/c it really needs to be synced up with the model and feels like it would change much more often than adding new models.

This is more of a product suggestion... but I wonder if we'd consider something like a field group.

Imagine something like:

```
class Profile:
   # Suppose this data could go in the policy too
   FIELD_GROUPS = {
       "anyone": ["name"],
       "friends_only": ["last_check_in_location"],
       "private": ["address"]
    }

# Oso enforcer hooks into field groups somehow, so you call
# authorized_fields in the same way
```

```
# polar

# anyone can read "anyone" fields
allow_field_group(_user, "read", _profile, "anyone");

# only friends can read friends fields
allow_field_group(user, "read", profile, "friends_only") if
    user in profile.friends;

# private fields
allow_field_group(user, "read", profile, "private") if
    user = profile;
```

Then, I can know to update `FIELD_GROUPS` when updating the model & adding a field and I don't need to write rules for every single field in my policy.

dhatch: Could also build the mapping of field to field group in the policy if we feel that's a better place for it.

docs/content/python/guides/enforcement/request.md
```diff
@@ -0,0 +1,102 @@
...
+If you'd prefer, you can also pass a string as the `request` argument. The
+string could hold the request's path or, if you're using GraphQL, the string
+could be the name of the current GraphQL query or mutation.
+
+Writing a request-level authorization policy means implementing an
+`allow_request` rule like this:
+
+```polar
+# Allow anyone to hit the login endpoint
+allow_request(_, Request{path: "/login"})
```
dhatch, line 64: This second parameter needs a name, and the rule needs a semicolon.
```suggestion

allow_request(_, _: Request{path: "/login"});
```

docs/content/python/guides/enforcement/request.md
```diff
@@ -0,0 +1,102 @@
...
+
+# Only allow access to payments by users with verified emails
+allow_request(user: User, request) if
```
dhatch, line 67: Use type on second param

docs/content/python/guides/enforcement/request.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Access token scopes
+
+You can use a request-level authorization policy to protect your endpoints using
+access token scopes. This can be used to build features like Github's [OAuth
+Scopes](https://docs.github.com/en/developers/apps/building-oauth-apps/scopes-for-oauth-apps).
+
+
+```polar
+allow_request(token: AccessToken, {method: "POST", path: "/repos"}) if
```
dhatch, line 90: Do you mean for the second arg to be a specializer?

```suggestion
allow_request(token: AccessToken, _ : Request{method: "POST", path: "/repos"}) if
```

dhatch (changes requested):
I didn't read the query guide since it's in draft.

Mostly small nits throughout these guides. The documentation is really well written and clear, and I found the examples you used throughout to be very relevant.

There were some errors in the code samples. Even though the docs are short, it might be good to extract these samples and use literalInclude so that you can test them (at minimum confirming the syntax for Python is valid & the policy loads). That would prevent these types of issues.

It would be good to link to API documentation throughout.

How are you planning on documenting the SQLAlchemy enforcer?

docs/content/python/guides/enforcement/_index.md
gkaemmer: Lol 🍰 

Good idea though

docs/content/python/guides/enforcement/resource.md
gkaemmer: Good point, hadn't really thought too much about it

docs/content/python/guides/enforcement/field.md
gkaemmer: I'll admit that I was hoping to avoid answering this question directly because it is a bit subtle. In general I would worry more about people using field-level rules where resource-level rules make more sense (as in this `Location` example). 

Perhaps an example with the `email` field would be better? It might more clearly suggest that it is about "fields on this object" instead of "objects related to this object".

docs/content/python/guides/enforcement/field.md
gkaemmer: Hmm this is a really good point and has me thinking. I don't think the current API is the final word on field-level enforcement.

For now, we could maybe add an example of an allow rule like this?
```
allow_field(user: User, "read", profile: Profile, field) if
  field in Profile.FRIENDS_ONLY_FIELDS and
  user in profile.friends;
```

gkaemmer:
> There were some errors in the code samples. Even though the docs are short, it might be good to extract these samples and use literalInclude so that you can test them (at minimum confirming the syntax for Python is valid & the policy loads). That would prevent these types of issues.

Generally I am a bit skeptical of the need to do this -- it makes the docs harder to tweak/view at a glance in markdown, and it splits our documentation code into N*M repos, where N is the number of languages we support and M is the number of guides.

> It would be good to link to API documentation throughout.

Agreed, will add. Didn't know about `apiDeepLink`.

> How are you planning on documenting the SQLAlchemy enforcer?

I kind of want to avoid documenting it until we have the data filtering API nailed down (i.e. it wouldn't make sense to document without documenting query-level enforcement, which I think is out of scope for now)

docs/content/python/guides/enforcement/_index.md
gkaemmer: Can't do this unless we redefine `callout` -- I was surprised to learn that hugo doesn't like using shortcodes from other shortcode definitions :(

https://discourse.gohugo.io/t/error-using-shortcodes-in-template-block-redefinition/22211

docs/content/python/guides/enforcement/field.md
dhatch: Yeah I think that example would bring up this question less.

docs/content/python/guides/enforcement/field.md
dhatch: yeah that's a cool example

dhatch:
> > There were some errors in the code samples. Even though the docs are short, it might be good to extract these samples and use literalInclude so that you can test them (at minimum confirming the syntax for Python is valid & the policy loads). That would prevent these types of issues.
> 
> Generally I am a bit skeptical of the need to do this -- it makes the docs harder to tweak/view at a glance in markdown, and it splits our documentation code into N*M repos, where N is the number of languages we support and M is the number of guides.
> 
> I'll give it a shot and see how I feel about it.

If you want to get this ready to merge first & then think about doing that that's cool.

> 
> > How are you planning on documenting the SQLAlchemy enforcer?
> 
> I kind of want to avoid documenting it until we have the data filtering API nailed down (i.e. it wouldn't make sense to document without documenting query-level enforcement, which I think is out of scope for now)

sounds good

dhatch (approved):
lgtm! I didn't run the docs & check the API links, just read md changes.

docs/content/python/guides/enforcement/field.md
gkaemmer: Added another section to this effect -- might appreciate a read-through in `field.md` if you have time.

PR #1094, opened 2021-08-19
JavaScript data filtering

saolsen:


languages/js/src/Polar.ts
```diff
@@ -208,23 +214,35 @@ export class Polar {
...
+    return new Query(ffiQuery, host, bindings).results;
   }
 
   /**
    * Query for a Polar rule.
    */
-  queryRule(name: string, ...args: unknown[]): QueryResult {
-    return this.query(new Predicate(name, args));
+  queryRule(name: string, bindings?: any, ...args: unknown[]): QueryResult {
+    return this.query(new Predicate(name, args), bindings);
```
gw000, line 224: this is a little confusing coming from the other host libraries. maybe a separate `queryRuleWithBindings` function?

languages/js/src/Polar.ts
```diff
@@ -235,6 +253,51 @@ export class Polar {
...
+    let jsonResults = queryResults.map(result => ({
+      // `Map<string, any> -> {[key: string]: PolarTerm}` b/c Maps aren't
+      // trivially `JSON.stringify()`-able.
+      bindings: [...result.entries()].reduce((obj: obj, [k, v]) => {
+        obj[k] = this.#host.toPolar(v);
+        return obj;
+      }, {}),
+    }));
+    let resultsStr = JSON.stringify(jsonResults);
+    console.log(this.#host.clsNames);
```
gw000, line 290: remember to take out

languages/js/src/Polar.ts
saolsen: Yeah I like that better

languages/js/src/dataFiltering.test.ts
```diff
@@ -0,0 +1,151 @@
...
+          break;
+        case 'In':
+          {
+            clause = `${name}.${c.field} IN (:...${c.field})`;
+          }
+          break;
+      }
+      let param: any = {};
+      param[c.field] = c.value;
+      if (!addedWhere) {
```
gkaemmer, line 78: Tiny note: I think you can use `andWhere` exclusively -- you don't necessarily need an original `where` call. 

I was curious about this during your demo and looked up how it works in the typeorm querybuilder, and they only add the string `AND` if it's not the first where clause: https://github.com/typeorm/typeorm/blob/51b2a63d91d4e6a935e83f47fb958f4867e82e9c/src/query-builder/QueryBuilder.ts#L685-L696

languages/js/src/Query.ts
```diff
@@ -116,15 +132,50 @@ export class Query {
...
+                const constraint = new Constraint(
+                  'Eq',
+                  fieldType.otherField,
+                  receiver[fieldType.myField]
+                );
+                const constraints = [constraint];
+                let results = otherClsFetcher(constraints);
+                results = await Promise.resolve(results);
+                if (fieldType.kind == 'parent') {
+                  value = results[0];
```
gw000, line 159: python & ruby assert there's only one parent

languages/js/src/dataFiltering.test.ts
```diff
@@ -0,0 +1,151 @@
...
+  await bars.save(helloBar);
+
+  let foos = connection.getRepository(Foo);
+
+  let anotherFoo = new Foo();
+  anotherFoo.id = 'Another';
+  anotherFoo.barId = 'hello';
+  anotherFoo.isFooey = true;
+
+  await foos.save(anotherFoo);
```
gw000, line 54: I'd have more confidence in the tests if there was some more data

languages/js/src/dataFiltering.test.ts
gw000: good find, changed it

languages/js/src/dataFiltering.test.ts
```diff
@@ -83,6 +109,11 @@ test('data filtering', async () => {
             clause = `${name}.${c.field} IN (:...${c.field})`;
           }
           break;
+        case 'Contains':
```
saolsen, line 112: Should we drop this if it doesn't work?

gw000 (approved):
☕

PR #1095, opened 2021-08-20
fields can be reserved words

gw000:
allowed: `x = {debug: 1, type: 2, new: 3} and x.new = x.debug + x.type`
not allowed:  `new(x);` `foo(debug);`

PR checklist:

- [x] Added changelog entry.

### DIFF
--- a/docs/content/any/project/changelogs/NEXT.md
+++ b/docs/content/any/project/changelogs/NEXT.md
@@ -48,6 +48,7 @@ Link to [relevant documentation section]().
 #### Other bugs & improvements
 
 - The debugger can now break on rule matches.
-- Reserved words can be used as field and method names in dictionaries and objects.
+- Polar reserved words (eg. `type`, `if`, `debug`) can be used as field and method names in
+  dictionaries and objects.
 
 #### TODO: enforcement changelog, if we decide to release it
--- a/docs/content/any/project/changelogs/NEXT.md
+++ b/docs/content/any/project/changelogs/NEXT.md
@@ -48,7 +48,7 @@ Link to [relevant documentation section]().
 #### Other bugs & improvements
 
 - The debugger can now break on rule matches.
-- Polar reserved words (eg. `type`, `if`, `debug`) can be used as field and method names in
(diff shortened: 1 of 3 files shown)


polar-core/src/polar.lalrpop
```diff
@@ -98,6 +114,11 @@ Boolean: Value = <b:"Boolean"> => {
 
 Name: Symbol = <s:"Symbol"> => s;
 
+ResName: Symbol = {
```
gkaemmer, line 117: Is this used anywhere?

gw000: 😮 no good catch ty!

polar-core/src/polar.lalrpop
```diff
@@ -126,6 +142,30 @@ Call: Value = {
     }
 };
 
+DotCall: Value = {
```
gj, line 145: I think we could probably reduce the duplication between `Call` and `DotCall` with a generic param, but I'm not sure it's worth the effort. Up to you whether you want to experiment with that or just ship it.

docs/content/any/project/changelogs/NEXT.md
```diff
@@ -48,5 +48,6 @@ Link to [relevant documentation section]().
 #### Other bugs & improvements
 
 - The debugger can now break on rule matches.
+- Reserved words can be used as field and method names in dictionaries and objects.
```
gj, line 51: Maybe give some examples of what reserved words are — I don't think we use that term anywhere else in the docs. Maybe "Reserved Polar keywords" would be more descriptive?

gj (approved):

docs/content/any/project/changelogs/NEXT.md
gw000: 👍  done

polar-core/src/polar.lalrpop
gw000: Good call, I'll revisit this later, not familiar enough with lalrpop to know how to do it right now.

PR #1099, opened 2021-08-24
data filtering docs (ruby)

gw000:


docs/content/any/guides/data_filtering/_index.md
```diff
@@ -178,10 +282,41 @@ def get_repos(constraints):
...
+          end
+        )
+      end
+
+      const_set(:FETCHER, lambda do |cons|
+        cons.reduce(self) { |q, con| kinds[con.kind][q, con] }
+      end)
+    end
+  end
+end
```
saolsen, line 309: Hmm, I don't really write much ruby so I dunno if this is idiomatic and obvious to ruby people but for me it was maybe a little too "clever".

Could we maybe just have it be one function and put the logic into the block of the reduce instead of a lambda for each kind (especially since they do the same thing) and then just set FETCHER to that function?

Also why do `Eq` and `In` do the same thing, is that an activerecord thing where a list automatically means `In`?

saolsen (approved):

docs/content/any/guides/data_filtering/_index.md
gw000: ok no problem & yes it's an active record thing

docs/content/any/guides/data_filtering/_index.md
```diff
@@ -290,19 +290,18 @@ def get_repos(constraints):
...
+          raise "Unsupported constraint kind: #{con.kind}" unless %w[Eq In].include? con.kind
+
+          q.where(
+            if con.field.nil?
+              { primary_key => con.value.send(primary_key) }
+            else
+              { con.field => con.value }
+            end
+          )
+        end
```
saolsen, line 304: beautiful

PR #1101, opened 2021-08-24
unify variables on rhs of match expressions

gw000:
this seems to fix the bug and might even be correct!


gj:
> this seems to fix the bug and might even be correct!

I love the optimism

gj (approved):

PR #1104, opened 2021-08-25
Data Filtering beta docs nits

gj:
Came across some nits while reading the new Data Filtering docs to respond to a community question.

Also spent a minute trying to figure out why a `<p>` tag is inserted before the Slack link but gave up:

image


gkaemmer:
Adding one commit to this, hold up

gkaemmer (approved):
I fixed that paragraph thing -- @saolsen and I discovered yesterday that hugo trips over nested shortcodes using the `{{%` syntax -- the markdown ends up being processed twice.

PR #1105, opened 2021-08-25
data filtering: support not equals operator

gw000:
- adds support for `!=` in data filtering
- "fixes" a bug ticket about loading multiple empty files, which will soon be obsolete anyway, sorry 😔 
- updates ruby to use the new data filtering API
- filters known values in rust, instead of the host

PR checklist:

- [x] Added changelog entry.


polar-core/src/data_filtering.rs
```diff
@@ -167,35 +165,14 @@ impl From<VarInfo> for Vars {
...
         let fields = info.field_relationships;
-        let triples = {
-            let mut assign_id = |item: &Symbol| get_var_id(&mut variables, item.clone(), &counter);
-            fields
-                .iter()
-                .map(|(p, f, c)| (assign_id(p), f, assign_id(c)))
-                .collect::<Vec<_>>()
-        };
-
-        // if x.y=a and x.y=b then a=b, so add a cycle.
```
gw000, line 178: we don't need to do this anymore because `dot_var` avoids creating redundant variables

polar-core/src/vm.rs
```diff
@@ -1418,16 +1412,18 @@ impl PolarVirtualMachine {
...
-                    self.push_goal(Goal::Query { term: val })?;
-                } else {
-                    // variable was unbound
-                    // apply a constraint to variable that it must be truthy
-                    self.push_goal(Goal::Unify {
-                        left: term.clone(),
-                        right: term!(true),
-                    })?;
-                }
+                self.push_goal(
```
dhatch, line 1415: I'm not really sure this is easier to read if that was the rationale behind this change.

polar-core/src/vm.rs
```diff
@@ -3123,15 +3120,15 @@ mod tests {
...
+        vm.push_goal(query!(op!(Unify, 1.into(), 1.into())))
             .unwrap();
 
         assert_query_events!(vm, [
             QueryEvent::Result{hashmap!{}},
             QueryEvent::Done { result: true }
         ]);
 
-        let q = op!(Unify, term!(1), term!(2));
+        let q = op!(Unify, 1.into(), 2.into());
```
dhatch, line 3131: imo `term!(..)` here clearly shows what type is being constructed, while `1.into()` is unclear.

languages/python/oso/polar/polar.py
```diff
@@ -312,61 +312,24 @@ def authorized_query(self, actor, action, cls):
...
-            new_result_sets = []
-            for c in complete:
-                constraints = []
-                typ = self.host.types[class_name]
-                if not typ.build_query:
-                    # Maybe a way around this if we make builtins for our builtin
-                    # classes but it'd be a hack just for this case and not worth it right now.
-                    assert False, "Can only filter registered classes"
-
-                for k, t in typ.fields.items():
```
dhatch, line 352: This got pushed into the core?

languages/python/oso/tests/test_data_filtering.py
```diff
@@ -293,6 +319,22 @@ def test_no_relationships(oso, t):
     assert len(results) == 3
 
 
+def test_neq(oso, t):
+    policy = """
+    allow("steve", "get", foo: Foo) if foo.bar.id != "hello";
```
dhatch, line 324: What about policies with multiple ne `foo.bar.id != 1 and foo.bar.id != 2`? What about `foo.bar.id = "hello" and foo.bar.id != "goodbye"` or `foo.bar.id = x and foo.bar.id != y`?

Seems like there's more cases we should test here.

languages/python/oso/tests/test_data_filtering.py
```diff
@@ -293,6 +319,22 @@ def test_no_relationships(oso, t):
     assert len(results) == 3
 
 
+def test_neq(oso, t):
```
dhatch, line 322: What's the approach on testing been?

Are we writing tests that confirm that the core produces the right filter plan from a policy in the core? Or are those all in Python?

If they're in Python how are we keeping track of which tests need to be ported to all languages and which only need to be in Python / one language?

dhatch:
Other thought... what happens for something like this:

`allow(actor, _, resource) if actor != resource;`

PR #1107, opened 2021-08-25
Enforcer API in ruby

gkaemmer:
PR checklist:

- [ ] Added changelog entry.


languages/ruby/lib/oso.rb
```diff
@@ -2,11 +2,14 @@
...
 require 'oso/oso'
 require 'oso/polar'
+require 'oso/errors'
 require 'oso/version'
 
 # Top-level namespace for oso authorization library.
 module Oso
   def self.new
+    # TODO: deprecate this method. Instead, users should use:
+    # > oso = Oso::Oso.new
```
gj, line 12: Why?

languages/ruby/lib/oso.rb
```diff
@@ -2,11 +2,14 @@
 
 require 'oso/oso'
 require 'oso/polar'
+require 'oso/errors'
 require 'oso/version'
 
 # Top-level namespace for oso authorization library.
```
gj, line 8:
```suggestion
# Top-level namespace for Oso authorization library.
```

gkaemmer: Heh good thing I'm touching all these files so we can find all the lowercase `oso`s -- will fix

We could maybe do a `s/oso/Oso/` and `s/polar/Polar` project-wide in one fell swoop soon too

languages/ruby/lib/oso.rb
gkaemmer: Whoops, this TODO made more sense when we were changing to using `Enforcer`

languages/ruby/lib/oso/enforcer.rb
```diff
@@ -0,0 +1,175 @@
...
+    # @param actor The actor for whom to collect allowed fields.
+    # @param action The action being taken on the field.
+    # @param resource The resource being accessed.
+    # @param allow_wildcard Flag to determine behavior if the policy \
+    #   includes a wildcard field. E.g., a rule allowing any field: \
+    #   +allow_field(_actor, _action, _resource, _field)+. If +true+, the \
+    #   method will return +["*"]+, if +false+, the method will raise an \
+    #   exception.
+    # @returns A list of the unique allowed fields.
+    def authorized_fields(actor, action, resource, allow_wildcard: false) # rubocop:disable Metrics/MethodLength
```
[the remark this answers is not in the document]
gkaemmer: I gave this a shot -- in order to make the error message match, we'd need to pass in a few extra parameters (`variable_name`, `method_name`). To me, it doesn't really clean up the code that much -- I could see these methods eventually diverging in subtle ways, and it's probably easier to introduce this abstraction later than it is to unwind it.

I could be wrong, but for now I think I'd prefer to leave as-is.

languages/ruby/lib/oso.rb
gj: Yeah we should do that, but unfortunately there are probably plenty of perfectly fine occurrences of lowercase `oso` in the project, like `const oso = new Oso();`-style code snippets

languages/ruby/lib/oso/oso.rb
```diff
@@ -19,5 +34,143 @@ def initialize
...
+    #
+    # Uses +allow_field+ rules in the policy to find all allowed fields.
+    #
+    # @param actor The actor for whom to collect allowed fields.
+    # @param action The action being taken on the field.
+    # @param resource The resource being accessed.
+    # @param allow_wildcard Flag to determine behavior if the policy \
+    #   includes a wildcard field. E.g., a rule allowing any field: \
+    #   +allow_field(_actor, _action, _resource, _field)+. If +true+, the \
+    #   method will return +["*"]+, if +false+, the method will raise an \
```
gj, line 153:
```suggestion
    #   method will return +Set["*"]+, if +false+, the method will raise an \
```

languages/ruby/lib/oso/oso.rb
```diff
@@ -19,5 +34,143 @@ def initialize
...
+    #   exception.
+    # @returns A set of the unique allowed fields.
+    def authorized_fields(actor, action, resource, allow_wildcard: false) # rubocop:disable Metrics/MethodLength
+      results = query_rule('allow_field', actor, action, resource, Polar::Variable.new('field'))
+      fields = Set.new
+      results.each do |result|
+        field = result['field']
+        if field.is_a?(Polar::Variable)
+          return ['*'] if allow_wildcard
```
gj, line 162:
```suggestion
          return Set['*'] if allow_wildcard
```

languages/ruby/lib/oso/oso.rb
```diff
@@ -19,5 +34,143 @@ def initialize
     def allowed?(actor:, action:, resource:)
       !query_rule('allow', actor, action, resource).first.nil?
```
gj, line 35: Update this to call `query_rule_once`?

languages/ruby/lib/oso/polar/polar.rb
```diff
@@ -227,6 +227,16 @@ def query_rule(name, *args, accept_expression: false, bindings: {})
         query(Predicate.new(name, args: args), host: host, bindings: bindings)
       end
 
+      # Query for a rule, returning true if it has any matches.
```
gj, line 230:
```suggestion
      # Query for a rule, returning true if it has any results.
```

languages/ruby/lib/oso/polar/polar.rb
```diff
@@ -227,6 +227,16 @@ def query_rule(name, *args, accept_expression: false, bindings: {})
...
+      #
+      # @param name [String]
+      # @param args [Array<Object>]
+      # @return [Boolean] of resulting bindings
```
gj, line 234:
```suggestion
      # @return [Boolean] indicating whether the query found at least one result.
```

languages/ruby/lib/oso/polar/query.rb
```diff
@@ -152,6 +152,8 @@ def handle_make_external(data) # rubocop:disable Metrics/AbcSize
         host.make_instance(cls_name, args: args, kwargs: kwargs, id: id)
       end
 
+      public
+
```
gj, line 156: Why does `Query#each` have to be public? If it does have to, I'd prefer to move it up with the other public method (`Query#initialize`) instead of having this random public block midway through the private section of the class

gj (changes requested):

languages/ruby/lib/oso/errors.rb
```diff
@@ -0,0 +1,44 @@
...
+  # methods when the action is not allowed.
+  #
+  # Most of the time, your app should handle this error by returning a 403 HTTP
+  # error to the client.
+  class ForbiddenError < AuthorizationError
+    def initialize
+      super(
+        'Oso ForbiddenError -- The requested action was not allowed for the ' \
+        'given resource. You should handle this error by returning a 403 error ' \
+        'to the client.'
```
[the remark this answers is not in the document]
gkaemmer: Lol it's apparently what rubocop wants

languages/ruby/lib/oso/oso.rb
```diff
@@ -19,5 +34,143 @@ def initialize
...
+    # @raise [Oso::ForbiddenError] Raised if the actor does not have permission
+    #   to perform this action on this resource, but _does_ have +"read"+
+    #   permission on the resource.
+    # @raise [Oso::NotFoundError] Raised if the actor does not have permission
+    #   to perform this action on this resource and additionally does not have
+    #   permission to +"read"+ the resource.
+    def authorize(actor, action, resource, check_read: true)
+      return if query_rule_once('allow', actor, action, resource)
+
+      if check_read && action == @read_action || !query_rule_once('allow', actor, @read_action, resource)
```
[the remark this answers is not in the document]
gkaemmer: Nice catch -- just added a test for this

languages/ruby/lib/oso/polar/query.rb
gkaemmer: It's kinda weird that you can't enumerate results using `.each` given that `Query` is an enumerable (the `.each` happens in `authorized_{actions,fields}`). We can call `.to_a.each`, but that feels weird.

I'll move this up to the public section -- will be big LOC change.

gkaemmer: I actually feel like `.each` should be public but there should be a kind of internal private `.run` method that's called by `.each`.

Just made that change -- what do you think?

languages/ruby/lib/oso/oso.rb
```diff
@@ -19,5 +34,143 @@ def initialize
...
+    #   can read the resource. Default is +true+.
+    #
+    # @raise [Oso::ForbiddenError] Raised if the actor does not have permission
+    #   to perform this action on this resource, but _does_ have +"read"+
+    #   permission on the resource.
+    # @raise [Oso::NotFoundError] Raised if the actor does not have permission
+    #   to perform this action on this resource and additionally does not have
+    #   permission to +"read"+ the resource.
+    def authorize(actor, action, resource, check_read: true)
+      return if query_rule_once('allow', actor, action, resource)
```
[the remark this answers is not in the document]
gkaemmer: I don't know about this. I suppose there is an argument to be made, but I'd rather not impose a dependency on that particular public API.

Also note that its actually more verbose because you have to do `allowed?(actor: actor, action: action, resource: resource)`

languages/ruby/lib/oso/errors.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+module Oso
+  class Error < StandardError
+  end
+
+  class AuthorizationError < StandardError
+  end
```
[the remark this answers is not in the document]
gkaemmer: I think it's pretty standard to extend `StandardError` instead of `RuntimeError`, but I don't have a strong opinion so I'll go with `RuntimeError` for consistency

languages/ruby/lib/oso/polar/query.rb
```diff
@@ -18,6 +18,16 @@ def initialize(ffi_query, host:, bindings: {})
         bindings.each { |k, v| ffi_query.bind k, host.to_polar(v) }
       end
 
+      # Create a generator that can be polled to advance the query loop. Yields
```
gj, line 21:
```suggestion
      # Create an enumerator that can be polled to advance the query loop. Yields
```

gj (approved):
Great work!

languages/ruby/lib/oso/polar/query.rb
gj: Don't feel strongly about it; :shipit:

Issue #824, opened 2021-04-05
[Rust] REPL logging needed to see tracing


Anniepoo:
The oso REPL  is documented as needing `POLAR_LOG=1` to get explanations

OBSERVED:
You actually need `POLAR_LOG=now` (which prints messages to stderr rather than going through the language binder, is intended as an internal tool for dev work on Polar).
To use `POLAR_LOG=1` you need to turn on `RUST_LOG=1` as well.
EXPECTED:
Agreement between the docs and the actual behavior. `POLAR_LOG=1` should just work.

gkaemmer:
A fix for this will go live in our next release: #1115 -- rust will use stdout to log traces when `POLAR_LOG=1` is set, making it consistent with our other host libraries.

PR #1111, opened 2021-08-27
Explicitly call out that rust requires 'POLAR_LOG=now'

gkaemmer:
There's a weird bug that's easy to get snagged on if you're using the rust library: #824 

As a temporary fix, this updates the rust docs to explicitly recommend that you use `POLAR_LOG=now`, which skips all of the messaging queue stuff.

At a quick glance I'm not sure why messages aren't coming through to stdout in rust. There appears to be code to that effect, but something is just not working with it.

### DIFF
--- a/docs/content/any/reference/tooling/tracing.md
+++ b/docs/content/any/reference/tooling/tracing.md
@@ -11,7 +11,14 @@ Polar tracing shows logs of how a query is evaluated.
 
 ## Enabling Tracing
 
+
+{{% ifLang not="rust" %}}
 Debug mode is enabled by setting an environment variable: `POLAR_LOG=1`.
+{{% /ifLang %}}
+
+{{% ifLang "rust" %}}
+Debug mode is enabled by setting an environment variable: `POLAR_LOG=now`.
+{{% /ifLang %}}
 
 It will print out `[debug]` messages during query evaluation that show how the query is executed.
 Notable things include:


gw000:
Will this affect the docs right away or would it be for next release?

gkaemmer:
> Will this affect the docs right away or would it be for next release?

Probably would go ahead and deploy this docs change immediately (it's pretty minor)

dhatch:
How much investigation have you done @gkaemmer on `POLAR_LOG=1` not working? Do we have a hypothesis why that's an issue?

`POLAR_LOG=now` should only be needed for internal development, so if there's a simple fix to get `POLAR_LOG=1` working in Rust I think we should just do it.

gkaemmer:
> How much investigation have you done @gkaemmer on `POLAR_LOG=1` not working? Do we have a hypothesis why that's an issue?
> 
> `POLAR_LOG=now` should only be needed for internal development, so if there's a simple fix to get `POLAR_LOG=1` working in Rust I think we should just do it.

Did a bit more digging on this on friday. In the rust library we use `tracing::debug!(...)` to print out messages instead of printing them directly to stdout like the other host libraries do. This means that they only show up if you have a `tracing` subscriber that's set up at the `debug` level.

I did a quick test, and subscribing at the `debug` level causes other events (from other libraries) to get emitted. Notably, `rustyline` emits debug events on every keystroke, which is a bit of a non-starter for the repl:

Screen Shot 2021-08-30 at 12 28 43 PM

At the moment I don't see why we don't print messages directly to stdout 🤔  -- do you think we could just do that?

gkaemmer:
Closing in favor of #1115 1115

PR #1114, opened 2021-08-30
Actor/Resource type registration

gj:
Alternative to #1084 that registers types as entities in Polar instead of in the host language libraries.

PR checklist:

- [ ] Added changelog entry.


leina05 (commented):
This all looks good, but I think it's missing the implementation for rule prototype matching so that rule prototypes with `Actor` and `Resource` are matched by classes that are registered as those entity types (https://github.com/osohq/oso/blob/e8f6017ea48eb21a2121939987d498b99367459d/polar-core/src/kb.rs#L199-L201). Let me know if you want to talk through that part/want to pair.

polar-core/src/sugar.rs
```diff
@@ -674,7 +729,7 @@ mod tests {
             .unwrap();
         assert_eq!(
             rewritten_role_role.to_polar(),
-            r#"has_role(actor, "reader", repo_instance: repo{}) if has_relation(org_instance, "parent", repo_instance) and has_role(actor, "member", org_instance);"#
+            r#"has_role(actor: Actor{}, "reader", repo_instance: repo{}) if has_relation(org_instance, "parent", repo_instance) and has_role(actor, "member", org_instance);"#
```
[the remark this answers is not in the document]
gj: It's how `to_polar()` formats instance patterns

polar-core/src/terms.rs
```diff
@@ -451,6 +451,16 @@ impl Term {
             None
         }
     }
+
+    // TODO(gj): what happens if the user registers a class as `Actor` or `Resource`?
```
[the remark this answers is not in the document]
gj: Agree that it's probably an uncommon case; I think the three options are:

1. Leave them as `Actor` / `Resource` and throw an error if folks try to register constants with those names.
2. Namespace as `Oso::Actor` / `Oso::Resource` or w/e to reduce the likelihood of collisions but otherwise same situation as (1).
3. Wait to name the actor/resource "unions" until after all application classes have been registered, then decide on a name for each that won't conflict with any application class names.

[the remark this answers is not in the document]
leina05: I think 1 is the most straight-forward.

polar-core/src/sugar.rs
```diff
@@ -219,8 +240,15 @@ impl Declaration {
...
+#[derive(Clone, Debug, PartialEq)]
+pub enum EntityType {
+    Actor,
+    Resource,
+}
+
 #[derive(Clone, Debug, PartialEq)]
 pub struct Namespace {
+    pub entity_type: EntityType,
     pub resource: Term,
```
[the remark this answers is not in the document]
gj: Renamed `entity_type` to `block_type` for now since the feature is just typed resource blocks. We'll be able to do away with `block_type`/`BlockType` once we have true union types.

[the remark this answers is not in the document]
gj: (also renamed `Namespace` -> `ResourceBlock`, so the `block_type` & `resource` fields make a bit more sense now)

polar-core/src/kb.rs
```diff
@@ -187,6 +187,73 @@ impl KnowledgeBase {
...
+
+                        if !success {
+                            return Ok(RuleParamMatch::False(format!("Rule specializer {} on parameter {} must be a member of prototype specializer {}", rule_instance.tag,index, prototype_instance.tag)));
+                        }
+                    }
+                    if !self.param_fields_match(&prototype_instance.fields, &rule_instance.fields) {
+                        RuleParamMatch::False(format!("Rule specializer {} on parameter {} did not match prototype specializer {} because the specializer fields did not match.", rule_instance.to_polar(), index, prototype_instance.to_polar()))
+                    } else {
+                        RuleParamMatch::True
+                    }
```
leina05, line 256: It seems like this could be broken into a `check_mro` function because the same logic is used in the next branch.

leina05 (changes requested):
Just one comment, then should be good.

polar-core/src/kb.rs
gj: 302b9ede

polar-core/src/kb.rs
```diff
@@ -168,6 +168,38 @@ impl KnowledgeBase {
             .all(|v| v);
     }
 
+    fn check_rule_instance_is_subclass_of_prototype_instance(
```
leina05, line 171:
```suggestion
    /// Use MRO lists passed in from host library to determine if one `InstanceLiteral` pattern is a subclass of another `InstanceLiteral` pattern. This function is used for Rule Type validation.
    fn check_rule_instance_is_subclass_of_prototype_instance(
```

polar-core/src/kb.rs
```diff
@@ -168,6 +168,38 @@ impl KnowledgeBase {
...
+        &self,
+        rule_instance: &InstanceLiteral,
+        prototype_instance: &InstanceLiteral,
+        index: usize,
+    ) -> PolarResult<RuleParamMatch> {
+        if let Some(Value::ExternalInstance(ExternalInstance { instance_id, .. })) = self
```
leina05, line 177:
```suggestion
        // Get the unique ID of the prototype instance pattern class
        if let Some(Value::ExternalInstance(ExternalInstance { instance_id, .. })) = self
```

leina05 (approved):
A couple suggestions for comments but then :shipit:

polar-core/src/kb.rs
gj: Added in #1138

polar-core/src/kb.rs
gj: Added in #1138

PR #1116, opened 2021-08-30
data filtering: new api for js

gw000:


languages/js/src/Host.ts
```diff
@@ -32,19 +31,40 @@ import {
 } from './types';
 import { map } from '../test/helpers';
 
+export class UserType {
+  name: string;
+  class: any;
+  fields: Map<string, any>;
+  buildQuery: undefined | ((_: any) => any);
```
gkaemmer, line 38: Do these make sense to go on each type? I feel like these should be global and should take a `class` as parameter where necessary.

When it was just one `fetcher` I was conflicted about this particular thing, but with the addition of `combineQuery`, type-specific fetchers don't feel right to me.

gw000: Hmm that's a good idea, I think you're probably right

gw000: I think what I'm going to do is have optional global defaults for them which will be used if one isn't given on the type. That way users can still be specific if they need to without having to write too much extra boilerplate. In practice every case I've seen with the new interface has different `buildQuery` implementations per type (which sometimes depend on values other than just the class) but use the same `combineQuery` and `execQuery` functions, so those wouldn't have to be passed to `registerClass` anymore.

gkaemmer:
Kinda confused what's going on here git-wise -- this PR seems like it includes lots of changes unrelated to js data filtering

PR #1102, opened 2021-08-24
Enforcer API for Node

gkaemmer:
PR checklist:

- [ ] Added changelog entry.

PR #1119, opened 2021-08-31
Remove Policy/Enforcer split in Python

gkaemmer:
PR checklist:

- [ ] Added changelog entry.


languages/python/oso/oso/oso.py
```diff
@@ -80,12 +157,75 @@ def get_allowed_actions(self, actor, resource, allow_wildcard=False) -> list:
...
+
     def _print_polar_log_message(self):
         if os.environ.get("POLAR_LOG", None):
             print(
                 "Polar tracing enabled. Get help with "
                 + "traces from our engineering team: https://help.osohq.com/trace"
             )
 
+    def _default_get_error(self, is_not_found):
+        return NotFoundError() if is_not_found else ForbiddenError()
```
gj, line 228: Same feedback as https://github.com/osohq/oso/pull/1102#discussion_r696307012 re: using a boolean to flip between two error variants

gkaemmer: Yep, making that change now to be consistent 👍

languages/python/oso/oso/oso.py
```diff
@@ -45,17 +73,76 @@ def is_allowed(self, actor, action, resource) -> bool:
...
 
         :param resource: The resource being accessed
 
         :param allow_wildcard: Flag to determine behavior if the policy \
-        includes a wildcard action. E.g., a rule allowing any action: \
-        ``allow(_actor, _action, _resource)``. If ``True``, the method will \
-        return ``["*"]``, if ``False``, the method will raise an exception.
+            includes a wildcard action. E.g., a rule allowing any action: \
+            ``allow(_actor, _action, _resource)``. If ``True``, the method will \
+            return ``["*"]``, if ``False``, the method will raise an exception.
```
gj, line 145: Maybe add an explanation of why the exception is raised? Something like "to avoid inadvertently indicating that all actions are permitted"... though now that I type that out it's really just alerting users to an issue with their policy, isn't it?

ETA: I see the explanation in the error message — maybe just align this description with that.

languages/python/oso/oso/oso.py
```diff
@@ -80,6 +167,66 @@ def get_allowed_actions(self, actor, resource, allow_wildcard=False) -> list:
...
+    def authorized_fields(
+        self, actor, action, resource, allow_wildcard=False
+    ) -> List[Any]:
+        """Determine the fields of ``resource`` on which ``actor`` is allowed to
+        perform  ``action``.
+
+        Uses ``allow_field`` rules in the policy to find all allowed fields.
+
+        :param actor: The actor for whom to collect allowed fields.
+        :param action: The action being taken on the field.
```
gj, line 196:
```suggestion
        :param action: The action being taken on the fields.
```

languages/python/oso/oso/oso.py
```diff
@@ -80,6 +167,66 @@ def get_allowed_actions(self, actor, resource, allow_wildcard=False) -> list:
...
+                        "unconstrained" field that could represent any
+                        field, but allow_wildcard was set to False. To fix,
+                        set allow_wildcard to True and compare with the "*"
+                        string."""
+                    )
+                else:
+                    return ["*"]
+            fields.add(field)
+
+        return list(fields)
```
gj, line 228: Same comment as https://github.com/osohq/oso/pull/1102#discussion_r696953414

languages/python/oso/oso/oso.py
```diff
@@ -80,6 +167,66 @@ def get_allowed_actions(self, actor, resource, allow_wildcard=False) -> list:
 
         return list(actions)
```
gj, line 168: Same comment about returning a Set instead of turning it back into a List

languages/python/oso/oso/oso.py
```diff
@@ -45,17 +73,76 @@ def is_allowed(self, actor, action, resource) -> bool:
     def get_allowed_actions(self, actor, resource, allow_wildcard=False) -> list:
```
gj, line 73: Same comment about returning a Set instead of turning it back into a List

gj: Also as a side note: wherever you have `list` in a type position in Python I think you probably want to use `List[ElementType]` so you can provide a type for the element too

gj (changes requested):
Only blocking thing is returning sets for some of the enforcement methods instead of deduplicating using a set and then turning the set back into a list

Looking good!

languages/python/oso/oso/oso.py
gkaemmer: Hmm this will make for a breaking change -- perhaps I'll have `get_allowed_actions` return `list(self.authorized_actions(...))`

gj: Up to you — we're making a bunch of other breaking changes in 0.20.0 so it wouldn't be terrible to add one more (small) one onto the pile.

languages/python/oso/oso/oso.py
gkaemmer: Left this one returning a list -- I don't want to break existing behavior and we're deprecating this method anyway.

Changed the new methods to return `set`s 👍

languages/python/oso/oso/oso.py
```diff
@@ -162,10 +163,10 @@ def authorized_actions(self, actor, resource, allow_wildcard=False) -> List[Any]
                         string."""
                     )
                 else:
-                    return ["*"]
+                    return set(["*"])
```
gj, line 166: pretty sure you can spell this:
```suggestion
                    return {"*"}
```

languages/python/oso/oso/exceptions.py
```diff
@@ -4,9 +4,9 @@ class AuthorizationError(Exception):
 
 class NotFoundError(AuthorizationError):
     """
-    Thrown by the ``authorize`` method of an ``Enforcer``. This error indicates
-    that the actor is not only not allowed to perform the given action, but also
-    is not allowed to ``"read"`` the given resource.
+    Thrown by the ``authorize`` method of an ``Oso`` instance. This error
+    indicates that the actor is not only not allowed to perform the given
+    action, but also is not allowed to ``"read"`` the given resource.
```
gj, line 9: nittiest nit, to account for cases where someone's specified a 'read' action that isn't `"read"`:

```suggestion
    action, but also is not allowed to read the given resource.
```

languages/python/oso/oso/exceptions.py
```diff
@@ -4,9 +4,9 @@ class AuthorizationError(Exception):
```
gj, line 4: Three thoughts:

1. Seems kinda weird to have `OsoError` defined in `polar/exceptions.py` instead of `oso/exceptions.py`. Maybe it has to be that way for module dependency resolution purposes?
2. Should `AuthorizationError` inherit from `OsoError` instead of `Exception`?
3. Should the three errors currently in this file be called `OsoAuthorizationError`, `OsoNotFoundError`, and `OsoForbiddenError`, or is that obvious/redundant info?

languages/python/oso/oso/oso.py
```diff
@@ -16,11 +19,36 @@ class Oso(Polar):
...
-        """Create an oso object."""
+    def __init__(
+        self,
+        *,
+        forbidden_error=ForbiddenError,
+        not_found_error=NotFoundError,
+        read_action="read"
+    ):
+        """
+        Create an oso object.
```
gj, line 30:
```suggestion
        Create an Oso object.
```

languages/python/oso/oso/oso.py
```diff
@@ -42,24 +70,84 @@ def is_allowed(self, actor, action, resource) -> bool:
...
+        if self.query_rule_once("allow", actor, action, resource):
+            return
+
+        is_not_found = False
+        if check_read:
+            if action == self.read_action:
+                is_not_found = True
+            elif not self.query_rule_once("allow", actor, self.read_action, resource):
+                is_not_found = True
+        raise self.not_found_error() if is_not_found else self.forbidden_error()
```
gj, line 112: nit: I think this is cleaner than assigning to a boolean, but I don't mind if you prefer it as-is:

```suggestion
        if check_read and (
            action == self.read_action
            or not self.query_rule_once("allow", actor, self.read_action, resource)
        ):
            raise self.not_found_error()
        raise self.forbidden_error()
```

languages/python/oso/tests/test_enforcer.py
```diff
@@ -1,78 +1,77 @@
 """Tests the Enforcement API"""
```
gj, line 1: nit: Worth renaming this file from `test_enforcer.py` -> `test_enforcement.py`?

languages/python/oso/oso/exceptions.py
```diff
@@ -4,9 +4,9 @@ class AuthorizationError(Exception):
...
     """
-    Thrown by the ``authorize`` method of an ``Enforcer``. This error indicates
-    that the actor is not only not allowed to perform the given action, but also
-    is not allowed to ``"read"`` the given resource.
+    Thrown by the ``authorize`` method of an ``Oso`` instance. This error
+    indicates that the actor is not only not allowed to perform the given
+    action, but also is not allowed to ``"read"`` the given resource.
 
     Most of the time, your app should handle this error by returning a 404 HTTP
     error to the client.
```
gj, line 12: Couple references to `Enforcer` below this line (GitHub doesn't allow me to leave this comment on the exact line)

gj (changes requested):
buncha nits

languages/python/oso/oso/exceptions.py
gkaemmer: Might keep this as-is. I kind of like defaulting to `"read"` in the docs, etc. Ideally, the custom read action is something most people shouldn't have to think about doing.

It is sort of analogous to how we've been documenting the errors: I typically write that `authorize` raises `NotFoundError` or `ForbiddenError` -- it's almost too complicated to explain that it raises "the not found error configured for your Oso instance".

languages/python/oso/oso/exceptions.py
gkaemmer: 1. Yeah, I think it makes sense to put `OsoError` into `oso/exceptions.py`. We do have to import it into `polar/exceptions.py` but I think that's okay
2. Yeah
3. Ehh I don't think so -- none of our other errors are prefixed with `Oso-` and IMO it is a bit too verbose. Not many libraries I've seen prefix their classes with the library's name. Sometimes I've seen it recommended to import the namespace directly like this:
```
import oso
try:
  g.oso.authorize(...)
except oso.ForbiddenError:
  # forbidden :|
```

languages/python/oso/oso/oso.py
gj: Good thinking 👍

languages/python/oso/oso/exceptions.py
gj: Cool, that makes sense

gj (approved):
Fix the import error & :shipit: 

Great work!

languages/python/oso/oso/exceptions.py
gkaemmer: Damn yeah unless we introduce a third `exceptions` module inside of the `languages/python/oso` folder, we can't put `OsoError` into `oso/exceptions.py` :(

PR #1120, opened 2021-08-31
Add `load_files` and deprecate `load_file`

leina05:
PR checklist:

- [ ] (maybe) improve the way we do warnings in Python so that it includes location of source of warning
- [ ] update documentation
- [ ] Added changelog entry.


polar-core/src/polar.rs
```diff
@@ -238,6 +234,15 @@ impl Polar {
...
     }
 
+    /// Validate rules in the knowledge base once all files have been loaded
+    pub fn validate_rules(&self) -> PolarResult<()> {
+        let mut kb = self.kb.write().unwrap();
+        // Rewrite namespace implications _before_ validating rule prototypes.
+        kb.rewrite_implications()?;
+        // check rules are valid against rule prototypes
+        kb.validate_rules()
+    }
```
gj, line 244: Why did you move this into a separate FFI call? What was wrong with calling both methods at the end of `load_source()`?

gj (changes requested):

polar-core/src/polar.rs
leina05: Since `load_source` is called once per file from the host libraries, if the validation were in there it would get run on every file instead of after all files are loaded. The way I implemented this, the host library is responsible for loading all the files and then calling `validate_rules()`. The alternative would have been adding a way to load multiple files in the core, which I thought would take longer to implement and be a bigger core change.

gj:
superseded by #1138

PR #1124, opened 2021-09-02
Await returning from the generator.

samscott89:
In order to complete the generator and trigger the `finally` block which frees the ffi query object.


gj (approved):
Good catch — the `return()` method on an async generator returns a Promise:

```ts
return(value: TReturn | PromiseLike<TReturn>): Promise<IteratorResult<T, TReturn>>;
```

PR #1125, opened 2021-09-02
use ActiveRecord 6.1.4

gw000:
for ruby 3+ compatibility


gj:
@gw000 can we try using activerecord 5.2.6 in our tests instead?

gw000:
5.2.6 has the same incompatibility. Why that version?

gj:
> @gj 5.2.6 has the same incompatibility. Why that version?

Oh, weird, [RubyGems](https://rubygems.org/gems/activerecord/versions/5.2.6) lists it as compatible with Ruby 2.4.

To rephrase the original question: can we try using the latest version of activerecord that's compatible with Ruby 2.4 in our tests instead?

PR #1126, opened 2021-09-03
Node: export Relationship class from top-level package

gkaemmer:


gj (approved):
Believe we're also going to rename it `Relation` before the next release, but that can come as a separate PR

PR #1130, opened 2021-09-03
Implement mro registration in JS.

dhatch:
There are some todos in this code that could use some insight from someone better at JS than me.

PR checklist:

- [ ] Added changelog entry.

### DIFF
--- a/languages/js/src/Host.ts
+++ b/languages/js/src/Host.ts
@@ -89,10 +89,9 @@ export class Host {
     return cls;
   }
 
-  isRegisteredType(cls: Class): boolean {
-    return this.clsNames.has(cls);
-  }
-
+  /**
+   * Get user type for `cls`.
+   */
   getType(cls: Class):
     | {
         name: string;
@@ -100,6 +99,7 @@ export class Host {
         cls: Class;
       }
     | undefined {
+    // TODO: Update when adding user type.
     const name = this.clsNames.get(cls);
     if (name === undefined) return undefined;
 
--- a/languages/js/src/Host.ts
+++ b/languages/js/src/Host.ts
@@ -30,7 +30,6 @@ import {
   isPolarStr,
   isPolarVariable,
 } from './types';
-import { map } from '../test/helpers';
 
 type UserType = {
   name: string;
@@ -122,7 +121,7 @@ export class Host {
     // TODO: Use UserType object once this is brought in line with Python.
     for (const [name, id] of this.#classIds.entries()) {
       const cls = this.#classes.get(name);
-      if (typeof cls === 'undefined') {
+      if (cls === undefined) {
         throw new Error('cls should not be undefined.');
       }
 
--- a/languages/js/src/Host.ts
+++ b/languages/js/src/Host.ts
@@ -207,9 +207,9 @@ export class Host {
     // NOTE: not ideal that the MRO gets updated each time loadStr is
     // called, but since we are planning to move to only calling load once
     // with the include feature, I think it's okay for now.
-    for (const typ of this.#classes.values()) {
+    for (const typ of this.distinctUserTypes()) {
       // Get MRO for type.
-      const mro = ancestors(typ)
+      const mro = ancestors(typ.cls)
         .map(c => this.getType(c as Class)?.id)
         .filter(id => id !== undefined);
 
--- a/languages/js/src/Host.ts
+++ b/languages/js/src/Host.ts
@@ -93,9 +93,7 @@ export class Host {
     return this.clsNames.has(cls);
   }
 
-  getType(
-    cls: Class
-  ):
+  getType(cls: Class):
     | {
         name: string;
         id: number;
@@ -205,6 +203,26 @@ export class Host {
     }
     this.#instances.set(instanceId, instance);
     return instanceId;
+  }
+
+  /**
+   * Register the MROs of all classes that are registered on `this`.
+   */
+  registerMros() {
+    // Get MRO of all registered classes
+    // NOTE: not ideal that the MRO gets updated each time loadStr is
+    // called, but since we are planning to move to only calling load once
+    // with the include feature, I think it's okay for now.
+    for (const typ of this.distinctUserTypes()) {
+      // Get MRO for type.
+      const mro = ancestors(typ.cls)
+        // TODO: as conversion ok?
+        .map(c => this.getType(c as Class)?.id)
+        .filter(id => !!id);
+
+      // Register with core.
+      this.#ffiPolar.registerMro(typ.name, mro);
+    }
   }
 
   /**
--- a/languages/js/src/Host.ts
+++ b/languages/js/src/Host.ts
@@ -89,6 +89,54 @@ export class Host {
     return cls;
   }
 
+  isRegisteredType(cls: Class): boolean {
+    return this.clsNames.has(cls);
+  }
+
+  getType(
+    cls: Class
+  ):
+    | {
+        name: string;
+        id: number;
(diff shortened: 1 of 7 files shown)


languages/js/src/Host.ts
```diff
@@ -89,6 +89,52 @@ export class Host {
...
+   * Get user type for `cls`.
+   */
+  getType(cls: Class):
+    | {
+        name: string;
+        id: number;
+        cls: Class;
+      }
+    | undefined {
+    // TODO: Update when adding user type.
```
gj, line 102: might be covered by the TODO comment, but I'd prefer something like

```suggestion
  getType(cls: Class): UserType? {
```

languages/js/src/Host.ts
```diff
@@ -89,6 +89,52 @@ export class Host {
     return cls;
   }
 
+  /**
+   * Get user type for `cls`.
+   */
+  getType(cls: Class):
```
leina05, line 95: I don't really get why we needed to add this if we're not adding the full `UserTypes` thing yet, since it's not actually removing the host attributes (`classIds`, `classNames`, etc.).

languages/js/src/Host.ts
```diff
@@ -158,6 +205,26 @@ export class Host {
...
+   */
+  registerMros() {
+    // Get MRO of all registered classes
+    // NOTE: not ideal that the MRO gets updated each time loadStr is
+    // called, but since we are planning to move to only calling load once
+    // with the include feature, I think it's okay for now.
+    for (const typ of this.distinctUserTypes()) {
+      // Get MRO for type.
+      const mro = ancestors(typ.cls)
+        // TODO: as conversion ok?
```
leina05, line 219: Is this TODO still relevant?

languages/js/src/Host.ts
```diff
@@ -158,6 +205,26 @@ export class Host {
...
+        .map(c => this.getType(c as Class)?.id)
+        .filter(id => !!id);
```
leina05, line 221: Why is the `!!` necessary?

languages/js/src/Host.ts
```diff
@@ -343,7 +410,13 @@ export class Host {
         );
         return { value: { Dictionary: { fields } } };
       default:
-        const instance_id = this.cacheInstance(v);
+        // TODO (dhatch): Reliable check that this is a class?
```
leina05, line 413: Is this TODO still relevant?

languages/js/src/Host.ts
```diff
@@ -343,7 +410,13 @@ export class Host {
...
+        let instanceId = undefined;
+        if (v instanceof Function) {
```
leina05, line 415: This clearly works but curious why the check is for `Function`?

languages/python/oso/polar/host.py
```diff
@@ -86,7 +90,8 @@ def get_class(self, name):
 
     def distinct_user_types(self):
         return map(
-            lambda k: self.types[k], filter(lambda k: type(k) is str, self.types.keys())
+            lambda k: self.types[k],
+            filter(lambda k: isinstance(k, str), self.types.keys()),
         )
```
leina05, line 95: Thanks for cleaning this file up.

polar-core/src/polar.rs
```diff
@@ -296,6 +296,12 @@ impl Polar {
         self.kb.write().unwrap().constant(name, value)
     }
 
+    /// Register MRO for `name` with `mro`.
+    ///
+    /// Params:
+    ///
+    /// - `mro`: Should go from `name`, `name`'s next superclass, `name's furthest away superclass.
+    ///          `mro` is a list of class ids.
```
leina05, line 304: Thanks!

leina05 (approved):
LGTM, a few comments but nothing major.

languages/js/src/Host.ts
gj: not sure this ever matters (do we ever assign `id=0` for anything, or do we start at 1?) but this will filter out `id=0`. Maybe add an invariant comment somewhere (maybe on the `getType` definition?) that ID will always be a positive integer?

Oh and if we're sure we're never going to encounter `id=0`, you can remove the `!!` (`.filter(id => id)`) or use `.filter(Number)` if you prefer that

languages/js/src/Host.ts
gj: It seems like we only use this in a single place and then immediately dump everything except the returned `id` field. Should this be `getId` instead of `getType`?

languages/js/src/Host.ts
gj: Even if we aren't introducing `UserType` yet, it would be nice to package `{ name: string; id: number; cls: Class; }` in a `type` that we can use in the return type for `getType` and `distinctUserTypes`

languages/js/src/Host.ts
```diff
@@ -89,6 +89,52 @@ export class Host {
...
+      }
+
+      const entryWithClass = {
+        name: entry[0],
+        id: entry[1],
+        cls,
+      };
+
+      yield entryWithClass;
+    }
```
gj, line 135:
```suggestion
    for (const [name, id] of this.#classIds.entries()) {
      const cls = this.#classes.get(name);

      if (cls === undefined) throw new Error('cls should not be undefined.');

      yield { name, id, cls };
    }
```

languages/js/src/Host.ts
```diff
@@ -158,6 +205,26 @@ export class Host {
     return instanceId;
   }
 
+  /**
+   * Register the MROs of all classes that are registered on `this`.
```
gj, line 209:
```suggestion
   * Register the MROs of all registered classes.
```

languages/js/src/Host.ts
```diff
@@ -158,6 +205,26 @@ export class Host {
...
+    // with the include feature, I think it's okay for now.
+    for (const typ of this.distinctUserTypes()) {
+      // Get MRO for type.
+      const mro = ancestors(typ.cls)
+        // TODO: as conversion ok?
+        .map(c => this.getType(c as Class)?.id)
+        .filter(id => !!id);
+
+      // Register with core.
+      this.#ffiPolar.registerMro(typ.name, mro);
```
gj, line 224: Why are we doing one FFI call per class/mro instead of a single call that batches everything?

languages/js/src/Host.ts
dhatch: I added this so it would be easier to implement with the API Gwen is adding in #1116.

languages/js/src/Host.ts
dhatch: Yeah... this coerces `c` which is a Function into the `Class` type. This is because `ancestors` returns an iterable of `Function`s, not the `Class` type that we've created.

dhatch: Not sure this is ok, could use @gj 's input

languages/js/src/Host.ts
dhatch: Changed this to explicitly check that id is not undefined.

languages/js/src/Host.ts
dhatch: Yeah, it is. I think this is the best way to check for this

languages/js/src/Host.ts
dhatch: A class is a function in Javascript. There is no class type in the actual language, only one in TypeScript (that we've supplied).

languages/js/src/Host.ts
gj: we talked — it's cool for now

PR #1132, opened 2021-09-03
Small refactor for Python MRO registration

leina05:


languages/python/oso/polar/host.py
```diff
@@ -112,6 +114,13 @@ def cache_class(
         )
         return name
 
+    def register_mros(self):
+        """Register the MRO of each registered class to be used for rule prototype validation."""
```
gj, line 118: you used "prototype" here but changed "prototype" to "type" in a diff file

gj (approved):
fix or ship, up 2 u

languages/python/oso/polar/host.py
leina05: ah
yeah
darn

leina05: gonna do a whole new PR to change prototype to type so i'm gonna ignore this for now.

gj: sgtm

PR #1134, opened 2021-09-03
Add MRO registration to Java to support Rule Types

leina05:


languages/java/oso/src/main/java/com/osohq/oso/Host.java
```diff
@@ -15,6 +15,7 @@
 public class Host implements Cloneable {
   private Ffi.Polar ffiPolar;
   private Map<String, Class<?>> classes;
+  private Map<Class<?>, Long> class_ids;
```
dhatch, line 18: Java uses camelCase

languages/java/oso/src/main/java/com/osohq/oso/Host.java
```diff
@@ -250,7 +272,14 @@ public JSONObject toPolarTerm(Object value) throws Exceptions.OsoException {
       }
     } else {
       JSONObject attrs = new JSONObject();
-      attrs.put("instance_id", cacheInstance(value, null));
+      Long instance_id = null;
```
dhatch, line 275: java uses camelcase

polar-core/src/error.rs
```diff
@@ -389,18 +389,22 @@ impl fmt::Display for RuntimeError {
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub enum OperationalError {
-    Unimplemented(String),
+    Unimplemented {
```
dhatch, line 392: This could change the serialization format of these errors (I think). Does that mean we need to change deserialization code in how libraries?

dhatch (changes requested):

dhatch (approved):

PR #1135, opened 2021-09-03
Remove Polar Roles

gj:


languages/python/oso/polar/errors.py
```diff
@@ -92,10 +89,6 @@ def _operational_error(subkind, message, details):
         return OperationalError(message, details)
 
 
-def _roles_validation_error(message, details):
```
dhatch, line 95: Do we need to remove these errors from java?

languages/ruby/spec/oso/polar/polar_spec.rb
```diff
@@ -901,7 +901,6 @@ def self.test
       subject.register_class(RolesHelpers::Issue, name: 'Issue')
       subject.register_class(RolesHelpers::User, name: 'User')
       subject.load_file(roles_file)
-      subject.enable_roles
```
dhatch, line 904: How is the rest of this test still valid?

polar-core/src/polar.rs
```diff
@@ -5,7 +5,6 @@ use super::kb::*;
 use super::messages::*;
```
dhatch, line 5: Do you need to remove roles errors from the core also? `RolesValidationError`?

languages/ruby/spec/oso/polar/polar_spec.rb
gj: I updated the 1 giant Polar Roles test in each language to the new syntax

languages/python/oso/polar/errors.py
gj: I don't see them in Java

polar-core/src/polar.rs
gj: Done in c507c66

dhatch (approved):
I did it

PR #1144, opened 2021-09-07
Run JS roles tests

gw000:
Gotta `expect` it `toBe` something

### DIFF
--- a/languages/js/src/Polar.test.ts
+++ b/languages/js/src/Polar.test.ts
@@ -868,7 +868,8 @@ describe('Oso Roles', () => {
         has_permission(actor, action, resource);
 
       has_role(user: User, name, resource) if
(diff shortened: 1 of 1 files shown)


languages/js/src/Polar.test.ts
```diff
@@ -868,7 +868,8 @@ describe('Oso Roles', () => {
         has_permission(actor, action, resource);
 
       has_role(user: User, name, resource) if
-        { name: name, resource: resource } in user.roles;
+        role in user.roles and
+        role matches { name: name, resource: resource };
```
leina05, line 872: Looks like it has to be written this way in JS too, @gj (not just Java).

leina05 (approved):
LGTM, thanks for catching this!

PR #1149, opened 2021-09-08
Remove enforcement aliases

gj:


gkaemmer (approved):
Ah thanks, missed this

PR #1158, opened 2021-09-09
Add rule types for resources that are Actors

leina05:


gw000 (approved):
👍

gj (approved):
For posterity: this is a temporary change until we have better union types in the core. When declaring actor/resource union members is disentangled from the resource block syntax, we'll be able to write `type Actor = User;` and `type Resource = User;` to declare `User` as both an actor and a resource.

PR #1159, opened 2021-09-09
Accept plain objects in js type decls

gw000:
Instead of requiring `Map` objects. For example:
```javascript
oso.registerClass(Foo, {
  types: { id: Number }
});
```
PR checklist:

### DIFF
--- a/languages/js/src/Host.ts
+++ b/languages/js/src/Host.ts
@@ -159,13 +159,18 @@ export class Host {
         existing,
       });
     }
+
+    // TODO(gw) maybe we only want to support plain objects?
+    let fields = types || {};
+    if (!(fields instanceof Map)) fields = new Map(Object.entries(fields));
+
     const userType = new UserType({
       name: clsName,
       cls: cls,
       buildQuery: buildQuery || this.buildQuery,
       execQuery: execQuery || this.execQuery,
       combineQuery: combineQuery || this.combineQuery,
-      fields: types || new Map(),
+      fields: fields,
       id: id || this.cacheInstance(cls, undefined),
     });
     this.types.set(cls, userType);
--- a/languages/js/src/Host.ts
+++ b/languages/js/src/Host.ts
@@ -166,11 +166,11 @@ export class Host {
 
     const userType = new UserType({
       name: clsName,
-      cls: cls,
+      cls,
+      fields,
       buildQuery: buildQuery || this.buildQuery,
       execQuery: execQuery || this.execQuery,
       combineQuery: combineQuery || this.combineQuery,
-      fields: fields,
       id: id || this.cacheInstance(cls, undefined),
     });
     this.types.set(cls, userType);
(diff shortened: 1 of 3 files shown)


languages/js/src/Host.ts
```diff
@@ -159,13 +159,18 @@ export class Host {
...
+    if (!(fields instanceof Map)) fields = new Map(Object.entries(fields));
+
     const userType = new UserType({
       name: clsName,
       cls: cls,
       buildQuery: buildQuery || this.buildQuery,
       execQuery: execQuery || this.execQuery,
       combineQuery: combineQuery || this.combineQuery,
-      fields: types || new Map(),
+      fields: fields,
```
gj, line 173: I think this works in JS? Been bouncing between all 6 langs this week so my head's a bit spinny

```suggestion
      fields,
```

if so could do it for `cls: cls,` above, too

languages/js/src/dataFiltering.test.ts
```diff
@@ -132,34 +132,34 @@ async function fixtures() {
...
-  numType.set('fooId', String);
-  numType.set('foo', new Relation('one', 'Foo', 'fooId', 'id'));
   oso.registerClass(Num, {
-    types: numType,
     buildQuery: fromRepo(nums, 'num'),
+    types: {
+      number: Number,
+      fooId: String,
+      foo: new Relation('one', 'Foo', 'fooId', 'id'),
+    },
```
gj, line 162: don't have enough context to approve the PR, but this looks like a big syntactic improvement

samscott89 (commented):
Thank you! This was one of my biggest complaints.

I'm a little surprised there's not a change needed to Polar.ts. Doesn't `registerClass` take in a `Map<string, any>`? Is typescript okay with object being passed in as a map?

languages/js/src/Host.ts
gw000: Oooh let me check, I sure hope that works...

gw000: 👍 thanks!

gw000:
@samscott89 `registerClass` used to work that way but the number of parameters got too large to manage positionally, so now it takes an optional object of params to simulate keyword arguments. `UserType` still requires a map but `cacheClass` makes sure it always gets one.

languages/js/src/Polar.ts
```diff
@@ -254,6 +254,14 @@ export class Polar {
 
   /**
    * Register a JavaScript class for use in Polar policies.
+   *
+   * @param cls The class to register.
+   * @param params An optional object with extra parameters.
+   *
+   * Accepted extra parameters are:
```
samscott89, line 261: Can we make a follow up issue to go back and add types for these?

samscott89: Or an interface, whatever the right typescript-ism is

gw000: Added it!

samscott89 (approved):
This looks good! Thanks @gw000! and thanks @gj for the TS suggestions :)

PR #1157, opened 2021-09-09
Add Python "Add to your App" guide to docs.

dhatch:
This is ready for a review on the writing & content. There is still some work
to do to get it ready to merge.

PR checklist:

- [ ] Fix or remove TODOs in docs.
- [x] Turn off broken links as warning.
- [x] Test some of the snippets (data filtering & policy in particular) with literal
      include because these aren't in the quickstart.
- [x] Potentially remove other add to your app guides, and convert this to a page
      with data entries so I can start writing other languages.
- [ ] Fix 1x1 with Oso engineer callout & figure out if there's better placement (I feel like the current placement is distracting).

PR #1161, opened 2021-09-10
Ruby implementation of add to your app.

dhatch:
PR checklist:

- [ ] Added changelog entry.
- [ ] Get code working (data filtering is still not working)
- [ ] Make sure ruby (the parts that will be in docs examples) is not horrible
- [ ] Add as literal include to have ruby add to app guide.


dhatch:
@gj or @gkaemmer could use a review from either of you on this code -- not ready to merge yet, but just to make sure it's not too ugly (only the parts that will go into docs)

docs/examples/add-to-your-application/ruby/Gemfile
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+source "https://rubygems.org"
+
+git_source(:github) { |repo_name| "https://github.com/#{repo_name}" }
+
+gem "oso-oso", path: "../../../../languages/ruby/"
```
gkaemmer, line 7: Make this `gem "oso-oso"` before merging

docs/examples/add-to-your-application/ruby/data_filtering.rb
```diff
@@ -0,0 +1,93 @@
...
+
+DB.create_table :repositories do
+  String :name
+  TrueClass :is_public
+end
+
+repositories = DB[:repositories]
+
+module DF
+class Repository < Sequel::Model(:repositories)
```
gkaemmer, line 17:
```suggestion
  class Repository < Sequel::Model(:repositories)
```

dhatch: I have this not indented intentionally because the literalInclude will include the indents I'm pretty sure (which I don't want).

docs/examples/add-to-your-application/ruby/data_filtering.rb
```diff
@@ -0,0 +1,93 @@
...
+  build_query: @get_repositories,
+  combine_query: @combine_query,
+  exec_query: @exec_query
+)
+
+OSO.load_files(["main.polar"])
+# docs: end-data-filtering
+
+
+end
```
gkaemmer, line 72: Does this work? What is this `end`ing?

docs/examples/add-to-your-application/ruby/data_filtering.rb
```diff
@@ -0,0 +1,93 @@
...
+DB = Sequel.sqlite
+
+DB.create_table :repositories do
+  String :name
+  TrueClass :is_public
+end
+
+repositories = DB[:repositories]
+
+module DF
```
gkaemmer, line 16: Does this module need to exist? Is it just to avoid name conflicts with the `Repository` defined in `models.rb`?

docs/examples/add-to-your-application/ruby/data_filtering.rb
gkaemmer: Oh this is ending the module

docs/examples/add-to-your-application/ruby/data_filtering.rb
```diff
@@ -0,0 +1,93 @@
...
+oso = DF::OSO
+
+# docs: begin-list-route
+get "/repos" do
+  repositories = oso.authorized_resources(
+    get_current_user(),
+    "read",
+    Repository)
+
+  serialize(repositories)
```
gkaemmer, line 91: Maybe we should inline the `serialize` method to shave LoC and number of concepts?

docs/examples/add-to-your-application/ruby/data_filtering.rb
```diff
@@ -0,0 +1,93 @@
...
+repositories = DB[:repositories]
+
+module DF
+class Repository < Sequel::Model(:repositories)
+end
+
+r = Repository.new(name: "gmail", is_public: false)
+r.save
+
+OSO = Oso.new
```
gkaemmer, line 23: I think you could just use an `oso` variable (not an `OSO` constant) to hold the oso instance.
```suggestion
oso = Oso.new
```

docs/examples/add-to-your-application/ruby/routes.rb
```diff
@@ -0,0 +1,27 @@
+require 'oso'
+require 'sinatra'
+
+require_relative './oso'
+require_relative './models'
+
+class User
```
gkaemmer, line 7: Maybe this doesn't need to be a class method? just `def get_current_user` at the base level of this file?

docs/examples/add-to-your-application/ruby/routes.rb
```diff
@@ -0,0 +1,27 @@
...
+  repo = Repository.get_by_name(params['name'])
+
+  begin
+    OSO.authorize(User.get_current_user,
+                  "read",
+                  repo)
+    "<h1>A Repo</h1><p>Welcome to repo #{repo.name}</p>"
+  rescue Oso::NotFoundError
+    # TODO can i have a message here
+    404
```
gkaemmer, line 24: I think you can do `status 404` and then return a string, but I'm not 100% sure

gkaemmer (changes requested):
Left some comments on the code pieces of this

docs/examples/add-to-your-application/ruby/data_filtering.rb
```diff
@@ -0,0 +1,93 @@
...
+      value = value.name
+      field = :name
+    else
+      field = filter.field.to_sym
+    end
+
+    if filter.kind == "Eq"
+      query = query.where(field => value)
+    else
+      raise "unimplemeneted constraint kind"
```
gw000, line 41: *unimplemented

docs/examples/add-to-your-application/ruby/data_filtering.rb
```diff
@@ -0,0 +1,93 @@
...
+OSO.register_class(User)
+OSO.register_class(
+  Repository,
+  name: "Repository",
+  fields: {
+    "is_public" => PolarBoolean
+  },
+  build_query: @get_repositories,
+  combine_query: @combine_query,
+  exec_query: @exec_query
```
gw000, line 65: Where are the instance variables defined? Are these supposed to be referencing the method definitions above? I don't think that will work...

Anyway if you define class methods on `Repository` with these names they'll be picked up automatically by `register_class`.

docs/examples/add-to-your-application/ruby/data_filtering.rb
```diff
@@ -0,0 +1,93 @@
...
+def exec_query(q)
+  q.all
+end
+
+OSO.register_class(User)
+OSO.register_class(
+  Repository,
+  name: "Repository",
+  fields: {
+    "is_public" => PolarBoolean
```
gw000, line 61: `"is_public"` can be a symbol if you want.

docs/examples/add-to-your-application/ruby/Gemfile
dhatch: Why? I intentionally want this to use the ruby code from the repo when being tested.

docs/examples/add-to-your-application/ruby/data_filtering.rb
dhatch: Yeah

docs/examples/add-to-your-application/ruby/data_filtering.rb
dhatch: this snippet won't be shown in the docs, I used a different function to serve as a placeholder in the docs for real serialization. I guess I could just call `to_s` directly there

docs/examples/add-to-your-application/ruby/data_filtering.rb
dhatch: I don't think it's accessible on L82 then.

docs/examples/add-to-your-application/ruby/routes.rb
dhatch: I was aiming to match the route from the quickstart... but maybe not necessary

dhatch:
Merged into #1157

PR #1164, opened 2021-09-10
Pass arguments to Oso constructor in Ruby

gw000:


gj (changes requested):
I think we're gonna need to do some testing of this across Ruby versions:

https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/#delegation

gw000:
I tested on 2.4.1 and 3.0.2 & it works

gj (approved):

PR #1169, opened 2021-09-12
Release 0.20.1 beta

samscott89:


languages/js/yarn.lock
```diff
@@ -2,29 +2,29 @@
 # yarn lockfile v1
```
gj, line 2: seems like the prevailing guidance is still to keep JS package manager lockfiles checked in, but man unnecessary ordering changes like this are annoying

samscott89: Oof, yeah. I didn't realise it was such a large change. I imagine this is from different yarn versions? Should we standardise on one?

PR #1170, opened 2021-09-12
Versioned docs

samscott89:
The main ideas here are:

- Deploy a specific docs "tag" to a subfolder `/v/<tag>`. This can be used for both versions `/v/0.15/` as well as for deploying previews `/v/<branch name>`. I'd expect the former for `docs.osohq.com`, and the latter for `docs-preview.oso.dev`.
- The docs-preview CI job should now push all changes live to `docs-preview.oso.dev/v/<tag>`
- Adds a version selector to the header. This could go somewhere else, I think it's pretty subtle right now.
- Version selector picks based on a dynamic JSON file `data.oso.dev/versions.json`. This means that we don't need to redeploy every single version whenever we add a version, since old versions will read the list from that JSON file.

Since this branch now gets pushed live by CI, you can see it here: https://docs-preview.oso.dev/v/sam-versioned-docs/index.html

Still TODO to make this fully functional:
- [ ] Setup cloudflare to redirect anything that _doesn't_ start with `/v/` to `/v/latest`. This means that publishing the docs live means pushing to `/v/latest/`. The reason we would need to do that, is otherwise if you try and sync the docs to `/`, then it will try and delete all the subfolders of `/v/`. AFAICT, the best way to work around this is to do this redirecting logic.
- [ ] (not blocking) Setup the CI workflow that will push a tag to `/v/<tag>` on docs.osohq.com
- [ ] (not blocking) something to help generate the `versions.json` and to deploy that to s3?

Another small thing in this PR:
- uses `s3deploy` https://github.com/bep/s3deploy to handle the update logic + cache clearing. It's a bit smarter about knowing when it needs to actually redeploy files. I'm hoping this will lessen the number of times where publishing docs leads to caching problems with stylesheets etc.


gneray:
@dhatch and I agreed to put version selector on releases page.

docs/themes/oso-webpack/getVersions.js
```diff
@@ -0,0 +1,44 @@
...
+
+    return await cachedResponse.json();
+}
+
+export async function setVersionList() {
+    try {
+        const data = await getVersions();
+        document.getElementById('version-list').innerHTML = data.versions.reduce((html, v) => {
+            return html + `
+          <a href="https://docs-preview.oso.dev/v/${v}/index.html"
```
samscott89, line 35:
```suggestion
          <a href="https://docs.osohq.com/v/${v}/index.html"
```

samscott89:
Went with a simpler approach:

- Banner at the top of the page for old docs to take you to latest
- Latest has list of releases

PR #1187, opened 2021-09-14
Move language selector out of header to top of page.

dhatch:
Screen Shot 2021-09-14 at 1 39 04 PM


docs/themes/oso-tailwind/assets/js/nav.js
```diff
@@ -26,13 +26,15 @@ if (sideBarButton) {
...
 }
 
-const langButton = document.getElementById('language-selector-toggle');
-const langsideBar = document.getElementById('language-selector-content');
-if (langButton) {
-  langButton.addEventListener('click', () =>
-    langsideBar.classList.toggle('hidden')
-  );
-}
+window.addEventListener('load', () => {
```
samscott89, line 29: Why is this changed to on load? 

`langButton` needs to be declared for the closing logic below (line 53).

docs/config.toml
```diff
@@ -4,7 +4,7 @@ title = "Oso Documentation"
 theme = ["oso-tailwind", "oso-webpack"]
 uglyurls = true
 
-defaultContentLanguage = "any"
+defaultContentLanguage = "python"
```
samscott89, line 7: This is a fantastic idea. Kudos to whoever suggested it

docs/themes/oso-tailwind/assets/js/nav.js
dhatch: Oh might be able to avoid this, it wasn't attaching the event listener but I think that was for a different reason

docs/config.toml
dhatch: @gneray

gkaemmer: Does this work? This has the potential to vastly simplify things

samscott89: It does

samscott89: https://docs-preview.oso.dev/v/dhatch-lang-selector-v2

samscott89: Other than the above comment which is breaking things

samscott89 (approved):

PR #1206, opened 2021-09-20
Re-enable ruby add to your app tests

gkaemmer:


.github/workflows/test.yml
```diff
@@ -348,6 +348,7 @@ jobs:
         uses: ruby/setup-ruby@v1
         with:
           ruby-version: 3.0
```
gj, line 350: Any idea why we're using Ruby 3 in the `docs_test` job but Ruby 2.4 in the `ruby_test` job?

gkaemmer: No idea. Seems not good. Will try downgrading here and see what happens. If it doesn't work, I may punt on it though

gkaemmer: Seems to work

docs/examples/add-to-your-application/Makefile
```diff
@@ -14,5 +18,5 @@ test-node:
 test-ruby:
 	# TODO disabled because these won't pass CI for some reason
```
dhatch, line 19: Remove comment.

dhatch (approved):
lgtm besides comment

PR #1208, opened 2021-09-20
Small updates to rule types tests

leina05:


polar-core/src/kb.rs
```diff
@@ -1100,6 +1168,17 @@ mod tests {
             )
             .unwrap()
             .is_true());
+        // test with *rest vars
+        // rule: f(x: [1, 2, 3]), rule_type: f(x: [1, 2, *rest]) => PASS
```
gj, line 1172:
```suggestion
        // rule: f(x: [1, 2, 3]), rule_type: f(x: [1, 2, *rest]) => ERROR
```

polar-core/src/kb.rs
```diff
@@ -943,6 +967,26 @@ mod tests {
...
             .unwrap()
             .is_true());
+        // rule: f(x: 6), rule_type: f(x: String) => FAIL
+        assert!(!kb
+            .rule_params_match(
+                &rule!("f", ["x"; value!(6)]),
+                &rule!("f", ["x"; instance!(sym!("String"))])
+            )
+            .unwrap()
+            .is_true());
```
gj, line 977: what's the deal with the new tests in here that are checking primitive values against built-in types w/o any fields? This one + `Float`, `Integer`, `Boolean`, `List`, and `Dictionary`

gj (approved):
one inconsequential question & one nit

polar-core/src/kb.rs
leina05: I realized that we didn't test that the check would fail if the rule type specializer is a primitive and the rule value is the wrong type. We were only checking if the value was a primitive and the specializer was Foo, which would be handled by a different piece of code since the function matches on the type of the rule type rather than the rule.

PR #1209, opened 2021-09-20
Keep unify constraints on dot lookups

gw000:
The simplifier was dropping unify constraints for dot lookups on free variables.  This fix just takes out the "is it a dot lookup?" special case. Seems to work, but there may be something I don't understand about why that case was needed to begin with?

https://www.notion.so/osohq/Losing-unification-constraint-when-querying-with-two-variables-60ecb0671c7f4e7781899b3462f0b31b

PR checklist:

- [ ] Added changelog entry.


polar-core/src/partial/simplify.rs
```diff
@@ -405,22 +391,14 @@ impl Simplifier {
                         // Replace variable with value if the value is ground
                         // or a dot output (x.foo) and variable is mentioned elsewhere
                         // in the expression.
```
gj, line 393:
```suggestion
                        // Replace unbound variable with ground value.
```

polar-core/src/partial/simplify.rs
```diff
@@ -405,22 +391,14 @@ impl Simplifier {
...
-                                && !self.is_bound(var)
-                                && !right.contains_variable(var) =>
-                        {
+                        (Value::Variable(var), val) if val.is_ground() && !self.is_bound(var) => {
                             simplify_debug!("*** 3");
                             MaybeDrop::Check(var.clone(), right.clone())
                         }
                         // Replace variable with value if the value is ground
                         // or a dot output (x.foo) and variable is mentioned elsewhere
                         // in the expression.
```
gj, line 400:
```suggestion
                        // Replace unbound variable with ground value.
```

polar-core/src/partial/partial.rs
```diff
@@ -588,8 +601,8 @@ mod test {
 
         let mut q = p.new_query_from_term(term!(call!("h", [sym!("x"), sym!("y")])), false);
         assert_partial_expressions!(next_binding(&mut q)?,
-            "x" => "y.x matches X{}",
-            "y" => "_this.x matches X{}"
+            "x" => "_this matches X{} and y.x = _this",
+            "y" => "x matches X{} and _this.x = x"
```
gj, line 605: little worried about changes like this trickling into the django-oso / sqlalchemy-oso constraint transformation code — do we have tests in those libraries for this behavior?

gw000: No, what should they test?

gj: At a minimum I'd feel more confident if we had analogous test cases for `h(x2: X, _: {x: x2});` and `foo(u: User, x: Post) if x.user = u;`, which seem to be the two cases in the core that were impacted by this change. What do you think?

gw000: But what are they actually testing? Are they identical to the core tests?

gw000: Ok, I've looked into this. SQLAlchemy and Django can't handle the partials that come out of rules like this, because they have multiple unbound variables. Since the adapters are intended to handle queries over one free variable, I don't think this is likely to cause any issues. But I added tests to each library to check that a partial with both an isa and a dot operation works & produces the expected result. What do you think?

gj: exceeds expectations

gj (approved):
when CI's happy, ship it

polar-core/src/partial/partial.rs
gw000: ty gj!!

PR #1211, opened 2021-09-20
Use ESLint & typescript-eslint to lint JS library

gj:


languages/js/src/Host.ts
```diff
@@ -117,38 +138,56 @@ export class Host {
...
-    return typ;
+  getType(cls: Class | string): UserType | undefined {
+    return this.types.get(cls);
   }
 
   /**
    * Return user types that are registered with Host.
    */
-  *distinctUserTypes(): IterableIterator<UserType> {
+  private *distinctUserTypes(): IterableIterator<UserType> {
```
gw000, line 151: what's the * do?

languages/js/src/Polar.ts
```diff
@@ -116,8 +108,8 @@ export class Polar {
    * @internal
    */
   private processMessages() {
-    while (true) {
-      let msg = this.#ffiPolar.nextMessage();
+    for (;;) {
```
gw000, line 111: I love this syntax, idk why

languages/js/src/helpers.ts
```diff
@@ -148,22 +142,14 @@ function parseNextExternal(d: obj): QueryEvent {
...
-    !Number.isSafeInteger(callId) ||
-    !isPolarTerm(instance) ||
-    typeof attribute !== 'string' ||
-    (args !== undefined &&
-      (!Array.isArray(args) || args.some((a: unknown) => !isPolarTerm(a))))
-  )
-    throw new Error();
+  if (typeof attribute !== 'string') throw new Error();
+  if (!isSafeInteger(callId)) throw new Error();
+  if (!isPolarTerm(instance)) throw new Error();
```
gw000, line 152: could cut down on repetition a little with something like
```javascript
type Pred<A> = (a: A) => boolean;
function validate<A>(x: A, ...ps: Pred<A>[]) {
  for (cons p of ps) if (!p(x)) throw new Error();
}

validate(x, isPolarTerm)
```

languages/js/src/helpers.ts
```diff
@@ -359,12 +324,63 @@ export function printError(e: Error) {
...
+
+/**
+ * Type guard to test if a value is an ES6 Map with string keys and PolarTerm
+ * values.
+ *
+ * @internal
+ */
+function isMapOfPolarTerms(x: unknown): x is Map<string, PolarTerm> {
+  if (!(x instanceof Map)) return false;
+  for (const [key, value] of x.entries()) {
```
gw000, line 355: can we take out the `.entries()`?

languages/js/src/Host.ts
gj: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function*

languages/js/src/Polar.ts
gj: is it b/c it looks like a crying face

languages/js/src/helpers.ts
gj: a7dce28

languages/js/src/Oso.ts
```diff
@@ -300,15 +319,19 @@ export class Oso<
...
+    action: Action,
+    resourceCls: Class<T>
+  ): Promise<T[]> {
+    const query = await this.authorizedQuery(actor, action, resourceCls);
+    if (!query) return [];
+    const userType = this.getHost().getType(resourceCls);
+    if (userType === undefined)
+      throw new UnregisteredClassError(resourceCls.name);
+    if (userType.execQuery === undefined)
+      throw new DataFilteringConfigurationError('execQuery');
```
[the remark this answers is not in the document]
gj: 7f83479c

languages/js/src/helpers.ts
```diff
@@ -359,12 +324,63 @@ export function printError(e: Error) {
...
+  return true;
+}
+
+/**
+ * Type guard to test if a value is an Array of PolarTerms.
+ *
+ * @internal
+ */
+function isArrayOfPolarTerms(x: unknown): x is Array<PolarTerm> {
+  return !Array.isArray(x) || x.some((v: unknown) => !isPolarTerm(v));
```
gw000, line 367: is this logic right? `isArrayOfPolarTerms("foo") = !Array.isArray("foo") || ... = !false || ... = true` ??

anyway it'd be nice to build these compound checks out of combinators
```javascript
const negate<A> = (pred: Pred<A>) => (a: A) => !pred(a);
const conjoin<A> = (p0: Pred<A>, p1: Pred<A>) => (a: A) => p0(a) && p1(a);
const disjoin<A> = (p0: Pred<A>, p1: Pred<A>) => (a: A) => p0(a) || p1(a);
const some<A> = (p: Pred<A>) => (x: A[]) => x.some(p);
const every<A> = (p: Pred<A>) => (x: A[]) => x.every(p);

const isArrayOfPolarTerms = conjoin(Array.isArray, every(isPolarTerm));
```

languages/js/src/Polar.ts
gw000: yes 😭

languages/js/src/Host.ts
gw000: huh, TIL 👍

languages/js/src/Oso.ts
```diff
@@ -131,7 +146,9 @@ export class Oso<
           return new Set(['*']);
         }
       }
-      actions.add(action);
+      // TODO(gj): do we need to handle the case where `action` is something
+      // other than a `Variable` or an `Action`? E.g., if it's an `Expression`?
```
gj, line 150: @gkaemmer any thoughts about this or the equivalent comment down below in `authorizedFields()`?

languages/js/src/Oso.ts
```diff
@@ -244,22 +263,25 @@ export class Oso<
    * @returns A query that selects authorized resources of type `resourceCls`
    */
   async authorizedQuery(
-    actor: any,
-    action: any,
-    resourceCls: any
-  ): Promise<any> {
+    actor: Actor,
+    action: Action,
+    resourceCls: Class
```
gj, line 268: Should this be `Resource`? Or maybe a type param that is a `Resource` but also extends `Class`? Not sure how to express that

languages/js/src/Oso.ts
```diff
@@ -244,22 +263,25 @@ export class Oso<
...
+  ): Promise<unknown> {
```
gj, line 269: Would be cool to link this up with the `Query` type param if we end up moving data filtering query params into the Oso constructor.

languages/js/src/Oso.ts
```diff
@@ -300,27 +321,25 @@ export class Oso<
...
-    resourceCls: any
-  ): Promise<any[]> {
-    const query = await this.authorizedQuery(actr, actn, resourceCls);
-    return !query
-      ? []
-      : this.getHost().types.get(resourceCls)!.execQuery!(query);
+  async authorizedResources<T>(
+    actor: Actor,
+    action: Action,
+    resourceCls: Class<T>
```
gj, line 327: Wonder if there's a way to tie this `Class<T>` to the `Resource` type param on the `Oso` class

gj: Also as an aside our `Class` type seems very similar to TypeScript's built-in [`InstanceType<Type>`](https://www.typescriptlang.org/docs/handbook/utility-types.html#instancetypetype)

languages/js/src/Query.ts
```diff
@@ -323,6 +340,11 @@ export class Query {
               this.processMessages();
             });
             break;
+          }
+          default: {
+            const _: never = event.kind;
+            return _;
```
gw000, line 346: is this is a typescript thing?

gj: https://www.typescriptlang.org/docs/handbook/2/narrowing.html#exhaustiveness-checking

languages/js/src/helpers.ts
```diff
@@ -106,19 +102,20 @@ function parseResult({ bindings }: obj): QueryEvent {
...
-    fields.some((v: unknown) => !isPolarTerm(v))
-  )
-    throw new Error();
+function parseMakeExternal(event: unknown): QueryEvent {
+  if (!isObj(event)) throw new Error();
+  const { instance_id: instanceId } = event;
+  if (!isSafeInteger(instanceId)) throw new Error();
+  const ctor = event['constructor'];
+  if (!isPolarTerm(ctor)) throw new Error();
+  if (!isPolarPredicate(ctor.value)) throw new Error();
```
gw000, line 111: there's so many of these I feel like we should just write `assert()`

languages/js/src/types.ts
```diff
@@ -617,5 +631,67 @@ export function isAsyncIterator(x: any): boolean {
...
+  /**
+   * Explicit name to use for the class in Polar. Defaults to the class's
+   * `name` property.
+   */
+  name?: string;
+  /**
+   * A Map or object with string keys containing types for fields. Used for
+   * data filtering.
+   */
+  fields?: obj<Class | Relation> | Map<string, Class | Relation>;
```
gw000, line 671: This is awesome 😄

languages/js/src/types.ts
```diff
@@ -591,17 +590,32 @@ export interface Options {
  *
  * @internal
  */
-export function isIterableIterator(x: any): boolean {
-  return typeof x?.next === 'function' && Symbol.iterator in Object(x);
+export function isIterableIterator(x: unknown): x is IterableIterator<unknown> {
```
gw000, line 593: so this `x is y` notation means that it adds a type constraint in the calling scope depending on if it returns true or false?

gj: yep: https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

languages/js/src/dataFiltering.ts
```diff
@@ -31,114 +69,106 @@ export class Field {
...
+  host: Host,
+  plan: FilterPlan
+): Promise<unknown> {
+  const queries = [];
+  let combine: CombineQueryFn | undefined;
+  for (const rs of plan.result_sets) {
+    const setResults: Map<number, unknown[]> = new Map();
+    for (const i of rs.resolve_order) {
+      const req = rs.requests.get(i)!;
+      const filters = await Promise.all(
```
gw000, line 139: nice I always wanted to know how to do this

languages/js/src/dataFiltering.ts
```diff
@@ -31,114 +69,106 @@ export class Field {
...
-    value = new Field(value);
-  }
+function groundFilter(results: Map<number, unknown[]>, filter: Filter): Filter {
+  const ref = filter.value;
+  if (!(ref instanceof Ref)) return filter;
 
-  return new Filter(kind, field, value);
-}
+  let value = results.get(ref.resultId);
+  value = ref.field ? value?.map(v => (v as any)[ref.field!]) : value; // eslint-disable-line @typescript-eslint/no-explicit-any
```
gw000, line 124: `!ref.field ? value : ...` has better end-weight

gw000 (approved):
🔥 

don't forget to call out the breaking API change!

languages/js/src/helpers.ts
gj: Tried this out:

1. Unfortunately it's not currently possible to parameterize a function with variadic generic types (w/o this I think we'd have to write a separate `validateN()` function for each arity we wanted to support since each generic type corresponds to a different predicate and the individual `x is X` assertions are ultimately `&`-ed together to create something like `x is A & B & C`).
2. I don't think it's possible to do this (but would love to be wrong about this!) b/c it's conflating the check that `a is A` with the enforcement of `throw new Error()`. The reason `if (!pred(a)) throw new Error();` works is b/c `pred(a)` asserts that `a is A` and `throw new Error()` basically acts as a "Do Not Pass Go; Do Not Collect $200" bomb-out, so that line tells the TS compiler, "if we make it past this line, `a is A`," i.e., the type of `a` has been narrowed to `A`.
3. We could probably write something a bit less generic and then do our own `as X` assertion if all of the predicates pass instead of trying to rely on the TS compiler's `is X` narrowing.

languages/js/src/dataFiltering.ts
gj: e0349238

languages/js/src/Oso.ts
gj: Verified that this works: [`8d49c50` (#1211)](https://github.com/osohq/oso/pull/1211/commits/8d49c505a9d6c96f586d98f977b0836cc46e27d3#diff-2178a05ab0077a949c52ac5f9d679b82d59967f2a2536708262b202d53dffee6R268)

languages/js/src/Oso.ts
gj: > Wonder if there's a way to tie this `Class<T>` to the `Resource` type param on the `Oso` class

turns out there is: [`8d49c50` (#1211)](https://github.com/osohq/oso/pull/1211/commits/8d49c505a9d6c96f586d98f977b0836cc46e27d3#diff-2178a05ab0077a949c52ac5f9d679b82d59967f2a2536708262b202d53dffee6R324)

languages/js/src/helpers.ts
gj: This would be cool; did a very novice version of this in c40377d0455c3daf4a161bf5e6cc8edd9e0977f8. Gonna punt for now on putting any more work into cleaning up the checks — a lot of (all of?) them are validating the structure of code emitted by the VM, so we control both ends of the pipe and could _probably_ delete most of this validation and not lose too much sleep. The validation checks were very helpful when I was building the library and figuring out what the shape of the data was after it'd gone through the wasm vortex, but we could probably just assert the shape of the data and move on.

languages/js/src/helpers.ts
gj: https://github.com/osohq/oso/pull/1211#discussion_r714507826

languages/js/src/Oso.ts
gw000: That would be cool but if we want to support different data sources then I think the query type would have to come through the resource.

languages/js/src/Oso.ts
gkaemmer: Oh yeah I suppose the return type should be `Set<Action | Expression | "*">`

PR #1212, opened 2021-09-20
Test RBAC guides against HEAD

gkaemmer:
The java, ruby, and python RBAC guides already seem to be using a local version of Oso. This updates the Go and Node RBAC guides to also run their tests against an installation at a relative path.

### DIFF
--- a/docs/examples/rbac/go/Makefile
+++ b/docs/examples/rbac/go/Makefile
@@ -1,4 +1,5 @@
 .PHONY: test
 
 test:
+	make -C ../../../../languages/go
 	go test
--- a/docs/examples/rbac/go/Makefile
+++ b/docs/examples/rbac/go/Makefile
@@ -1,5 +1,5 @@
 .PHONY: test
 
 test:
-	make -C ../../../../languages/go
+	make -C ../../../../languages/go copy_lib
 	go test
--- a/docs/examples/rbac/go/app.go
+++ b/docs/examples/rbac/go/app.go
@@ -46,7 +46,9 @@ o.RegisterClass(reflect.TypeOf(Repository{}), nil)
 o.RegisterClass(reflect.TypeOf(User{}), NewUser)
 // docs: end-register
 
-o.LoadFiles([]string{"main.polar"})
+if err := o.LoadFiles([]string{"main.polar"}); err != nil {
+	panic(err)
+}
 // docs: end-setup
 
 return o
--- a/docs/examples/rbac/go/app.go
+++ b/docs/examples/rbac/go/app.go
@@ -46,9 +46,7 @@ o.RegisterClass(reflect.TypeOf(Repository{}), nil)
 o.RegisterClass(reflect.TypeOf(User{}), NewUser)
 // docs: end-register
 
-if err := o.LoadFiles([]string{"main.polar"}); err != nil {
-	panic(err)
-}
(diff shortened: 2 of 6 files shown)


docs/examples/rbac/nodejs/Makefile
```diff
@@ -5,4 +5,6 @@ test: node_modules
 
 node_modules: package.json
 	yarn
+	cd ../../../../languages/js; yarn link
```
gj, line 8: Is this going to leave us in `../../../../languages/js` for the following steps in this task or will we pop back to `docs/examples/rbac/nodejs`?

docs/examples/rbac/go/Makefile
```diff
@@ -1,4 +1,5 @@
 .PHONY: test
 
 test:
+	make -C ../../../../languages/go
```
gj, line 4: Maybe worth specifying the (current) default task in case we change it? Looks like it's `copy_lib`:
```suggestion
	make -C ../../../../languages/go copy_lib
```
Also begs the question... what happens if we run `make -C docs/examples/rbac/go test` before we build the core lib? Should we add a task in the Go makefile that builds the Rust core?

docs/examples/rbac/ruby/Makefile
```diff
@@ -1,3 +1,5 @@
+unexport BUNDLE_GEMFILE
```
gj, line 1: I saw this in another PR as well, but could you add an inline comment explaining what it's doing?

gj (changes requested):

docs/examples/rbac/go/app.go
```diff
@@ -46,7 +46,9 @@ o.RegisterClass(reflect.TypeOf(Repository{}), nil)
 o.RegisterClass(reflect.TypeOf(User{}), NewUser)
 // docs: end-register
 
-o.LoadFiles([]string{"main.polar"})
+if err := o.LoadFiles([]string{"main.polar"}); err != nil {
+	panic(err)
+}
```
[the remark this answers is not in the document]
gkaemmer: Ah yeah I added this actually because I was getting a silent error here (which had to do with me not running `make` in the go directory for a while). Happy to revert this piece for simplicity/conciseness.

docs/examples/rbac/nodejs/Makefile
gkaemmer: No, it'll bring you back. TIL a number of things about making, including that it runs each command as a subprocess, so your `cd`s only affect that one line.

docs/examples/rbac/go/Makefile
gkaemmer: Oh yeah actually wasn't sure about this. Will change to `make copy_lib`. 

And yeah we probably should be explicit somehow that this depends on the rust core being built. In this case it happens to be built by the earlier python build.

Through this change I have been feeling like we should do a little audit of all of our make dependencies. I bet we could clean a lot up and also improve build times if we really nail down what depends on what, especially in the docs tests.

docs/examples/rbac/ruby/Makefile
gkaemmer: Yeah good idea, this took me a while to figure out

gj (approved):
Add a follow-up task for https://github.com/osohq/oso/pull/1212#discussion_r712555323 and :shipit:

PR #1214, opened 2021-09-20
Fix quickstart tests by updating Node version

leina05:


gkaemmer (approved):
Makes sense

gj (changes requested):
Why is this necessary?

We use 12.20 in other parts of the CI process b/c certain development dependencies require it (Jest, IIRC), but we test against 10.0.0 here because that's the minimum Node version the library supports.

leina05:
> Why is this necessary?
> 
> We use 12.20 in other parts of the CI process b/c certain development dependencies require it (Jest, IIRC), but we test against 10.0.0 here because that's the minimum Node version the library supports.

The minimum version is actually 12.20 according to the README.

.github/workflows/publish.yml
```diff
@@ -117,10 +117,10 @@ jobs:
         run: |
           mkdir -p oso-js-node-${{ github.event.inputs.version }}
           tar -xzf oso-js-node-${{ github.event.inputs.version }}.tgz -C oso-js-node-${{ github.event.inputs.version }} --strip-components=1
-      - name: Use Node.js 12.20.0
+      - name: Use Node.js 10.24.0
         uses: actions/setup-node@v1
         with:
-          node-version: 12.20.0
+          node-version: 10.24.0
```
gj, line 123: Not consequential here, but a general note: I think YAML removes trailing `.0`s from numbers, so this will actually get parsed as `10.24` and the latest version that matches `10.24.x` will be installed instead of `10.24.0`.

gj (approved):

PR #1215, opened 2021-09-21
Throw an error if the attribute is truly undefined

### LINKED ISSUE #749: Non-existent attributes still match in a rule head specializer in Node.JS library
A rule like:
```
allow(actor, action, resource: {attr: attr});
```
still matches even if the object passed in as `resource` does not have the `attr` attribute. This has to do with how the library handles `undefined` (it passed it back into Polar instead of treating it as a failed match).

samscott89:
Differentiates between `{x: undefined}.x = y` and `{}.x = y`. The former is fine with `y = undefined`, the latter raises an error.

Fixes #749 

Discussed this approach with @gj a few weeks back. Check for the attribute with `"x" in d`.

### DIFF
--- a/languages/js/src/Polar.test.ts
+++ b/languages/js/src/Polar.test.ts
@@ -833,6 +833,7 @@ describe('iterators', () => {
       await qvar(p, 'x = new BarIterator([1, 2, 3]).sum()', 'x', true)
     ).toBe(6);
   });
+
 });
 
 test('handles expressions', async () => {
@@ -947,10 +948,10 @@ describe('Oso Roles', () => {
   });
 
   test('rule types correctly check subclasses', async () => {
-    class Foo {}
-    class Bar extends Foo {}
-    class Baz extends Bar {}
-    class Bad {}
+    class Foo { }
+    class Bar extends Foo { }
+    class Baz extends Bar { }
+    class Bad { }
 
     // NOTE: keep this order of registering classes--confirms that MROs are added at the correct time
     const p = new Polar();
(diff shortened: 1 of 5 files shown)


samscott89:
@gj what do you think? This would be a breaking change. Do you think anyone would depend on the old behaviour?

languages/js/src/Query.ts
```diff
@@ -174,6 +175,10 @@ export class Query {
             // Error on attempt to call non-function.
             throw new InvalidCallError(receiver, attr);
           }
+        } else {
+          if (value === undefined && !(attr in receiver)) {
```
gj, line 179:
```suggestion
          // If value isn't a property anywhere in receiver's prototype chain,
          // throw an error.
          if (value === undefined && !(attr in receiver)) {
```

languages/js/src/Query.ts
```diff
@@ -174,6 +175,10 @@ export class Query {
...
+            throw new InvalidAttributeError(receiver, attr)
+          }
         }
       }
     } catch (e) {
```
gj, line 184: Should we also check `e instanceof InvalidAttributeError` on the line below this one?

languages/js/src/Polar.test.ts
```diff
@@ -992,3 +992,13 @@ describe('Oso Roles', () => {
...
 });
+
+test('can specialize on a dict with undefineds', async () => {
+  const p = new Polar();
+  p.registerClass(User);
+  await p.loadStr('f(_: {x: 1});');
+  const result1 = await query(p, pred('f', { x: 1 }));
+  expect(result1).toStrictEqual([map()]);
+  const result2 = await query(p, pred('f', {}));
+  expect(result2).toStrictEqual([]);
```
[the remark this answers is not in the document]
samscott89: I have no idea what this means.

PR #1216, opened 2021-09-21
Fix JS data filtering initial constraint bug

gw000:
PR checklist:

- [x] Added changelog entry.


languages/js/src/Query.ts
```diff
@@ -50,9 +50,10 @@ export class Query {
     this.#calls = new Map();
     this.#host = host;
 
-    for (const k in bindings) {
-      this.bind(k, bindings.get(k));
-    }
+    if (bindings)
+      for (const [k, v] of bindings) {
+        this.bind(k, v);
+      }
```
gw000, line 56: This was the source of the bug. The loop wasn't actually iterating over anything, so the initial type constraint on `resource` wasn't being set, so it was matching a rule it shouldn't have matched which was inducing an impossible constraint.

samscott89: WOW

languages/js/src/Query.ts
```diff
@@ -51,9 +51,7 @@ export class Query {
     this.#calls = new Map();
     this.#host = host;
 
-    for (const k in bindings) {
-      this.bind(k, bindings.get(k));
-    }
+    if (bindings) for (const [k, v] of bindings) this.bind(k, v);
```
gw000, line 54: This is the fix for the bug.

languages/js/tsconfig.json
```diff
@@ -12,9 +12,10 @@
     "rootDir": ".",
     "experimentalDecorators": true,
     "emitDecoratorMetadata": true,
+    "strictPropertyInitialization": false
```
gj, line 15: This seems not great; what's the case where we can't set a sensible default in the constructor and then update it sometime thereafter?

languages/js/src/Query.ts
```diff
@@ -136,31 +137,24 @@ export class Query {
...
+          receiver[rel.myField]
+        );
+        let query = await Promise.resolve(typ.buildQuery!([constraint]));
+        let results = await Promise.resolve(typ.execQuery!(query));
+        if (rel.kind == 'one') {
+          if (results.length != 1)
+            throw new Error('Wrong number of parents: ' + results.length);
+          value = results[0];
+        } else {
+          value = results;
```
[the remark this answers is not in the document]
samscott89: Would it make sense to change like 158 to be `else`? Is there ever a case where we want to look up a value as an attribute even if it was registered as a relation?

languages/js/src/Query.ts
```diff
@@ -137,31 +135,24 @@ export class Query {
...
+        const typ = userTypes.get(rel.otherType)!;
+        // Use the fetcher for the other type to traverse
+        // the relationship.
+        const constraint = new Filter(
+          'Eq',
+          rel.otherField,
+          receiver[rel.myField]
+        );
+        let query = await Promise.resolve(typ.buildQuery!([constraint]));
+        let results = await Promise.resolve(typ.execQuery!(query));
```
samscott89, line 149:
```suggestion
        const query = await Promise.resolve(typ.buildQuery!([constraint]));
        const results = await Promise.resolve(typ.execQuery!(query));
```

languages/js/src/dataFiltering.test.ts
```diff
@@ -35,14 +43,77 @@ class Num {
   number!: number;
 }
 
+@Entity()
+export class Issue {
```
samscott89, line 47: It would be good to include the relationships here too, to make sure there aren't any clashes between Polar's relations and the typeorm relations?

languages/js/src/Query.ts
gj: pls not until #1211 goes in 😄

languages/js/src/Query.ts
```diff
@@ -137,31 +135,24 @@ export class Query {
...
+        if (rel.kind == 'one') {
```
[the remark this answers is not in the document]
gj: yeah would prefer to get this PR in and then I'll update #1211 accordingly

languages/js/src/dataFiltering.test.ts
gw000: 👍

languages/js/tsconfig.json
gw000: fixed!

languages/js/src/Query.ts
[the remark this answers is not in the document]
gw000: ✅

samscott89 (approved):
LGTM!

gj (approved):

PR #1218, opened 2021-09-21
Add changelog for #1215

samscott89:
PR checklist:

- [ ] Added changelog entry.

### DIFF
--- a/docs/content/any/project/changelogs/2021-09-29.md
+++ b/docs/content/any/project/changelogs/2021-09-29.md
@@ -24,10 +24,10 @@ description: >-
   before upgrading.
 {{% /callout %}}
 
-##### Undefined attributes now error
+##### Accessing a nonexistent property on an object now errors
 
-To help with catching erroneous policies, accessing an undefined attribute
-on an object will now return an error.
+To help with catching erroneous policies, accessing a property that doesn't
+exist on an object or anywhere in its prototype chain will now return an error.
 
 Policy:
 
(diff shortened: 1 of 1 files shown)


docs/content/any/project/changelogs/2021-09-29.md
```diff
@@ -15,37 +15,51 @@ description: >-
...
 {{% callout "Warning" "orange" %}}
   This release contains breaking changes. Be sure to follow migration steps
   before upgrading.
 {{% /callout %}}
 
-##### Breaking change 1
+##### Undefined attributes now error
+
+To help with catching erroneous policies, accessing an undefined attribute
+on an object will now return an error.
```
gj, line 30:
```suggestion
To help with catching erroneous policies, accessing a property that doesn't
exist on an object or anywhere in its prototype chain will now return an error.
```

docs/content/any/project/changelogs/2021-09-29.md
```diff
@@ -15,37 +15,51 @@ description: >-
...
 
-<!-- TODO: remove warning and replace with "None" if no breaking changes. -->
-
 {{% callout "Warning" "orange" %}}
   This release contains breaking changes. Be sure to follow migration steps
   before upgrading.
 {{% /callout %}}
 
-##### Breaking change 1
+##### Undefined attributes now error
```
gj, line 27:
```suggestion
##### Accessing a nonexistent property on an object now errors
```

docs/content/any/project/changelogs/2021-09-29.md
```diff
@@ -15,37 +15,51 @@ description: >-
...
 
-##### Feature 1
+```js
+const user = {name: "notanadmin"};
+await oso.isAllowed(user, "read", "document");
+```
 
-Summary of user-facing changes.
+Before this would silently fail, since `user.isAdmin = undefined` in JavaScript, and
+`undefined` is not `true`.
```
gj, line 46:
```suggestion
Before this change, this `isAllowed()` call would silently fail. Oso would look up the
nonexistent `isAdmin` property on the `{name: "notanadmin"}` object, which returns
`undefined`. Since `undefined` is not equivalent to `true`, the `isAllowed()` call would
fail, but the developer wouldn't have any idea that they'd written a policy that referenced
a nonexistent property.
```

docs/content/any/project/changelogs/2021-09-29.md
```diff
@@ -15,37 +15,51 @@ description: >-
...
 
-Link to [relevant documentation section]().
+Now, this will error since `isAdmin` is not a property on `user`.
```
gj, line 48:
```suggestion
Now, this will error since `isAdmin` is not a property on the `user` object.
```

docs/content/any/project/changelogs/2021-09-29.md
```diff
@@ -15,37 +15,51 @@ description: >-
...
+
+This will still work (1) if `user = {isAdmin: undefined}`,
+and (2) by instead using the policy
```
gj, line 51:
```suggestion
This will still work if (1) the `isAdmin` property is explicitly set to `undefined`
(e.g., `user = {isAdmin: undefined}`) or (2) the property check is moved into the head
of the rule:
```

docs/content/any/project/changelogs/2021-09-29.md
```diff
@@ -15,37 +15,51 @@ description: >-
...
+
+```polar
+allow(user: {isAdmin: true}, "read", "document");
+```
+since specializers will not error if a field is missing.
```
gj, line 56:
```suggestion
since specializers will not error if a property doesn't exist (but the `{name: "notanadmin"}`
object will fail to match the specializer and the `isAllowed()` call will still correctly
return `false`).
```

gj (approved):

PR #1220, opened 2021-09-22
add py.typed to sqlalchemy-oso

leina05:


dhatch:
Did you test this by confirming users can see type annotations?

dhatch (approved):

gj (changes requested):
@dhatch:

> Doesn’t need another review but it would be great to test by building the wheel & confirming it has annotations
>
> Getting the settings right in setup.py can impact whether py.typed actually gets built in the wheel

@leina05 you can dismiss this review once you've confirmed the above

leina05:
Confirmed

PR #1222, opened 2021-09-22
Make docs header bigger.

dhatch:


docs/layouts/home.html
```diff
@@ -2,7 +2,7 @@
 {{- partial "sidebar.html" . -}}
 <div class="md:w-4/5 lg:w-3/5 lg:mx-0 mx-auto pt-12">
     <div class="w-full md:mt-6 lg:mt-0 lg:mx-10 text-black">
-        <h1 class="text-xl font-bold pb-4 text-center">Batteries-Included Framework for Application Authorization</h1>
+        <h1 class="text-2xl font-bold pb-4 text-center">Batteries-Included Framework for Application Authorization</h1>
```
gkaemmer, line 5: Can we also drop the `text-center` here?

I think there's not much of a reason for this to be centered on the page, and gives it an appearance of being arbitrarily indented on bigger screens.

dhatch: ya, Stephie had the same suggestion

gkaemmer (approved):

PR #1224, opened 2021-09-23
Disable debug mode with POLAR_LOG=0 or POLAR_LOG=off

### LINKED ISSUE #1121: POLAR_LOG=0 should disable tracing
The documentation indicates that POLAR_LOG=1 enables tracing. Intuitively you would think that POLAR_LOG=0 disables tracing, particularly if you use the setting in a dotfile/docker-compose.yaml etc.

However, it seems that POLAR_LOG just needs to be set to anything to enable tracing, and to be unset to disable tracing. I've tested this with the go, node, python and rust libraries, and it's the same with all four:

Here's an example using the node library:

```
❯ which oso
/Users/alexhafner/.volta/bin/oso
❯ echo $POLAR_LOG
0
❯ oso
query> 1+2=3
[debug]   QUERY: 1 + 2 = _op_1 and _op_1 = 3, BINDINGS: {}
[debug]     QUERY: 1 + 2 = _op_1, BINDINGS: {}
[debug]     QUERY: _op_1 = 3, BINDINGS: {_op_1 = 3}
true
query> %
❯ unset POLAR_LOG
❯ oso
query> 1+2=3
true
query> %
❯ export POLAR_LOG=foo
❯ oso
query> 1+2=3
[debug]   QUERY: 1 + 2 = _op_1 and _op_1 = 3, BINDINGS: {}
[debug]     QUERY: 1 + 2 = _op_1, BINDINGS: {}
[debug]     QUERY: _op_1 = 3, BINDINGS: {_op_1 = 3}
true
query>
```

This was tested on main while reviewing the changes from https://github.com/osohq/oso/pull/1115

It would be helpful for developers if tracing could be disabled by setting POLAR_LOG=0.

alexhafner:
This PR allows the user to disable POLAR_LOG without having to unset and resolves #1121. Only in Python, a message "Polar tracing enabled. Get help..." is currently being displayed when debug mode is active. The logic there has been updated as well. I have tested expected behaviour with https://github.com/appsmode/oso-docker-polyglot/blob/main/examples/python/test_polar_log_0.py which logs into the REPLs in each language and tests a number of scenarios as to whether [debug] is in the output.

PR checklist:

- [x] Added changelog entry.


docs/content/any/project/changelogs/2021-09-29.md
```diff
@@ -14,6 +14,7 @@ description: >-
 #### Other bugs & improvements
 - Unification constraints on dot properties are no longer elided when emitting 
   partial results for queries over multiple unbound variables.
+- Debug mode can now be disabled by setting `POLAR_LOG=0` or `POLAR_LOG=off`.
```
gj, line 17:
```suggestion
- Debug mode can now be disabled by setting `POLAR_LOG=0` or `POLAR_LOG=off`. Thanks to [`@alexhafner`](https://github.com/alexhafner) for the contribution!
```

gj (approved):
Tested locally; thanks again @alexhafner!

PR #1225, opened 2021-09-23
Data Filtering ❤️ Specializers

gw000:
- Adds comprehensive tests for data filtering with specializers to host libraries
- Adds extra tests for specializers to core
- Fixes a bug that caused disparity between specializers and dot lookups in JS library

PR checklist:

- [x] Added changelog entry.


languages/js/src/dataFiltering.test.ts
```diff
@@ -461,6 +486,31 @@ async function fixtures() {
...
+      allow(_: Foo{logs: logs},   "PtnNone",  log) if log in logs;
+      allow(_: Foo{logs: logs},   "PtnCls",   log: Log) if log in logs;
+      allow(foo: Foo{logs: logs}, "PtnDict",  log: {foo: foo}) if log in logs;
+      allow(foo: Foo{logs: logs}, "PtnPtn",   log: Log{foo: foo}) if log in logs;
+    `);
+
+    const parts = ['None', 'Cls', 'Dict', 'Ptn'];
+    for (const p1 of parts) for (const p2 of parts)
+      await checkAuthz(aFoo, p1+p2, Log, [aLog]);
+  });
```
gj: this is a good test

languages/js/src/helpers.ts
```diff
@@ -225,16 +226,15 @@ function parseExternalIsa(event: unknown): QueryEvent {
...
-    call_id: callId,
-    base_tag: baseTag,
-    path,
-    class_tag: classTag,
-  } = event;
+  const { call_id: callId, base_tag: baseTag, class_tag: classTag } = event;
+  const path: string[] = (event.path as PolarTerm[]).map(x => {
+    const val = x.value;
+    if (!isPolarStr(val)) throw new Error();
+    return val.String;
```
gj, line 233: I think this would be the only place we go from Polar -> JS w/o going through `Host#toJs`. Do you like this a lot more than changing the type of `ExternalIsaWithPath.path` to `PolarTerm[]` and then mapping those PolarTerms (all of which will be Strings) in `Host#isaWithPath`? Basically what we do here: https://github.com/osohq/oso/blob/3f45dcd9561ce148cfb79f10fccbe0a5772f36f7/languages/js/src/Host.ts#L302

gw000: No, I'm into that idea, one minute

gw000: ok this is done, good idea!

languages/js/src/Polar.test.ts
```diff
@@ -750,7 +750,7 @@ Type error: can only use \`in\` on an iterable value, this is Number(Integer(2))
     undefined.foo
   in query at line 1, column 1
     undefined.foo
-Application error: Cannot read properties of undefined (reading 'foo') at line 1, column 1`
```
gj: What Node version are you on? We run tests on 12.20.0 in CI

languages/js/src/Host.ts
```diff
@@ -359,9 +359,15 @@ export class Host implements Required<DataFilteringQueryParams> {
...
+  async isaWithPath(
+    baseTag: string,
+    path: PolarTerm[],
+    classTag: string
+  ): Promise<boolean> {
     let tag = baseTag;
-    for (const field of path) {
+    for (const fld of path) {
+      const field = await this.toJs(fld);
+      if (!isString(field)) throw new Error();
```
samscott89, line 370: Should provide an error message here.

languages/js/src/Polar.test.ts
gw000: 16.9.1 ??? I changed it to the longest common prefix of the two messages ...

languages/js/src/Host.ts
gj: Wouldn't hitting this error mean we'd broken the `isa_with_path` API in the core? I.e., I don't think users should ever see this error. I guess it isn't a bad idea to add a message for ourselves; just not something we do w/ much (any?) of our other validation code in the JS lib that's validating data that we're in control of.

gw000: @gj that's correct. I'm indifferent 🤷‍♀️

languages/js/src/dataFiltering.test.ts
```diff
@@ -461,6 +483,31 @@ async function fixtures() {
 }
 
 describe('Data filtering using typeorm/sqlite', () => {
+  test('specializers', async () => {
+    const { oso, checkAuthz, aFoo, aLog } = await fixtures();
+    oso.loadStr(`
```
gj, line 488:
```suggestion
    await oso.loadStr(`
```

languages/js/src/dataFiltering.test.ts
```diff
@@ -517,6 +564,41 @@ describe('Data filtering using typeorm/sqlite', () => {
     expect(result).toEqual(expect.arrayContaining([aFoo]));
   });
 
+  test('a roles policy', async () => {
+    const { oso, checkAuthz, aFoo, anotherFoo, helloBar } = await fixtures();
+    oso.loadStr(`
```
gj, line 569:
```suggestion
    await oso.loadStr(`
```

languages/js/src/dataFiltering.test.ts
```diff
@@ -606,7 +688,8 @@ resource Issue {
   "read" if "reader" on "parent";
 }
 
-has_relation(repo: Repo, "parent", issue: Issue) if repo = issue.repo;
+#has_relation(repo: Repo, "parent", issue: Issue) if issue.repo = repo;
+has_relation(repo: Repo, "parent", _issue: Issue{repo:repo});
```
gj, line 692: Delete one of these two? I assume the former matches the other language tests and the latter tests the bug you fixed in this PR?

polar-core/tests/integration_tests.rs
```diff
@@ -1675,6 +1675,83 @@ fn test_circular_data() -> TestResult {
     Ok(())
 }
 
+fn commute_ops(u: &Value, v: &Value) -> bool {
```
gj, line 1678: Pls add a comment explaining what this does / how it works for us mere mortals. I'm having trouble following the control flow

gj (changes requested):

languages/js/src/dataFiltering.test.ts
gw000: 😱

gj (approved):

PR #1226, opened 2021-09-24
Zip artifacts to speed up Docs Diff workflow

gj:
Docs Diff workflow currently takes ~7-10 minutes, but ~3 of those minutes are spent uploading (x2, in parallel) and downloading (x2, in sequence) ~1300 files.

## Before

image

## After

image


dhatch (approved):
nice

PR #1227, opened 2021-09-24
ESLint Part 2: Now With More Linting

gj:


languages/js/src/Query.ts
```diff
@@ -1,13 +1,15 @@
-import type { Query as FfiQuery } from './polar_wasm_api';
+// TODO(gj): Make sure this is cool w/ Oso in the browser.
```
gj, line 1:
```suggestion
```

I tested this by updating `package.json` in https://github.com/osohq/oso-demo-widget as follows:

```diff
diff --git a/package.json b/package.json
index f262ffe..b7bfdf7 100644
--- a/package.json
+++ b/package.json
@@ -18 +18 @@
-    "oso": "^0.20.0-beta",
+    "oso": "../oso/main/languages/js",
```

and then following the steps in the README: https://github.com/osohq/oso-demo-widget/blob/d58d3335ddff7d297e0160aa84fec5aae23675af/README.md#L9-L13

cc: @gkaemmer

languages/js/src/Query.ts
```diff
@@ -135,22 +139,27 @@ export class Query {
     // have a nice hook where we know every class has been registered
     // (e.g., once we enforce that all registerCalls() have to happen
     // before loadFiles()).
-    const typ = this.#host.getType(rel.otherType)!;
+    const typ = this.#host.getType(rel.otherType);
+    if (typ === undefined) throw new UnregisteredClassError(rel.otherType);
+
+    if (!isObj(receiver))
```
gj, line 145: Think I'm using `isObj()` a bit too liberally, potentially leading to false negatives. Think I really want to check if the value is _indexable_; I don't particularly care that it responds to `typeof` with `'object'`.

polar-wasm-api/src/query.rs
```diff
@@ -73,9 +73,8 @@ impl Query {
     }
 
     #[wasm_bindgen(js_class = Query, js_name = source)]
-    pub fn wasm_source(&self) -> JsResult<JsValue> {
-        let source = self.0.source_info();
-        serde_wasm_bindgen::to_value(&source).map_err(|e| serialization_error(e.to_string()))
+    pub fn wasm_source(&self) -> String {
```
gj, line 76: This cleanup is unrelated to the rest of the PR; just noticed that the type of `Query.source()` was `any` instead of `string` and wondered why.

languages/js/src/helpers.ts
```diff
@@ -1,6 +1,6 @@
 import { inspect } from 'util';
-
-const _readFile = require('fs')?.readFile;
+// TODO(gj): verify this doesn't break things when using Oso in the browser.
```
gj, line 2: https://github.com/osohq/oso/pull/1227#discussion_r715787835

```suggestion
```

languages/js/src/helpers.test.ts
```diff
@@ -0,0 +1,29 @@
+import { ancestors } from './helpers';
```
gj, line 1: Not really related to this PR; just noticed we didn't have any tests for `ancestors()` as I was fiddling with the types on it

languages/js/src/dataFiltering.ts
```diff
@@ -119,22 +120,34 @@ function groundFilter(results: Map<number, unknown[]>, filter: Filter): Filter {
...
-  let value = results.get(ref.resultId);
-  value = !ref.field ? value : value?.map(v => (v as any)[ref.field!]); // eslint-disable-line @typescript-eslint/no-explicit-any
+  const { field, resultId } = ref;
+  let value = results.get(resultId);
+
+  if (field !== undefined) {
+    value = value?.map(v => {
+      // TODO(gj): what should we do here if `v` isn't indexable? Without
+      // handling the case where `v` can't be indexed by `field`, it'll blow up
+      // at runtime.
```
gj, line 130: @gw000 any thoughts?

languages/js/src/Oso.test.ts
```diff
@@ -3,7 +3,7 @@ import { map, query } from '../test/helpers';
 
 test('#isAllowed', async () => {
   const o = new Oso();
-  o.loadStr('allow(1, 2, 3);');
+  await o.loadStr('allow(1, 2, 3);');
```
gw000, line 6: was this hypothetically a race condition before? did the test ever fail?

gj: I'm not sure I've ever seen this particular one fail, but yeah unless Jest has some wild mechanism that subverts the JS event loop then this (and any of the other places in the test suite where we were missing `await`s) could've raced

languages/js/src/dataFiltering.ts
gw000: Undefined behavior I'd say. If this happens it's because something is wrong, either with the data filtering configuration or with the user's ORM layer or whatever. Blowing up at runtime sounds about right to me.

languages/js/src/enforcement.test.ts
```diff
@@ -3,7 +3,7 @@ import { BaseActor, User, Widget } from '../test/classes';
 import { ForbiddenError, NotFoundError, OsoError } from './errors';
 
 describe(Oso, () => {
-  let oso: Oso<BaseActor, String>;
+  let oso: Oso<BaseActor, string>;
```
gw000, line 6: what effect does this have?

languages/js/src/types.ts
```diff
@@ -612,10 +609,7 @@ export function isIterable(x: unknown): x is Iterable<unknown> {
  * @internal
  */
 export function isAsyncIterable(x: unknown): x is AsyncIterable<unknown> {
-  return (
-    Symbol.asyncIterator in Object(x) &&
-    typeof Object(x)[Symbol.asyncIterator] === 'function'
-  );
+  return isObj(x) && Symbol.asyncIterator in x;
```
gw000, line 612: do we need to check `typeof x[Symbol.asyncIterator] === 'function'`?

languages/js/src/types.ts
```diff
@@ -600,10 +600,7 @@ export function isIterableIterator(x: unknown): x is IterableIterator<unknown> {
  * @internal
  */
 export function isIterable(x: unknown): x is Iterable<unknown> {
-  return (
-    Symbol.iterator in Object(x) &&
-    typeof Object(x)[Symbol.iterator] === 'function'
-  );
+  return isObj(x) && Symbol.iterator in x;
```
gw000, line 603: same as `asyncIterator` below

languages/js/src/dataFiltering.ts
gj: 70db50cf

languages/js/src/enforcement.test.ts
gj: Ran afoul of this rule: https://github.com/typescript-eslint/typescript-eslint/blob/9b13a86025860c71c4d805178833ec0902320d06/packages/eslint-plugin/docs/rules/ban-types.md

Looks like [the relevant guidance](https://github.com/typescript-eslint/typescript-eslint/blob/9b13a86025860c71c4d805178833ec0902320d06/packages/eslint-plugin/docs/rules/ban-types.md#default-options) is:

> Don't use the upper-case primitive types, you should use the lower-case types for consistency.

languages/js/src/types.ts
gj: I'm not sure... the only reason it wouldn't be a function is if someone fiddled around with it manually. If they fiddled with it, then the runtime error would be `TypeError: x is not iterable` if we tried to use it in a `for...of` or `for await...of` loop or `yield` from it. I'm kind of inclined to let it blow up since there's only so much validation we can be expected to do. What do you think?

languages/js/src/types.ts
gj: Responded in https://github.com/osohq/oso/pull/1227#discussion_r715900928; let's discuss there.

languages/js/src/types.ts
gj: Updated these type guards a bit in 959afd09, but the update didn't address your question.

languages/js/src/Query.ts
gj: c83d92a8

languages/js/src/Host.ts
```diff
@@ -317,16 +315,27 @@ export class Host implements Required<DataFilteringQueryParams> {
...
-    } else {
-      return leftIndex < rightIndex;
+    try {
+      // NOTE(gj): TS is worried about looking up the `.constructor` property
+      // on an `unknown` type. `instance.constructor` will throw a `TypeError`
+      // if `instance` is `null` or `undefined`, but we catch that error down
+      // below and return `false`.
+      //
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
```
gw000, line 325: just want to make sure I understand ... this is a comment telling the linter not to care about a comment telling the type checker not to care about the next line?

gw000: Can you get the same effect with
```javascript
if (instance === null || instance === undefined) throw new Error();
```

gw000 (approved):
Looks ok, the only thing is I'm not a huge fan of the
```javascript
// linter: ignore the next line
// typescript: ignore the next line
```
comments. Maybe there's a better way? Not a huge deal though.

languages/js/src/Host.ts
gj: > just want to make sure I understand ... this is a comment telling the linter not to care about a comment telling the type checker not to care about the next line?

Yeah, that's it. We have [this typescript-eslint rule](https://github.com/typescript-eslint/typescript-eslint/blob/02c6ff3c5a558f9308d7166d524156dc12e32759/packages/eslint-plugin/docs/rules/ban-ts-comment.md) enabled.

> Can you get the same effect with
> 
> ```js
> if (instance === null || instance === undefined) throw new Error();
> ```

I did something similar to this in 2725ad6. What do you think?

gj: Got confirmation that this is 👍

PR #1233, opened 2021-09-27
Ensure all guides are tested against Oso 0.20.1

gj:
### TODO

- [x] Update Rust quickstart once https://github.com/osohq/oso-rust-quickstart/pull/4 ships
- [x] Update Java quickstart submodule once https://github.com/osohq/oso-java-quickstart/pull/3 ships
- [x] Align Oso version constraints across all guides? Might want to make them all exact version matches instead of, e.g., `>=version` or `^version` or `~> version`.
  - [ ] Add non-submodule example Oso versions to `scripts/bump_versions.py`?
  - [ ] Add quickstart Oso versions to `scripts/bump_versions.py`? Is this possible w/ submodules?


dhatch:
> -  Align Oso version constraints across all guides? Might want to make them all exact version matches instead of, e.g., >=version or ^version or ~> version.

Exact matches sounds right

> -  Add non-submodule example Oso versions to scripts/bump_versions.py?

Good idea.


> -  Add quickstart Oso versions to scripts/bump_versions.py? Is this possible w/ submodules?

Also good idea, we should be able to do this with submodules, but it would require them being checked out, and then you'd need to go into each submodule and make a commit to bump the version and push that to the submodule remote. That could be scripted probably.

dhatch (approved):
lgtm

gj:
> > * Align Oso version constraints across all guides? Might want to make them all exact version matches instead of, e.g., >=version or ^version or ~> version.
> 
> Exact matches sounds right

Did this

> > * Add non-submodule example Oso versions to scripts/bump_versions.py?
> 
> Good idea.
> 
> > * Add quickstart Oso versions to scripts/bump_versions.py? Is this possible w/ submodules?
> 
> Also good idea, we should be able to do this with submodules, but it would require them being checked out, and then you'd need to go into each submodule and make a commit to bump the version and push that to the submodule remote. That could be scripted probably.

Created a follow-up task for updating `bump_versions.py`

PR #1237, opened 2021-09-28
Release 0.21.0

gj:


docs/content/any/project/changelogs/2021-09-29.md
```diff
@@ -66,24 +69,25 @@ const user = {name: "notanadmin"};
...
 ```
-since specializers will not error if a property doesn't exist (but the `{name: "notanadmin"}`
-object will fail to match the specializer and the `isAllowed()` call will still correctly
-return `false`).
+
+since specializers will not error if a property doesn't exist (but the `{name:
+"notanadmin"}` object will fail to match the specializer and the `isAllowed()`
+call will still correctly return `false`).
 
 ##### Optional `types` parameter to `registerClass()` renamed to `fields`
```
gw000, line 92: We also made this change in the Python library, we need to call that out too. Also maybe move this one up to the top of the list, since it's relatively short & otherwise folks might miss it

gw000 (changes requested):
Changelog needs an update, otherwise 👍

docs/content/any/project/changelogs/2021-09-29.md
gj: Good catch — gonna do this as part of a follow-up PR that adds a couple small changelog entries.

PR #1245, opened 2021-10-01
Warn when loading a policy that does not contain an allow rule

gkaemmer:
PR checklist:
- [ ] Added changelog entry.
Screen Shot 2021-10-01 at 10 15 36 AM

### DIFF
--- a/polar-core/src/polar.rs
+++ b/polar-core/src/polar.rs
@@ -9,7 +9,7 @@ use super::runnable::Runnable;
 use super::sources::*;
 use super::terms::*;
 use super::vm::*;
-use super::warnings::{check_ambiguous_precedence, check_singletons};
+use super::warnings::{check_ambiguous_precedence, check_singletons, full_policy_warnings};
 
 use std::sync::{Arc, RwLock};
 
@@ -249,6 +249,16 @@ impl Polar {
             kb.clear_rules();
             return Err(e);
         }
+
+        // TODO(@gkaemmer) is this the right place for this?
+        let mut warnings = vec![];
+        warnings.append(&mut full_policy_warnings(&*kb)?);
+        let warnings = warnings.into_iter().map(|msg| Message {
+            kind: MessageKind::Warning,
+            msg,
+        });
+        self.messages.extend(warnings);
+
         Ok(())
     }
 
--- a/polar-core/src/polar.rs
+++ b/polar-core/src/polar.rs
@@ -9,7 +9,7 @@ use super::runnable::Runnable;
 use super::sources::*;
 use super::terms::*;
 use super::vm::*;
-use super::warnings::{check_ambiguous_precedence, check_singletons, full_policy_warnings};
+use super::warnings::{check_ambiguous_precedence, check_no_allow_rule, check_singletons};
 
 use std::sync::{Arc, RwLock};
 
@@ -252,7 +252,7 @@ impl Polar {
 
         // TODO(@gkaemmer) is this the right place for this?
         let mut warnings = vec![];
-        warnings.append(&mut full_policy_warnings(&*kb)?);
+        warnings.append(&mut check_no_allow_rule(&*kb)?);
         let warnings = warnings.into_iter().map(|msg| Message {
             kind: MessageKind::Warning,
             msg,
--- a/polar-core/src/warnings.rs
+++ b/polar-core/src/warnings.rs
@@ -204,3 +204,27 @@ pub fn check_ambiguous_precedence(rule: &Rule, kb: &KnowledgeBase) -> PolarResul
     walk_rule(&mut visitor, rule);
     visitor.warnings()
 }
+
+pub fn full_policy_warnings(kb: &KnowledgeBase) -> PolarResult<Vec<String>> {
+    let has_allow_rule = kb
+        .get_rules()
+        .iter()
+        .find(|(name, _)| name.0 == "allow")
+        .is_some();
+    if has_allow_rule {
+        return Ok(vec![]);
+    } else {
+        return Ok(vec![
+            "Your policy does not contain an allow rule, which means \
+that no actions are allowed. Did you mean to add an allow rule to \
+the top of your policy?
+
+  allow(actor, action, resource) if ...
+
+For more information about allow rules, see:
+
+  https://docs.osohq.com/reference/polar/builtin_rule_types.html#allow"
+                .to_string(),
+        ]);
+    }
+}
--- a/polar-core/src/warnings.rs
+++ b/polar-core/src/warnings.rs
@@ -205,7 +205,7 @@ pub fn check_ambiguous_precedence(rule: &Rule, kb: &KnowledgeBase) -> PolarResul
     visitor.warnings()
 }
 
-pub fn full_policy_warnings(kb: &KnowledgeBase) -> PolarResult<Vec<String>> {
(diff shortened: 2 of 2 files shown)


polar-core/src/polar.rs
```diff
@@ -249,6 +249,16 @@ impl Polar {
             kb.clear_rules();
             return Err(e);
         }
+
+        // TODO(@gkaemmer) is this the right place for this?
```
gkaemmer, line 253: I can't think of a better place really, but at some point we should probably pull our errors/warnings into a shared pipeline and data structure

polar-core/src/warnings.rs
```diff
@@ -204,3 +204,53 @@ pub fn check_ambiguous_precedence(rule: &Rule, kb: &KnowledgeBase) -> PolarResul
     walk_rule(&mut visitor, rule);
     visitor.warnings()
 }
+
+pub fn check_no_allow_rule(kb: &KnowledgeBase) -> PolarResult<Vec<String>> {
+    let has_allow_rule = kb.get_rules().iter().any(|(name, _)| name.0 == "allow");
```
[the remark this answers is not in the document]
gj: or

```suggestion
    let has_allow_rule = kb.get_generic_rule(&sym!("allow")).is_some();
```

or

```suggestion
    let has_allow_rule = kb.get_rules().contains_key(&sym!("allow"));
```

a lot of ways to skin this cat lol

polar-core/src/warnings.rs
```diff
@@ -204,3 +204,53 @@ pub fn check_ambiguous_precedence(rule: &Rule, kb: &KnowledgeBase) -> PolarResul
...
+    if has_allow_rule {
+        Ok(vec![])
+    } else {
+        Ok(vec![
+            "Your policy does not contain an allow rule, which means \
+that no actions are allowed. Did you mean to add an allow rule to \
```
gj, line 215: What if someone intentionally isn't using `is_allowed/3` or `authorize/3` in their policy?

Case 1: using `query_rule()` to query a non-`allow` entrypoint rule.

Case 2: using `authorize_request/2` instead of `authorize/3`, so `allow_request/2` is the entrypoint rule instead of `allow/3`. Same thing for folks authorizing fields instead of requests.

polar-core/tests/integration_tests.rs
```diff
@@ -1600,21 +1600,29 @@ fn test_and_or_warning() -> TestResult {
...
-        "Expression without parentheses could be ambiguous. \nPrior to 0.20, `x and y or z` would parse as `x and (y or z)`. \nAs of 0.20, it parses as `(x and y) or z`, matching other languages. \n\n\n001: f(x) if x = 1 and x > 1 or x < 3;\n             ^"
-    );
+    let mut messages = vec![];
+    while let Some(msg) = p.next_message() {
+        messages.push(msg);
+    }
+    assert!(messages.iter().any(|msg| {
+      matches!(&msg.kind, MessageKind::Warning) &&
+      (msg.msg ==
+        "Expression without parentheses could be ambiguous. \nPrior to 0.20, `x and y or z` would parse as `x and (y or z)`. \nAs of 0.20, it parses as `(x and y) or z`, matching other languages. \n\n\n001: f(x) if x = 1 and x > 1 or x < 3;\n             ^")
```
gj, line 1610: nit: Might be mistaken, but I think `assert_eq!()` on a pair of strings (what we used to have here) will give a better error message than this equality check inside an `.any()` inside an `assert!()`

polar-core/src/polar.rs
```diff
@@ -249,6 +249,16 @@ impl Polar {
...
+        let mut warnings = vec![];
+        warnings.append(&mut check_no_allow_rule(&*kb)?);
+        let warnings = warnings.into_iter().map(|msg| Message {
+            kind: MessageKind::Warning,
+            msg,
+        });
```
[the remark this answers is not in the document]
gkaemmer: I like the idea of `Message::warning` -- I think I'll add that.

On the second comment, I think I'll keep the method signature as-is for now. Right now, our warning methods (in `warnings.rs`) all return `PolarResult<Vec<String>>`, so this would probably mean changing all of them. That's fine, but IMO it'd be even more ideal if these methods would each return a list of specific `Warning` or `Error` objects, which the caller would then gather up and convert into messages. A couple benefits to this approach:
- I imagine it'd be easy to "promote" warnings to errors (like we did with singleton warnings -> errors)
- With a standard for warning types, users could tell Oso to ignore certain warnings
- We could use the same pipeline for validating policies during `load_files` and also in an LSP context, where the output format is different (likely not using the `Message` type).

Anyway, I feel like there's a chunk of refactoring work here to do, but don't want to do it in this change. For now, I think keeping the same method signature as the other `check_*` methods (`PolarResult<Vec<String>>`) is the simplest thing to do.

polar-core/src/warnings.rs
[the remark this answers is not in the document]
gkaemmer: Haha this always happens when I write rust -- I had been using an even longer variant (`iter().find(...).is_some()`), but clippy had me replace it with `iter().any(...)`

I think I'll go with `contains_key`

polar-core/src/warnings.rs
gkaemmer: Hmm yeah this is a good point. I don't think it's the end of the world to have a bit of a noisy warning in that case, but it would be great if users could tell us to ignore specific warnings (see my other comment). I think right now making this warning ignore-able would introduce more complexity than its worth (probably would have to use a random environment variable). 

To make this warning more "semantically correct", perhaps I could add a "usually" to the warning message?
```suggestion
            "Your policy does not contain an allow rule, which usually means \
that no actions are allowed. Did you mean to add an allow rule to \
```

polar-core/tests/integration_tests.rs
gkaemmer: Yeah that's definitely true. Can you think of a more idiomatic way to assert that there is "at least one thing in this list meeting this condition"?

polar-core/src/polar.rs
gkaemmer: Yeah, that's kinda why I was thinking this was the best place for now. If there are other errors during resource block rewriting/validation, then this check will not be performed. That being said, I think we do always run `check_messages` regardless of whether or not the FFI call returns an error response.

On your last question, I added some thoughts to another comment that might address it.

polar-core/src/warnings.rs
gj: What do you think about updating the check to look for `allow` or `allow_request` or `allow_field`?

polar-core/src/polar.rs
```diff
@@ -249,6 +246,13 @@ impl Polar {
             kb.clear_rules();
             return Err(e);
         }
+
+        // TODO(@gkaemmer) is this the right place for this?
+        let mut warnings = vec![];
+        warnings.append(&mut check_no_allow_rule(&*kb)?);
```
gj, line 252: I think this deref will happen automatically

```suggestion
        warnings.append(&mut check_no_allow_rule(&kb)?);
```

polar-core/src/warnings.rs
gkaemmer: Good idea, updated

gkaemmer: Note that I still made the warning message primarily about `allow` -- I feel like that's the common thing to point people towards, but it still mentions `allow_request` and `allow_field` for correctness

gj (approved):

PR #1249, opened 2021-10-04
Trim trailing newline from polar code snippets

gkaemmer:
Before:
Screen Shot 2021-10-04 at 11 54 34 AM

After:
Screen Shot 2021-10-04 at 11 54 45 AM


gj (approved):
seems legit

PR #1250, opened 2021-10-04
Don't embed c runtime library in static lib

saolsen:
Currently the musl runtime is embedded in our static lib for linux. This works if the kernel it's run on is compatible with the musl version but doesn't work otherwise.
Now we allow the application to link dynamically with the c runtime library and that should work on more platforms.


.github/workflows/release.yml
```diff
@@ -5,6 +5,7 @@ on:
       - "v*" # Push events to matching v*, i.e. v1.0, v20.15.10
     branches:
       - main
+      - steve/gooooo
```
samscott89, line 8:
```suggestion
```
Reminder to delete before merge

samscott89 (approved):

docs/content/any/project/changelogs/NEXT.md
```diff
@@ -11,6 +11,15 @@ draft: true
 
 ## `RELEASED_PACKAGE_1` NEW_VERSION
 
+### Core
+
+#### Other bugs & improvements
+
+- Fixed the way we build our static library on Linux so it doesn't embed
+  musl and instead links to the system c runtime library.
+  Language that depend on the static lib Linux build such as python and go
```
gj, line 20:
```suggestion
  Languages that depend on the static lib Linux build such as python and go
```

.github/workflows/release.yml
```diff
@@ -497,6 +497,10 @@ jobs:
         if: runner.os == 'Linux' && matrix.go-version != '1.12'
         run: docker run --rm -v `pwd`:/oso tetafro/golang-gcc:${{ matrix.go-version }}-alpine /bin/sh -c 'cd /oso && go build && go test -v ./tests/'
         working-directory: go
+      - name: "test buster"
+        if: runner.os == 'Linux' && matrix.go-version != '1.12'
```
gj, line 501: Do we need to skip 1.12 here?

.github/workflows/release.yml
```diff
@@ -497,6 +497,10 @@ jobs:
...
+        run: docker run --rm -v `pwd`:/oso golang:${{ matrix.go-version }}-buster /bin/sh -c 'cd /oso && go build && go test -v ./tests/'
```
gj, line 502: Is `go build` necessary before `go test`?

PR #1255, opened 2021-10-06
Use `_.isEqual` as default equality function

2178309:
This pulls in `lodash.isequal` as a dependency which we might not want, however, the current default equality behavior needs to change:
Screen Shot 2021-10-06 at 1 53 52 PM

PR checklist:

- [x] Added changelog entry.

### DIFF
--- a/docs/content/any/project/changelogs/NEXT.md
+++ b/docs/content/any/project/changelogs/NEXT.md
@@ -21,6 +21,8 @@ draft: true
   should support more platforms now.
 - Oso will now issue a warning if there is no `allow` rule in your policy (and
   also no `allow_request` or `allow_field` rules).
+- Oso will propose a suggested fix if you forget to write an actor block when
+  using resource blocks.
 
 ### Node.js
 
--- a/.github/workflows/cla.yml
+++ b/.github/workflows/cla.yml
@@ -22,7 +22,7 @@ jobs:
           remote-repository-name: cla-signatures
           path-to-signatures: individual.json
           branch: main
-          allowlist: dhatch,gj,gkaemmer,gneray,leina05,litonico,samscott89,saolsen,ssglaser
+          allowlist: dhatch,gj,gkaemmer,gneray,leina05,litonico,samscott89,saolsen,ssglaser,patrickod,gw000
           use-dco-flag: false
           create-file-commit-message: 'Creating file for storing CLA signatures'
           signed-commit-message: '$contributorName has signed the CLA in osohq/oso#$pullRequestNo'
--- a/polar-core/src/kb.rs
+++ b/polar-core/src/kb.rs
@@ -265,7 +265,16 @@ impl KnowledgeBase {
                             }
                         }
                         if !success {
-                            return Ok(RuleParamMatch::False(format!("Rule specializer {} on parameter {} must be a member of rule type specializer {}", rule_instance.tag,index, rule_type_instance.tag)));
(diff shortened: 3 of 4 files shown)


docs/content/any/project/changelogs/NEXT.md
```diff
@@ -32,6 +32,8 @@ draft: true
   Many thanks to [`@rradczewski`](https://github.com/rradczewski) for
   [raising](https://github.com/osohq/oso/issues/1242) and reproducing
   the issue, and confirming the fix!
+- Oso now defaults to using Lodash's `isEqual` function compare JavaScript values
```
gj, line 35:
```suggestion
- Oso now defaults to using Lodash's `isEqual` function when comparing JavaScript values
```

languages/js/package.json
```diff
@@ -39,6 +39,8 @@
   },
   "devDependencies": {
     "@types/jest": "^26.0.9",
+    "@types/lodash": "^4.14.175",
```
gj, line 42: why do we need this?

languages/js/src/helpers.ts
```diff
@@ -9,6 +9,7 @@ import {
   QueryEventKind,
 } from './types';
 import type { Class, obj, PolarTerm, QueryEvent } from './types';
+import isEqual = require('lodash.isequal');
```
gj, line 12: I don't think I've ever seen an `import ... = require(...)` statement — what's the difference between this and `import ... from ...`?

gj (changes requested):
Except for the addition of `lodash.isequal`, the `yarn.lock` changes look to be almost entirely reorderings w/ no difference in package versions or SHAs. It's an unfortunate amount of noise for this otherwise pretty small PR. Do you know whether that's down to using a different version of Yarn, perhaps? If that's the case, it would be great to standardize on a Yarn version to avoid this going forward.

(the above isn't a blocking concern — I'm requesting changes for reasons laid out in other comments)

languages/js/src/helpers.ts
gw000: I'd never seen it either ... according to error messages from node we would need to add an config option to use the normal syntax with `lodash.isequal` for some reason.

languages/js/package.json
gw000: can't import from an external package without it.

gw000:
can we even just add `yarn.lock` to the gitignore?

gj:
> can we even just add `yarn.lock` to the gitignore?

I think it's supposed to be in version control (even for a library) to ensure that all of us are developing on the library w/ the same dependency versions for reproduce-ability of behavior and whatnot.

From some recent poking around I did, it seems like we should migrate to Yarn 2/3 at some point, and that might help the problem since it seems to use a project-local Yarn w/ a specific version instead of having every developer install Yarn globally.

languages/js/package.json
gj: like we need the base `@types/lodash` in order to import `@types/lodash.isequal`?

languages/js/src/helpers.ts
gkaemmer: Did you try `import * as isEqual from "lodash.isequal"`? I think that might be a more idiomatic way of doing this

languages/js/package.json
gj: ah I see:
```
+  dependencies:
+    "@types/lodash" "*"
```

gj (approved):

gj (changes requested):
need to update `EqualityFn` jsdoc comment in src/types.ts

PR #1281, opened 2021-10-13
Fix recursive relations in resource blocks

leina05:
PR checklist:

- [ ] Added changelog entry.
- [ ] Probably want to move tests to the core, but recursive types in Rust are annoying/confusing

### DIFF
--- a/languages/go/tests/unit_test.go
+++ b/languages/go/tests/unit_test.go
@@ -242,20 +242,6 @@ type Foo struct {
 	Num  int
 }
 
-type RecursiveType struct {
-	Parent *RecursiveType
-	User User
-}
-
-func (r RecursiveType) GetParent() RecursiveType {
-	if r.Parent != nil {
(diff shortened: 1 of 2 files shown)


gj:
> Probably want to move tests to the core, but recursive types in Rust are annoying/confusing

I don't think we need an actual recursive test in the core (or in any of the host languages). I added a test for the rewrite behavior (the source of the bug) in e17791a. What do you think?

gw000:
> > Looks fine. We should make a task to move the Go tests to the core. One of the python tests is failing, lmk after it's fixed & the changelog is added & I'll approve
> 
> I think we can remove the Go & Python tests before shipping this PR. Relevant: [#1281 (comment)](https://github.com/osohq/oso/pull/1281#issuecomment-944326121)

Sounds chill to me 😄 Any other potential cases with name collisions in autogenerated rules that we might want to preemptively fix while we're here?

PR #1285, opened 2021-10-14
Add suggested type specializer for rules without sufficient type constraints

patrickod:
PR checklist:
- [x] Added changelog entry.


leina05:
One thing: it might be easier to add a quick addition to the changelog now, before we forget all the changes.

PR #1302, opened 2021-10-20
Support multiple resource blocks for the same resource

gj:
PR checklist:

- [x] Added changelog entry.


polar-core/src/error.rs
```diff
@@ -460,6 +464,17 @@ impl fmt::Display for ValidationError {
             Self::UnregisteredClass { term } => {
                 write!(f, "Unregistered class: {}", term)
             }
+            Self::DuplicateShorthandRule {
+                resource,
+                shorthand_rule,
+            } => {
+                write!(
+                    f,
+                    "Duplicate shorthand rule `{}` for resource {}",
```
patrickod, line 473:
```suggestion
                    "Duplicate shorthand rule `{}` declared for resource {}",
```

polar-core/src/resource_block.rs
```diff
@@ -872,15 +903,197 @@ mod tests {
...
+              roles = ["reader", "writer"];
+
+              "reader" if "writer";
+            }
+        "#;
         expect_error(
             &p,
             invalid_policy,
-            "Duplicate declaration of 'Org' resource block.",
+            r#"Duplicate shorthand rule `"reader" if "writer";` for resource Repo"#,
```
patrickod, line 1096:
```suggestion
            r#"Duplicate shorthand rule `"reader" if "writer";` declared for resource Repo"#,
```

polar-core/src/resource_block.rs
```diff
@@ -298,18 +300,43 @@ impl ResourceBlocks {
         resource: Term,
         declarations: Declarations,
         shorthand_rules: Vec<ShorthandRule>,
-    ) {
-        self.declarations.insert(resource.clone(), declarations);
-        self.shorthand_rules
-            .insert(resource.clone(), shorthand_rules);
+    ) -> PolarResult<()> {
+        // Check for duplicate existing shorthand rules and return a
+        // ValidationError if we discover one
```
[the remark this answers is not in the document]
gj: Talked offline — we're gonna punt on the duplication check for this PR since it'd be best to implement it as a policy-wide check (i.e., not specific to the resource blocks feature).

polar-core/src/resource_block.rs
```diff
@@ -298,18 +300,43 @@ impl ResourceBlocks {
...
+                        shorthand_rule: rule.clone(),
+                    }
+                    .into());
+                }
+            }
+        }
+
+        // Merge existing declarations if we are reopening a resource block, otherwise add new
+        if let Some(existing) = self.declarations.get_mut(&resource) {
+            existing.extend(declarations.into_iter());
```
dhatch, line 321: I don't know this code very well, but when you were presenting was thinking about this - 

What happens if I declare the same term twice?

```
resource Org {
    roles = ["foo"];
}

resource Org {
    permissions = ["foo"];
}
```

I don't think this would be caught with the current code -- seems like this could lead to bugs.

polar-core/src/resource_block.rs
```diff
@@ -909,7 +1123,7 @@ mod tests {
                 roles=["member"];
                 "member" if "owner";
             }"#,
```
dhatch, line 1125: Think about edge cases more in these tests:

conflicting declarations (`actor Org`, `resource Org`) - seems like fine from what you said but good to test
conflicting role / permission `roles = ["foo"];`, `permissions = ["foo"];`

polar-core/src/resource_block.rs
patrickod: good shout. will add a check for this & corresponding unit test

polar-core/src/resource_block.rs
patrickod: updated the merging code to only permit overwriting declarations if they are of the same type, otherwise we will raise a validation error.

polar-core/src/error.rs
```diff
@@ -428,6 +431,17 @@ pub enum ValidationError {
         /// Term<Symbol> where the error arose, tracked for lexical context.
         term: Term,
     },
+    DuplicateShorthandRule {
+        resource: Term,
+        shorthand_rule: ShorthandRule,
+    },
```
gj, line 437: I thought we were punting on this for now?

polar-core/src/error.rs
```diff
@@ -3,6 +3,9 @@ use std::fmt;
 use indoc::formatdoc;
 use serde::{Deserialize, Serialize};
 
+use crate::resource_block::{Declaration, ShorthandRule};
+use crate::terms::ToPolarString;
```
gj, line 7: should go into the `use super::{` block below

polar-core/src/error.rs
patrickod: you are correct. I misinterpreted the "punt" in the quill thread. just backed the shorthand rule specific deduplication out.

polar-core/src/error.rs
patrickod: 👍🏻

polar-core/src/resource_block.rs
```diff
@@ -167,15 +170,31 @@ pub fn resource_block_from_productions(
...
+            }
+            Self::Permission => {
+                write!(f, "permission")
+            }
+            Self::Relation(_) => {
+                write!(f, "relation")
+            }
+        }
+    }
+}
```
gj, line 195: this should probably go in `formatting.rs` along with (many of) the existing `Display` impls

polar-core/src/resource_block.rs
```diff
@@ -298,7 +355,7 @@ impl ResourceBlocks {
         if let Some(declaration) = maybe_declaration {
             Ok(declaration)
         } else {
-            let msg = format!("Undeclared term {} referenced in rule in the '{}' resource block. Did you mean to declare it as a role, permission, or relation?", declaration, resource_name);
+            let msg = format!("Undeclared term {} referenced in rule in '{}' resource block. Did you mean to declare it as a role, permission, or relation?", declaration.to_polar(), resource_name);
```
gj, line 358: `.to_polar()` usually (always?) defaults to the `Display` impl — you shouldn't need to add an explicit call to `.to_polar()` when you're already formatting the thing as `{}` (which uses the thing's `Display` impl)

polar-core/src/resource_block.rs
```diff
@@ -583,26 +640,22 @@ fn shorthand_rule_head_to_params(head: &Term, resource: &Term) -> Vec<Parameter>
...
-    if blocks.exists(resource) {
-        let msg = format!("Duplicate declaration of '{}' resource block.", resource);
-        let term = resource.clone();
-        return Err(ValidationError::ResourceBlock { msg, term });
-    }
-    Ok(())
-}
-
-fn check_that_shorthand_rule_heads_are_declared_locally(
+fn check_that_shorthand_rule_heads_are_declared_globally(
```
gj, line 643: This function was removed in [`8d414fb` (#1400)](https://github.com/osohq/oso/pull/1400/commits/8d414fbb087bd495e4bffb2010186cd608089e9a) b/c it was duplicative of the check in `ResourceBlocks::get_declaration_in_resource_block`

polar-core/src/resource_block.rs
```diff
@@ -167,15 +170,31 @@ pub fn resource_block_from_productions(
...
+                write!(f, "permission")
+            }
+            Self::Relation(_) => {
+                write!(f, "relation")
+            }
+        }
+    }
+}
+
+#[derive(Clone, Serialize, Deserialize, Debug, Hash, PartialEq, Eq)]
```
gj, line 197: Why do we need all these new `derive`s? Maybe leftover from when you had that other new error variant? In general adding `derive`s increases compile time, so not much reason to add them until we need them.

polar-core/src/resource_block.rs
```diff
@@ -271,18 +290,56 @@ impl ResourceBlocks {
...
+                        existing.insert(key.clone(), new.clone());
+                    }
+                    // disallow all other combinations (mismatched declaration types)
+                    _ => {
+                        return Err(ValidationError::DuplicateResourceBlockDeclaration {
+                            resource,
+                            declaration: key.clone(),
+                            existing: existing_declaration.unwrap().clone(),
+                            new: new.clone(),
+                        })
```
gj, line 320: In the spirit of surfacing multiple errors at once, it would be nice to not bomb out when encountering the first error here. Instead, we should collect errors up and then _always_ return a `Vec<ValidationError>` from `ResourceBlocks::add`

patrickod: great shout. updated the function signature here to return `Vec<ValidationError>`

polar-core/src/resource_block.rs
patrickod: `Serialize, Deserialize` are ✂️ they were necessary only for the duplicate shorthand rule error variant (now gone)

polar-core/src/resource_block.rs
patrickod: ✂️

polar-core/src/resource_block.rs
patrickod: ✂️ `.to_polar()`

polar-core/src/resource_block.rs
patrickod: moved into the `display` mod alongside the other `fmt::Display` impls

polar-core/src/resource_block.rs
```diff
@@ -280,21 +282,61 @@ impl ResourceBlocks {
...
+            for (key, new) in &declarations {
+                let existing_declaration = existing.get(key);
+                // allow
+                // 1. overwriting declarations of the **same** type
+                // 2. inserting declarations where none previously existed
+                //
+                // deny overwriting of declarations with mismatched types
+                match (new, existing_declaration) {
+                    // new or same-type declarations are accepted
+                    (Declaration::Relation(_), Some(Declaration::Relation(_)))
```
gj, line 299: Doesn't this miss the case where I've declared `relations = { a: A };` in one block and `relations = { a: B };` in another?

polar-core/src/resource_block.rs
```diff
@@ -280,21 +282,61 @@ impl ResourceBlocks {
...
+            self.declarations.insert(resource.clone(), declarations);
+        }
+
+        // Merge existing shorthand rules if we are reopening a resource block, otherwise add new
+        if let Some(existing) = self.shorthand_rules.get_mut(&resource) {
+            existing.extend(shorthand_rules.into_iter());
+        } else {
+            self.shorthand_rules
+                .insert(resource.clone(), shorthand_rules);
+        }
```
gj, line 329: This is a fine way to write this (and avoids an unnecessary `.clone()` in the "merge" case), but just FYI in case you haven't encountered [the Entry API](https://doc.rust-lang.org/std/collections/hash_map/enum.Entry.html) yet:

```rust
        self.shorthand_rules
            .entry(resource.clone())
            .and_modify(|existing| existing.extend(shorthand_rules.clone()))
            .or_insert(shorthand_rules);
```

polar-core/src/resource_block.rs
gj: This is handled in dd9dd3f4 but it'd still be good to add a test case for this

polar-core/src/resource_block.rs
patrickod: TIL about the Entry API 🎉 updated to use the chained form

polar-core/src/resource_block.rs
patrickod: Added a test case for relations of differing types in 160e72aa91ae3bbb7fb06be1d51491190775166c

polar-core/src/resource_block.rs
```diff
@@ -280,21 +282,43 @@ impl ResourceBlocks {
...
+            // or insert a new set of declarations for the resource type
+            self.declarations.insert(resource.clone(), declarations);
+        }
+
+        // Merge existing shorthand rules if we are reopening a resource block, otherwise add new
         self.shorthand_rules
-            .insert(resource.clone(), shorthand_rules);
+            .entry(resource.clone())
+            .and_modify(|existing| existing.extend(shorthand_rules.clone()))
+            .or_insert(shorthand_rules);
```
gj, line 307: Just a note that this merge might result in duplicate shorthand rules that'll get rewritten into duplicate longhand rules. That behavior's more correct than silently de-duping, but just a reminder that we'll want to circle back and error (or at least warn) on duplicate rules in a follow-up task.

polar-core/src/error.rs
```diff
@@ -439,6 +447,9 @@ impl ValidationError {
             ResourceBlock { term, .. }
             | SingletonVariable { term, .. }
             | UndefinedRuleCall { term }
+            | DuplicateResourceBlockDeclaration {
+                declaration: term, ..
```
gj, line 452: Not an issue introduced in this PR, but for 'relation' declarations the lexical context is going to point to the relation type (the value `Bar` in the `relations = { foo: Bar };` declaration; not the key `foo`).

gj (commented):
Approved ✅

patrickod (approved):

PR #1310, opened 2021-10-21
Replace standard rule types with dynamically generated ones based on resource blocks

patrickod:
traverse the set of declarations in a given knowledge base to construct
a set of unique relationship tuples and then use these to template
has_relation rules.

PR checklist:

- [ ] Added changelog entry.


gkaemmer (approved):
Will be good to revisit the error message thrown for missing required rules, but this looks good to me.

polar-core/src/debugger.rs
```diff
@@ -65,6 +65,7 @@ enum Step {
 /// [`Goal::Query`](../vm/enum.Goal.html). When either breakpoint is hit, we check the
 /// [`Debugger`](struct.Debugger.html)'s internal [`step`](struct.Debugger.html#structfield.step)
 /// field to determine how evaluation should proceed.
+#[allow(clippy::large_enum_variant)]
```
gj, line 68: Is this needed?

polar-core/src/kb.rs
```diff
@@ -757,6 +792,160 @@ impl KnowledgeBase {
...
+        }
+
+        // Iterate through resource block shorthand rules and create *required*
+        // rule types for each relation which is traversed in the rules.
+        for (object, shorthand_rules) in &self.resource_blocks.shorthand_rules {
+            for shorthand_rule in shorthand_rules {
+                // We create rule types from shorthand rules in the following scenarios:
+                // 1. when the the third "relation" term points to a related Resource
+                // 2. when the second "implier" term points to a related Resource
+                // 3. when the second "implier" term ponits to a related Actor
```
gj, line 841:
```suggestion
                // 3. when the second "implier" term points to a related Actor
```

polar-core/src/kb.rs
```diff
@@ -757,6 +792,160 @@ impl KnowledgeBase {
...
+                                        instance!(sym!("String")),
+                                        subject.value().as_symbol()?.clone(),
+                                        false,
+                                    ));
+                                    has_role_rule_types_to_create.insert((
+                                        sym!("Actor"),
+                                        instance!(sym!("String")),
+                                        subject.value().as_symbol()?.clone(),
+                                        true,
+                                    ));
```
gj, line 888: `HashSet::insert` will overwrite the previous entry — I don't think you need to call `HashSet::remove` first

gj (changes requested):

polar-core/src/kb.rs
patrickod: they're identical **except** for the last `required` argument which is being flipped here.

polar-core/src/debugger.rs
patrickod: yep the introduction of our `MissingRequiredRule { Rule }` struct caused this to balloon and clippy complain.

polar-core/src/kb.rs
gj: Oh... Hash *Set* 😆 

In that case, what about splitting it out into a `HashMap<(Symbol, Instance, Symbol), bool>`?

gj: Basically, `remove` & `insert` where everything but a `bool` remains constant feels like a code smell to me

polar-core/src/error.rs
```diff
@@ -477,6 +480,14 @@ impl fmt::Display for ValidationError {
             Self::UndefinedRule { rule_name } => {
                 write!(f, r#"Call to undefined rule "{}""#, rule_name)
             }
+            Self::MissingRequiredRule { rule } => {
+                write!(
+                    f,
+                    r#"Missing implementation for required rule {}({});"#,
+                    rule.name,
+                    format_params(&rule.params, ", ")
```
gj, line 488: Unless I'm missing something, it seems like we could use the existing `Display` impl:

https://github.com/osohq/oso/blob/f0b7156e2a9bcbacd314d90b391c5e9e4f8fe3e3/polar-core/src/formatting.rs#L317-L323

polar-core/src/kb.rs
```diff
@@ -162,6 +162,37 @@ impl KnowledgeBase {
...
+        // resource block definitions.
+        for rule_type in self.rule_types.required_rule_types() {
+            if let Some(GenericRule { rules, .. }) = self.rules.get(&rule_type.name) {
+                let found_match = rules.values().any(|rule| {
+                    self.rule_params_match(rule.as_ref(), rule_type)
+                        .map(|r| match r {
+                            RuleParamMatch::True => true,
+                            RuleParamMatch::False(_) => false,
+                        })
+                        .unwrap()
```
gj, line 176: Is this `.unwrap()` cool?

patrickod: I think this should be safe, the reason being
1. the only `Err` case this function can encounter relates checking subtypes for specializers
2. the rule types we're checking here were created with specializer symbols lifted directly from the policy which we require to be registered.

polar-core/src/error.rs
patrickod: yep 💯 correct. just trimmed this up.

polar-core/src/kb.rs
```diff
@@ -193,8 +224,12 @@ impl KnowledgeBase {
         rule_type_instance: &InstanceLiteral,
         index: usize,
     ) -> PolarResult<RuleParamMatch> {
+        // TODO @patrickod remove this hacky comparison edge-case by making Actor and Resource "real" classes
+        if (rule_instance.tag.0 == RESOURCE_UNION_NAME) || (rule_instance.tag.0 == ACTOR_UNION_NAME)
```
gj, line 228: I'm not sure this logic is correct.

At this call site, `rule_instance` isn't a union b/c we would've fallen into the case on 281 and `rule_type_instance` isn't a union b/c a union can't currently be a member of another union, and we've constructed `rule_type_instance` from the members of the `rule_type_instance` union value that caused us to enter this `else if` block:

https://github.com/osohq/oso/blob/62ff5aeb85e4cacb800e64faea6fea2e7c7a4cfb/polar-core/src/kb.rs#L311

So that call site won't ever hit this `if` case.

At the other call site, `rule_instance` *might* be a union and `rule_type_instance` will never be a union (b/c it would've fallen into the branch on 280):

https://github.com/osohq/oso/blob/62ff5aeb85e4cacb800e64faea6fea2e7c7a4cfb/polar-core/src/kb.rs#L337

So if `rule_instance` is a union here, which will hit this `if` case, then `rule_type_instance` *won't* be a union, and unions are less specific than members of unions, so it should be a `RuleParamMatch::False`, no?

gj (changes requested):
Need some help parsing whether logic is correct.

As as aside, I think additional test cases that test the branches we're modifying/adding in this PR would make reviewing the logic *much* easier vs. trying to trace through the hairy rule type validation code sans examples.

polar-core/src/resource_block.rs
```diff
@@ -266,7 +266,7 @@ pub struct ResourceBlock {
 #[derive(Clone, Default)]
 pub struct ResourceBlocks {
     /// Map from resource (`Symbol`) to the declarations in that resource's block.
-    declarations: HashMap<Term, Declarations>,
+    pub declarations: HashMap<Term, Declarations>,
```
gj, line 269: We might not've followed this advice previously given the other `pub` fields here, but in general if you want to expose read-only access to a field a good strategy is to make a getter method that returns a reference to the field instead of making the field `pub`.

patrickod: 👍🏻 refactored to use a getter.

polar-core/src/error.rs
```diff
@@ -479,6 +481,9 @@ impl fmt::Display for ValidationError {
             Self::UndefinedRule { rule_name } => {
                 write!(f, r#"Call to undefined rule "{}""#, rule_name)
             }
+            Self::MissingRequiredRule { rule } => {
+                write!(f, r#"Missing implementation for required rule {}"#, rule)
```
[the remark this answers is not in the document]
patrickod: 👍🏻

polar-core/src/kb.rs
```diff
@@ -162,6 +162,34 @@ impl KnowledgeBase {
...
+                }
+            } else {
+                return Err(self.set_error_context(
+                    &rule_type.body,
+                    error::ValidationError::MissingRequiredRule {
+                        rule: rule_type.clone(),
+                    },
+                ));
+            }
+        }
```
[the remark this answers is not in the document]
patrickod: https://github.com/osohq/oso/blob/patrickod/validate-relations-roles/polar-core/tests/integration_tests.rs#L2508 added this integration test which exercises the missing required rule case.

polar-core/src/macros.rs
```diff
@@ -198,6 +199,20 @@ macro_rules! rule {
...
+    // this macro variant is used exclusively to create rule *types*
+    // TODO: @patrickod break into specific-purpose rule_type! macro and RuleType struct
+    ($name:expr, [$($args:tt)*], $required:expr) => {{
+        let mut params = args!($($args)*);
+        params.reverse();
+        Rule {
+            name: sym!($name),
+            params,
+            body: term!(op!(And)),
+            source_info: $crate::sources::SourceInfo::Test,
```
dhatch, line 214: This source info doesn't seem correct. This isn't coming from a Test.

languages/rust/oso/tests/test_resource_blocks.rs
```diff
@@ -78,10 +78,15 @@ fn test_resource_blocks() {
       allow(actor, action, resource) if
         has_permission(actor, action, resource);
 
-      has_role(user: User, name: String, resource: Resource) if
+      has_role(user: User, name: String, repo: Repo) if
```
dhatch, line 81: Does writing a general `has_role` rule over `Resource` no longer work?

polar-core/src/macros.rs
samscott89: This is a problem with all the macros. They were originally built for tests, but are a little too convenient and have grown to other uses. I created a task to address this.

polar-core/src/kb.rs
```diff
@@ -96,10 +96,13 @@ impl KnowledgeBase {
...
 
     pub fn validate_rules(&self) -> Vec<Diagnostic> {
-        let mut diagnostics = self.validate_rule_calls();
+        let mut diagnostics = vec![];
         if let Err(e) = self.validate_rule_types() {
             diagnostics.push(Diagnostic::Error(e));
         }
+
+        diagnostics.append(&mut self.validate_rule_calls());
+
```
gj, line 105: why this change?

polar-core/src/kb.rs
```diff
@@ -178,7 +209,6 @@ impl KnowledgeBase {
         rule_type_instance: &InstanceLiteral,
         index: usize,
     ) -> PolarResult<RuleParamMatch> {
-        // Get the unique ID of the prototype instance pattern class.
```
gj, line 181: Why's this deleted?

polar-core/src/kb.rs
```diff
@@ -272,6 +302,12 @@ impl KnowledgeBase {
 \tPerhaps you meant to add an actor block to the top of your policy, like this:
 
 \t  actor {} {{}}", rule_instance.tag));
+                            } else if rule_type_instance.tag == sym!("Resource") {
```
gj, line 305: Something like this should work, with `RESOURCE_UNION_NAME` imported from `resource_block.rs`:

```suggestion
                            } else if rule_type_instance.tag.0 == RESOURCE_UNION_NAME {
```

polar-core/src/macros.rs
gj: I understand not wanting to fix the existing issue as part of this PR, but I'm not sure how I feel about exacerbating it (adding a new macro with `SourceInfo::Test` and using it outside of tests)

docs/examples/rbac/go/main.polar
```diff
@@ -4,10 +4,15 @@ allow(actor, action, resource) if
 # docs: end-allow
 
 # docs: begin-has_role
-has_role(user: User, name: String, resource: Resource) if
+has_role(user: User, name: String, repo: Repository) if
```
gj, line 7: Can these be reverted now?

polar-core/src/kb.rs
patrickod: because rule types are not dynamically generated validating rule calls before types with malformed resources/relations can produce an opaque "call to undefined rule `has_relation"`

patrickod: rather than special casing undefined rule calls for "built ins" re-ordering to validate types first will surface the missing required rule implementation error instead

polar-core/src/kb.rs
patrickod: accidental. thanks

docs/examples/rbac/go/main.polar
patrickod: yep! will do.

polar-core/src/kb.rs
patrickod: 👍🏻 updated this and the other actor check on the previous line

docs/examples/rbac/go/main.polar
patrickod: reverted 😄

polar-core/src/kb.rs
```diff
@@ -688,6 +723,125 @@ impl KnowledgeBase {
...
+
+        // Iterate through all resource block declarations and create
+        // non-required rule types for each relation declaration we observe.
+        //
+        // We create non-required rule types to gracefully account for the case
+        // where users have declared relations ahead of time that are used in
+        // rule or resource definitions.
+        for (object, declarations) in self.resource_blocks.declarations() {
+            for (name, declaration) in declarations.iter() {
+                if let Declaration::Relation(subject) = declaration {
```
gj, line 742: It would be nice to move this logic requiring intimate knowledge of the inner structure of `ResourceBlocks` into a method on `ResourceBlocks`:

```diff
diff --git a/polar-core/src/kb.rs b/polar-core/src/kb.rs
index 80737271..9799c131 100644
--- a/polar-core/src/kb.rs
+++ b/polar-core/src/kb.rs
@@ -739,13 +739,9 @@ impl KnowledgeBase {
-        for (object, declarations) in self.resource_blocks.declarations() {
-            for (name, declaration) in declarations.iter() {
-                if let Declaration::Relation(subject) = declaration {
-                    has_relation_rule_types_to_create.insert(
-                        (
-                            subject.value().as_symbol()?,
-                            name.value().as_string()?,
-                            object.value().as_symbol()?,
-                        ),
-                        false,
-                    );
-                }
-            }
+        for (subject, name, object) in self.resource_blocks.relation_tuples() {
+            has_relation_rule_types_to_create.insert(
+                (
+                    subject.value().as_symbol()?,
+                    name.value().as_string()?,
+                    object.value().as_symbol()?,
+                ),
+                false,
+            );
diff --git a/polar-core/src/resource_block.rs b/polar-core/src/resource_block.rs
index 66a8e510..ed8670da 100644
--- a/polar-core/src/resource_block.rs
+++ b/polar-core/src/resource_block.rs
@@ -391,0 +392,12 @@ impl ResourceBlocks {
+
+    pub fn relation_tuples(&self) -> Vec<(&Term, &Term, &Term)> {
+        let mut tuples = vec![];
+        for (object, declarations) in self.declarations() {
+            for (name, declaration) in declarations {
+                if let Declaration::Relation(subject) = declaration {
+                    tuples.push((subject, name, object));
+                }
+            }
+        }
+        tuples
+    }
```

polar-core/src/kb.rs
gj: Great catch; would you mind adding comments to the code that capture that nuance? Otherwise, I am _definitely_ not going to remember that when I next come bulldozing through here 😆

polar-core/src/kb.rs
```diff
@@ -266,12 +297,18 @@ impl KnowledgeBase {
                         }
                         if !success {
                             let mut err = format!("Rule specializer {} on parameter {} must be a member of rule type specializer {}", rule_instance.tag,index, rule_type_instance.tag);
-                            if rule_type_instance.tag == sym!("Actor") {
+                            if rule_type_instance.tag.0 == ACTOR_UNION_NAME {
```
gj, line 300: good catch

polar-core/src/kb.rs
```diff
@@ -147,6 +152,34 @@ impl KnowledgeBase {
                 }
             }
         }
+
+        // For every rule type that is *required*, see that there is at least one corresponding
+        // implementation.
+        for rule_type in self.rule_types.required_rule_types() {
```
gj, line 158: nit: this whole iteration could probably be a `validate_required_rule_types()` method that we call here w/ the `?` operator to propagate errors northward

polar-core/src/kb.rs
gj: we talked about this; gonna try using `?` instead

polar-core/src/kb.rs
```diff
@@ -711,6 +893,7 @@ impl KnowledgeBase {
 mod tests {
     use super::*;
     use crate::error::*;
+    use crate::terms::InstanceLiteral;
```
gj, line 896: remove if we don't need this

polar-core/src/macros.rs
gj: follow up

polar-core/src/kb.rs
patrickod: updated w/ comment reflecting the need to order rule type validation before call validation.

polar-core/src/kb.rs
gj: I don't think we need it

gj (approved):
Amazing work & perseverance!

PR #1315, opened 2021-10-22
[POC] Remove sorting and isa-based filtering.

samscott89:


samscott89:
Made an idea doc for this

PR #1320, opened 2021-10-26
Improve dot lookup binding behavior

gw000:
This includes and extends the rewrite/grounder/simplifier changes in #1276

While breaking them out into a separate PR I discovered that not only is the original fix not as general as it needs to be (it works within but not across rules), but also that [we have tests](https://github.com/osohq/oso/blob/fa77c630dfa26af487b5eaecfa4aafc509ad62e1/polar-core/src/partial/partial.rs#L751) that check explicitly for the current incorrect behavior 😭 So this must have been on our radar at some point, but I guess we never prioritized fixing it.

Anyway the fix now basically has two parts:
- unify equivalent dot lookups within a rule body by assigning them the same output var names
- use the simplifier to ensure bindings are consistent before passing them back out of the VM

The old tests for incorrect behavior are removed. @gkaemmer 's test is added to the JS and Python libs.

This is definitely not optimized for performance, there must be a faster way.

PR checklist:

- [x] Added changelog entry.


samscott89:
Closing this in favour of #1323 which makes this error. The changes here had too bad perf implications.

PR #1321, opened 2021-10-26
Don't send resource block errors through LALRPOP

gj:


polar-core/src/resource_block.rs
```diff
@@ -60,128 +44,147 @@ fn term_source_range(term: &Term) -> Range<usize> {
...
+        ("roles", Value::List(_)) => Ok(ParsedDeclaration::Roles(term)),
+        ("permissions", Value::List(_)) => Ok(ParsedDeclaration::Permissions(term)),
+        ("relations", Value::Dictionary(_)) => Ok(ParsedDeclaration::Relations(term)),
 
         ("roles", Value::Dictionary(_)) | ("permissions", Value::Dictionary(_)) => {
             let (loc, ranges) = (term.offset(), vec![term_source_range(&term)]);
-            let msg = format!("Expected '{}' declaration to be a list of strings; found a dictionary:\n", name);
-            Err(LalrpopError::User { error: ParseError::ResourceBlock { loc, msg, ranges } })
+            let msg = format!("Expected '{}' declaration to be a list of strings; found a dictionary:\n", name.to_polar());
+            Err(ValidationError::ResourceBlock { loc, msg, ranges }.into())
```
samscott89, line 63: What if we stored the offending term on the resource block error instead? And then we could move all the location-finding logic elsewhere?

polar-core/src/error.rs
```diff
@@ -86,13 +85,21 @@ impl PolarError {
...
         // Augment ResourceBlock errors with relevant snippets of parsed Polar policy.
-        if let ErrorKind::Parse(ParseError::ResourceBlock {
+        if let ErrorKind::Validation(ValidationError::ResourceBlock {
             ref mut msg,
             ref ranges,
-            ..
+            ref loc,
         }) = self.kind
         {
             if let Some(source) = source {
```
samscott89, line 94: Is `source` ever set on the context for resource block errors? Where's that happening? 

Related to my other comment about carrying through `term` instead of `loc`. I'm wondering if we would even know which file the error comes from if we throw away the term

polar-core/src/error.rs
```diff
@@ -86,13 +85,21 @@ impl PolarError {
...
+                let (row, column) = crate::lexer::loc_to_pos(&source.src, *loc);
+                self.context.replace(ErrorContext {
+                    source: source.clone(),
+                    row,
+                    column,
+                    include_location: false,
```
samscott89, line 100: This should be unnecessary. This will never be set for any of these error types.

polar-core/src/resource_block.rs
gj: 82dd969

polar-core/src/error.rs
gj: Removed in 82dd969

polar-core/src/error.rs
gj: 82dd969 sets context & carries around the relevant term for its lexical info

polar-core/src/error.rs
```diff
@@ -69,7 +68,12 @@ impl PolarError {
                 }
                 _ => {}
             },
-            (e, Some(source), Some(term)) => {
+            (e, Some(source), Some(term))
+            | (
```
samscott89, line 72: I think you can drop the arm, and then drop the places where `.set_context(Some(source), None)` are called. It looks to me like the one place that does `set_error_context` on the `kb` should be sufficient.

polar-core/src/polar.rs
```diff
@@ -216,9 +217,25 @@ impl Polar {
...
+                    parser::Line::ResourceBlock {
+                        keyword,
+                        resource,
+                        productions,
+                    } => match resource_block_from_productions(keyword, resource, productions) {
+                        Ok(block) => {
+                            let errors = block
+                                .add_to_kb(kb)
+                                .into_iter()
+                                .map(|e| e.set_context(Some(source), None));
```
samscott89, line 229: This and the one below it is what I'm referring to in this comment: https://github.com/osohq/oso/pull/1321/files#r740529825

polar-core/src/kb.rs
```diff
@@ -699,7 +699,20 @@ impl KnowledgeBase {
...
 
-        errors.into_iter().map(Diagnostic::Error).collect()
+        errors
+            .into_iter()
+            // Attach context to ResourceBlock errors.
+            //
+            // TODO(gj): can we attach context to *all* errors here since all errors will be
+            // parse-time errors and so will have some source context to attach?
+            .map(|e| match e.kind {
+                ErrorKind::Validation(ref e @ ValidationError::ResourceBlock { ref term, .. }) => {
```
samscott89, line 709: I think this is the only context-setting you need to keep. And yeah, I think this is potentially a reasonable place to pull in any additional context from the knowledge base.

gj: This only sets context for errors from `check_all_relation_types_have_been_registered()` and `as_rule()`.

samscott89 (approved):
Nice!

PR #1325, opened 2021-10-27
Statically build the parser, and conditionally update it

samscott89:
This removes lalrpop from the happy path. Added a CI job to make sure we don't forget to rebuild it if we update the parser.

TODO: 
- [ ] Check this properly fails CI if the parser needs updating


samscott89:
Closing pending further perf analysis

PR #1329, opened 2021-10-29
Add parity tests for unhandled partial error + update Go types

samscott89:
The main change here was supposed to be adding tests for the error in #1323.

But then I realised I was missing the error type for Go... And a bunch of others.

Before a bunch of the errors would raise complaining that the error couldn't be deserialized. I think this was related to the segfaults somehow?!


gw000:
Is `polar_types.go` fully autogenerated? It should have a comment saying that if so. Otherwise this looks fine ...

samscott89:
> 
> 
> Is `polar_types.go` fully autogenerated? It should have a comment saying that if so. Otherwise this looks fine ...

Sort of... Not officially. Creating a follow-up task to handle this

gw000 (approved):
🦔

PR #1330, opened 2021-10-29
Error on unregistered fields in data filtering

gw000:
Kind of like the recent JS change. Maybe this could be configurable?

```
oso.config({ missingAttributeIsError: false })
```

Screen Shot 2021-10-29 at 10 05 39 AM

PR checklist:

- [x] Added changelog entry.


polar-core/src/data_filtering.rs
```diff
@@ -316,6 +318,30 @@ impl VarInfo {
...
+    Err(RuntimeError::DataFilteringError(msg).into())
+}
+
+fn unregistered_field_error<A>(var_type: &str, field: &str) -> PolarResult<A> {
+    let msg = formatdoc!(
+        r#"Unregistered field or relation: {}.{}
+        
+        Please include `{}` as a field in your `register_class` call:
+
+        oso.register_class({}, fields={{
```
samscott89, line 331: Love the error message, but this could be a bit confusing for non-Python people.
Maybe we could say, "e.g. in Python" and link to docs?

samscott89: Alternatively, this would be one reason to actually handle the error variant differently.

polar-core/src/error.rs
```diff
@@ -298,6 +298,7 @@ pub enum RuntimeError {
         var: Symbol,
         term: Term,
     },
+    DataFilteringError(String),
```
samscott89, line 301: We're trying to get a little better with how we construct errors. Instead of constructing the error message at the callsite, you should store the relevant info in the error itself, and construct the message as part of `impl fmt::Display for RuntimeError`

So this would be:

```
DataFilteringError { var_type: String, field: String }
```
and the message formatting would be handled on 361: https://github.com/osohq/oso/pull/1330/files#diff-20681b7129ace19298fd33795d95bbf7c5b118c4be2bf33844520b21d8798d45R361

docs/content/any/project/changelogs/NEXT.md
```diff
@@ -18,6 +18,12 @@ draft: true
 
 ## `RELEASED_PACKAGE_1` NEW_VERSION
 
+### JavaScript
+
+#### Other bugs & improvements
+
+- Fixed an error in JavaScript data filtering documentation (thanks to Ken Fukuyama for the change)
```
samscott89, line 25: This was from last release?

polar-core/src/data_filtering.rs
gw000: Good point. I think adding "e.g. in Python" & a docs link would be ok for now, but esp. as we add more long explicit errors it'd make sense to have a standard way for the host to do its own formatting etc.

samscott89: @gj -- potentially relevant to your interests

gj: oof, another angle to think about

docs/content/any/project/changelogs/NEXT.md
gw000: Oops thanks

polar-core/src/error.rs
```diff
@@ -344,6 +349,29 @@ The expression is: {expr}
...
+
+                    For more information please refer to our documentation:
+                        https://docs.osohq.com/guides/data_filtering.html
+                    "#,
+                    var_type,
+                    field,
+                    field,
+                    var_type,
+                    var_type,
+                    field
```
gj, line 371: it'd be easier to match all these variables w/ the corresponding format directive if we used the named syntax, e.g., `"{var_type}", var_type=var_type` instead of `"{}", var_type`

gw000: Ooooh cool I did not know you could do this, thanks!

languages/go/internal/ffi/native/polar.h
```diff
@@ -13,27 +13,29 @@ typedef struct polar_Query polar_Query;
 
 const char *polar_get_error(void);
 
-polar_Polar *polar_new(void);
+struct polar_Polar *polar_new(void);
```
gj, line 16: what's up with the changes in this file?

gw000: not sure, I thought they were generated by the polar-crossgen. @samscott89 do you know if this was supposed to happen?

samscott89: That shouldn't be a side effect of polar-crossgen, no... Although I have also started to see these popping up in my `polar-c-api/polar.h` file lately... I assumed it was something to do with Rust/cbindgen versions changing

gw000: ok replaced it with `polar.h` from `main`

saolsen: Yeah its probably just an update to cbindgen/rust. It also generates typedefs so it's valid with or without saying `struct`. Fine to just commit whatever comes out.

samscott89 (approved):

Issue #1264, opened 2021-10-08
Ensure all languages have an 'accept_expression'-style flag


gj:
## TODO

- [x] JS
  https://github.com/osohq/oso/blob/7602f25204b5790231dcadb9ea5fd5bd4e37a55d/languages/js/src/Host.ts#L558-L559
- [x] Go
  https://github.com/osohq/oso/blob/7602f25204b5790231dcadb9ea5fd5bd4e37a55d/languages/go/internal/host/host.go#L390-L393

## Done

- [x] Python
  https://github.com/osohq/oso/blob/7602f25204b5790231dcadb9ea5fd5bd4e37a55d/languages/python/oso/polar/host.py#L334-L336
- [x] Ruby
  https://github.com/osohq/oso/blob/7602f25204b5790231dcadb9ea5fd5bd4e37a55d/languages/ruby/lib/oso/polar/host.rb#L357-L358
- [x] Rust
  https://github.com/osohq/oso/blob/7602f25204b5790231dcadb9ea5fd5bd4e37a55d/languages/rust/oso/src/query.rs#L269-L270
- [x] Java
  https://github.com/osohq/oso/blob/7602f25204b5790231dcadb9ea5fd5bd4e37a55d/languages/java/oso/src/main/java/com/osohq/oso/Host.java#L383-L386

gj:
All done! Thanks to @joshrotenberg (#1335) & @Kn99HN (#1412) for adding the final pair.

PR #1335, opened 2021-10-31
Add acceptExpression to host

joshrotenberg:
Adding acceptExpression to Go. See #1264.

PR checklist:

- [ ] Added changelog entry.


gj:
Hi @joshrotenberg, thanks for the contribution! Do you need any help from us on pushing this forward? Would love to get it into the next release (~November 10th)

joshrotenberg:
> Hi @joshrotenberg, thanks for the contribution! Do you need any help from us on pushing this forward? Would love to get it into the next release (~November 10th)

I think I did the easy part, looking at `polar.go` now to see what's next.

languages/go/errors/error.go
```diff
@@ -210,3 +210,13 @@ func (e *ForbiddenError) Error() string {
...
 		"403 error to the client."
 }
+
+type UnexpectedExpressionError struct{}
+
+func (e *UnexpectedExpressionError) Error() string {
+	return "Received Expression from Polar VM. The Expression type is only supported when " +
+		"using data filtering features. Did you perform an operation over an unbound variable " +  
+		"in your policy?\n\n" + 
+		"To silence this error and receive an Expression result, pass acceptExpression as true to Oso.query"
```
samscott89, line 220:
```suggestion
		"To silence this error and receive an Expression result, call Oso.SetAcceptExpression(true)" 
```

samscott89 (dismissed):
Thanks @joshrotenberg!

A couple of small things and this looks good to go.

1. I left a comment to update the error message
2. Please can you run the formatter (either `make fmt` in the go/ directory, or `make fmt-go` from the root

Thanks!

languages/go/oso.go
```diff
@@ -72,6 +72,16 @@ func (o *Oso) SetNotFoundError(notFoundError func() error) {
 	o.notFoundError = notFoundError
 }
 
+/*
+Set whether the Host accepts Expression types from Polar, or raises an error.
+
+	o, _ = oso.NewOso()
+	o.SetAcceptExpression(true)
+*/
+func (o *Oso) SetAcceptExpression(acceptExpression bool) {
```
samscott89, line 81: Hey! Sorry, one more quick change. The other libraries expose this method via the `Query` object instead. Please can we move it there to be consistent?

languages/go/errors/error.go
```diff
@@ -210,3 +210,13 @@ func (e *ForbiddenError) Error() string {
...
 		"403 error to the client."
 }
+
+type UnexpectedExpressionError struct{}
+
+func (e *UnexpectedExpressionError) Error() string {
+	return "Received Expression from Polar VM. The Expression type is only supported when " +
+		"using data filtering features. Did you perform an operation over an unbound variable " +
+		"in your policy?\n\n" +
+		"To silence this error and receive an Expression result, call Oso.SetAcceptExpression(true)"
```
samscott89, line 220:
```suggestion
		"To silence this error and receive an Expression result, call query.SetAcceptExpression(true)
		 on a query (e.g. created by calling Oso.NewQueryFromRule)."
```

joshrotenberg: Argh missed that.

languages/go/internal/host/host.go
```diff
@@ -388,10 +390,14 @@ func (h Host) ToGo(v types.Term) (interface{}, error) {
 	case ValueVariable:
 		return inner, nil
 	case ValueExpression:
-		return nil, fmt.Errorf(
-			"Received Expression from Polar VM. The Expression type is not yet supported in this language.\n" +
-				"This may mean you performed an operation in your policy over an unbound variable.")
+		if !h.acceptExpression {
+			return nil, &errors.UnexpectedExpressionError{}
+		}
```
samscott89, line 395: Just realised this needs a `return inner, nil`. There's not much you can do with the expression at the moment, but this would give someone a chance!

languages/go/tests/unit_test.go
```diff
@@ -443,3 +443,86 @@ func TestFailingALot(t *testing.T) {
...
+
+	if o.LoadString("f(1); f(x) if x = 1 and x = 2;") != nil {
+		t.Fatalf("Load string failed: %v", err)
+	}
+
+	q, err := o.NewQueryFromRule("f", Variable("x"))
+	if err != nil {
+		t.Fatalf("Failed to get result: %s", err)
+	}
+	q.SetAcceptExpression(true)
```
gj, line 463: Removing this has no effect on the test b/c the first result doesn't return an expression (it returns `1` after matching `f(1);`) and the second result doesn't return an expression (`x = 1 and x = 2` are incompatible constraints, so no result is returned). Is this section of the test actually testing anything worthwhile?

https://github.com/osohq/oso/blob/75c6f608072ea1bb2e1b6bed42fce64d4b35a780/languages/go/tests/unit_test.go#L455-L476

samscott89: I copied this test from the python tests 

My assumption was that this is testing that setting allow expression doesn't suddenly break everything else?

gj: Okay I suppose that's fine. Would also be nice to add a test that _doesn't_ set the flag and asserts we receive the error — maybe even combine it with the final test case here but modify the policy to produce a pair of expression results, and we frob the flag in between the results and see one of them error & the other return the `Expression`.

samscott89: Okay, made the test case reusable, and checked that it works for:
- not setting
- setting to false
- setting to true

gj (approved):
Thanks @joshrotenberg & @samscott89 !

joshrotenberg:
Haha I feel like I just made more work for @samscott89

PR #1336, opened 2021-10-31
Remove unused stuff from #1035

gj:


polar-core/src/rules.rs
```diff
@@ -18,66 +16,17 @@ impl Parameter {
...
 }
 
-#[derive(Debug, Clone, Deserialize, Serialize)]
+#[derive(Debug, Clone, Deserialize, PartialEq, Serialize)]
 pub struct Rule {
     pub name: Symbol,
     pub params: Vec<Parameter>,
     pub body: Term,
-    #[serde(skip, default = "SourceInfo::ffi")]
-    pub source_info: SourceInfo,
```
samscott89, line 27: Why are you getting rid of source info? Don't we need that to return good errors?

gj: I'm only getting rid of it from rules — not from terms. We don't currently use it anywhere for rules.

samscott89: It seems to me like we'll probably want to pretty soon, no? I don't know why we'd make that harder on ourselves by undoing the work to enable it

samscott89 (changes requested):
Apart from the source info part (see comment thread). Other removals look good. We no longer support the concept of loading individual files, so makes sense to remove the ability to remove individual files.

polar-core/src/rules.rs
gj: Maybe. It's just as easy to bring it back when we need it instead of having it hang around now, unused, but I don't really care either way. I will add it back.

samscott89 (approved):

PR #1338, opened 2021-11-01
Export Class type in JS library

gw000:
The type `Class<T>` appears in the arguments to data filtering API functions, where it represents a constructor for type `T`: eg. in `oso.authorizedQuery(actor, action, Resource)` `Resource` has type `Class<T>`.

`Class` is defined in `types.ts`. This PR exports it from `index.ts` so users can access it more easily.

PR checklist:

- [x] Added changelog entry.


languages/js/src/index.ts
```diff
@@ -3,3 +3,4 @@ export { Variable } from './Variable';
 export { AuthorizationError, ForbiddenError, NotFoundError } from './errors';
 export { Relation } from './dataFiltering';
 export { defaultEqualityFn } from './helpers';
+export { Class } from './types';
```
gj, line 6: Is this okay?
```suggestion
export type { Class } from './types';
```

docs/content/any/project/changelogs/NEXT.md
```diff
@@ -16,6 +16,12 @@ draft: true
 
 - Fixed a bug that caused the Go library to crash when using Go 1.17 on macOS.
 
+### Node.js
+
+#### Other bugs & improvements
+- The `Class` type for representing abstract resources for data filtering is
```
gj, line 22: This isn't really what the `Class` type is for, at least internally. Is this the only context in which someone would encounter the `Class` type while using the library?

languages/js/src/index.ts
gj: Also, if we're making it a public export, we should add/update the doc string on the type: https://github.com/osohq/oso/blob/054b3ddae78a1f02e26fc74090a15ca95f4c493c/languages/js/src/types.ts#L316-L320

gj (changes requested):
How did you test this change? Do you think we should add a note to the docs about when/how to use this newly exported type? Is this the only change required across the whole JS lib public API, or is this just the first fix and you're still evaluating the non-`authorizedResources()` APIs?

languages/js/src/index.ts
gw000: I have no idea what the difference is between `export` and `export type`, can you explain

gw000:
> How did you test this change? Do you think we should add a note to the docs about when/how to use this newly exported type? Is this the only change required across the whole JS lib public API, or is this just the first fix and you're still evaluating the non-`authorizedResources()` APIs?

I didn't test it, I'm just assuming JavaScript's module system works.

This is the only thing I've looked at. I'm still a little fuzzy on the motivation for this change honestly.

docs/content/any/project/changelogs/NEXT.md
gw000: It's the only example on record afaik ...

gj (approved):
:shipit:

PR #1339, opened 2021-11-01
Fix node parity test

samscott89:
Negated the correct logic in #1329 but missed it since failed promises weren't failing the CI test.


gj (approved):
Tested locally

PR #1342, opened 2021-11-01
print the debugging message in handleDebug event handler

patrickod:
Fix REPL functionality by printing the result messages from the core in
the Golang debug event handler.


patrickod:
Creating draft at the moment to follow up with the equivalent "enrichment" pretty printing translating instance IDs to their class names.

samscott89:
> 
> 
> Creating draft at the moment to follow up with the equivalent "enrichment" pretty printing translating instance IDs to their class names.

I saw you requested a review. Should I wait for the follow up?

samscott89 (approved):
Verified locally that this works.

Issue #1279, opened 2021-10-13
Data filtering caused panic


kamal:
oso-oso rubygem 0.21.0

```
thread '<unnamed>' panicked at 'invalid query', polar-core/src/vm.rs:1760:17
stack backtrace:
   0: std::panicking::begin_panic
   1: polar_core::vm::PolarVirtualMachine::query_for_operation
   2: polar_core::vm::PolarVirtualMachine::next
   3: <polar_core::vm::PolarVirtualMachine as polar_core::runnable::Runnable>::run
   4: polar_core::polar::Query::next_event
   5: polar_core::polar::Query::next_event
   6: polar_core::polar::Query::next_event
   7: polar_core::polar::Query::next_event
   8: polar_core::polar::Query::next_event
   9: polar_core::polar::Query::next_event
  10: polar_core::polar::Query::next_event
  11: polar_core::polar::Query::next_event
  12: polar_core::polar::Query::next_event
  13: <std::panic::AssertUnwindSafe<F> as core::ops::function::FnOnce<()>>::call_once
  14: _polar_next_query_event
  15: _ffi_call_unix64
  16: _ffi_call_int
  17: _ffi_call
  18: rbffi_CallFunction
             at /Users/kamal/.asdf/installs/ruby/2.5.3/lib/ruby/gems/2.5.0/gems/ffi-1.9.25/ext/ffi_c/Call.c:433:9
  19: custom_trampoline
             at /Users/kamal/.asdf/installs/ruby/2.5.3/lib/ruby/gems/2.5.0/gems/ffi-1.9.25/ext/ffi_c/MethodHandle.c:232:34
  20: _call_cfunc_m1
  21: _vm_call_cfunc_with_frame
  22: _vm_call_cfunc
  23: _vm_exec_core
  24: _vm_exec
  25: _invoke_block
  26: _invoke_iseq_block_from_c
  27: _invoke_block_from_c_bh
  28: _vm_yield
  29: _rb_yield_0
  30: _loop_i
  31: _rb_rescue2
  32: _rb_f_loop
  33: _call_cfunc_0
  34: _vm_call_cfunc_with_frame
  35: _vm_call_cfunc
  36: _vm_call_method_each_type
  37: _vm_call_method
  38: _vm_call_general
  39: _vm_exec_core
  40: _vm_exec
  41: _vm_call0_body
  42: _vm_call0
  43: _rb_call0
  44: _rb_call
  45: _iterate_method
  46: _rb_iterate0
[...]

PR #1345, opened 2021-11-02
Handle simplified input in `query_op_helper`

gw000:
The inverter uses the simplifier to construct the constraints it passes out to the host VM. The VM has its own ideas about acceptable input, which don't always match up with the simplifier's output. For example, in [this case](https://github.com/osohq/oso/issues/1279) a negated comparison on a dot lookup in a partial led to the VM encountering an illegal expression when the constraint was reevaluated after one of its variables was bound.

This change adds special cases to handle rewritten dot ops in `query_op_helper`, much like we already do in `unify`. Changing the inverter not to rely on the simplifier, or changing the VM and simplifier so they agree on how to interpret expressions, would be more robust ways to fix the problem.

This PR also replaces the other `panic!`s and `assert!`s in `vm.rs` with typed errors.

PR checklist:

- [x] Added changelog entry.
- [x] Test if nested dots work (they don't, future work will address)


polar-core/src/vm.rs
```diff
@@ -394,18 +399,16 @@ impl PolarVirtualMachine {
         self.stack_limit = limit;
     }
 
+    fn kb(&self) -> RwLockReadGuard<KnowledgeBase> {
```
samscott89, line 402: Nice

polar-core/src/vm.rs
```diff
@@ -523,31 +526,24 @@ impl PolarVirtualMachine {
...
-            .into());
-        }
+        use VariableState::Unbound;
         match goal {
-            Goal::LookupExternal { call_id, .. } | Goal::NextExternal { call_id, .. } => {
-                assert!(matches!(
-                    self.variable_state(self.get_call_sym(call_id)),
-                    VariableState::Unbound
-                ), "The call_id result variables for LookupExternal and NextExternal goals must be unbound.");
+            _ if self.goals.len() >= self.stack_limit =>
```
samscott89, line 533: Stylistic change or functional change? If its the former, I find this more confusing.

polar-core/src/vm.rs
```diff
@@ -571,15 +567,19 @@ impl PolarVirtualMachine {
...
-        self.choices.push(Choice {
-            alternatives,
-            bsp: self.bsp(),
-            goals: self.goals.clone(),
-            queries: self.queries.clone(),
-            trace: self.trace.clone(),
-            trace_stack: self.trace_stack.clone(),
-        });
+        if self.choices.len() >= self.stack_limit {
+            invalid_state("Too many choices.".to_string())
```
samscott89, line 571: 👏

samscott89 (approved):
Nice! Good, pragmatic fix.

polar-core/src/vm.rs
```diff
@@ -1758,16 +1781,68 @@ impl PolarVirtualMachine {
...
         let mut args = args.clone();
-        assert!(args.len() >= 2);
+        if args.len() < 2 {
+            return invalid_state(format!("query_op_helper: wrong arity: {}", term.to_polar()));
+        }
         let left = &args[0];
         let right = &args[1];
 
         match (left.value(), right.value()) {
+            // We may be querying a partial from the simplifier, which can contain
```
samscott89, line 1791: Worth adding a TODO to rip this out once we fix the inverter?

polar-core/src/vm.rs
gw000: Stylistic. There's three non overlapping paths through the function so I thought it was more clear to have them all be in one block at the same indentation level

samscott89: Yeah, that's fair. It seems a little weird in that the conditions are unrelated to the thing that is actually being matched on, but I don't feel that strongly about it

gw000: I changed it to an `if/else` chain because I agree using a `match` like that is kind of odd

samscott89: Oh I love it. Good solution

PR #1346, opened 2021-11-02
Update django-oso and sqlalchemy-oso docs language

leina05:


gneray:
Didn't realize this was still here. Unless I'm missing something, let's remove this warning altogether

dhatch (approved):

PR #1347, opened 2021-11-02
Build and test python wheels for M1 macs.

saolsen:
A successful test run https://github.com/osohq/oso/actions/runs/1414187611


.github/workflows/release.yml
```diff
@@ -614,6 +646,42 @@ jobs:
...
+          arch -arm64 ./venv38/bin/python test.py
+        working-directory: test
+      - name: "test python 3.9"
+        run: |
+          arch -arm64 /opt/homebrew/opt/python@3.9/bin/python3 -m venv venv39
+          arch -arm64 ./venv39/bin/pip install oso==${{ steps.version.outputs.oso_version }} -f ../wheel --no-deps --no-index
+          arch -arm64 ./venv39/bin/pip install oso==${{ steps.version.outputs.oso_version }} -f ../wheel
+          arch -arm64 ./venv39/bin/python test.py
+        working-directory: test
+      - name: "test python 3.10"
```
leina05, line 677: If we validate 3.10 here should we do it on the other ones as well?

leina05 (approved):

.github/workflows/release.yml
saolsen: yeah, we need to build 3.10 for the other ones too, gonna do that seperately

PR #1348, opened 2021-11-03
[PLS] Ignore errors dependent on app data

gj:


polar-core/src/error.rs
```diff
@@ -411,7 +398,17 @@ pub enum ValidationError {
...
+    SingletonVariable {
+        /// Term<Symbol> where the error arose, tracked for lexical context.
+        term: Term,
+        /// Variable name.
+        name: String,
+    },
+    UnregisteredConstant {
+        /// Term<Symbol> where the error arose, tracked for lexical context.
+        term: Term,
+        msg: String,
```
dhatch, line 410: I think it's probably best going forward to include the information necessary to generate the message in the Error struct, but not the message itself. This way, if the error is generated at multiple sites the message will be consistent.

I realize we probably have not followed that convention so far, but maybe we can add a note to this file about it.

polar-core/src/kb.rs
```diff
@@ -551,6 +551,28 @@ impl KnowledgeBase {
...
+    pub fn is_constant(&self, name: &Symbol) -> bool {
+        self.constants.contains_key(name)
+    }
+
+    /// Getter for `constants` map without exposing it for mutation.
+    pub fn get_registered_constants(&self) -> &Bindings {
+        &self.constants
+    }
+
+    pub fn get_registered_constant(&self, constant: &Term) -> PolarResult<&Term> {
```
dhatch, line 564: Having APIs that only accept one term type but don't say so in the type description is unfortunate. You are probably already thinking this from your comments, but we should rethink how we handle source information & types so that it can be attached to more specific types so that our APIs can be more clear.

Could you write a clean-up task about this (assuming we're on the same page)?

polar-core/src/kb.rs
```diff
@@ -536,7 +536,7 @@ impl KnowledgeBase {
     }
 
     /// Define a constant variable.
-    pub fn constant(&mut self, name: Symbol, value: Term) -> PolarResult<()> {
+    pub fn register_constant(&mut self, name: Symbol, value: Term) -> PolarResult<()> {
```
dhatch, line 539: +1

polar-core/src/polar.rs
```diff
@@ -255,24 +255,26 @@ impl Polar {
...
-                    {
-                        return Diagnostic::Error(kb.set_error_context(term, e.clone()));
+        for diagnostic in &mut diagnostics {
+            if let Diagnostic::Error(e) = diagnostic {
+                use {ErrorKind::Validation, ValidationError::*};
+                match e.kind {
+                    Validation(ResourceBlock { ref term, .. })
+                    | Validation(SingletonVariable { ref term, .. })
+                    | Validation(UnregisteredConstant { ref term, .. }) => {
+                        *e = kb.set_error_context(term, e.clone());
```
dhatch, line 269: Conceptually we shouldn't need a clone here. Why doesn't `set_error_context` take a mut ref?

polar-core/src/resource_block.rs
```diff
@@ -54,7 +54,7 @@ fn validate_parsed_declaration((name, term): (Term, Term)) -> PolarResult<Parsed
 
         ("roles", Value::Dictionary(_)) | ("permissions", Value::Dictionary(_)) => {
             let msg = format!("Expected '{}' declaration to be a list of strings; found a dictionary", name.to_polar());
-            Err(ValidationError::ResourceBlock { msg, term}.into())
```
dhatch, line 57: how'd this happen with format check 😕

polar-core/src/resource_block.rs
```diff
@@ -226,6 +226,7 @@ impl Declaration {
         if let Declaration::Relation(relation) = self {
             Ok(relation)
         } else {
+            // TODO(gj): why isn't this a ResourceBlock error?
```
dhatch, line 229: Seems small enough to address or remove

polar-wasm-api/src/errors.rs
```diff
@@ -49,6 +48,8 @@ fn kind(err: &PolarError) -> String {
         Validation(InvalidRuleType { .. }) => "ValidationError::InvalidRuleType",
         Validation(ResourceBlock { .. }) => "ValidationError::ResourceBlock",
         Validation(UndefinedRule { .. }) => "ValidationError::UndefinedRule",
+        Validation(SingletonVariable { .. }) => "ValidationError::SingletonVariable",
+        Validation(UnregisteredConstant { .. }) => "ValidationError::UnregisteredConstant",
```
dhatch, line 52: Any changes required in JS & other libraries to handle this new error variant?

dhatch (changes requested):
Some comments throughout approach looks good though

polar-core/src/resource_block.rs
gj: Not sure; `rustfmt` seems pretty apathetic about how I format that line

polar-wasm-api/src/errors.rs
samscott89: Go types will need updating.

polar-core/src/resource_block.rs
gj: b365bf5e

polar-core/src/polar.rs
gj: dunno; preexisting API that I would like to change in a follow-up

polar-core/src/error.rs
gj: I generally agree (e.g., see https://github.com/osohq/oso/commit/204d04917dfa8d8cb128567d8f348b8d92a9d8b0#diff-20681b7129ace19298fd33795d95bbf7c5b118c4be2bf33844520b21d8798d45R385-R398). I don't want to update errors per that convention as part of this PR, but I would like to standardize our errors in a follow-up.

polar-core/src/kb.rs
gj: https://www.notion.so/osohq/Allow-Term-to-be-specific-over-the-inner-Value-type-0c6fa761e84b4b03a1a5f1be83c1d137

polar-wasm-api/src/errors.rs
gj: > Go types will need updating.

7bd53f89

gj: > Any changes required in JS & other libraries to handle this new error variant?

- JS: none — all errors crossing the WASM boundary from the core are regular old `Error` instances that have the stringified variant name from the core attached as the `name` field.
- Ruby: none — the code for handling errors received across the FFI boundary is fairly out-of-sync with the core, but not in a way that will currently break anything. All four error variants (parse, runtime, validation, operational) have fallback translations that will handle new errors fine. Users won't be able to specifically pattern match on the new error type (just the generic variant, e.g., `ValidationError` in this case), but that's something I'd like to address in another error cleanup follow-up.
- Python: none — same as Ruby, including being out-of-sync with the core.
- Go: none — errors have been generated (see above comment).
- Rust: none.
- Java: none — same as Python & Ruby, including being out-of-sync with the core.

polar-core/src/kb.rs
dhatch: o

polar-core/src/error.rs
dhatch: This is a new error though right? Can we store the constant name on it instead of the `msg`?

dhatch (approved):
lgtm besides this comment could potentially use addressing: https://github.com/osohq/oso/pull/1348#discussion_r743037503

polar-core/src/error.rs
gj: We store the constant name in the `term: Term` field on line 409. The `msg` field is used to provide some unique context about the situations in which this error can arise:

https://github.com/osohq/oso/blob/17e7f43315fe19d9368bff810740f930d250c94e/polar-core/src/kb.rs#L218

https://github.com/osohq/oso/blob/17e7f43315fe19d9368bff810740f930d250c94e/polar-core/src/resource_block.rs#L650

https://github.com/osohq/oso/blob/17e7f43315fe19d9368bff810740f930d250c94e/polar-core/src/resource_block.rs#L671

https://github.com/osohq/oso/blob/17e7f43315fe19d9368bff810740f930d250c94e/polar-core/src/resource_block.rs#L687

gj: Actually these are a bit different... thinking about how to group them into different error types that don't require `msg` differentiation

dhatch: Oh - I didn't realize that. 🤦 Not sure it's necessary to split up then.

gj: Doing this cleanup as a follow-up.

PR #1350, opened 2021-11-03
test python 3.10

saolsen:
Adds python 3.10 as a build and test target.
Removes musl specific wheels, python wheels now work for both glibc and musl based distros.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -308,7 +308,7 @@ jobs:
           python-version: "3.7"
       - name: Install cibuildwheel
         run: |
-          python -m pip install cibuildwheel==1.6.2
+          python -m pip install cibuildwheel==2.2.2
       - name: Download static libs
         uses: actions/download-artifact@v1
         with:
@@ -316,7 +316,7 @@ jobs:
       - name: Copy static library.
         run: |
           mkdir -p languages/python/oso/native
-          cp -r oso_static_library/libpolar-Linux.a languages/python/oso/native/libpolar.a
+          cp -r oso_static_library/libpolar-musl.a languages/python/oso/native/libpolar.a
           cp -r oso_static_library/polar.h languages/python/oso/native/polar.h
       - name: Copy in readme
         run: |
@@ -353,7 +353,7 @@ jobs:
           python-version: "3.7"
       - name: Install cibuildwheel
         run: |
-          python -m pip install cibuildwheel==1.6.2
+          python -m pip install cibuildwheel==2.2.2
       - name: Download static libs
         uses: actions/download-artifact@v1
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -593,7 +567,7 @@ jobs:
     runs-on: macos-10.15
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
```
samscott89, line 570: lol yaml

.github/workflows/release.yml
```diff
@@ -564,7 +538,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: ["3.6", "3.7", "3.8", "3.9"]
```
samscott89, line 541:
```suggestion
        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
```

samscott89 (commented):
Take it or leave it: was looking at the changes between 1.6 and 2.2 for cibuildwheel. Just checking if we care about any of the following:

- Dropped support for 3.5 (don't care)
- Can configure cibuildwheel through [config file instead](https://cibuildwheel.readthedocs.io/en/stable/options/#setting-options) of env variables -- worth doing?
- It now support `musllinux` I didn't realise that was a thing... anything to do there? https://cibuildwheel.readthedocs.io/en/stable/changelog/#v220 we might currently filter them out accidentally
- Seems like there is a GitHub action we could be using directly instead: https://cibuildwheel.readthedocs.io/en/stable/setup/#github-actions

.github/workflows/release.yml
saolsen: yah

saolsen:
> * Can configure cibuildwheel through [config file instead](https://cibuildwheel.readthedocs.io/en/stable/options/#setting-options) of env variables -- worth doing?

meh, maybe later

> * It now support `musllinux` I didn't realise that was a thing... anything to do there? https://cibuildwheel.readthedocs.io/en/stable/changelog/#v220 we might currently filter them out accidentally

These work, I see them in there.

> * Seems like there is a GitHub action we could be using directly instead: https://cibuildwheel.readthedocs.io/en/stable/setup/#github-actions

meh, maybe later

samscott89 (approved):

.github/workflows/release.yml
```diff
@@ -16,7 +16,7 @@ jobs:
       - name: Set version env
         run: echo "oso_version=$(cat VERSION)" >> $GITHUB_ENV
       - name: Check github ref matches
-        if: github.ref != 'refs/heads/main'
+        if: github.ref != 'refs/heads/main
```
samscott89, line 19:
```suggestion
        if: github.ref != 'refs/heads/main'
```

PR #1356, opened 2021-11-04
Disallow parenthesized specializers in rule heads

gj:
Effectively reverts #61.

Makes all of these cases parse errors:

```py
polar.register_constant(1, name="Foo")
polar.load_str("f(_: (Foo));")
assert qeval("f(1)")
assert not qeval("f(2)")
```

```py
polar.load_str("""
    type f(_: (Foo));
    f(_: (Foo));
""")
```

```py
with pytest.raises(exceptions.ValidationError):
    polar.load_str("""
        type f(_: (Foo));
        f(_: Foo);
    """)
```

```py
polar.register_constant(1, name="Foo")
polar.register_constant(1, name="Bar")
with pytest.raises(exceptions.ValidationError):
    polar.load_str("""
        type f(_: (Bar));
        f(_: (Foo));
    """)
```

PR checklist:

- [x] Added changelog entry.

### DIFF
--- a/docs/content/any/project/changelogs/NEXT.md
+++ b/docs/content/any/project/changelogs/NEXT.md
@@ -12,7 +12,29 @@ draft: true
 
 ### Core
 
+#### Breaking changes
+
+{{% callout "Warning" "orange" %}}
+  This release contains breaking changes. Be sure to follow migration steps
+  before upgrading.
+{{% /callout %}}
+
+##### Parenthesized variable dereferencing in specializer position is no longer permitted
+
+Previously, Oso would treat parenthesized specializers as variables. This made
+it possible to write rules with parameters that referenced each other, like so:
+
+```polar
+isa(x, y, x: (y));
+```
+
+Now, that parenthesized `(y)` will be a parse error.
+
+If you discovered this feature and were using it, please get in touch. We'd
+love to hear about your use case!
+
 #### Other bugs & improvements
+
 - Fixed a bug where a negated constraint on a dot lookup could cause Polar to crash
   when the underlying variable became bound.
 
--- a/docs/content/any/project/changelogs/NEXT.md
+++ b/docs/content/any/project/changelogs/NEXT.md
@@ -28,7 +28,12 @@ it possible to write rules with parameters that referenced each other, like so:
 isa(x, y, x: (y));
 ```
 
-Now, that parenthesized `(y)` will be a parse error.
+Now, that parenthesized `(y)` will be a parse error. The same logic can be
+written in the body of the rule with the `matches` operator:
+
+```polar
+isa(x, y) if x matches y;
+```
 
 If you discovered this feature and were using it, please get in touch. We'd
 love to hear about your use case!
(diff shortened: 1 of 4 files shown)


samscott89 (commented):
Draft approved

docs/content/any/project/changelogs/NEXT.md
```diff
@@ -12,7 +12,29 @@ draft: true
...
+  before upgrading.
+{{% /callout %}}
+
+##### Parenthesized variable dereferencing in specializer position is no longer permitted
+
+Previously, Oso would treat parenthesized specializers as variables. This made
+it possible to write rules with parameters that referenced each other, like so:
+
+```polar
+isa(x, y, x: (y));
```
gw000, line 28: Is the current behavior equivalent to `isa(x, y, x) if x matches y;`? If so we could call that out here so anyone who's affected by this change will know how to fix it.

gj: I think it would be equivalent to `isa(x, y) if x matches y;` — the `isa/3` rule in the changelog is complemented by an `isa(x, y) if isa(x, y, x);` in @samscott89's original PR (#61)

gj: de38bb57

samscott89: So `isa` is actually a token in the lexer. So that example doesn't even work -- you get a parse error on unexpected token

Also, the rule doesn't even work any more. We'll try and unify x with y (since `y` isn't picked up as a tag, it's an external instance).

And the replacement suggestion doesn't work either. `isa(x, y) if x matches y` will actually do `x matches y{}` and you get `unregistered class y`.

And we also don't seem to support `x matches (y)` so that alternative doesn't work.

In conclusion... doing `rule(x, y, x: (y))` maybe _is_ pathological... it's a _really_ horrible way to write `rule(x, y, x)`. And there is nothing useful you can currently do with that syntax.

samscott89: Concretely. I'd remove the existing changelog entry, and replace it with a super minor entry like: "removed syntax for `f(x: (y))` which does not currently achieve anything".

samscott89: Oh, fun fact. The test cases with `x: (y)` in them output `x matches y` as constraints. But if you ground either side it'll ultimately end up attempting to do `x = y`.

samscott89 (changes requested):
All looks fine, other than the changelog entry: https://github.com/osohq/oso/pull/1356/files#r743918189

docs/content/any/project/changelogs/NEXT.md
gj: > Concretely. I'd remove the existing changelog entry, and replace it with a super minor entry like: "removed syntax for `f(x: (y))` which does not currently achieve anything".

256831d5

samscott89 (approved):

PR #1357, opened 2021-11-04
No more panics during rule type validation

gj:


patrickod (approved):

patrickod:
🚢 idy do da!

patrickod (approved):
💯 love the test suite. looks solid 👌🏻

polar-core/src/kb.rs
```diff
@@ -239,15 +239,17 @@ impl KnowledgeBase {
                     Ok(RuleParamMatch::True)
                 }
             } else {
-                Err(error::OperationalError::InvalidState{msg: format!(
-                    "All registered classes must have a registered MRO. Class {} does not have a registered MRO.",
-                    &rule_instance.tag
-                )}.into())
+                // If `rule_instance.tag` were registered as a class, it would have an MRO.
+                Ok(RuleParamMatch::False(format!(
```
dhatch, line 243: Why convert this from an Error?

polar-core/src/kb.rs
```diff
@@ -239,15 +239,17 @@ impl KnowledgeBase {
...
+                    "Rule specializer {} on parameter {} is not registered as a class.",
+                    rule_instance.tag, index
+                )))
             }
         } else {
-            // TODO(gj): `rule_type_instance.tag` was registered as something other than an
-            // external instance. What should we do here?
-            unreachable!("Unregistered specializer classes should be caught before this point.");
+            Ok(RuleParamMatch::False(format!(
```
dhatch, line 249: Same question here.

dhatch: Could this message say `instead registered as a constant with value: {}`? That's the only other possible case, right?

polar-core/src/kb.rs
```diff
@@ -1479,11 +1483,167 @@ mod tests {
...
+    fn test_rule_type_validation_errors_for_non_class_specializers() {
+        let mut kb = KnowledgeBase::new();
+
+        kb.register_constant(sym!("String1"), term!("not an external instance"))
+            .unwrap();
+        kb.register_constant(sym!("String2"), term!("also not an external instance"))
+            .unwrap();
+        kb.register_constant(
+            sym!("ExternalInstanceWithoutMRO1"),
+            term!(Value::ExternalInstance(ExternalInstance {
```
dhatch, line 1506: This can't actually happen in a host library, right?

polar-core/src/kb.rs
```diff
@@ -1479,11 +1483,167 @@ mod tests {
...
+    }
+
+    #[test]
+    fn test_rule_type_validation_errors_for_non_class_specializers() {
+        let mut kb = KnowledgeBase::new();
+
+        kb.register_constant(sym!("String1"), term!("not an external instance"))
+            .unwrap();
+        kb.register_constant(sym!("String2"), term!("also not an external instance"))
+            .unwrap();
```
dhatch, line 1503: No need to make any changes but you can have a test return an error type and then use the `?` operator which can sometimes shorten all these `unwrap` s in tests a bit (you may know this).

gj: Yeah, I use that sometimes. In this case I'm not too bothered b/c the only `unwrap()`s sprinkled into the actual meat of the test (as opposed to the setup up here) are unwrapping `Option` instead of `Result`

polar-core/src/kb.rs
gj: Won't it happen if you register a class/instance with `register_constant` instead of `register_class`?

dhatch: yes nvm

polar-core/src/kb.rs
dhatch: Yeah, I'm just asking whether this should be a failure of the rule to match (`RuleParamMatch::False`) or another type of error indicating that the rule itself is invalid.

If you have registered a constant and then used that as a specializer, isn't that different type of error than just writing a rule that doesn't match a rule type? It's an invalid rule.

gj: Talked about this; message is good enough.

polar-core/src/kb.rs
gj: 5b365d63

dhatch (approved):

PR #1358, opened 2021-11-05
Make it possible to extend the isa check.

samscott89:
PR checklist:

- [ ] Added changelog entry.


samscott89:
Example usage: https://github.com/osohq/gitclub/blob/sam/prisma/backends/express-prisma/src/oso.ts#L32

languages/js/src/Host.ts
```diff
@@ -350,8 +359,14 @@ export class Host implements Required<DataFilteringQueryParams> {
     const instance = (await this.toJs(
       polarInstance
     )) as NullishOrHasConstructor;
-    const cls = this.getClass(name);
-    return instance instanceof cls || instance?.constructor === cls;
+
+    const userType = this.types.get(name);
+    if (userType !== undefined) {
```
gkaemmer, line 364: Is there a need to introduce this branch here? The old code assumed `getClass(name)` is never undefined -- the same assumption would hold for `types.get(name)`, no?

gkaemmer (commented):
This looks good to me, pretty simple. Just have one question.

languages/js/src/Host.ts
saolsen: Well, `getClass` throws the error if the class isn't registered, If we didn't fall through we'd have to handle the error here I guess. It wouldn't be valid to pass in a non-registered thing but you'd want the right error if you did.

gkaemmer: Makes sense. I suppose ideally `getClass(name)` would be replaced with `getUserType(name)`, but let's get this out.

gkaemmer (approved):

PR #1360, opened 2021-11-05
implement ExternalIsSubclass for Rust

davepacheco:
This change implements `ExternalIsSubclass` for Rust following similar behavior as for Golang.  This essentially says that a type is only a subclass of itself.  The patch is from @gj in Slack (thanks Gabe!).

This fixes a panic I ran into:

```
thread 'main' panicked at 'not implemented: Unhandled event ExternalIsSubclass { call_id: 2527, left_class_tag: Symbol("Instance"), right_class_tag: Symbol("Fleet") }', /home/dap/.cargo/registry/src/github.com-1ecc6299db9ec823/oso-0.22.0/src/query.rs:89:26
```

I'm not sure what else might be needed for this PR -- please let me know!


davepacheco:
I'm happy to sign the CLA (I'm waiting for the bot to say what I need to do), but this contribution is probably better attributed to @gk since this is the patch he sent me in Slack as-is.

davepacheco:

samscott89 (commented):
Awesome! Thanks for opening this PR.

A test here would be super helpful. This case comes about when we have a variable for which we know it's type, but we don't have a concrete version of it yet, and we need to see if it's a subclass of another type.

For example:

```
allow(user, "read", document: Document) if
   has_relation(project, "parent", document);

has_relation(project: Project, "parent", document: Document) if
   is_parent(project) and project = document.project;

is_parent(project: Project) if project.parent = nil;

is_parent(_: Document) if false;
```

Where:

```rust
struct Document {
   project: Project
}

struct Project {
   parent: Option<Project>
}
```

Bit of a strange example off the top of my head. But that should exercise the code path.

The reason is that Polar will try and check `is_parent` for the variable `project` and all it knows if `project` is a `Project`. So it needs to check whether `Projects` can be `Documents`.

davepacheco:
Thanks!  I will look at adding a test for it and reach out on Slack if I run into trouble.

davepacheco:
I took a first swing at writing a new test called test_subclass.rs.  I tried to stick to your suggestion @samscott89 but I found that the test succeeded even without my fix so I don't think it exercises the code path.  This might be a result of changes I made to your suggested policy.  The main one was that I changed Project.parent to just Project.public, since `struct Project` cannot contain another `Project` directly.  (I also tried `Option<Box<Project>>`, but `ToPolar` isn't implemented for `Box<T>`.)

I can instead try to start with the case where I originally hit this and simplify it as much as possible but I figured I'd check first to see if I'm way off the mark.  Even if we find an example that works, will we have sufficient confidence that it will continue to exercise the right code path?

davepacheco:
I've pushed commit 3664b31e, which changes the test to more closely match the most recent case where I hit this.  It triggers the `unimplemented` panic prior to my fix.  But I can't say I understand why it hits that code path or whether it's a good test case for that code path.  Note that this policy uses `==` instead of `=`, which is wrong even with the panic fixed.  But using `=` prevents this code path from being hit altogether.

Sorry for the mess of test cases.  I don't feel I understand Oso well enough to generate a simple test case here, but I'm hoping you can point me in a better direction.

samscott89:
> 
> 
> I took a first swing at writing a new test called test_subclass.rs. I tried to stick to your suggestion @samscott89 but I found that the test succeeded even without my fix so I don't think it exercises the code path. This might be a result of changes I made to your suggested policy. The main one was that I changed Project.parent to just Project.public, since `struct Project` cannot contain another `Project` directly. (I also tried `Option<Box<Project>>`, but `ToPolar` isn't implemented for `Box<T>`.)
> 
> I can instead try to start with the case where I originally hit this and simplify it as much as possible but I figured I'd check first to see if I'm way off the mark. Even if we find an example that works, will we have sufficient confidence that it will continue to exercise the right code path?

Ah, that makes sense! Yeah, it's a bit of a non trivial example in an attempt to trigger this with a real, working policy.

> 
> 
> I've pushed commit [3664b31](https://github.com/osohq/oso/commit/3664b31ef7691d928c94f7936d5815dbe258bc50), which changes the test to more closely match the most recent case where I hit this. It triggers the `unimplemented` panic prior to my fix. But I can't say I understand why it hits that code path or whether it's a good test case for that code path. Note that this policy uses `==` instead of `=`, which is wrong even with the panic fixed. But using `=` prevents this code path from being hit altogether.
> 
> Sorry for the mess of test cases. I don't feel I understand Oso well enough to generate a simple test case here, but I'm hoping you can point me in a better direction.

Thanks for trying a few things! Yeah, I see the problem with the first one. It seems like this is actually kind of a hard test case to hit in regular (non data filtering) usage. 

I think we can probably just push up a less realistic use case. I'll try one now.

languages/rust/oso/tests/test_polar_rust.rs
```diff
@@ -736,6 +736,30 @@ fn test_option() {
...
+            Foo::get_polar_class_builder()
+                .set_constructor(Foo::default)
+                .with_equality_check()
+                .build(),
+        )
+        .unwrap();
+    test.oso.register_class(Bar::get_polar_class()).unwrap();
+    test.qeval("x matches Foo and x matches Foo and x = new Foo()");
+    // should fail by checking that Foo != Bar so not a subclass
+    test.qnull("x matches Foo and x matches Bar");
```
samscott89, line 760: Went with more like a unit test in the end. A bit less inspiring, but tests the behaviour directly. I appreciate the effort to get this tested with something more realistic, thanks @davepacheco !

samscott89 (approved):
Thanks @davepacheco !

PR #1362, opened 2021-11-06
Only return early on parse errors

gj:


polar-core/src/polar.rs
```diff
@@ -255,42 +253,23 @@ impl Polar {
...
-
         // TODO(gj): need to bomb out before rule type validation in case additional rule types
         // were defined later on in the file that encountered the `ParseError`. Those additional
         // rule types might extend the valid shapes for a rule type defined in a different,
         // well-parsed file that also contains rules that don't conform to the shapes laid out in
         // the well-parsed file but *would have* conformed to the shapes laid out in the file that
         // failed to parse.
-        if diagnostics.iter().any(Diagnostic::is_error) {
+        if diagnostics.iter().any(Diagnostic::is_parse_error) {
+            // NOTE(gj): need to set context _before_ clearing the KB so we still have source info.
```
samscott89, line 263: Actually, why do we need to clear the KB? What can someone do with a partially-loaded KB?

gj: Probably don't _need_ to, but it feels slightly more coherent to end up with an empty KB in the presence of unrecoverable errors instead of ending up with some indeterminate subset of your policy loaded.

samscott89: Yeah, okay that's probably fine.

samscott89 (approved):

PR #1364, opened 2021-11-06
Clean up SingletonVariable error

gj:


polar-core/src/vm.rs
```diff
@@ -790,8 +790,7 @@ impl PolarVirtualMachine {
     }
 
     pub fn source(&self, term: &Term) -> Option<Source> {
-        term.get_source_id()
-            .and_then(|id| self.kb.read().unwrap().sources.get_source(id))
+        self.kb.read().unwrap().get_term_source(term)
```
samscott89, line 793:
```suggestion
        self.kb().get_term_source(term)
```

gj: nice, forgot we recently added that helper

samscott89 (approved):

PR #1368, opened 2021-11-09
Track context for more errors / Expand error context to a range / (gently) enforce adding context to errors

gj:


polar-core/src/parser.rs
```diff
@@ -32,37 +32,31 @@ pub enum Line {
...
         },
         ParseError::User { error } => error,
     }
 }
 
-pub fn parse_term(src: &str) -> PolarResult<Term> {
-    polar::TermParser::new()
-        .parse(0, Lexer::new(src))
-        .map_err(|e| to_parse_error(e).into())
-}
```
gj, line 64: This is the only actual change in this file. We were only using this function in the `tests` mod down below, so I merged it with `tests::parse_term`, which was previously calling this function.

polar-core/src/kb.rs
```diff
@@ -861,18 +854,30 @@ impl KnowledgeBase {
...
             let object_specializer = pattern!(instance!(&object.value().as_symbol().expect("must be symbol").0));
-            rule!("has_relation", ["subject"; subject_specializer, relation_name, "object"; object_specializer], required)
+
+            let src_id = relation.get_source_id().expect("must be parsed");
+            let (left, right) = relation.span().expect("must be parsed");
+
+            let mut params = args!("subject"; subject_specializer, relation_name, "object"; object_specializer);
+            params.reverse();
+            let body = term!(op!(And));
+            let mut rule = Rule::new_from_parser(src_id, left, right, sym!("has_relation"), params, body);
```
samscott89, line 868: We could add a `clone_with_value` method like we have for `Term` that carries across source info?

polar-core/src/error.rs
```diff
@@ -35,74 +41,182 @@ pub enum ErrorKind {
...
-                        column,
-                        include_location: false,
-                    });
+    pub fn set_context(&mut self, source: Source) {
+        if let Some(span) = self.span() {
+            let range = Range::from_span(&source.src, span);
+            self.context.replace(ErrorContext { source, range });
+        } else {
+            panic!("set_context w/o span: {}", self);
+        }
```
gj, line 84: Will remove prior to shipping

polar-core/src/resource_block.rs
```diff
@@ -46,68 +43,63 @@ enum ParsedDeclaration {
     Relations(Term),   // Dict<Symbol, Symbol>
 }
 
-fn validate_parsed_declaration((name, term): (Term, Term)) -> PolarResult<ParsedDeclaration> {
-    match (name.value().as_symbol()?.0.as_ref(), term.value()) {
+fn validate_parsed_declaration(
+    (name, term): (Term, Term),
+) -> Result<ParsedDeclaration, ValidationError> {
+    match (name.value().as_symbol().expect("parsed as symbol").0.as_ref(), term.value()) {
```
gj, line 49: Same as previous comment in `kb.rs` — we _know_ `name` is a `Term<Variable>` but we can't (yet) express that certainty via the type.

dhatch:
(Copying some context from @gj from quill)


1. We enforce more rigidity w/r/t only creating validation errors while validating or parse errors while parsing by changing return signatures from Result<_, PolarError> to Result<_, ValidationError>. This makes it harder to (accidentally or otherwise) use the wrong type of error for a particular situation.
2. We add error context when elevating an error variant into a PolarError. The error variants don't contain source context, but PolarError does, so I think it's a logical point to add source context.
3. Defining the From impls to take a Source (in the case of ParseError) or a &KnowledgeBase (in the case of RuntimeError & ValidationError) forces folks to think about adding error context when elevating an error variant into a PolarError (which must happen prior to emission via the public APIs, which all expect PolarError).
4. Defining the From impl for each individual error variant provides a uniform structure for defining how to elevate each variant into a PolarError. Previously, there were two different flows. For ParseError, we already had the relevant Source in hand (from, you know, parsing it), so all we needed was to calculate the relevant span from the error and then use the source & span to create an ErrorContext. For RuntimeError & ValidationError, we would immediately turn them into PolarError with empty context, and then at some point between then and when the error was emitted through one of the public APIs we'd hopefully attach context for it. If we remembered to attach context before emitting it, we'd have to construct that context using SourceInfo::Parser info attached to the error via a relevant term or rule. That source info contains left & right lexical bookends and a src_id, which we need a KB to turn into an actual source. The impl From<(ParseError, Source)> for PolarError and impl From<(ValidationError, &KnowledgeBase)> for PolarError signatures tell you exactly what each variant needs in order to be elevated into a PolarError.

dhatch (changes requested):
To preface this, I'm not completely sure about the goal of this PR, so without that understanding my comments may be off base.


It seems like this change has the assumption (in moving away from `PolarResult`) that each error variant is going to be 1:1 tied to a part of the code. But, `OperationalError` is supposed to capture errors that should never occur for users and indicate a programming error on our part, and `InvalidState` could occur anywhere. (I think adding `Runtime::Unsupported` to distinguish from `Unimplemented` is a good idea, although I'm not sure if we need `Unimplemented` then.)

`InvalidState` could occur anywhere, parsing, validation, or at runtime. If we want to preserve `OperationalError` as another top level hierarchy then using `Result<..., RuntimeError>` instead of`PolarResult` doesn't make too much sense.

I think another goal of this change was to try to ensure we always attach context to errors. But there's nothing in the type system to help us know whether an error has context, especially because `PolarError` still allows context `None`. We still have `Into` implemented from `RuntimeError` => `PolarError`, so there isn't much to keep us from inadvertently converting an error & not including source information.

Why not do something similar to what we do with Term? A new type that indicates an error with contextual information.

```
pub struct ErrorWithContext<E = PolarError> {
    inner: E
    context: ErrorContext
}

# Make conversions easy
impl <A, B> From<ErrorWithContext<A>> for ErrorContext<B> where
A: Into<B> {
    ...
}
```

Unless I am missing something, I'm not sure we even need to keep the lazy addition of error context (`with_context`) rather than immediately including context in the construction of every error. For `RuntimeError` it looks like this would be easy for every error construction (we always have access to the KB when the error is constructed) besides those coming from a Runnable. Looks easy for `ValidationError` as well. For the parser, we just need to pass `Source` to `parse_lines` and `parse_query` and use it in `to_parse_error`. If it's impossible to construct an error without proper context, we don't need to worry about any of the above.

polar-core/src/bindings.rs
```diff
@@ -157,12 +157,16 @@ impl BindingManager {
     ///
     /// If the binding succeeds, the new expression is returned as a goal. Otherwise,
     /// an error is returned.
-    fn partial_bind(&mut self, partial: Operation, var: &Symbol, val: Term) -> PolarResult<Goal> {
+    fn partial_bind(
+        &mut self,
+        partial: Operation,
+        var: &Symbol,
+        val: Term,
+    ) -> Result<Goal, RuntimeError> {
```
samscott89, line 165: Nit: is there any benefit to seeing the runtime error here? I think most files only expose one error type, so perhaps `type Result<T> = Result<T, RuntimeError>` etc. at the top of the file?

polar-core/src/kb.rs
```diff
@@ -300,7 +286,7 @@ impl KnowledgeBase {
                             // Turn `member` into an `InstanceLiteral` by copying fields from
                             // `rule_type_instance`.
                             let rule_type_instance = InstanceLiteral {
-                                tag: member.value().as_symbol()?.clone(),
+                                tag: member.value().as_symbol().expect("parsed as symbol").clone(),
```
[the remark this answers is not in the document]
samscott89: Okay... as a rule I way prefer us to handle the panics. But I see that this is in the loading stage (kinda cool that scanning to see it was a ValidationError told me where this was happening).

I'd be okay with making an exception for these panics if in parsing/file loading. The rationale being that typically any of these being invalid here is a serious problem. Compared to them occurring at runtime (which is a little more common sadly).

polar-core/src/kb.rs
```diff
@@ -476,11 +462,17 @@ impl KnowledgeBase {
...
+                                    // past the parser or is it unreachable? Prior to #1356 we
+                                    // could hit this branch with a `Value::Variable` if the
+                                    // specializer in the rule head was parenthesized.
+                                    return Err(ValidationError::InvalidRule {
+                                        msg: format!(
+                                            "Value variant {} cannot be a specializer",
+                                            rule_value
+                                        ),
+                                        rule: rule.clone(),
+                                    });
```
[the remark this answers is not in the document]
samscott89: Hmm, yeah can't think of anything that could go here.

polar-core/src/kb.rs
```diff
@@ -664,86 +666,48 @@ impl KnowledgeBase {
...
-                }
-                .into());
+                })
             }
             _ => {}
         }
         Ok(())
     }
 
-    pub fn set_error_context(&self, term: &Term, error: impl Into<PolarError>) -> PolarError {
```
samscott89, line 701: 👋

polar-core/src/parser.rs
```diff
@@ -32,65 +32,60 @@ pub enum Line {
 }
 
 fn to_parse_error(e: ParseError<usize, lexer::Token, error::ParseError>) -> error::ParseError {
+    use error::ParseError::*;
```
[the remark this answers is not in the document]
samscott89: > Also why isn't this just an Into implementation?

This actually works better. You end up with `parse_things(...).map_err(to_parse_error)...`

instead of `.map_err(std::into::Into::into)` or `.map_err(|e| e.into())`.

Since we wouldn't see any parse errors outside of this file, the value of implementing a trait seems low.

polar-core/src/kb.rs
samscott89: Ah, sadly we might still need some of this.

For the unhandled partial error. The term that's in the error doesn't itself have source context, since it's a partial constructed in the core. This GetSource visitor was walking the partial looking for the first term that had source info (for example, the original variable that was created).

That context was nice... could we bring it back?

polar-core/src/polar.rs
```diff
@@ -176,9 +175,11 @@ impl Polar {
 
     /// Load `Source`s into the KB.
     pub fn load(&self, sources: Vec<Source>) -> PolarResult<()> {
-        if self.kb.read().unwrap().has_rules() {
-            let msg = MULTIPLE_LOAD_ERROR_MSG.to_owned();
-            return Err(RuntimeError::FileLoading { msg }.into());
+        if let Ok(kb) = self.kb.read() {
+            if kb.has_rules() {
+                let msg = MULTIPLE_LOAD_ERROR_MSG.to_owned();
+                return Err(RuntimeError::FileLoading { msg }.with_context(&*kb));
```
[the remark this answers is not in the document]
samscott89: Realistically, this isn't a runtime error now that files only get loaded once.

Also, agree with this decision to have one API.

gj:
> But, `OperationalError` is supposed to capture errors that should never occur for users and indicate a programming error on our part and `InvalidState` could occur anywhere. (I think adding `Runtime::Unsupported` to distinguish from `Unimplemented` is a good idea, although I'm not sure if we need `Unimplemented` then.)

Did you look at the PR that this one is branched off of, #1389? 

> I think another goal of this change was to try to ensure we always attach context to errors. But there's nothing in the type system to help us know whether an error has context, especially because `PolarError` still allows context `None`.

The assurance comes from (gently) forcing folks to attach context when they upgrade an error variant into a `PolarError`. As of this PR, there are two ways to go from `RuntimeError`, `ValidationError`, and `ParseError` to `PolarError`:

1. Manually construct a `PolarError` from the variant.
2. Use the `with_context()` API, which adds context.

> We still have `Into` implemented from `RuntimeError` => `PolarError`, so there isn't much to keep us from inadvertently converting an error & not including source information.

Where are you seeing `Into` implemented from `RuntimeError` => `PolarError`?

> Unless I am missing something, I'm not sure we even need to keep the lazy addition of error context (`with_context`) rather than immediately including context in the construction of every error.

We might be able to do this; I thought about it but decided to punt since it would've involved more extensive changes.

> For `RuntimeError` it looks like this would be easy for every error construction (we always have access to the KB when the error is constructed) besides those coming from a Runnable.

"besides" is doing a lot of work in that sentence 😄

polar-language-server/src/lib.rs
```diff
@@ -284,7 +295,9 @@ impl PolarLanguageServer {
                     range: range_from_polar_error_context(e),
                     severity: Some(DiagnosticSeverity::Error),
                     source: Some("Polar Language Server".to_owned()),
-                    message: e.to_string(),
+                    // NOTE(gj): We stringify the `ErrorKind` instead of the full `PolarError`
```
samscott89, line 298: smart

polar-wasm-api/tests/polar.rs
```diff
@@ -33,10 +33,10 @@ fn load_file_errors() {
     let err = polar.wasm_load(sources).unwrap_err();
     let err: Error = err.dyn_into().unwrap();
     assert_eq!(err.name(), "ParseError::UnrecognizedToken");
-    assert_eq!(
-        err.message(),
-        "did not expect to find the token ';' at line 1, column 1"
-    );
+    assert!(err.message().starts_with(
+        "did not expect to find the token ';' at line 1, column 1",
+        0
```
samscott89, line 38: what's this 0 doing here?

samscott89: Oh, it's calling the JS method! Huh

samscott89 (changes requested):
Super minor changes, then this is good to go IMO. Although didn't see the giant review from @dhatch ...

polar-wasm-api/tests/polar.rs
gj: https://docs.rs/js-sys/0.3.44/js_sys/struct.JsString.html#method.starts_with

dhatch:
> "besides" is doing a lot of work in that sentence 😄

I thought so initially, but I think we could just require that all Runnables have a ref to the KB if they want to construct errors.
We could also get rid of `sources` map on KB & just give terms from the parser a pointer to the source with an Arc.

dhatch:
> Did you look at the PR that this one is branched off of, #1389?

Just the description. I don't think RuntimeErrors should be all errors that occur when running queries.

`RuntimeError`s are user caused (invalid policy, stack overflow, etc).
`OperationalError`s are caused by us (we wrote incorrect serialization code, we have a bug leading to InvalidState, we didn't support an operations, etc.)

polar-core/src/kb.rs
gj: https://github.com/osohq/oso/commit/d2644423b47bb562b94fc88edbac7207da2b0ef8

polar-core/src/parser.rs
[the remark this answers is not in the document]
gj: > I find this function confusing because of all the enum variants with the same names but from different namespaces, it was less confusing before when the `error::` namespace on the RHS was more explicit.

9ae45835

samscott89 (approved):

polar-core/src/bindings.rs
gj: 5bdc9985

PR #1369, opened 2021-11-09
Remove sqlalchemy-oso version.

dhatch:
The version sometimes drifts from the version of the `oso` package.

This PR removes the version from the docs page. `pip install sqlalchemy-oso`
will install the latest version, so the version pin is unnecessary.


gj (changes requested):
What about Flask & Django? Should we just remove the `{{< version >}}` specifier entirely?

dhatch:
Removed from both of those. It's fine to have it for the `oso` package, it just gets out of date with the sub packages.

gj (approved):

PR #1374, opened 2021-11-10
Add language to title on Quickstart, Add to App, RBAC & Hierarchies pages

dhatch:
- Adds a new `metaTitle` param to page front matter that will substitute LANG
  for the current pages language, allowing us to customize the title used by
  search to include the language.
- Adds to H1 the current language on some pages that we want to appear for SEO.

### DIFF
--- a/docs/content/any/getting-started/application/enforce/index.md
+++ b/docs/content/any/getting-started/application/enforce/index.md
@@ -1,12 +1,13 @@
 ---
 title: Enforce authorization
+metaTitle: Enforce authorization in LANG
 description: |
     Add authorization enforcement throughout your application using the
     authorize API to reject or accept requests that users make.
 weight: 2
 ---
 
-# Enforce authorization
+# Enforce authorization in {{% lang %}}
 
 In [Model your authorization logic](model) you defined an authorization
 policy. In this guide, we will cover using Oso's enforcement API to
--- a/docs/content/any/getting-started/application/enforce/index.md
+++ b/docs/content/any/getting-started/application/enforce/index.md
@@ -1,6 +1,6 @@
 ---
 title: Enforce authorization
-metaTitle: Enforce authorization in LANG
+metaTitle: Enforce authorization in $LANG
 description: |
     Add authorization enforcement throughout your application using the
     authorize API to reject or accept requests that users make.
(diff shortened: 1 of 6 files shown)


docs/content/any/getting-started/application/enforce/index.md
```diff
@@ -1,12 +1,13 @@
 ---
 title: Enforce authorization
+metaTitle: Enforce authorization in LANG
```
samscott89, line 3: Maybe `$LANG` to make it more obvious it's interpolated?

docs/themes/oso-tailwind/layouts/partials/head.html
```diff
@@ -2,11 +2,16 @@
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
     <title>{{ block "title" . }}
+        {{- if .Params.metaTitle -}}
+        {{- replace .Params.metaTitle "LANG" .Page.Language.LanguageName }} - {{ .Site.Title -}}
+        {{- else -}}
         {{- if .Title -}}
```
samscott89, line 8: Make this an `else if`

samscott89 (commented):
Looking at the source for the docs, we also have twitter + open graph titles (used when creating link snippets). I don't think those will currently be updated based on the change. But that might not matter

docs/themes/oso-tailwind/layouts/partials/head.html
```diff
@@ -2,11 +2,16 @@
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
     <title>{{ block "title" . }}
+        {{- if .Params.metaTitle -}}
+        {{- replace .Params.metaTitle "LANG" .Page.Language.LanguageName }} - {{ .Site.Title -}}
```
samscott89, line 6: I wonder whether we actually want the end of this " - Oso Documentation". Is this good for SEO?

samscott89: (Same for non-meta title changes)

dhatch: I think it's fine but I'll check with Stephie

dhatch:
> Looking at the source for the docs, we also have twitter + open graph titles (used when creating link snippets). I don't think those will currently be updated based on the change. But that might not matter

Yeah no way to do those as far as I can tell unless we vendor those templates & add the string replacement to them.

samscott89 (approved):

PR #1376, opened 2021-11-11
Fix bug in formatting::source_lines & add tests

gj:


patrickod (approved):

polar-core/src/formatting.rs
```diff
@@ -74,10 +74,15 @@ pub fn source_lines(source: &Source, offset: usize, num_lines: usize) -> String
     let mut lines = Vec::new();
     let mut target = None;
     let prefix_len = "123: ".len();
+    // TODO(gj): test that this line walking code below works for terms/rules parsed from a file
+    // with \r\n line endings. Worried the term/rule's offset might take the extra \r character
+    // into account whereas str::lines makes no distinction between \r\n and \n.
+    //
+    // TODO(gj): There's probably a better way to rewrite this by walking `source.src.chars()`.
```
gj, line 81:
```suggestion
    // TODO(gj): There's probably a better way to rewrite this by walking `source.src.char_indices()`.
```

polar-core/src/formatting.rs
```diff
@@ -57,36 +54,27 @@ impl Trace {
...
-        }
-        lines.push(line);
-    };
-    let mut index = 0;
-    let mut lines = Vec::new();
-    let mut target = None;
+    let (target_line, target_column) = loc_to_pos(&source.src, offset);
+    // Skip everything up to the first line of requested context (`target_line - num_lines`), but
+    // don't overflow if `num_lines > target_line`.
+    let skipped_lines = target_line.saturating_sub(num_lines);
```
patrickod, line 65: TIL about `saturating_sub`. neat.

patrickod: I tripped up reading this at first because I had in my head that `num_lines` was the _total number_ of lines to be returned, rather than the number of _lines of context either side_ of the match to be returned. maybe `s/num_lines/context_lines/` ?

gj: Same! I was overflowing (underflowing? lol) and needed to figure out a way to spell "subtract down to zero but no further" in Rust

gj: > I tripped up reading this at first because I had in my head that `num_lines` was the _total number_ of lines to be returned, rather than the number of _lines of context either side_ of the match to be returned. maybe `s/num_lines/context_lines/` ?

That's great feedback; thanks — I'll push a change.

gj: 5a4a8643

polar-core/src/formatting.rs
```diff
@@ -57,36 +54,27 @@ impl Trace {
...
-            push_line(&mut lines, format!("{}^", spaces));
-        }
-        index = end;
-        if target.is_some() && lineno == target.unwrap() + num_lines {
-            break;
-        }
-    }
+    // Insert 'indicator' line pointing at `target_column`.
+    let indicator_line = format!("{}^", " ".repeat(prefix_len + target_column));
+    lines.insert(target_line - skipped_lines + 1, indicator_line);
```
patrickod, line 77: I might be mistaken, but could this be simplified to `lines.insert(num_context_lines + 1, indicator_line)` ?

patrickod: indexing the shorter `Vec` which is only `(num_context_lines * 2) + 1` items long by using the indices of the wider `lines` Vec was confusing

gj: > I might be mistaken, but could this be simplified to `lines.insert(num_context_lines + 1, indicator_line)` ?

Is `num_context_lines` the third function argument (now called `context_lines`)? If so, I don't think that holds true for the case where you say you'd like 2 lines of context on either side of the target line but the source text only contains 1 line. You'd be trying to insert at position `3` into a `Vec` of length `1`.

gj: > indexing the shorter `Vec` which is only `(num_context_lines * 2) + 1` items long by using the indices of the wider `lines` Vec was confusing

But it might not actually be `(context_lines * 2) + 1` items long — that's just the _maximum_ size the window can be

patrickod: ah yes! of course.

gj: (to be clear: I definitely agree that it's confusing 😆)

patrickod (approved):

PR #1378, opened 2021-11-12
Fix filtering logic when using `in`

samscott89:


languages/python/sqlalchemy-oso/tests/test_post_relationship.py
```diff
@@ -784,6 +792,15 @@ def test_partial_isa_with_path(session, oso, tag_nested_many_many_test_fixture):
 
     assert len(posts) == 5
 
+    tags = session.query(Tag).filter(authorize_model(oso, user, "read", session, Tag))
```
samscott89, line 795: Against main, this test fails with: 

```
field_name = 'is_public', value = True, operator = 'Unify', session = <sqlalchemy.orm.session.Session object at 0x7fbd382e2e20>, model = <class 'tests.models.Post'>

    def emit_compare(field_name, value, operator, session, model):
        """Emit a comparison operation comparing the value of ``field_name`` on ``model`` to ``value``."""
        assert not isinstance(value, Variable), "value is a variable"
>       property = getattr(model, field_name)
E       AttributeError: type object 'Post' has no attribute 'is_public'

sqlalchemy_oso/partial.py:352: AttributeError
```

Since all the `check` rules will pass type checking, and it incorrectly attempts to check `is_public` on a post.

polar-core/src/partial/partial.rs
```diff
@@ -1007,6 +1023,17 @@ mod test {
                i(y: C) if y.c > 0;
                i(y: B{bar: 2}) if y.b > 0;
 
+               # traversing `in` is tough!
+               r(a: A) if b in a.b and t(b);
+               s(a: A) if b in a.b and b matches B and u(b.c);
+               # This is still incorrect
+               # this should be equivalent to `s`
+               s_bad(a: A) if b in a.b and u(b.c);
```
samscott89, line 1031: This is only a partial fix. If we don't know the type of `b`, then we can't do the isa with path check to find if `b.c matches C`. The problem is that the partial will look like: `_this matches A{} and b in _this.b` and we are checking `b.c matches C{}`. But we won't be able to "traverse" that `in` to check if `a.b.c matches C`. 

On the other hand, it'll work fine for `b in _this.b` checking if `b matches B{}`, because the new logic here will substitute `_this.b` for `b`. and so check whether `_this.b matches B`. Which currently works in the sqlalchemy logic.

gw000 (approved):
:this_is_fine.gif:

languages/python/sqlalchemy-oso/tests/test_post_relationship.py
samscott89: This is what this PR is addressing. Previously `post` from `post in tag.posts` would be "untyped", so it would match all of the `check` rules. With this PR, we will check `tag.posts matches Post`, etc. So only the right rule gets applied.

PR #1380, opened 2021-11-12
Clean up more runtime errors

gj:


polar-core/src/vm.rs
```diff
@@ -1401,17 +1401,14 @@ impl PolarVirtualMachine {
         }
     }
 
-    pub fn check_error(&self) -> PolarResult<QueryEvent> {
-        if let Some(error) = &self.external_error {
+    pub fn check_error(&mut self) -> PolarResult<QueryEvent> {
+        if let Some(msg) = self.external_error.take() {
```
gj, line 1405: I inspected the other uses of `self.external_error`, and this move seems chill. Could probably use a second pair of 👀

samscott89 (approved):

polar-core/src/vm.rs
gj: Talked offline. @samscott89 "trusts my judgment" — only time will tell how poor his own is

PR #1383, opened 2021-11-12
Remove reliance on KB::set_error_context for runtime errors in VM

gj:


samscott89 (approved):

polar-core/src/vm.rs
```diff
@@ -178,7 +178,12 @@ fn invalid_state<A>(msg: String) -> PolarResult<A> {
     Err(OperationalError::InvalidState { msg }.into())
 }
 
-pub fn compare(op: Operator, left: &Term, right: &Term) -> PolarResult<bool> {
+pub fn compare(
+    op: Operator,
+    left: &Term,
+    right: &Term,
+    context: Option<&Term>,
```
samscott89, line 185: Not a huge fan of needing this extra param, but also don't really mind

gj: Yeah I agree. It's the type of thing that having a less-opaque `Term` type would improve a lot. Being able to refine an opaque `Term<Value>` to a `Term<BinaryOperation>` and then pass the latter around would be a much better experience than this snippet, where we've already pattern-matched a `Term` down to a binary op and want to pass around the constituent parts (instead of having to re-extract them) and the original term (for source info).

PR #1387, opened 2021-11-12
Fix Go test flake

samscott89:
Pretty sure this fixed the "End of JSON input" test flake. I was able to repro it locally.

Seems like the problem was some race condition with multiple threads accessing the error in the refcell. So we were returning a null error here: https://github.com/osohq/oso/compare/sam/fix-go-flake?expand=1#diff-42d3477e8e5600703b52d91104159a7691531532173e0e7e41ead2a819e4c5c6R71

(I added a panic there and confirmed it was reaching that line).

RefCells aren't thread safe, and it seems like Go running tests multithreaded is hitting a race condition here.

### DIFF
--- a/polar-c-api/src/lib.rs
+++ b/polar-c-api/src/lib.rs
@@ -2,11 +2,11 @@ pub use polar_core::polar::Polar;
 pub use polar_core::query::Query;
 use polar_core::{error, terms};
 
-use std::cell::RefCell;
 use std::ffi::{CStr, CString};
 use std::os::raw::c_char;
 use std::panic::{catch_unwind, AssertUnwindSafe};
 use std::ptr::{null, null_mut};
+use std::sync::{Arc, Mutex};
 
 /// Get a reference to an object from a pointer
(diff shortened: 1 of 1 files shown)


polar-c-api/src/lib.rs
```diff
@@ -2,11 +2,11 @@ pub use polar_core::polar::Polar;
 pub use polar_core::query::Query;
 use polar_core::{error, terms};
 
-use std::cell::RefCell;
 use std::ffi::{CStr, CString};
 use std::os::raw::c_char;
 use std::panic::{catch_unwind, AssertUnwindSafe};
 use std::ptr::{null, null_mut};
+use std::sync::{Arc, Mutex};
```
samscott89, line 9:
```suggestion
use std::sync::Mutex;
```

saolsen (approved):
Nice find.
Alternatively we could put the error on the polar object so it's not a global thing at all, would be a bit more work probably.

PR #1390, opened 2021-11-15
Make FFI errors not thread local

samscott89:


languages/go/internal/ffi/native/polar.h
```diff
@@ -11,31 +11,60 @@ typedef struct polar_Polar polar_Polar;
 
 typedef struct polar_Query polar_Query;
 
-const char *polar_get_error(void);
+/**
+ * Wrapper struct to help us return errors
+ * We Go now
```
gj, line 16: We Go now?

languages/python/oso/polar/ffi.py
```diff
@@ -127,7 +127,7 @@ def __del__(self):
     def call_result(self, call_id, value):
         """Make an external call and propagate FFI errors."""
         if value is None:
-            value = ffi.NULL
+            value = to_c_str("null")
```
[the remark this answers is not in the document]
samscott89: Yeah. On the Rust side, `value` is an optional value. We were previously doing a bit of weird logic in the polar-c-api to check if `value == nil` then make it `None`, otherwise deserialize.

But `null` deserializes to `None`, so this handles this directly.

Oh, `None` already serialializes to "null", so this if can be removed entirely.

samscott89 (commented):
Initial approach here doesn't seem great. Returning the struct and accessing the members on the ffi side is a bit awkward (doesn't seem to work for Java atm)

saolsen:
> Initial approach here doesn't seem great. Returning the struct and accessing the members on the ffi side is a bit awkward (doesn't seem to work for Java atm)

Yeah, I was a little worried that would be hard. Could instead serialize the result type itself to json. Then every function would just return a json result.

samscott89:
> Yeah, I was a little worried that would be hard. Could instead serialize the result type itself to json. Then every function would just return a json result.

Yeah, that's a decent idea. It's a little awkward since we have methods that return `*mut Query`, so those pointers would need to be deserialized and then... casted?

I tried to make it return pointers to the result struct instead. The result is pretty ugly still.

languages/java/oso/src/main/java/com/osohq/oso/Ffi.java
```diff
@@ -15,7 +16,52 @@
...
+    private final Pointer error = new Pointer();
+
+    // Necessary constructor that takes a Runtime
+    public CResultVoid(jnr.ffi.Runtime runtime) {
+      super(runtime);
+    }
+
+    public void check() throws Exceptions.OsoException {
+      jnr.ffi.Pointer e = this.error.get();
+      long r = this.result.get();
```
samscott89, line 55:
```suggestion
      long r = this.result.get();
      polarLib.result_free(getMemory(this));
```

samscott89: Crap

languages/java/oso/src/main/java/com/osohq/oso/Ffi.java
```diff
@@ -68,16 +113,21 @@ protected Query nextInlineQuery() throws Exceptions.OsoException {
...
     }
 
-    protected int registerConstant(String value, String name) throws Exceptions.OsoException {
-      return checkResult(polarLib.polar_register_constant(ptr, name, value));
+    protected void registerConstant(String value, String name) throws Exceptions.OsoException {
+      CResultVoid result = polarLib.polar_register_constant(ptr, name, value);
+      // System.out.println(result);
+      // if(result == null) {
+
+      // }
```
samscott89, line 121:
```suggestion
```

languages/ruby/lib/oso/polar/ffi/message.rb
```diff
@@ -1,40 +0,0 @@
-# frozen_string_literal: true
```
samscott89, line 1: @gj I can back out these changes. But I couldn't see what these were achieving other than obscuring what the types were? These are all just strings at the end of the day. Any objections/rationale I'm missing?

gj: works for me

languages/python/oso/polar/ffi.py
```diff
@@ -183,43 +186,28 @@ def process_messages(self):
...
+
+    r = result.result
+    e = result.error
+    # This is a bit of a hack. All CResult struct types
+    # store two pointers: one to a result, one to an error
+    # result_free doesn't care about what the pointers actually
+    # point to though, so we can use the same result_free method
+    lib.result_free(ffi.cast("polar_CResult_c_void *", result))
+    if is_null(e):
+        return r
```
dhatch, line 206: Assert result is non null?

languages/python/oso/polar/ffi.py
```diff
@@ -183,43 +186,28 @@ def process_messages(self):
...
+    else:
+        error_str = read_c_str(e)
```
dhatch, line 208: Assert result is null

polar-c-api/src/lib.rs
```diff
@@ -199,54 +175,40 @@ pub extern "C" fn polar_new_query(
...
-                null_mut()
-            }
-        }
+        polar.new_query(&s, trace).map(|q| box_ptr!(q))
     })
 }
 
 #[no_mangle]
-pub extern "C" fn polar_next_polar_message(polar_ptr: *mut Polar) -> *const c_char {
+pub extern "C" fn polar_next_polar_message(polar_ptr: *mut Polar) -> *mut CResult<c_char> {
```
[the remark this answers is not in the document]
dhatch: Why not return a `CResult` by value rather than as a pointer? Then you can avoid worrying about `result_free`.

languages/go/internal/ffi/ffi.go
```diff
@@ -23,10 +23,13 @@ import (
...
 */
-func readStr(cStr *C.char) string {
+func readStr(cStr *C.char) *string {
+	if cStr == nil {
+		return nil
+	}
 	goStr := C.GoString(cStr)
 	C.string_free(cStr)
-	return goStr
+	return &goStr
```
dhatch, line 32: why?

languages/ruby/lib/oso/polar/ffi.rb
```diff
@@ -28,39 +36,34 @@ def self.release(ptr)
...
-        def self.release(ptr)
-          Rust.free(ptr) unless ptr.null?
-        end
-      end
+      # Defines the result type version of
+      # each of these structs
+      # result(T) => { result: T, error: string }
+      #
+      # We have a bunch more here than in the other language
+      # because
```
dhatch, line 57: because?

dhatch (changes requested):
Looks pretty good. High level, I am wondering why we can't just return `CResult` rather than `*mut CResult` from all these functions. Returning a small struct like this is pretty fast & would take one more free out of the equation.

If we keep `*mut CResult` we could think about adding two functions `c_result_get_err(*const CResult) *string`, `c_result_get_result(...) *void` to avoid requiring all the libraries to understand the memory layout of the Result struct. That way we can change the layout in the future without changing every library.

samscott89:
> Looks pretty good. High level, I am wondering why we can't just return `CResult` rather than `*mut CResult` from all these functions. Returning a small struct like this is pretty fast & would take one more free out of the equation.

Tried that first. But the Java library doesn't support them. 

https://github.com/jnr/jnr-ffi/issues/14

You're right though, it was a lot easier. 

> If we keep `*mut CResult` we could think about adding two functions `c_result_get_err(*const CResult) *string`, `c_result_get_result(...) *void` to avoid requiring all the libraries to understand the memory layout of the Result struct. That way we can change the layout in the future without changing every library.

I like that. Current approach is still based on returning by value

polar-c-api/src/lib.rs
```diff
@@ -1,12 +1,49 @@
+// everything in this file is unsafe, clippy
+#![allow(clippy::not_unsafe_ptr_arg_deref)]
```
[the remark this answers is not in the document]
samscott89: Yeah. We can mark all the functions as unsafe. Just seems a little redundant since these are all extern anyway.

polar-c-api/src/lib.rs
```diff
@@ -1,12 +1,49 @@
...
+use std::ffi::{c_void, CStr, CString};
 use std::os::raw::c_char;
 use std::panic::{catch_unwind, AssertUnwindSafe};
 use std::ptr::{null, null_mut};
-use std::sync::Mutex;
+
+/// Wrapper struct to help us return errors
+#[repr(C)]
+pub struct CResult<T> {
+    pub result: *mut T,
```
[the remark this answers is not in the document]
samscott89: Short answer is because we are passing ownership of the memory back to the host. E.g. when you create a query. I doubt it actually changes anything since no other language will do anything with that information?

languages/python/oso/polar/ffi.py
samscott89: There is a weird case where we return a null pointer back to represent none. I think it's in next message

languages/ruby/lib/oso/polar/ffi.rb
samscott89: Oh lol I think halfway through the thought I realized we could just delete the other classes :)

samscott89:
> > If we keep `*mut CResult` we could think about adding two functions `c_result_get_err(*const CResult) *string`, `c_result_get_result(...) *void` to avoid requiring all the libraries to understand the memory layout of the Result struct. That way we can change the layout in the future without changing every library.
> 
> I like that. Current approach is still based on returning by value

Tried this. Don't love it. We either need to add methods for each generic variant of `CResult` (void, string, query), or to do more casting shenanigans when we get a pointer _back_ from the result method. The former is extra code, and it's harder to ensure you've implemented all the right ones. The latter is even less robust in the host libraries (those that can actually use the pointer types from the header to do type-checking, like Go).

What kind of layout changes are you picturing that wouldn't require rewriting the libraries anyway?

polar-c-api/src/lib.rs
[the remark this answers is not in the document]
dhatch: These are all considered unsafe based on Rust's definition of unsafe (potential for undefined behavior) from what I can tell since they access potentially uninitialized memory.

polar-c-api/src/lib.rs
```diff
@@ -377,62 +309,52 @@ pub extern "C" fn polar_bind(
...
 pub extern "C" fn query_free(query: *mut Query) -> i32 {
-    ffi_try!({
-        std::mem::drop(unsafe { Box::from_raw(query) });
-        POLAR_SUCCESS
-    })
+    std::mem::drop(unsafe { Box::from_raw(query) });
+    POLAR_SUCCESS
+}
+
+/// Recovers the original boxed version of `query` so that
```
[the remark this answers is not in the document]
dhatch: Still need to update this comment

dhatch:
> Tried this. Don't love it. We either need to add methods for each generic variant of CResult (void, string, query), or to do more casting shenanigans when we get a pointer back from the result method. The former is extra code, and it's harder to ensure you've implemented all the right ones. The latter is even less robust in the host libraries (those that can actually use the pointer types from the header to do type-checking, like Go).

Makes sense.

languages/go/internal/ffi/ffi.go
samscott89: Because there's the next message API which returns an optional message, and so sometimes returns `nil`. So we need this codepath to make that correctly map `Option<C.char>` to `Option<string>`

polar-c-api/src/lib.rs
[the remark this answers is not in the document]
samscott89: Done

dhatch (commented):
Mostly LGTM.

Does this need a changelog entry?
And are the GitClub submodule changes intentional or should those be reverted before merging? (It looks like this changes the submodule commit)

dhatch (approved):

samscott89:
> Mostly LGTM.
> 
> Does this need a changelog entry? And are the GitClub submodule changes intentional or should those be reverted before merging? (It looks like this changes the submodule commit)

Argh, good catch. That was a mistake. Nope, no user-facing changes. I can add a short note on the bug it fixes though

PR #1393, opened 2021-11-15
Structured warnings

gj:


polar-core/src/warning.rs
```diff
@@ -0,0 +1,149 @@
+use std::fmt;
+
+use indoc::indoc;
+
+use super::diagnostic::{Context, Range};
+use super::kb::KnowledgeBase;
+use super::terms::{InstanceLiteral, Pattern, Symbol, Term, Value};
+
+#[derive(Debug)]
+pub struct PolarWarning {
```
dhatch, line 10: If this type has the same structure as `Error` why is it a different type? Is this the intended final state, or just a stepping stone?

gj: Stepping stone. All we need for the current LSP work is enough structure on warnings to translate them into VSCode diagnostics. After #1368, we have enough structure on errors to translate them into VSCode diagnostics, so this PR just aligns warnings with errors. I think we can deal with the duplication (via a trait or something) in some future work on diagnostics that isn't specific to the LSP project. Sound good?

polar-core/src/validations.rs
```diff
@@ -218,16 +166,10 @@ impl ResourceBlocksMissingHasPermissionVisitor {
...
-
-  allow(actor, action, resource) if
-      has_permission(actor, action, resource);
-
-For more information about resource blocks, see https://docs.osohq.com/any/reference/polar/polar-syntax.html#actor-and-resource-blocks".to_string()
-            ));
+            return Some(Diagnostic::Warning(PolarWarning {
+                kind: ValidationWarning::MissingHasPermissionRule,
+                context: None,
+            }));
```
gj, line 172: @dhatch this is the lone instance of manually constructing a `PolarWarning` that I mentioned on the call this morning. We know the `MissingHasPermissionRule` won't ever have source context and we don't have access to a KB in this context, so rather than pass a KB in I decided to manually construct the warning w/o context.

samscott89: Wouldn't the typical approach like with errors to be to return this as a ValidationWarning instead? And do the wrapping elsewhere. This is totally unimportant, just curious if there is a distinction between how we treat warnings

polar-core/Cargo.toml
```diff
@@ -32,7 +32,6 @@ permute = "0.1.0"
 pipe = "0.4.0"
 pretty_assertions = "1.0.0"
 maplit = "1.0.2"
-indoc = "1.0.3"
```
samscott89, line 35: Is this right? It looks like you're still using this?

gj: This is removing it from dev deps — it's still in regular deps

gj: here: https://github.com/osohq/oso/pull/1393/files/ecd33230343e4a34684dc6151f303e8a83757175#diff-21e05eb50ab6604a97135d0c32a948d7e8408b5997f8fcb818029f8cf716c215L23

polar-core/src/validations.rs
gj: derp; thanks @samscott89 

6e9dcea6

samscott89 (approved):
Nice, this is looking clean. Love to see all the manual source line construction code dropped from the validation logic.

polar-core/src/validations.rs
```diff
@@ -75,11 +23,25 @@ impl<'kb> SingletonVisitor<'kb> {
...
+            .into_iter()
+            .flat_map(|(sym, term)| term.map(|t| (sym, t)))
+            .collect::<Vec<_>>();
+        singletons.sort_by_key(|(_, term)| term.offset());
         singletons
             .into_iter()
-            .map(|term| diagnostic_from_singleton(term, self.kb))
+            .map(|(sym, term)| {
+                if let Value::Pattern(_) = term.value() {
+                    Diagnostic::Warning(
```
samscott89, line 36: Perhaps we should promote this to a full-blown error at this point?

gj: Was that the plan at some point? Off the top of my head I can't think of any reason why we shouldn't, but I'd rather make a follow-up task instead of stapling it onto this PR

PR #1394, opened 2021-11-16
Add actors to both resource & actor union hashsets

### LINKED ISSUE #1386: Invalid rule error when using roles on an actor resource
Using oso 0.23.0, I have the following policy (note that `Team` is an `actor`, not a resource):
```
actor User {}

actor Team {
  roles = ["owner", "member"];

  "member" if "owner";
}
```

When I load that policy using `oso.load_files`, I get the following ValidationError:
```
polar.exceptions.ValidationError: Invalid rule: has_role(actor: Actor{}, "member", team: Team{}) if has_role(actor, "owner", team); Must match one of the following rule types:

has_role(actor: Actor{}, role: String{}, resource: Resource{});
	Failed to match because: Rule specializer Team on parameter 3 must be a member of rule type specializer Resource

	Perhaps you meant to add a resource block to your policy, like this:

	  resource Team { .. }
 at line 7, column 15 in file main.polar
	Get help with Oso from our engineers: https://help.osohq.com/error/ValidationError
```

It looks like the shorthand rule (`"member" if "owner"`) is resulting in a rule that does not pass the type validation. The reason it's not passing type validation is that `Team` is not considered a `Resource` for the purposes of type checking. I think a good fix would be always to consider `Actor` types to be `Resource` types also.

patrickod:
Allow the `Resource` specializer to match both actor and resources by adding actors to both hashsets. 

Fixes #1386


gj (changes requested):
I believe this default rule type is now redundant:

https://github.com/osohq/oso/blob/e327d453c45ed58f0208702b7ff7b807179019e4/polar-core/src/rules.rs#L110-L111

polar-core/src/resource_block.rs
```diff
@@ -1518,4 +1521,54 @@ mod tests {
...
         Ok(())
     }
+
+    // Test creation of rule types for actor roles
+    //   - has_role created because at least one resource block has roles declared
+    #[test]
+    fn test_create_resource_specific_rule_types_actor_roles() -> core::result::Result<(), PolarError>
+    {
+        let policy = r#"
+            actor User {}
```
gj, line 1531: what's the point of having the `actor User {}` block in this test?

patrickod: this arrived via me copying in the policy from the original issue. I can ✂️ it out as only the `actor Team` block is crucial to testing the `has_role` rule type creation here.

patrickod: trimmed. thanks for the sohut.

gj (approved):

PR #1396, opened 2021-11-17
formatting::source_lines context bug

gj:


polar-core/src/formatting.rs
```diff
@@ -688,6 +692,14 @@ mod tests {
             007:        seven"};
         assert_eq!(lines, expected, "\n{}", lines);
 
+        let lines = source_lines(&source, 1, 2);
+        let expected = indoc! {"
+            001:  one
+                  ^
+            002:   two
+            003:    three"};
```
gj, line 700: previously this would print
```
001:  one
      ^
002:   two
003:    three
004:     four
005:      five
```
b/c we were unconditionally asking for up to `context_lines * 2` lines of context. If `target_line` were line 0 and we'd asked for `2` lines of context on either side of the target line, we'd print 5 lines instead of 3.

samscott89 (commented):
We could potentially avoid doing all the arithmetic + line counting if we built the lines up as an iterator.

```rust
let skipped_lines = ... // still need this

let prefix = lines.skip(skipped_lines).take(std::cmp::min(context_lines, target_line))
let target = lines.nth(target_line);
let suffix = lines.skip(target_line).take(context_lines);

let lines = prefix.map(format_line).chain(target).chain(suffix.map(format_lines));
```

Something like that?

gj:
> Something like that?

yeah something like it; I don't think `.nth(target_line)` will work b/c we'll have already skipped `skipped_lines`, so `target_line` usize won't point to the correct line anymore. But I'll figure it out

samscott89:
> 
> 
> > Something like that?
> 
> yeah something like it; I don't think `.nth(target_line)` will work b/c we'll have already skipped `skipped_lines`, so `target_line` usize won't point to the correct line anymore. But I'll figure it out

The example was three different iterators, and then combining them together at the end.

gj:
> The example was three different iterators, and then combining them together at the end.

f540c8a9

samscott89 (approved):

samscott89 (approved):

PR #1398, opened 2021-11-18
Packaging for new VSCode extension bits

gj:


vscode/oso/Makefile
```diff
@@ -1,20 +1,28 @@
-.PHONY: build wasm fmtcheck lint test
+.PHONY: build package wasm fmtcheck lint test clean
 
-build: wasm node_modules
+build: clean wasm node_modules
 	@yarn compile
 
+package: clean wasm node_modules
+	@yarn vsce package
```
dhatch, line 7: I don't think you need to use `@yarn` here (and elsewhere). It's intentional that `make` prints out the command being executed so you can understand what's going on for inspection reasons. Usually `@` is only used to silence logging commands like `@echo "Running"`.

See: https://www.gnu.org/software/make/manual/html_node/Echoing.html#Echoing

vscode/oso/package.json
```diff
@@ -75,10 +74,17 @@
     "fmtcheck": "prettier --check 'client/**/*.ts' 'server/**/*.ts'",
     "fix": "gts fix",
     "postinstall": "cd client && yarn install && cd ../server && yarn install && cd ..",
-    "test": "yarn compile && node ./client/out/test/runTest.js"
+    "test": "yarn compile && yarn esbuild && node ./client/out/test/runTest.js",
+
+    "vscode:prepublish": "yarn esbuild-server --minify && yarn esbuild-client --minify",
+    "esbuild-server": "esbuild ./server/src/server.ts --bundle --outfile=server/out/server.js --external:vscode --format=cjs --platform=node",
```
dhatch, line 80: This is pretty hardcoded to the specific structure of the extension today. Would it be easy to generalize this a bit more? For example build everything in the `client/src` directory for client & similar for server? That will prevent us from having to update the packaging process in the future.

(I know nothing about `esbuild`)


I'm not sure how you plan to set up testing, but I think it's important to test the released version in case we accidentally don't include a file in the release build.

dhatch (commented):
I have really limited knowledge about Typescript, Yarn, or `esbuild`. So, if you want someone else, maybe @gkaemmer to review this could be good but I'm also fine to approve & we can improve as needed later.

vscode/oso/Makefile
gj: Ah okay that's good info — I've been adding it everywhere b/c I didn't find any value from printing the commands out. Will remove

vscode/oso/package.json
gj: The release version is tested as of this PR — note the `yarn esbuild` command here:

https://github.com/osohq/oso/blob/d5ef25773f0a73742a21a8140d3cf0dec1a2941b/vscode/oso/package.json#L77

gj: > This is pretty hardcoded to the specific structure of the extension today. Would it be easy to generalize this a bit more? For example build everything in the `client/src` directory for client & similar for server? That will prevent us from having to update the packaging process in the future.

I think I'd rather keep it like this for now b/c I don't imagine we're going to have to change this much, if ever. `client/src/extension.ts` & `server/src/server.ts` are the entrypoint files for the client & server, respectively, so the only time we'll need to update this is if we add additional _entrypoints_, which I don't foresee. We might split some functionality out of one of those into separate files (or add new files when adding new functionality) that are then included in the entrypoint, but we wouldn't need to update this build process in that case.

dhatch: cool thanks for that clarification

dhatch (approved):
🚀

vscode/oso/Makefile
gj: d34bb70a

PR #1400, opened 2021-11-18
Check for unrecoverable diagnostics during file loading & validation

gj:


polar-core/src/resource_block.rs
```diff
@@ -595,26 +607,6 @@ fn check_for_duplicate_resource_blocks(blocks: &ResourceBlocks, resource: &Term)
     Ok(())
 }
 
-fn check_that_shorthand_rule_heads_are_declared_locally(
```
gj, line 598: This check is duplicative of the more general check in `get_declaration_in_resource_block`. The error messages are almost identical: https://github.com/osohq/oso/blob/e327d453c45ed58f0208702b7ff7b807179019e4/polar-core/src/resource_block.rs#L301

dhatch (approved):
lgtm

PR #1401, opened 2021-11-18
Display warnings in VSCode

gj:
Added structure in #1393 but neglected to actually do anything with the received warnings on the VSCode side 😆. There's a bunch of duplication between handling errors & warnings that should be obviated when we get to refactoring them (via a `HasSource` trait or similar).


polar-language-server/src/lib.rs
```diff
@@ -264,45 +272,74 @@ impl PolarLanguageServer {
...
-            _ => self.document_from_polar_error_context(e).map(|doc| {
+        // NOTE(gj): We stringify the error / warning variant instead of the full `PolarError` /
+        // `PolarWarning` because we don't want source context as part of the error message.
+        let (message, severity) = match &d {
+            PolarDiagnostic::Error(e) => (e.kind.to_string(), DiagnosticSeverity::Error),
+            PolarDiagnostic::Warning(w) => (w.kind.to_string(), DiagnosticSeverity::Warning),
+        };
+
+        // If the diagnostic applies to a single doc, use it; otherwise, default to emitting a
+        // duplicate diagnostic for all docs.
```
gj, line 325: Calling out this behavioral change

polar-language-server/src/lib.rs
```diff
@@ -264,45 +272,74 @@ impl PolarLanguageServer {
...
                 None
             }
         })
     }
 
-    /// Create a `Diagnostic` from a `PolarError`, filtering out "ignored" errors.
-    fn diagnostic_from_polar_error(
+    /// Create `Diagnostic` from `polar_core::diagnostic::Diagnostic`, filtering out "ignored"
+    /// diagnostics.
+    fn diagnostic_from_polar_diagnostic(
```
dhatch, line 296:
```suggestion
    fn diagnostics_from_polar_diagnostic(
```

polar-language-server/src/lib.rs
```diff
@@ -264,45 +272,74 @@ impl PolarLanguageServer {
...
         &self,
-        e: &PolarError,
-    ) -> Option<(TextDocumentItem, Diagnostic)> {
+        d: PolarDiagnostic,
```
dhatch, line 298: function is a bit long to use use `d` throughout and you also have `docs` maybe make this the full name

polar-language-server/src/lib.rs
```diff
@@ -264,45 +272,74 @@ impl PolarLanguageServer {
...
+        );
+
+        docs.into_iter()
+            .map(|doc| {
                 let diagnostic = Diagnostic {
-                    range: range_from_polar_error_context(e),
-                    severity: Some(DiagnosticSeverity::Error),
+                    range: range_from_polar_error_context(&d),
+                    severity: Some(severity),
                     source: Some("Polar Language Server".to_owned()),
```
dhatch, line 336: I wonder if this should just say `Polar`? Is `rust-analyzer`'s error labeled with `rust` or `rust-analyzer`?

dhatch (approved):
lgtm, small changes if you want

polar-language-server/src/lib.rs
gj: image

polar-language-server/src/lib.rs
gj: 03f62ac2

polar-language-server/src/lib.rs
gj: 29b8207b

polar-language-server/src/lib.rs
gj: This is the description of the `Diagnostic.source` field: https://github.com/microsoft/vscode/blob/c15cb13a383dc9ff2dc0828152e374a6b9ecc2b3/src/vs/workbench/contrib/tasks/common/problemMatcher.ts#L754-L758

I chose `"Polar Language Server"` b/c it seemed analogous to their example of `"super lint"`, but I don't have strong feelings either way

PR #1402, opened 2021-11-21
python datafiltering adapters

saolsen:
new adapter api for data filtering that uses the new stuff


languages/python/oso/polar/data.py
```diff
@@ -0,0 +1,78 @@
...
+            v = val[kind]
+            return host.to_python({'value': v})
+
+
+    def build_query(self, host, types, filter):
+        assert filter['root'] in types
+        typ = filter['root']
+
+        # todo: joins
+        assert filter['relations'] == []
```
dhatch, line 50: We might want to include a nice helper in the adapter API for unsupported operations. That way when we explain how to implement this we can point users to that if they don't want to implement part of the API.

languages/python/oso/polar/query.py
```diff
@@ -105,24 +105,44 @@ def handle_external_call(self, data):
...
-                        assert exec_query is not None
-                        constraint = Filter(
-                            kind="Eq",
-                            field=rel.other_field,
-                            value=getattr(instance, rel.my_field),
-                        )
-                        constraints = [constraint]
-                        query = build_query(constraints)
-                        results = exec_query(query)
+                        adapter = self.host.adapter
```
dhatch, line 108: prob need to switch here too based on the same logic of whether we are using new / old data filtering

languages/python/oso/polar/data.py
```diff
@@ -0,0 +1,78 @@
+# New interface for an adapter.
+# Only does things needed for authorized_resources and authorized_query
+# TODO: Authorized session/database
+
+# todo: build a filter object instead of passing the json through
+# that way we can parse everything to python.
+class Filter:
```
dhatch, line 7: Would we go further adding more structure kind of like what @gw000 added in the Ruby lib?

Something like:

`Filter`
`Relations`
`Condition`

`Filter.to_query => Query`
`Filter.to_base_query => Query`
`Relations.add_to_query(query) => Query`
`Condition.add_to_query(query) => Query`
`Filter.add_condition(query)`

```
class Filter:
    def to_query(self):
        # make query w/ from
        query = self.to_base_query()

        
        query = self.relations.add_to_query(query)
        query = reduce(lambda q, cond: cond.add_to_query(q), self.conditions)

        return query
```

(meant to be purely illustrative)

languages/python/oso/oso/oso.py
```diff
@@ -268,7 +268,50 @@ def authorized_query(self, actor, action, resource_cls):
...
+        query is built by using the ``build_query`` and ``combine_query``
+        functions registered for the ``resource_cls``.
+
+        :param actor: The actor for whom to collect allowed resources.
+        :param action: The action that user wants to perform.
+        :param resource_cls: The type of the resources.
+
+        :return: A query to fetch the resources,
+        """
+        # todo: maybe call the old function if the old config is registered so we dont break existing code.
```
dhatch, line 283: good idea

dhatch: & error if you configure both?

languages/python/oso/polar/data.py
```diff
@@ -0,0 +1,117 @@
...
+                test = True
+                for (lhs, op, rhs) in conditions:
+                    kind = next(iter(lhs))
+                    if kind == 'Field':
+                        (t, f) = lhs[kind]
+                        i = rec[t]
+                        obj = self.type_arrays[t][i]
+                        v = getattr(obj, f)
+                        l = v
+                    elif kind == 'Imm':
```
gw000, line 85: This name changed in rust

PR #1407, opened 2021-11-24
VSCode Extension Telemetry

gj:


polar-language-server/src/lib.rs
```diff
@@ -361,9 +385,19 @@ impl PolarLanguageServer {
...
-        self.load_documents()
+        let diagnostics = self
+            .load_documents()
             .into_iter()
             .flat_map(|diagnostic| self.diagnostics_from_polar_diagnostic(diagnostic))
+            .collect::<Vec<_>>();
+
+        self.send_telemetry(DiagnosticLoadTelemetryEvent {
+            diagnostics: diagnostics.clone().into_iter().map(|(_, d)| d).collect(),
+            has_resource_blocks: false,
```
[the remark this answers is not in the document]
gj: #1409

vscode/oso/Makefile
```diff
@@ -15,14 +15,15 @@ node_modules: package.json client/package.json server/package.json
 	yarn install
 	@touch $@
 
-fmtcheck: clean node_modules
+fmtcheck: clean
 	yarn fmtcheck
 
-lint: clean wasm node_modules
+lint: clean wasm
```
dhatch, line 21: I guess this was already here but why clean on every operation?

vscode/oso/server/src/server.ts
```diff
@@ -11,7 +11,9 @@ const connection = createConnection(ProposedFeatures.all);
 
 const sendDiagnosticsCallback = (params: PublishDiagnosticsParams) =>
   connection.sendDiagnostics(params);
-const pls = new PolarLanguageServer(sendDiagnosticsCallback);
+const telemetryCallback = (event: unknown) =>
```
dhatch, line 14: Any more specific type here?

vscode/oso/client/src/extension.ts
```diff
@@ -154,6 +159,9 @@ async function startClient(folder: WorkspaceFolder, context: ExtensionContext) {
   context.subscriptions.push(deleteWatcher);
   context.subscriptions.push(createChangeWatcher);
 
+  // Synchronize `events` state across devices.
+  context.globalState.setKeysForSync([telemetryEventsKey]);
```
dhatch, line 163: Are there any docs you were looking at for this synchronization API?

I'm not sure this is the intended usage, because it [looks like](https://code.visualstudio.com/api/extension-capabilities/common-capabilities#setkeysforsync-example) it's designed for settings sync (infrequently changing data, rather than high frequency like telemetry).

It also will only be enabled if the user logs into VSCode & turns on settings sync. Not sure whether that's an issue for us.

vscode/oso/client/src/telemetry.ts
```diff
@@ -0,0 +1,135 @@
...
+      };
+
+      const diagnosticEvents: MixpanelEvent[] = diagnostics.map(({ code }) => ({
+        event: 'diagnostic',
+        properties: { code, ...metadata },
+      }));
+
+      // TODO(gj): race condition?
+      const old = state.get<MixpanelEvent[]>(telemetryEventsKey, []);
+      const events: MixpanelEvent[] = [...old, loadEvent, ...diagnosticEvents];
```
dhatch, line 128: Is this going to copy everything from old? This code is going to run a lot so may be a bit performance sensitive --- although it is async.

vscode/oso/client/src/telemetry.ts
```diff
@@ -0,0 +1,135 @@
...
+          ...metadata,
+        },
+      };
+
+      const diagnosticEvents: MixpanelEvent[] = diagnostics.map(({ code }) => ({
+        event: 'diagnostic',
+        properties: { code, ...metadata },
+      }));
+
+      // TODO(gj): race condition?
```
dhatch, line 126: Yes, right? The get & update are not atomic so another async task could come in & do a complete update which would not be included in our `old` list and would then be overwritten.

vscode/oso/client/src/telemetry.ts
```diff
@@ -0,0 +1,135 @@
...
+    void (async () => {
+      try {
+        // Retrieve all queued events.
+        const events = state.get<MixpanelEvent[]>(telemetryEventsKey, []);
+
+        if (events.length === 0) return;
+
+        // Clear events queue.
+        outputChannel.appendLine(`Flushing ${events.length.toString()} events`);
+        await state.update(telemetryEventsKey, []);
```
dhatch, line 72: race condition

vscode/oso/client/src/telemetry.ts
```diff
@@ -0,0 +1,135 @@
...
+
+      const diagnosticEvents: MixpanelEvent[] = diagnostics.map(({ code }) => ({
+        event: 'diagnostic',
+        properties: { code, ...metadata },
+      }));
+
+      // TODO(gj): race condition?
+      const old = state.get<MixpanelEvent[]>(telemetryEventsKey, []);
+      const events: MixpanelEvent[] = [...old, loadEvent, ...diagnosticEvents];
+      await state.update(telemetryEventsKey, events);
```
dhatch, line 129: Same question here about memento API, is it meant for frequent updates? Will this have any performance impact?

vscode/oso/client/src/extension.ts
```diff
@@ -17,12 +17,17 @@ import {
   TransportKind,
 } from 'vscode-languageclient/node';
 
+import {
```
dhatch, line 20: I think we need to implement opt out:

From https://code.visualstudio.com/docs/getstarted/telemetry

> If you would prefer to not utilize the npm module, it is still recommended that extension authors respect the user's choice by utilizing the isTelemetryEnabled and onDidChangeTelemetryEnabled API.

dhatch (changes requested):
Questions about race conditions & opt out

vscode/oso/client/src/extension.ts
dhatch: Ignore this, I See #1410

vscode/oso/Makefile
gj: Make sure there aren't any old compiled files hanging out — any reason not to?

vscode/oso/client/src/telemetry.ts
```diff
@@ -0,0 +1,135 @@
...
+  Diagnostic,
+  DiagnosticSeverity as Severity,
+} from 'vscode-languageclient';
+
+export const telemetryEventsKey = 'events';
+// Flush telemetry events in batches every hour.
+export const TELEMETRY_INTERVAL = 1000 * 60 * 60;
+
+const hash = (contents: string) =>
+  createHash('sha256').update(contents).digest('base64');
```
patrickod, line 15: potentially overparanoid security engineer speaking here: 
we might consider prefixing the `contents` here with a domain-specific prefix `oso-telemetry:$CONTENTS` to prevent someone from leveraging existing rainbow table SHA256 datasets to work backwards from a hashed identifier.

vscode/oso/client/src/telemetry.ts
gj: Anecdotally I haven't noticed any lag, but I haven't seen anything in the VSCode docs that mentions the performance characteristics of the pair of memento objects exposed by VSCode.

vscode/oso/client/src/extension.ts
gj: > Are there any docs you were looking at for this synchronization API?

The one you linked, plus docs linked off of it like [this](https://code.visualstudio.com/docs/editor/settings-sync#_sync-user-global-state-between-machines) and looking through Microsoft's org for [examples](https://github.com/search?p=1&q=org%3Amicrosoft+setKeysForSync&type=Code).

> I'm not sure this is the intended usage, because it looks like it's designed for settings sync (infrequently changing data, rather than high frequency like telemetry).
>
> It also will only be enabled if the user logs into VSCode & turns on settings sync. Not sure whether that's an issue for us.

Might not be designed for this use case; I'm not sure. I haven't noticed any performance impact from using it. Most of this telemetry stuff is happening asynchronously in the background so not super worried about it.

vscode/oso/server/src/server.ts
gj: Not in any way that would be helpful; the `event` param is any "JSON serializable object"

https://github.com/microsoft/vscode-languageserver-node/blob/1fe84c5b656d9db07f812cb573d2c0d82f9d8366/server/src/common/server.ts#L898-L900

vscode/oso/client/src/telemetry.ts
gj: Yeah, I didn't really care when I wrote this since it isn't the end of the world if we don't capture every single piece of telemetry. (Relevant: many telemetry implementations out there seem to set a "filter" that will only send a sample of X% of collected telemetry.) Especially now that #1414 sends telemetry on VSCode shutdown instead of on startup, there's no reason to use the Memento API at all. I'd prefer to ship this as-is and rip it out in #1414.

vscode/oso/client/src/extension.ts
gj: Removed in #1414

vscode/oso/client/src/telemetry.ts
gj: Removed in #1414

vscode/oso/client/src/telemetry.ts
gj: Removed in #1414

vscode/oso/client/src/telemetry.ts
gj: Removed in #1414

vscode/oso/client/src/telemetry.ts
gj: Removed in #1414

vscode/oso/client/src/telemetry.ts
gj: 65ce867

vscode/oso/Makefile
dhatch: Clean = delete intermediate files that could speed up build for me. i.e. `cargo clean` torches all the intermediate state. So running it between every build was surprising to me for that reason. But I guess in this case it's just removing the compiled files so if you feel that's necessary than seems okay.

dhatch (approved):
lgtm

responded to your comment re: clean. If you want to remove it based on that you can or leave it whatever you want.

vscode/oso/Makefile
gj: > Clean = delete intermediate files that could speed up build for me. i.e. cargo clean torches all the intermediate state.

Isn't that going to slow your build down since you're going to have to recompile everything instead of just the incremental change since your last compilation?

> But I guess in this case it's just removing the compiled files so if you feel that's necessary than seems okay.

🤷 it doesn't really affect anything — I think TypeScript & ESBuild both support some form of incremental compilation but I'm not sure I have the knobs fiddled correctly to take advantage of that and it also doesn't really matter since compilation w/ either tool takes less than a second

PR #1410, opened 2021-11-25
Make VSCode telemetry configurable

gj:


vscode/oso/client/src/telemetry.ts
```diff
@@ -20,6 +20,28 @@ const distinct_id = hash(env.machineId);
...
+  // Check if user explicitly enabled or disabled telemetry.
+  if (setting === 'on') return true;
+  if (setting === 'off') return false;
+
+  // Otherwise, default to VSCode's telemetry setting.
+
+  // VSCode >=1.55
+  //
+  // https://code.visualstudio.com/updates/v1_55#_telemetry-enablement-api
+  if (env.isTelemetryEnabled !== undefined) return env.isTelemetryEnabled;
```
dhatch, line 37: I think you also need to look at the telemetry change event: https://code.visualstudio.com/api/references/vscode-api

See (`onDidChangeTelemetryEnabled`)

dhatch (changes requested):

vscode/oso/client/src/telemetry.ts
gj: This check happens every time before we record or send telemetry. That `onDidChangeTelemetryEnabled` event only matters if you're maintaining local state about whether or not telemetry is enabled & need to update that every time the user changes their telemetry settings. Since we check every time before we do anything, we get updates immediately.

dhatch (approved):
ok lgtm

PR #1412, opened 2021-11-28
[Node.js] Add acceptExpression / update queryRule & query APIs

Kn99HN:
Since `queryRulesWithBinding` takes rest parameters, I wasn't able to add `acceptExpression` as the last arguments. I also don't know if it's good idea to break the function signature. My workaround is to add other functions that would take `acceptExpression` as a parameter and call `queryRulesWithBinding`. The work mirrors the work in Go and Java libraries.

Pls let me know if there's anything needed changing : )

PR checklist:

- [x] Added changelog entry.


Kn99HN:

gj:
Hi @Kn99HN thanks for the contribution! I'm sorry it's taken us a few days to get to it — we've been pretty heads-down on getting ready for tomorrow's release. I'll review this PR by tomorrow and make sure it gets into the next release

languages/js/src/errors.ts
```diff
@@ -129,3 +129,16 @@ export class DataFilteringConfigurationError extends PolarError {
...
+    super(`${message}`);
+  }
+}
+
+export const UNEXPECTED_EXPRESSION_MESSAGE = `Received Expression from Polar VM. The Expression type is only supported when
+  using data filtering features. Did you perform an
+  operation over an unbound variable in your policy?
+
+  To silence this error and receive an Expression result, pass
+  accept_expression as true to Oso.Query.`;
```
gj, line 144: Since we only construct this error in a single place, I'd prefer to either:
- Raise a `PolarError` w/ `UNEXPECTED_EXPRESSION_MESSAGE` as the `message`, removing the new `UnexpectedPolarTypeError` class
- Rename `UnexpectedPolarTypeError` -> something like `UnexpectedExpressionError` and directly call `super(UNEXPECTED_EXPRESSION_MESSAGE);` in the `constructor`

languages/js/src/Polar.ts
```diff
@@ -185,8 +185,13 @@ export class Polar {
   /**
    * Query for a Polar predicate or string.
    */
-  query(q: Predicate | string, bindings?: Map<string, unknown>): QueryResult {
+  query(
+    q: Predicate | string,
+    bindings?: Map<string, unknown>,
+    acceptExpression?: boolean
```
gj, line 191: What do you think about an `options` object as the final parameter here? So the call looks like

```ts
query('f(x)', {
  bindings: myBindings,
  acceptExpression: true
})
```

languages/js/src/errors.ts
Kn99HN: I decided to go with option 2 since I think it differentiates the error nicely from a generic PolarError. Lmk if you want to change it!

languages/js/src/Polar.ts
Kn99HN: I think that makes a lot of sense. I have refactored the code with this. Thanks for your suggestions!

gw000 (commented):
This looks fine, but we could change `queryRule` in a backwards-compatible way that might arguably be more convenient (see the comment on `queryRule`). What do you think?

gw000 (approved):
🍎

PR #1413, opened 2021-11-28
[POC] Added Support for AsyncSession in SQLAlchemy 1.4+

zikphil:
Im adding a PoC for AsyncSession support in the latest sqlalchemy-oso module.

2 functions are added: authorized_async_sessionmaker and async_scoped_session, both are expecting AsyncSessions as opposed to normal Session.

SQLAlchemy converts all code to Sync when time comes to process signals so there shouldn't be a need for any modifications there. I have tested with the following code from your tutorial and got the expected output:

[https://gist.github.com/zikphil/9b0c7d70b789796bcbcc56528e41ea51](https://gist.github.com/zikphil/9b0c7d70b789796bcbcc56528e41ea51)

Feel free to adapt this code as you see fit.


zikphil:

gj:
Hi @zikphil thanks for the contribution! I'm sorry it's taken us a few days to get to it — we've been pretty heads-down on getting ready for tomorrow's release. I'll make sure this PR gets reviewed ASAP

languages/python/sqlalchemy-oso/sqlalchemy_oso/session.py
```diff
@@ -180,6 +253,70 @@ def _scopefunc():
         get_oso, get_user, get_checked_permissions, **kwargs
     )
 
+    return async_scoped_session(factory, scopefunc=_scopefunc)
```
gkaemmer, line 256: Hmm, we want to keep using a normal, synchronous `scoped_session` here, right?

languages/python/sqlalchemy-oso/sqlalchemy_oso/session.py
```diff
@@ -291,41 +429,26 @@ def do_orm_execute(execute_state):
         if checked_permissions is None:
             return
 
-        def entities_in_statement(statement):
```
gkaemmer, line 294: I think we'll need to keep this method around for sqlalchemy <1.4. Perhaps we can add this old implementation to the pre-1.4 version of `all_entities_in_statement` in `sqlalchemy_utils.py`? (right now it raises an NotImplementedError)

@dhatch do you have any thoughts on this?

languages/python/sqlalchemy-oso/sqlalchemy_oso/session.py
```diff
@@ -1,8 +1,10 @@
 """SQLAlchemy session classes and factories for oso."""
-from typing import Any, Callable, Dict, Optional, Type
+from typing import Any, Callable, Dict, Optional, Type, Union
+import logging
 
 from sqlalchemy import event, inspect
 from sqlalchemy.orm import sessionmaker, Session
+from sqlalchemy.ext.asyncio import AsyncSession, async_scoped_session
```
gkaemmer, line 7: I think we'll want to move this import into a method so that this file won't cause errors using pre-1.4 sqlalchemy. @dhatch it looks like you did something to this effect a little while ago when you were adjusting the way `joinedload`s get authorized. Do you have any pointers here?

gkaemmer (changes requested):
Thanks for this, this is awesome -- glad its not too bad to get this working.

To get this merged, I think we'll need to do some testing in sqlalchemy `1.3`, which we need to continue supporting. Right now I'm pretty sure there are a few ways this breaks compatibility.

Do you think you'd have time to do that testing? We'd love to get this out in the next release (in two weeks)

zikphil:
Please review the new implementation. I have split it out in separate files and added a condition for SQLAlchemy => 1.4

languages/python/sqlalchemy-oso/sqlalchemy_oso/async_session.py
```diff
@@ -0,0 +1,144 @@
...
+    All other keyword arguments are passed through to
+    :py:func:`sqlalchemy.orm.session.sessionmaker` unchanged.
+
+    **Invariant**: the values returned by the `get_oso()`, `get_user()`, and
+    `get_checked_permissions()` callables provided to this function *must
+    remain fixed for a given session*. This prevents authorization responses
+    from changing, ensuring that the session's identity map never contains
+    unauthorized objects.
+
+    NOTE: _baked_queries are disabled on SQLAlchemy 1.3 since the caching
```
dhatch, line 56: We can drop this note since this API will only be used on 1.4.

languages/python/sqlalchemy-oso/sqlalchemy_oso/signal.py
```diff
@@ -0,0 +1,118 @@
+"""SQLAlchemy signal for processing queries."""
```
dhatch, line 1: Just confirming, has this been moved unmodified?

zikphil: That is correct

languages/python/sqlalchemy-oso/sqlalchemy_oso/async_session.py
zikphil: Removed

gkaemmer (changes requested):
I think from an implementation standpoint this looks good to me. Do you think you could add a few tests, one for each of `async_authorized_sessionmaker` and `async_scoped_session` -- I would think you could copy some tests verbatim from `test_sqlalchemy.py` and add `@pytest.mark.async` to make them asynchronous.

You'd also probably need to skip the test file during the sqlalchemy 1.3 tests (which the test suite runs in parallel with `tox`). You should be able to use `pytest.mark.skip` with `allow_module_level=True` to do that (see: https://docs.pytest.org/en/latest/how-to/skipping.html#skip).

Lemme know if you need any tips on that. Thanks again for this contribution -- very excited to have this in.

alexhafner:
Hi @zikphil - do you have a chance to move forward with this PR? We had used it with a handful of calls so far and it worked fine, would be great to have it operating on top of `oso` 0.26.0   😇

PR #1415, opened 2021-11-29
New DF pre-release fixes

gw000:
PR checklist:

- [x] adapter API
- [x] changelog


languages/ruby/lib/oso/polar/host.rb
```diff
@@ -67,11 +69,12 @@ class Host # rubocop:disable Metrics/ClassLength
       public
 
       attr_writer :accept_expression
-      attr_accessor :build_query, :combine_query, :exec_query
+      attr_accessor :build_query, :combine_query, :exec_query, :adapter
 
       DEFAULT_COMBINE_QUERY = proc { raise 'implement combine_query to use data filtering' }
       DEFAULT_BUILD_QUERY = proc { raise 'implement build_query to use data filtering' }
       DEFAULT_EXEC_QUERY = proc { raise 'implement exec_query to use data filtering' }
+      DEFAULT_ADAPTER = ::Oso::Polar::Data::Adapter.new
```
saolsen, line 77: so, this defaults to just one that throws errors? I guess that's cool actually.

saolsen (approved):

polar-core/src/filter.rs
```diff
@@ -357,8 +354,24 @@ impl FilterInfo {
         self.add_condition(left, Comparison::In, right)
     }
 
+    /// Validate FilterInfo before constructing a Filter
+    fn validate(self, root: &str) -> FilterResult<Self> {
+        let mut set = singleton(root);
+        for Relation(_, _, dst) in self.relations.iter() {
+            if set.contains(dst as &str) {
+                return invalid_state_error(format!(
```
dhatch, line 362: Should this be an unsupported error? I thought invalid state errors are for when we have an internal bug that ends up breaking one of our invariants? This could be user caused though.

languages/ruby/lib/oso/polar/host.rb
gw000: took this out & made it `nil` to be easier to test if it's configured

polar-core/src/filter.rs
gw000: Unsupported errors need a `term`, which we don't have here.

PR #1418, opened 2021-11-30
VSCode extension metrics documentation

gj:


vscode/oso/README.md
```diff
@@ -19,3 +19,15 @@ will be displayed inline in the editor and in the **Problems** pane.
   VSCode's file tree, the simplest remediation is to restart the Oso extension.
 
 [60813]: https://github.com/microsoft/vscode/issues/60813
+
+## Telemetry
+
+The extension collects **non-identifiable** telemetry data that we use to
+improve Oso. We collect data into un-timestamped batches instead of sending it
+on every policy load since we care about aggregate statistics; not tracking
```
gneray, line 27: Change semicolon to comma (here and in other instance)

gneray (commented):
Generally lgtm!

Let's use the term "metrics" instead of "telemetry" though. The difference is nuanced, but it emphasizes more the aggregate nature, which is what we care about. Thanks.

vscode/oso/README.md
gj: I think that makes it a comma splice but ur the boss

samscott89 (approved):

PR #1419, opened 2021-11-30
new data filtering sqlalchemy example adapter

gw000:
PR checklist:

- [x] Added changelog entry.


languages/python/oso/polar/data/filter.py
```diff
@@ -0,0 +1,60 @@
+class Fil:
```
gw000, line 1: TODO longer type names

saolsen (approved):

PR #1422, opened 2021-11-30
More specific new DF config doc

gw000:


docs/content/ruby/guides/data_filtering_next/index.md
```diff
@@ -1,104 +1,68 @@
 ---
-title: "Filter Data - Adapter API (Preview)"
+title: "Filter Data (Preview)"
 weight: 1
 showContentForAnyLanguage: true
-no_nav: true
 ---
 
-# Filter Data - Adapter API (Preview)
+# New Data Filtering for Ruby
```
dhatch, line 7: Can we call this something other than "New Data Filtering"? I changed the title to "Adapter API" previously. I think we need a clearer way to disambiguate the two other than new / old for users.

docs/content/ruby/guides/data_filtering_next/index.md
```diff
@@ -1,104 +1,68 @@
 ---
-title: "Filter Data - Adapter API (Preview)"
+title: "Filter Data (Preview)"
 weight: 1
 showContentForAnyLanguage: true
-no_nav: true
```
dhatch, line 5: I added this intentionally so that the page is only accessible from Ruby data filtering documentation, rather than adding a second "Filter Data" entry at the top level.

docs/content/ruby/guides/data_filtering_next/index.md
```diff
@@ -1,104 +1,68 @@
...
-Often you need data that is not contained on the object to make
-authorization decisions. This comes up when the role required to
-do something is implied by a role on it's parent object. For instance,
-you want to check the organization for a repository but that data isn't
-embedded on the repository object. You can add a `Relation` type to the type
-definition that states how the other resource is related to this one. Then
-you can access this field in the policy like any other field and it will
-fetch the data when it needs it (via the query functions).
+The four attributes of a `Filter` are:
+- `root`: the class of the queried resource
```
dhatch, line 51: Is this a string, or the actual class?

docs/content/ruby/guides/data_filtering_next/index.md
```diff
@@ -1,104 +1,68 @@
...
+- `relations`: a list of `[from_class, rel_name, to_class]` triples,
+  with values corresponding to your globally registered relations. We apologize
```
dhatch, line 53: This second sentence "We apologize..." is a bit unclear to me. Could you make it more specific? I'm not sure what it's referring to.

Are we parsing relations into objects on the Ruby side? Or are they just lists?
Same question for conditions.

You had this at one point in the POC, I think that could make things easier to interact with.

docs/content/ruby/guides/data_filtering_next/index.md
```diff
@@ -124,23 +88,14 @@ know what fields to match up with building a query for the other type.
...
+Like previous releases, the new data filtering can't perform method calls on
+unbound variables, and excludes Polar operators and expressions. See the [general
+data filtering docs][docs] for more details.
 
-Some Polar expressions are not supported. `not`, `cut` and `forall` are not
-allowed in policies that want to use data filtering. Numeric comparisons with
-the `<` `>` `<=` and `>=` are not currently supported either.
+This release doesn't include support for queries where a single resource occurs
+more than once. Therefore direct or indirect relations from a type to itself are
+prohibited. We plan to remove this limitation will in an upcoming release.
```
dhatch, line 99:
```suggestion
prohibited. We plan to remove this limitation in an upcoming release.
```

docs/content/ruby/guides/data_filtering_next/index.md
```diff
@@ -1,104 +1,68 @@
...
-authorization decisions. This comes up when the role required to
-do something is implied by a role on it's parent object. For instance,
-you want to check the organization for a repository but that data isn't
-embedded on the repository object. You can add a `Relation` type to the type
-definition that states how the other resource is related to this one. Then
-you can access this field in the policy like any other field and it will
-fetch the data when it needs it (via the query functions).
+The four attributes of a `Filter` are:
+- `root`: the class of the queried resource
+- `relations`: a list of `[from_class, rel_name, to_class]` triples,
```
dhatch, line 52: Need to explain more what the values in this triple mean, and how to use them.

docs/content/ruby/guides/data_filtering_next/index.md
dhatch: In the code this property is called `model`.

docs/content/ruby/guides/data_filtering_next/index.md
```diff
@@ -1,104 +1,68 @@
...
+  with values corresponding to your globally registered relations. We apologize
+  for the number of different things we designate a `relation`. This will be
+  addressed in an upcoming release.
+- `conditions`: a list of lists of `[lhs, op, rhs]` triples, with `op` in
+  `%w[Eq Neq In Nin]`, and `lhs` and `rhs` being `Projection` objects having a
```
dhatch, line 57: Sometimes LHS / RHS will be an immediate value.

docs/content/ruby/guides/data_filtering_next/index.md
```diff
@@ -1,104 +1,68 @@
...
-- `kind` is either "one" or "many". "one" means there is one related object and
-      "many" means there is a list of related objects.
-- `other_type` is the class of the related objects.
-- `my_field` Is the field on this object that matches `other_field`.
-- `other_field` Is the field on the other object that matches `this_field`.
-
-The `my_field` / `other_field` relationship is similar to a foreign key. It lets Oso
-know what fields to match up with building a query for the other type.
 
 ## Example
```
dhatch, line 67: This example doesn't really show how to implement an Adapter since it just uses the pre-built active record adapter.

dhatch (changes requested):

docs/content/ruby/guides/data_filtering_next/index.md
gw000: - data filtering `Relation` passed to `register_class`
- data filtering `Relation` object in a `Filter`
- `relations` in resource blocks and `has_relation` rules

docs/content/ruby/guides/data_filtering_next/index.md
dhatch: may still need to address this

docs/content/ruby/guides/data_filtering_next/index.md
dhatch: oh lol right

dhatch: Maybe add a note that these are distinct from the `relations` & `has_relation` used in resource blocks, and describe relationships in your data model (rather than in your authorization model)

PR #1434, opened 2021-12-08
Fix Node Debugger

dhatch:
The debugger in Node had a few issues:

1. We weren't waiting for input before continuing the run loop
2. We were creating a new readline session for every debug event without destroying the old one, so when you typed a character it would be duplicated a bunch of times
3. If we already had a readline instance for the REPL we didn't use it.
4. The REPL doesn't immediately print results, it collects them all and prints when the query is done. If you were debugging a query that outputs results, you wouldn't see them until the end, which is confusing.

This draft PR mostly fixes this. The code is pretty janky so could use some clean up before merging in. It has an issue when run from the node REPL. It prints `undefined` after each query because the node repl expects a return from the `eval` callback.

PR checklist:

- [ ] Added changelog entry.

### DIFF
--- a/languages/js/src/Polar.ts
+++ b/languages/js/src/Polar.ts
@@ -269,6 +269,8 @@ export class Polar {
     const repl = global.repl?.repl as REPLServer | undefined; // eslint-disable-line @typescript-eslint/no-unsafe-member-access
 
     if (repl) {
+      // @ts-ignore
+      global.oso_repl_rl = repl;
       repl.setPrompt(PROMPT);
       const evalQuery = this.evalReplInput.bind(this);
       // eslint-disable-next-line @typescript-eslint/ban-ts-comment
@@ -292,6 +294,10 @@ export class Polar {
         prompt: PROMPT,
         tabSize: 4,
       });
+
+      // @ts-ignore
+      global.oso_repl_rl = rl;
+
       rl.prompt();
       // eslint-disable-next-line @typescript-eslint/no-misused-promises
       rl.on('line', async (line: string) => {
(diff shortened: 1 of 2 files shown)


languages/js/src/Polar.ts
```diff
@@ -269,6 +269,8 @@ export class Polar {
     const repl = global.repl?.repl as REPLServer | undefined; // eslint-disable-line @typescript-eslint/no-unsafe-member-access
 
     if (repl) {
+      // @ts-ignore
+      global.oso_repl_rl = repl;
```
dhatch, line 273: This could probably be passed into query from Polar.

languages/js/src/Polar.ts
```diff
@@ -314,18 +320,23 @@ export class Polar {
...
-          results.push(result);
-        }
-        if (results.length === 0) {
-          return false;
-        } else {
-          for (const result of results) {
-            for (const [variable, value] of result) {
-              console.log(variable + ' = ' + repr(value));
-            }
+          hadResults = true;
```
dhatch, line 326: I updated this to print results as the query executes, but it doesn't quite work correctly if run from within the node repl.

languages/js/src/Query.ts
```diff
@@ -286,6 +288,8 @@ export class Query {
             for (const [k, v] of bindings) {
               transformed.set(k, await this.#host.toJs(v));
             }
+            rl?.emit('exit');
```
dhatch, line 291: This is kind of a mess but works. We need to "close" out our interaction with readline by removing the event handlers, so we fire the exit event which then adds back the handlers for the REPL. This call (`rl?.emit('exit')`) must be made any time this function yields, so could easily be messed up. Makes me think that debug should maybe be handled by a separate async generator that wraps this one.

languages/js/src/Query.ts
```diff
@@ -359,17 +363,44 @@ export class Query {
...
-              const trimmed = line.trim().replace(/;+$/, '');
-              const command = this.#host.toPolar(trimmed);
-              this.#ffiQuery.debugCommand(JSON.stringify(command));
-              this.processMessages();
+
+            if (typeof rl === 'undefined') {
+              // @ts-ignore
+              rl = global.oso_repl_rl;
+              if (rl) {
+                const exitListeners = rl.listeners('exit') as (() => void)[];
```
dhatch, line 371: Messy stuff to swap event handlers from the REPL ones to debugger ones and back. Kind of like what we do for the node repl.

gkaemmer: Took me a second to understand what this is doing -- can you make this a comment? And maybe flesh it out a bit, like `If we hit a debug event during a REPL session, we need to re-use the existing readline interface to prevent double inputs. This code updates the readline interface to use a debug prompt, until the debug prompt exits.`

languages/js/src/Query.ts
```diff
@@ -359,17 +363,44 @@ export class Query {
...
+            }
+
+            rl!.prompt();
+            const command: string = await new Promise(resolve => {
+              rl!.once('line', (line: string) => {
+                const trimmed = line.trim().replace(/;+$/, '');
+                resolve(trimmed)
+              });
             });
+            this.#ffiQuery.debugCommand(command);
```
gkaemmer, line 402: Could you emit the `exit` event on `rl` here instead of in the `Result` branch of the switch statement?

PR #1436, opened 2021-12-08
Fix temporary variable scope in ForAll expressions

gw000:
Currently `forall(x in y, x.n < z)` gets rewritten to something like `forall(x in y, _value_1 < z) and _value_1 = x.n`. This PR changes it to `forall(x in y, _value_1 < z and _value_1 = x.n)`.

PR checklist:

- [x] Added changelog entry.


polar-core/src/rewrites.rs
```diff
@@ -184,14 +184,23 @@ impl<'kb> Folder for Rewriter<'kb> {
...
+                                    operator: ForAll,
+                                    args: vec![args[0].clone(), term!(Operation {
+                                        operator: And,
+                                        args: rewrites
+                                    })],
+                                }));
                             }
-                        } else {
-                            for rewrite in rewrites.drain(..).rev() {
+                            _ => for rewrite in rewrites.drain(..).rev() {
```
dhatch, line 203: I think it would be helpful to add a rewrite test for the `ForAll` case. I've looked at those in the past when working on the rewriter both to figure out how things are rewritten when I forget (as a form of documentation), and to make it easier to refactor / make changes to the code rather than running an integration test.

what do you think?

dhatch (approved):
LGTM, but I suggest you add a quick unit test to `rewrites.rs` before merging (see my inline comment).

polar-core/src/rewrites.rs
gw000: Hmm... I'm somewhat not in favor because I think we should be testing for meaning not format here. In the simplifier it would be useful because we kind of use that as an interchange format, but here all that matters is if the Polar program runs correctly. We don't really care about, or need to document whether it's `(and (and (and a b) c) d)` or `(and a b c d)` at this point.

gw000: Actually never mind, you were right, doing this helped improve the fix a lot, thank you 😄

polar-core/src/rewrites.rs
```diff
@@ -245,22 +236,13 @@ fn only_dots(rewrites: &[Term]) -> bool {
...
-        operator: Operator::And,
-        args: vec![right, left.clone()],
-    });
-    left.replace_value(new_value);
-}
-
-/// Replace the left value with And(left, right).
-fn and_append(left: &mut Term, right: Term) {
-    let new_value = Value::Expression(Operation {
+fn and_(left: Term, right: Term) -> Term {
```
dhatch, line 239: +1 on making this not mutate in place

dhatch (approved):
lgtm still

PR #1437, opened 2021-12-09
Beginning of discrete log levels in VM

patrickod:
* cut new `LogLevel` enum which will replace the `self.log` bool used to denote
logging output level
* extend `log_with` with a log level argument denoting the priority of the msg

### DIFF
--- a/polar-core/src/diagnostic/context.rs
+++ b/polar-core/src/diagnostic/context.rs
@@ -57,3 +57,14 @@ impl fmt::Display for Context {
         Ok(())
     }
 }
+
+impl Context {
+    pub fn source_file_and_line(&self) -> String {
+        let Position { row, .. } = self.range.start;
+        if let Some(filename) = &self.source.filename {
+            format!("{}:{}", filename, row)
+        } else {
+            format!(":{}", row)
+        }
+    }
+}
--- a/polar-core/src/diagnostic/context.rs
+++ b/polar-core/src/diagnostic/context.rs
@@ -46,11 +46,8 @@ fn pos_to_loc(src: &str, row: usize, column: usize) -> usize {
 
 impl fmt::Display for Context {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}", self.source_file_and_line())?;
         let Position { row, column } = self.range.start;
-        write!(f, " at line {}, column {}", row + 1, column + 1)?;
-        if let Some(ref filename) = self.source.filename {
-            write!(f, " of file {}", filename)?;
-        }
         let loc = pos_to_loc(&self.source.src, row, column);
         let lines = source_lines(&self.source, loc, 0).replace('\n', "\n\t");
         writeln!(f, ":\n\t{}", lines)?;
@@ -60,11 +57,12 @@ impl fmt::Display for Context {
 
 impl Context {
     pub fn source_file_and_line(&self) -> String {
-        let Position { row, .. } = self.range.start;
-        if let Some(filename) = &self.source.filename {
-            format!("{}:{}", filename, row)
-        } else {
-            format!(":{}", row)
+        let mut f = String::new();
+        let Position { row, column } = self.range.start;
+        f += &format!(" at line {}, column {}", row + 1, column + 1);
+        if let Some(ref filename) = &self.source.filename {
+            f += &format!(" of file {}", filename);
         }
+        f
     }
 }
--- a/polar-core/src/diagnostic/context.rs
+++ b/polar-core/src/diagnostic/context.rs
@@ -56,11 +56,11 @@ impl fmt::Display for Context {
 }
 
 impl Context {
-    pub fn source_file_and_line(&self) -> String {
+    pub(crate) fn source_file_and_line(&self) -> String {
         let mut f = String::new();
         let Position { row, column } = self.range.start;
         f += &format!(" at line {}, column {}", row + 1, column + 1);
-        if let Some(ref filename) = &self.source.filename {
+        if let Some(ref filename) = self.source.filename {
             f += &format!(" of file {}", filename);
         }
         f
--- a/polar-core/src/formatting.rs
+++ b/polar-core/src/formatting.rs
@@ -312,6 +312,16 @@ pub mod display {
         }
     }
 
+    impl Rule {
+        pub fn head_as_string(&self) -> String {
+            format!(
+                "{}({}",
+                self.name.to_polar(),
+                format_params(&self.params, ", ")
+            )
+        }
+    }
+
     impl fmt::Display for Rule {
         fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
             match &self.body.value() {
@@ -320,18 +330,12 @@ pub mod display {
                     args,
                 }) => {
                     if args.is_empty() {
-                        write!(
-                            fmt,
-                            "{}({});",
-                            self.name.to_polar(),
-                            format_params(&self.params, ", ")
-                        )
+                        write!(fmt, "{};", self.head_as_string())
                     } else {
                         write!(
                             fmt,
-                            "{}({}) if {};",
-                            self.name.to_polar(),
-                            format_params(&self.params, ", "),
+                            "{} if {};",
+                            self.head_as_string(),
                             format_args(Operator::And, args, ",\n  "),
                         )
                     }
--- a/polar-core/src/formatting.rs
+++ b/polar-core/src/formatting.rs
@@ -358,6 +358,16 @@ pub mod display {
             }
         }
     }
+
+    impl fmt::Display for LogLevel {
+        fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+            match self {
+                Self::Trace => write!(f, "trace"),
+                Self::Debug => write!(f, "debug"),
+                Self::Info => write!(f, "info"),
+            }
+        }
+    }
 }
 
 pub mod to_polar {
--- a/polar-core/src/formatting.rs
+++ b/polar-core/src/formatting.rs
@@ -315,7 +315,7 @@ pub mod display {
     impl Rule {
         pub fn head_as_string(&self) -> String {
             format!(
-                "{}({}",
+                "{}({})",
                 self.name.to_polar(),
                 format_params(&self.params, ", ")
             )
--- a/polar-core/src/formatting.rs
+++ b/polar-core/src/formatting.rs
@@ -314,11 +314,7 @@ pub mod display {
 
     impl Rule {
         pub fn head_as_string(&self) -> String {
-            format!(
-                "{}({})",
-                self.name.to_polar(),
-                format_params(&self.params, ", ")
-            )
+            format!("{}({})", self.name, format_params(&self.params, ", "))
         }
     }
 
--- a/polar-core/src/formatting.rs
+++ b/polar-core/src/formatting.rs
@@ -172,7 +172,6 @@ pub fn to_polar_parens(op: Operator, t: &Term) -> String {
 }
 
 pub mod display {
-    use crate::formatting::format_params;
     use std::fmt;
     use std::sync::Arc;
 
@@ -324,12 +323,6 @@ pub mod display {
         }
     }
 
-    impl Rule {
-        pub fn head_as_string(&self) -> String {
-            format!("{}({})", self.name, format_params(&self.params, ", "))
-        }
-    }
-
     impl fmt::Display for Rule {
         fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
             write!(fmt, "{}", self.to_polar())
@@ -367,7 +360,7 @@ pub mod display {
 }
 
 pub mod to_polar {
-    use crate::formatting::{format_args, to_polar_parens};
+    use crate::formatting::{format_args, format_params, to_polar_parens};
     use crate::resource_block::{BlockType, ResourceBlock, ShorthandRule};
     use crate::rules::*;
     use crate::terms::*;
@@ -563,6 +556,12 @@ pub mod to_polar {
                 None => args,
             };
             format!("{}({})", self.name.to_polar(), combined_args)
+        }
+    }
+
+    impl Rule {
+        pub fn head_as_string(&self) -> String {
+            format!("{}({})", self.name, format_params(&self.params, ", "))
         }
     }
 
--- a/polar-core/src/formatting.rs
+++ b/polar-core/src/formatting.rs
@@ -172,7 +172,7 @@ pub fn to_polar_parens(op: Operator, t: &Term) -> String {
 }
 
 pub mod display {
-    use crate::formatting::{format_args, format_params};
+    use crate::formatting::format_params;
     use std::fmt;
     use std::sync::Arc;
 
@@ -181,7 +181,7 @@ pub mod display {
     use crate::numerics::Numeric;
     use crate::resource_block::Declaration;
     use crate::rules::Rule;
-    use crate::terms::{Operation, Operator, Symbol, Term, Value};
+    use crate::terms::{Symbol, Term};
     use crate::vm::*;
 
     impl fmt::Display for Binding {
@@ -320,24 +320,7 @@ pub mod display {
 
     impl fmt::Display for Rule {
         fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
-            match &self.body.value() {
-                Value::Expression(Operation {
-                    operator: Operator::And,
-                    args,
-                }) => {
-                    if args.is_empty() {
-                        write!(fmt, "{};", self.head_as_string())
-                    } else {
-                        write!(
-                            fmt,
-                            "{} if {};",
-                            self.head_as_string(),
-                            format_args(Operator::And, args, ",\n  "),
-                        )
-                    }
-                }
-                _ => panic!("Not any sorta rule I parsed"),
-            }
+            write!(fmt, "{}", self.to_polar())
         }
     }
 
@@ -372,7 +355,7 @@ pub mod display {
 }
 
 pub mod to_polar {
-    use crate::formatting::{format_args, format_params, to_polar_parens};
+    use crate::formatting::{format_args, to_polar_parens};
     use crate::resource_block::{BlockType, ResourceBlock, ShorthandRule};
     use crate::rules::*;
     use crate::terms::*;
@@ -578,20 +561,12 @@ pub mod to_polar {
                     operator: Operator::And,
                     args,
                 }) => {
-                    // rule type (parameters only; no arguments defined))
                     if args.is_empty() {
-                        format!(
-                            "{}({});",
-                            self.name.to_polar(),
-                            format_params(&self.params, ", ")
-                        )
-
-                    // rule (both parameters and arguments defined)
+                        format!("{};", self.head_as_string())
                     } else {
                         format!(
-                            "{}({}) if {};",
-                            self.name.to_polar(),
-                            format_params(&self.params, ", "),
+                            "{} if {};",
+                            self.head_as_string(),
                             format_args(Operator::And, args, " and "),
                         )
                     }
--- a/polar-core/src/formatting.rs
+++ b/polar-core/src/formatting.rs
@@ -353,6 +353,7 @@ pub mod display {
             }
         }
     }
+
     impl fmt::Display for Declaration {
         fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
             match self {
--- a/polar-core/src/formatting.rs
+++ b/polar-core/src/formatting.rs
@@ -181,7 +181,7 @@ pub mod display {
     use crate::numerics::Numeric;
     use crate::resource_block::Declaration;
     use crate::rules::Rule;
-    use crate::terms::{Symbol, Term};
+    use crate::terms::{Call, Operator, Symbol, Term};
     use crate::vm::*;
 
     impl fmt::Display for Binding {
@@ -197,6 +197,18 @@ pub mod display {
     }
 
     impl fmt::Display for Term {
+        fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
+            write!(fmt, "{}", self.to_polar())
+        }
+    }
+
+    impl fmt::Display for Call {
+        fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
+            write!(fmt, "{}", self.to_polar())
+        }
+    }
+
+    impl fmt::Display for Operator {
         fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
             write!(fmt, "{}", self.to_polar())
         }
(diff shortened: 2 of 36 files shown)


polar-core/src/formatting.rs
```diff
@@ -312,6 +312,16 @@ pub mod display {
...
+            format!(
+                "{}({})",
+                self.name.to_polar(),
+                format_params(&self.params, ", ")
+            )
+        }
+    }
+
     impl fmt::Display for Rule {
         fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
```
gj, line 326: I think this function should shell out to `Rule::to_polar`:

```rs
    impl fmt::Display for Rule {
        fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
            write!(fmt, "{}", self.to_polar())
        }
    }
```

and we should make the formatting updates in that function

polar-core/src/formatting.rs
```diff
@@ -567,12 +582,15 @@ pub mod to_polar {
                     operator: Operator::And,
                     args,
                 }) => {
+                    // rule type (parameters only; no arguments defined))
```
gj, line 585: This is the format for bodiless rules as well — not just rule types.

polar-core/src/vm.rs
```diff
@@ -758,37 +795,46 @@ impl PolarVirtualMachine {
     }
 
     fn log(&self, message: &str, terms: &[&Term]) {
```
gj, line 797: This seems like it's only used once, so I'd rather use this delightfully short name for `log_with`

polar-core/src/vm.rs
```diff
@@ -758,37 +795,46 @@ impl PolarVirtualMachine {
...
-                        relevant_bindings
-                            .iter()
-                            .map(|(var, val)| format!("{} => {}", var.0, val.to_polar()))
-                            .collect::<Vec<String>>()
-                            .join(", ")
-                    ));
+        if let Some(configured_log_level) = self.log_level {
+            // preserve the old `polar_log_mute` behavior which omits parameter
+            // specialization checking Unify, IsA and other events from the log
+            if level == LogLevel::Trace && self.polar_trace_mute {
```
gj, line 809: This feels odd to me — `Trace` is supposed to be the most verbose level, so why would we ever want to skip logs for `Trace` but not for the other levels?

polar-core/src/vm.rs
```diff
@@ -2944,6 +3026,23 @@ impl Runnable for PolarVirtualMachine {
...
+            || {
+                if bindings.is_empty() {
+                    "RESULT: SUCCESS".to_string()
+                } else {
+                    let mut out = "RESULT: { ".to_string(); // open with single right-padded curly
+                    for (key, value) in &bindings {
+                        out.push_str(&format!("{}: {} ", key, value)); // write right-padded key: value pairs
+                    }
+                    out.push('}'); // closing curly
+                    out
```
gj, line 3040: Probably best to put each binding on a new line since the expressions can get pretty long

gj (changes requested):

polar-core/src/diagnostic/context.rs
```diff
@@ -46,14 +46,23 @@ fn pos_to_loc(src: &str, row: usize, column: usize) -> usize {
...
-        }
         let loc = pos_to_loc(&self.source.src, row, column);
         let lines = source_lines(&self.source, loc, 0).replace('\n', "\n\t");
         writeln!(f, ":\n\t{}", lines)?;
         Ok(())
     }
 }
+
+impl Context {
+    pub fn source_file_and_line(&self) -> String {
```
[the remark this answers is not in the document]
patrickod: 👍🏻

polar-core/src/formatting.rs
```diff
@@ -312,6 +312,16 @@ pub mod display {
         }
     }
 
+    impl Rule {
+        pub fn head_as_string(&self) -> String {
+            format!(
+                "{}({})",
+                self.name.to_polar(),
+                format_params(&self.params, ", ")
+            )
```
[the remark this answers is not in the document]
patrickod: ✂️ 'd things up.

polar-core/src/formatting.rs
patrickod: refactored to use `to_polar` and DRY'd up `to_polar` with the `head_as_string` method.

polar-core/src/vm.rs
patrickod: I had the same thought and snipped it in #1444 !

polar-core/src/vm.rs
patrickod: we currently temporarily mute the `POLAR_LOG=trace` output for a period of the VM execution starting before rule filtering and resuming again in `sort_rules` once we have filtered & sorted all applicable rules. I understand from @saolsen that this is intended to prevent the log from being cluttered with these parameter related outputs.
https://github.com/osohq/oso/blob/main/polar-core/src/vm.rs#L1520
https://github.com/osohq/oso/blob/main/polar-core/src/vm.rs#L2582

gj: Right, I understand why we want to early return if `self.polar_trace_mute` is `true`. I'm wondering whether we need the `level == LogLevel::Trace` condition here. Maybe the answer is that currently we'll only ever encounter `log_with(LogLevel::Trace, ...)` calls while the `polar_trace_mute` bit is flipped, but maybe that won't always be the case.

Have you done any thinking on whether any sort of non-trace logging would be useful during rule selection and sorting? I guess that's kind of related to the "why a rule was filtered out" logic that's coming up soon. What are your thoughts?

polar-core/src/vm.rs
```diff
@@ -758,37 +795,46 @@ impl PolarVirtualMachine {
...
+                for _ in 0..=self.queries.len() {
+                    indent.push_str("  ");
                 }
-                self.print(msg);
-                for line in &lines[1..] {
-                    self.print(format!("[debug] {}{}", &indent, line));
+                let message = message_fn();
+                let lines = message.as_ref().split('\n').collect::<Vec<&str>>();
+                if let Some(line) = lines.first() {
+                    let mut msg = format!("[{}] {}{}", level, &indent, line);
```
gj, line 820: I wonder if we should identify our logs somehow, like `[oso - debug] ...`. Thinking about the case where you're running your web server in development mode w/ `POLAR_LOG=info` and Oso's logs are getting interspersed with the server's logs.

polar-core/src/vm.rs
```diff
@@ -650,9 +683,11 @@ impl PolarVirtualMachine {
 
     /// Push a binding onto the binding stack.
     pub fn bind(&mut self, var: &Symbol, val: Term) -> Result<()> {
-        if self.log {
-            self.print(&format!("⇒ bind: {} ← {}", var.to_polar(), val.to_polar()));
-        }
+        self.log_with(
+            LogLevel::Trace,
+            || format!("⇒ bind: {} ← {}", var.to_polar(), val.to_polar()),
```
gj, line 688: nit: there are a bunch of `format!("{}", x.to_polar())` expressions across these `log_with()` calls where the `.to_polar()` call is redundant since the `.to_string()` impl shells out to `.to_polar()`

polar-core/src/vm.rs
patrickod: great call! will add a prefix here

polar-core/src/formatting.rs
```diff
@@ -312,32 +312,15 @@ pub mod display {
         }
     }
 
+    impl Rule {
+        pub fn head_as_string(&self) -> String {
+            format!("{}({})", self.name, format_params(&self.params, ", "))
+        }
+    }
```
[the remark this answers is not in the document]
patrickod: 👍🏻

polar-core/src/vm.rs
```diff
@@ -310,6 +324,18 @@ impl PolarVirtualMachine {
...
+
+        let mut log_level = None;
+        if !polar_log_vars.is_empty() {
+            log_level = if polar_log_vars.contains(&"trace") {
+                Some(LogLevel::Trace)
+            } else if polar_log_vars.contains(&"debug") {
+                Some(LogLevel::Debug)
+            } else {
+                Some(LogLevel::Info)
+            }
```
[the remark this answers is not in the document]
patrickod: 👍🏻 good spot. thanks! updated to have `off` and `0` win over any other values.

polar-core/src/vm.rs
patrickod: I believe you are correct that we only currently encounter `log_with(LogLevel::Trace)`, however I can definitely see us expanding the set of logs to include "why a rule didn't match" style messages in the not too distant future. My thinking on that was to use the DEBUG level for these logs, in which case we'd want to retain the `level == LogLevel::Trace` condition and allow them to pass.

polar-core/src/vm.rs
patrickod: cleaned up `.to_polar` calls and added some `fmt::Display` impls to fill in the gaps.

polar-core/src/formatting.rs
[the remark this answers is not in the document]
gj: Not sure if GitHub's playing tricks on me, but it looks like this convo was resolved but the code is still in the same place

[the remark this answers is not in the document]
patrickod: womp I think this was a `git stash` mistake on my part.

polar-core/src/vm.rs
```diff
@@ -895,10 +945,7 @@ impl PolarVirtualMachine {
     /// Remove all bindings after the last choice point, and try the
     /// next available alternative. If no choice is possible, halt.
     fn backtrack(&mut self) -> Result<()> {
-        if self.log {
-            self.print("⇒ backtrack");
-        }
-        self.log("BACKTRACK", &[]);
+        self.log_with(LogLevel::Trace, || "BACKTRACK", &[]);
```
gj, line 948: Looks like we probably logged `BACKTRACK` _and_ `⇒ backtrack` on `POLAR_LOG=trace` for some period of time 🤦. That probably happened when we cleaned up the RUST_LOG/trace stuff a little while ago.

Anyway, just verifying that before this PR `POLAR_LOG=on` would spit out a `BACKTRACK` and after this PR it'll be silent unless `POLAR_LOG=trace`. Is that the desired behavior?

polar-core/src/vm.rs
```diff
@@ -963,7 +1010,7 @@ impl PolarVirtualMachine {
 
     /// Halt the VM by clearing all goals and choices.
     fn halt(&mut self) -> QueryEvent {
-        self.log("HALT", &[]);
+        self.log_with(LogLevel::Trace, || "HALT", &[]);
```
gj, line 1013: Same question as above — this log is being moved from `Info` -> `Trace`. Is that intentional?

polar-core/src/vm.rs
```diff
@@ -438,9 +476,7 @@ impl PolarVirtualMachine {
     /// Try to achieve one goal. Return `Some(QueryEvent)` if an external
     /// result is needed to achieve it, or `None` if it can run internally.
     fn next(&mut self, goal: Rc<Goal>) -> Result<QueryEvent> {
-        if self.log {
-            self.print(&format!("{}", goal));
-        }
+        self.log_with(LogLevel::Trace, || format!("{}", goal), &[]);
```
gj, line 479: super nit: I think `goal.to_string()` is a nicer way to spell `format!("{}", goal)`

polar-core/src/vm.rs
```diff
@@ -509,6 +545,7 @@ impl PolarVirtualMachine {
             Goal::TraceRule { trace } => {
                 if let Node::Rule(rule) = &trace.node {
                     self.log_with(
+                        LogLevel::Trace,
```
gj, line 548: Another "downgrade" from Info -> Trace. Highlighting all of these to make sure they're intentional shifts

polar-core/src/vm.rs
```diff
@@ -973,7 +1020,8 @@ impl PolarVirtualMachine {
     #[allow(clippy::many_single_char_names)]
     pub fn isa(&mut self, left: &Term, right: &Term) -> Result<()> {
         self.log_with(
-            || format!("MATCHES: {} matches {}", left.to_polar(), right.to_polar()),
+            LogLevel::Trace,
```
gj, line 1023: Another `Info` -> `Trace` shift

polar-core/src/vm.rs
```diff
@@ -1341,6 +1389,7 @@ impl PolarVirtualMachine {
         self.push_choice(vec![])?;
 
         self.log_with(
+            LogLevel::Trace,
```
gj, line 1392: `Info` -> `Trace`

polar-core/src/vm.rs
```diff
@@ -1432,14 +1481,30 @@ impl PolarVirtualMachine {
...
+                    );
+                }
+            }
             Value::Expression(Operation {
                 operator: Operator::And,
                 args,
             }) if args.len() < 2 => (),
             _ => {
-                self.log_with(|| format!("QUERY: {}", term.to_polar()), &[term]);
+                self.log_with(LogLevel::Trace, || format!("QUERY: {}", term), &[term]);
```
gj, line 1507: Info -> Trace

polar-core/src/vm.rs
```diff
@@ -1432,14 +1481,30 @@ impl PolarVirtualMachine {
...
     fn query(&mut self, term: &Term) -> Result<QueryEvent> {
-        // Don't log if it's just a single element AND like lots of rule bodies tend to be.
+        // print INFO event for the first call `query` with a rule
+        // (`Call(predicate)`) in the policy execution, ignoring subsequent
+        // calls nested in the execution.
+        //
+        // print TRACE (a superset of INFO) event for all the subsequent calls
+        // We filter out "single element AND" which many rule bodies
+        // take the form of to instead log only their inner operations for
+        // readibility|brevity reasons.
```
gj, line 1491: Took a stab at cleaning up these comments a bit:

```suggestion
        // - Print INFO event for the top-level query for a rule.
        // - Print TRACE (a superset of INFO) event for all subsequent calls.
        // - We filter out single-element ANDs, which many rule bodies take the form of, to instead
        //   log only their inner operations for readibility|brevity reasons.
```

polar-core/src/vm.rs
```diff
@@ -1485,7 +1550,7 @@ impl PolarVirtualMachine {
                     term,
                     format!(
                         "{} isn't something that is true or false so can't be a condition",
-                        term.value().to_polar()
+                        term.value()
```
gj, line 1553: This is probably what caused the test suite regression noted elsewhere. Do we need the `.value()` call for some reason I'm not seeing?

```suggestion
                        term
```

polar-core/src/vm.rs
```diff
@@ -2579,12 +2638,29 @@ impl PolarVirtualMachine {
             // We're done; the rules are sorted.
             // Make alternatives for calling them.
 
-            self.polar_log_mute = false;
+            self.polar_trace_mute = false;
+            // as with the "QUERY RULE:" event print the first of these events
+            // to INFO and subsequent queries to TRACE (a superset of INFO)
```
gj, line 2643: minor cleanup

```suggestion
            // As with the logging of queries in `PolarVirtualMachine::query`, print applicable
            // rules for the top-level query to INFO and applicable rules for all subsequent
            // queries to TRACE, a superset of INFO.
```

polar-core/src/vm.rs
```diff
@@ -2853,13 +2929,9 @@ impl Runnable for PolarVirtualMachine {
...
-        if self.log {
-            self.print("⇒ result");
-            if self.tracing {
-                for t in &self.trace {
-                    self.print(&format!("trace\n{}", t.draw(self)));
-                }
-            }
+        self.log_with(LogLevel::Trace, || &"⇒ result", &[]);
+        for t in &self.trace {
+            self.log_with(LogLevel::Trace, || format!("trace\n{}", t.draw(self)), &[]);
```
gj, line 2934: This was previously gated by `self.tracing` — was removing the gate intentional?

polar-core/src/vm.rs
```diff
@@ -2853,13 +2929,9 @@ impl Runnable for PolarVirtualMachine {
...
         }
 
-        if self.log {
-            self.print("⇒ result");
-            if self.tracing {
-                for t in &self.trace {
-                    self.print(&format!("trace\n{}", t.draw(self)));
-                }
-            }
+        self.log_with(LogLevel::Trace, || &"⇒ result", &[]);
```
gj, line 2932: This seems kind of duplicative in light of the new `Info`-level `RESULT: ...` log on 3019 below

polar-core/src/vm.rs
```diff
@@ -2977,7 +3066,7 @@ impl Runnable for PolarVirtualMachine {
         // For example what happens if the call asked for a field that doesn't exist?
 
         if let Some(value) = term {
-            self.log_with(|| format!("=> {}", value), &[]);
+            self.log_with(LogLevel::Trace, || format!("=> {}", value), &[]);
```
gj, line 3069: Info -> Trace

polar-core/src/vm.rs
```diff
@@ -3934,4 +4023,35 @@ mod tests {
             QueryEvent::Done { result: true }
         ]);
     }
+
+    #[test]
+    fn test_log_level_should_print_for_level() {
+        let trace = LogLevel::Trace;
+        let info = LogLevel::Info;
+        let debug = LogLevel::Debug;
```
gj, line 4031: nit
```suggestion
        use LogLevel::*;
```
and then reference them as `Trace` instead of `trace`, etc.

polar-core/src/vm.rs
```diff
@@ -3934,4 +4023,35 @@ mod tests {
...
+        // INFO
+        assert!(
+            !info.should_print_on_level(trace),
+            "info does not print trace"
+        );
+        assert!(
+            !info.should_print_on_level(debug),
+            "info does not print debug"
+        );
+        assert!(info.should_print_on_level(info), "info prints info");
```
gj, line 4055: truly the smallest of nits: the "on assertion error" messages aren't really making it much easier to understand each individual case, and I think having them all lined up like 
```suggestion
        assert!(!info.should_print_on_level(trace));
        assert!(!info.should_print_on_level(debug));
        assert!(info.should_print_on_level(info));
```
would be easier to quickly visually grep

polar-core/src/vm.rs
```diff
@@ -2579,12 +2638,29 @@ impl PolarVirtualMachine {
...
+                        let context = self.kb.read().unwrap().get_rule_source(rule).map_or_else(
+                            || "".to_string(),
+                            |source| {
+                                let range = Range::from_span(&source.src, rule.span().unwrap());
+                                let context = Context { source, range };
+                                context.source_file_and_line()
+                            },
+                        );
+
+                        rule_strs.push_str(&format!("\n  {}{}", rule.head_as_string(), context));
```
gj, line 2663: Don't necessarily need to figure this out for this PR, but it's slightly unfortunate that this is going to print out the rewritten ("longhand") head of a shorthand rule, like `has_permission(actor: Actor, "blorp", bleep: Bleep)`... although I guess it wouldn't be much clearer if we printed out the entire shorthand rule (`"blorp" if "blip";`). Maybe if we printed out the shorthand rule and underlined its head? Idk

polar-core/src/vm.rs
```diff
@@ -2579,12 +2638,29 @@ impl PolarVirtualMachine {
...
+            } else {
+                LogLevel::Trace
+            };
             self.log_with(
+                level,
                 || {
                     let mut rule_strs = "APPLICABLE_RULES:".to_owned();
                     for rule in rules {
-                        rule_strs.push_str(&format!("\n  {}", self.rule_source(rule)));
+                        let context = self.kb.read().unwrap().get_rule_source(rule).map_or_else(
```
gj, line 2654:
```suggestion
                        let context = self.kb().get_rule_source(rule).map_or_else(
```

gj (changes requested):

polar-core/src/vm.rs
```diff
@@ -345,14 +372,25 @@ impl PolarVirtualMachine {
...
-                None | Some("0") | Some("off") => false,
-                _ => true,
+        let polar_log = polar_log.unwrap_or("".to_string());
+        let polar_log_vars = polar_log.split(",").collect::<Vec<&str>>();
+
+        self.polar_log_stderr = polar_log_vars.contains(&"now");
+
+        // TODO: @patrickod consolidate log level evaluation here & in VM create
+        self.log_level = if rust_log.is_some() {
+            Some(LogLevel::Info)
```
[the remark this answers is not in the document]
patrickod:
```
(oso) 17:11:55 patrickod@kimkilwhan oso patrickod/log-levels ? ag -l rust_log
docs/content/any/project/changelogs/2021-10-19.md
languages/js/src/Query.ts
languages/js/src/polar_wasm_api.d.ts
languages/js/src/polar_wasm_api.js
polar-core/src/query.rs
polar-wasm-api/src/query.rs
polar-core/src/vm.rs
polar-wasm-api/pkg/polar_wasm_api.d.ts
polar-wasm-api/pkg/polar_wasm_api_bg.js
```

I believe that is correct & we should clean it up (I wasn't aware that it previously existed in the other language libraries and we were deprecating it)

polar-core/src/vm.rs
```diff
@@ -345,14 +372,25 @@ impl PolarVirtualMachine {
 
     #[cfg(target_arch = "wasm32")]
     pub fn set_logging_options(&mut self, rust_log: Option<String>, polar_log: Option<String>) {
```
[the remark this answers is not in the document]
patrickod: since updated! used communally now between the `PolarVirtualMachine::new` and the WASM routes.

polar-core/src/vm.rs
```diff
@@ -304,12 +318,28 @@ impl PolarVirtualMachine {
...
-            .flat_map(|pl| pl.split(','))
-            .collect::<Vec<&str>>();
+        let polar_log_vars: HashSet<&str> = polar_log.iter().flat_map(|pl| pl.split(',')).collect();
+
+        if polar_log_vars
+            .intersection(&HashSet::from(["off", "0"]))
+            .cloned()
+            .collect::<HashSet<&str>>()
+            .is_empty()
+        {
```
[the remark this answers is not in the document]
patrickod: 👍🏻 since tidied up with the use of `is_disjoint`

polar-core/src/vm.rs
gj: expected behavior

polar-core/src/vm.rs
```diff
@@ -1432,14 +1467,27 @@ impl PolarVirtualMachine {
...
     /// querying for each body clause.
     fn query(&mut self, term: &Term) -> Result<QueryEvent> {
-        // Don't log if it's just a single element AND like lots of rule bodies tend to be.
+        // print INFO event for call `query` with a rule (`Call(predicate)`) in
+        // the policy execution
+        //
+        // print TRACE (a superset of INFO) event for other queries
+        // We filter out "single element AND" which many rule bodies
+        // take the form of to instead log only their inner operations for
+        // readibility|brevity reasons.
```
gj, line 1476:
```suggestion
        // - Print INFO event for queries for rules.
        // - Print TRACE (a superset of INFO) event for all other queries.
        // - We filter out single-element ANDs, which many rule bodies take the form of, to instead
        //   log only their inner operations for readibility|brevity reasons.
```

gj (approved):

PR #1441, opened 2021-12-10
Pin versions of postcss etc for docs.

samscott89:
Grabbed these versions from the last successful run of the docs preview job.


patrickod (approved):

gj (changes requested):
package-lock.json changes

gj (approved):

PR #1443, opened 2021-12-13
Remove unneeded advapi32 lib

gj:
Used to be needed when Rust sometimes relied on RtlGenRandom instead of BCryptGenRandom, but that changed as of https://github.com/rust-lang/rust/pull/84096

advapi32 is still used in Rust core, but only in the deprecated `env::home_dir` API, which we don't use:

- https://github.com/rust-lang/rust/blob/22f8bde876f2fa9c5c4e95be1bce29cc271f2b51/library/std/src/sys/windows/c.rs#L681-L689
- https://github.com/rust-lang/rust/blob/22f8bde876f2fa9c5c4e95be1bce29cc271f2b51/library/std/src/sys/windows/os.rs#L292
- https://github.com/rust-lang/rust/blob/22f8bde876f2fa9c5c4e95be1bce29cc271f2b51/library/std/src/sys/windows/os.rs#L319
- https://github.com/rust-lang/rust/blob/22f8bde876f2fa9c5c4e95be1bce29cc271f2b51/library/std/src/env.rs#L566-L575

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,138 +5,139 @@ on:
       - "v*" # Push events to matching v*, i.e. v1.0, v20.15.10
     branches:
       - main
+      - gj/remove-advapi
 
 jobs:
-  version:
-    name: Compute and verify the version number
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v2
-      - uses: ./.github/actions/check_version
-      - name: Set version env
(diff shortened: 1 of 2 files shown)


gj:
Guess it is needed. I don't think we use `env::home_dir` anywhere, but maybe I don't understand our linking requirements well enough and `std` needs to link against advapi32 for the `__imp_OpenProcessToken` symbol (even though we don't use the one API that needs it) for... reasons

PR #1444, opened 2021-12-14
Print type or UNKNOWN alongside query arguments in INFO log

patrickod:
```(oso) 20:13:13 patrickod@kimkilwhan oso-org POLAR_LOG=info python scenarios/unknown_class.py      130 ↵
Polar tracing enabled. Get help with traces from our engineering team: https://help.osohq.com/trace
[info]   QUERY RULE: allow(<__main__.User object at 0x7fe206175280> TYPE `User`, "view", <__main__.Repository object at 0x7fe206175790> TYPE `Repository`)
[info]     APPLICABLE_RULES:
[info]       allow(actor, action, resource) at line 11, column 5
[info]   RESULT: SUCCESS
True
[info]   QUERY RULE: allow(<__main__.User object at 0x7fe206175280> TYPE `User`, "view", <__main__.Project object at 0x7fe206175430> UNREGISTERED TYPE `Project`)
[info]     APPLICABLE_RULES:
[info]       allow(actor, action, resource) at line 11, column 5
[info]         No matching rules found
False
```


languages/python/oso/polar/host.py
```diff
@@ -225,7 +225,8 @@ def replace_repr(match):
             instance_id = int(match[1])
             try:
                 instance = self.get_instance(instance_id)
-                return repr(instance)
+                unregistered = not self.ffi_polar.is_registered_constant(type(instance).__name__)
```
gj, line 228: I'm not sure you need to go to the core for this — host language is aware of registered types in the `Host.types` map (or in languages that haven't gotten the new DF stuff yet, might be stored in a `Host.classes` map or similar)

patrickod: of course! 🤦🏻 that is much much cleaner.

dhatch (commented):
Just some thoughts on the approach:
- Enrich message only exists in Python & Ruby, but I hadn't thought about using that over sending the information into the core to format messages with. The reason we added that API is not to reduce the size of FFI messages, but rather to eliminate the need to stringify every object before sending it to the core. Stringification can be pretty expensive depending on the type (some ORMs evaluate database queries or compile to SQL when performing stringification for example). Getting the name of a type's class is probably much cheaper.
- it would generally be useful to have type info in the core as it would help us evaluate matches without an external event, but that's not the main goal of what we're doing now. Having the class name would let us add a fast path for `x matches Foo` where x.name == "Foo" but not for superclasses or matches during data filtering.
- The memory hit isn't really a big deal we are already not optimal in this area in many dimensions (just using JSON). It's not what's going to slow us down or speed us up right now.

I think you have a choice to make between continuing this approach & implementing enrich message everywhere or sending the class name into the core & using that in `to_polar` in the core when formatting external instances.

languages/python/oso/polar/build.py
```diff
@@ -21,6 +21,7 @@
     libs.append("Ws2_32.lib")
     libs.append("advapi32.lib")
     libs.append("userenv.lib")
+    libs.append("bcrypt.lib")
```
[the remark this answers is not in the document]
gj: from #1442 — looks like that commit was merged into this PR but not #1439

polar-core/src/terms.rs
```diff
@@ -43,6 +43,7 @@ pub struct ExternalInstance {
     pub instance_id: u64,
     pub constructor: Option<Term>,
     pub repr: Option<String>,
+    pub class_repr: Option<String>,
```
gkaemmer, line 46: Is there a reason not to have this reference the instance ID of the class instance itself, instead of a string representation? I feel like that is a direction we've talked about moving before (i.e. it would make it trivial to prevent `isa` checks from crossing the FFI boundary), and it feels like it'd be essentially this same change with only some minor tweaks.

patrickod: the reason for choosing `Option<String>` over `class_id` was one of implementation expediency. I originally started with that plan but chose the string alternative as it could be implemented without necessitating FFI changes in all of the host languages which turned out to be somewhat tricker than I expected.

gj: We're going to make the name -> ID shift in a follow-up

polar-core/src/formatting.rs
```diff
@@ -386,12 +386,20 @@ pub mod to_polar {
...
             } else {
                 // Print out external instances like ^{id: 123}
                 // NOTE: this format is used by host libraries to enrich output
                 // messages with native representations of the instances.
-                format!("^{{id: {}}}", self.instance_id)
+                format!(
+                    "^{{id: {}}} TYPE `{}`",
+                    self.instance_id,
+                    self.class_repr.as_ref().unwrap_or(&"UNKNOWN".to_string())
+                )
```
gj, line 402: This assumption might already be broken by the existing `ExternalInstance::to_polar` implementation, but the goal of the `ToPolarString` trait is that it should be a "reverse-parser" that turns parsed values back into valid Polar strings. If it's already broken, then I don't mind shipping this change w/ an inline comment explaining the divergence from the rest of the `ToPolarString` impls.

gj: On second thought, I think we should leave the `ExternalInstance::to_polar` impl exactly how it was before this PR and then create a separate function that's only used in our logging code that invokes `ExternalInstance::to_polar` and then adds the

```
 TYPE `{}`
```

suffix. What do you think?

polar-core/src/vm.rs
```diff
@@ -783,10 +783,6 @@ impl PolarVirtualMachine {
         }
     }
 
-    fn log(&self, message: &str, terms: &[&Term]) {
-        self.log_with(LogLevel::Info, || message, terms)
-    }
-
     fn log_with<F, R>(&self, level: LogLevel, message_fn: F, terms: &[&Term])
```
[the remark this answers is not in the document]
patrickod: 👍🏻

polar-core/src/vm.rs
```diff
@@ -1710,11 +1706,19 @@ impl PolarVirtualMachine {
                 let constructor = args.pop().unwrap();
 
                 let instance_id = self.new_id();
+
+                let class = &constructor.value().as_call().unwrap().name;
+                let class_repr = if self.kb.read().unwrap().is_constant(class) {
+                    Some(constructor.to_polar())
```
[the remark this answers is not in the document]
patrickod: 👍🏻

docs/content/any/project/changelogs/NEXT.md
```diff
@@ -30,6 +30,10 @@ draft: true
   before upgrading.
 {{% /callout %}}
 
+* Previously we supported `POLAR_LOG=trace` and `POLAR_LOG=1` which enabled verbose "TRACE"-level logging of the execution of queries and their constituent goals within the Polar VM.
+* `POLAR_LOG=on` was slightly less verbose than `POLAR_LOG=trace` but still produced a voluminous output which made it hard to parse and follow the execution of a particular query.
+* To enable easier query debugging we have broken out `POLAR_LOG` into new discrete `INFO` and `TRACE` levels. Specifying `POLAR_LOG=info` will cause Polar to emit a more concise log output intended to be consumed by developers as they build and debug their Polar policies. The more verbose TRACE output is still available through `POLAR_LOG=trace`. Check out our [documentation](/reference/tooling/tracing.html) for more information on tracing.
```
gj, line 35: Random thought: do we care about casing? since we're referring to the levels as `INFO` and `TRACE` might someone expect that `POLAR_LOG=TRACE` would give them trace output instead of info?

docs/content/any/project/changelogs/NEXT.md
```diff
@@ -30,6 +30,10 @@ draft: true
   before upgrading.
 {{% /callout %}}
 
+* Previously we supported `POLAR_LOG=trace` and `POLAR_LOG=1` which enabled verbose "TRACE"-level logging of the execution of queries and their constituent goals within the Polar VM.
```
gj, line 33:
```suggestion
* Previously we supported `POLAR_LOG=trace` and `POLAR_LOG=on`, both of which enabled slightly different but verbose "TRACE"-level logging of the execution of queries and their constituent goals within the Polar VM.
```

polar-core/src/formatting.rs
patrickod: as discussed captured this as a follow-on notion task https://www.notion.so/osohq/Resolve-format-changes-to-ExternalInstance-to_polar-9585fe6e6216477b8179e5c215533108

languages/js/src/types.ts
```diff
@@ -179,6 +179,7 @@ interface PolarInstance {
   ExternalInstance: {
     instance_id: number;
     repr: string;
+    class_repr: string | undefined;
```
gj, line 182: nit

```suggestion
    class_repr?: string;
```

docs/content/any/project/changelogs/NEXT.md
patrickod: updated to be case agnostic with respect to specifying log levels

languages/python/oso/tests/test_repr.py
```diff
@@ -20,7 +20,7 @@ def test_repr_when_logging(polar, capsys):
     polar.load_str("f(_foo: Foo) if 1 = 1;")
     list(polar.query_rule("f", Foo()))
     captured = capsys.readouterr()
-    assert f"QUERY RULE: f({FOO_REPR})" in captured.out
+    assert f"QUERY RULE: f({FOO_REPR} TYPE `Foo`)" in captured.out
```
gj, line 23: I think it makes sense to print `Foo` instead of `{FOO_REPR}` here since `__repr__` is overridden at the instance level

docs/content/any/project/changelogs/NEXT.md
```diff
@@ -30,6 +30,10 @@ draft: true
   before upgrading.
 {{% /callout %}}
```
gj, line 32: Missing a `#####` heading here — peep the one below for an example (`##### Second parameter...`)

languages/rust/oso/src/host/value.rs
```diff
@@ -110,8 +110,9 @@ This may mean you performed an operation in your policy over an unbound variable
                 let id = host.cache_instance(instance.clone(), None);
                 Value::ExternalInstance(ExternalInstance {
                     constructor: None,
-                    repr: Some(std::any::type_name::<Self>().to_owned()),
                     instance_id: id,
+                    repr: Some(std::any::type_name::<Self>().to_owned()),
+                    class_repr: None,
```
gj, line 115: Should this be the same as `repr` on the line above?

polar-core/src/vm.rs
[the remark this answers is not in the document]
gj: did you push the rename? I'm not seeing it on GitHub

languages/go/types/polar_types.go
```diff
@@ -201,6 +201,8 @@ type ExternalInstance struct {
 	Constructor *Term `json:"constructor"`
 	// Repr
 	Repr *string `json:"repr"`
+	// Class Repr
```
gj, line 204: Need to run `polar-crossgen` before shipping

docs/content/any/reference/tooling/tracing.md
```diff
@@ -7,67 +7,36 @@ description: Polar tracing shows logs of how a query is evaluated.
...
-[debug]           BACKTRACK
-[debug]           HALT
-False
-```
+{{<literalInclude path="reference/tooling/info" tabGroup="tracing">}}
+{{<literalInclude path="reference/tooling/trace" tabGroup="tracing">}}
 
-Debug mode can be disabled by setting `POLAR_LOG=0` or `POLAR_LOG=off`.
\ No newline at end of file
+All trace output can be disabled by setting `POLAR_LOG=0` or `POLAR_LOG=off`.
```
gj, line 42: to make it clear that these will disable all info/debug output in addition to all trace output
```suggestion
All `POLAR_LOG` output can be disabled by setting `POLAR_LOG=0` or `POLAR_LOG=off`.
```

docs/content/any/reference/tooling/tracing.md
```diff
@@ -7,67 +7,36 @@ description: Polar tracing shows logs of how a query is evaluated.
...
 # Tracing
 
-Polar tracing shows logs of how a query is evaluated.
+Polar's tracing support prints logs of how a query is evaluated. It is available in both `INFO` and `TRACE` level outputs.
 
-## Enabling Tracing
+## Enabling `INFO` output 
 
-Debug mode is enabled by setting an environment variable: `POLAR_LOG=1`.
+`INFO` traces are viewed by setting the environment variable: `POLAR_LOG=info`. The `INFO` output is a more concise subset of `TRACE`, and is intended to by used by those developing and debugging polar policies in their local environment.
```
gj, line 14: nit: comma splice

```suggestion
`INFO` traces are viewed by setting the environment variable: `POLAR_LOG=info`. The `INFO` output is a more concise subset of `TRACE` and is intended to by used by those developing and debugging polar policies in their local environment. 
```

docs/content/any/reference/tooling/tracing.md
```diff
@@ -7,67 +7,36 @@ description: Polar tracing shows logs of how a query is evaluated.
...
+## Enabling `INFO` output 
 
-Debug mode is enabled by setting an environment variable: `POLAR_LOG=1`.
+`INFO` traces are viewed by setting the environment variable: `POLAR_LOG=info`. The `INFO` output is a more concise subset of `TRACE`, and is intended to by used by those developing and debugging polar policies in their local environment. 
 
-It will print out `[debug]` messages during query evaluation that show how the query is executed.
-Notable things include:
+It will print out `[oso][info]` messages for many notable points of the query execution, for example:
 
+ * Queries and sub-query evaluations.
```
gj, line 18: right?

```suggestion
 * Queries for rules.
```

docs/content/any/reference/tooling/tracing.md
```diff
@@ -7,67 +7,36 @@ description: Polar tracing shows logs of how a query is evaluated.
...
-* Queries and sub-query evaluations.
+`TRACE` logs are viewed by setting an environment variable: `POLAR_LOG=trace`.
 
+`POLAR_LOG=trace` exposes a more verbose view of a query execution in Polar and extends the contents of `POLAR_LOG=info`. The `POLAR_LOG=trace` traces will emit a log for every Goal encountered by the Polar virtual machine during the query execution. Notable examples include:
 
-* Showing values that are bound to variables for each expression.
-
-
-* Showing which rules a Polar predicate call will evaluate.
+* Showing parameter type checking information during query applicability tests
```
gj, line 30: nit

```suggestion
* Showing parameter type checking information during query applicability tests.
```

docs/content/any/reference/tooling/tracing.md
```diff
@@ -7,67 +7,36 @@ description: Polar tracing shows logs of how a query is evaluated.
...
 
+* Showing values that are bound to variables for each expression.
 
 * Showing values returned from calls into the application.
```
gj, line 35:
```suggestion

* All query evaluations; not limited to rules like `POLAR_LOG=info`.
```

languages/js/src/types.ts
patrickod: 👌🏻

polar-core/src/vm.rs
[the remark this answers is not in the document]
patrickod: pushed!

docs/content/any/project/changelogs/NEXT.md
patrickod: updated w/ heading

gj (approved):
:shipit:

PR #1445, opened 2021-12-14
isa/cc commutes with unify

gw000:
Before:
```
x = y and x matches Integer and y matches String # fails
x matches Integer and y matches String and x = y # succeeds
```

After:
```
x = y and x matches Integer and y matches String # fails
x matches Integer and y matches String and x = y # fails
```

Changes:
- When two partially bound variables are unified for the first time, a binding is created & the existing partials are rewritten & queried (similar to the grounding bug fix).

PR checklist:

- [x] Added changelog entry.

### DIFF
--- a/languages/rust/oso/tests/test_polar.rs
+++ b/languages/rust/oso/tests/test_polar.rs
@@ -969,3 +969,26 @@ fn test_rule_types() {
         })
     ));
 }
+
+#[test]
+fn test_exclusive_isa() {
+    common::setup();
+    let mut oso = test_oso();
+    oso.qnull("x matches Integer and x matches String");
+    oso.qnull("x.a matches Integer and x.a matches String");
+    let policy = r#"
+        foo(x: Integer) if x matches String;
+        boo(x) if x.a matches String and x.a matches Integer;
+        moo(x: Integer) if y matches String and y = x;
+        goo(x) if x.a matches String and y.a matches Integer and x = y;
+
+    "#;
+    oso.load_str(policy);
+    oso.qnull("foo(x)");
+    oso.qnull("boo(x)");
+    oso.qnull("goo(x)");
+    oso.qnull("moo(x)");
+    // TODO(gw) support expressions
+    // oso.qeval("x matches Integer and y matches String");
+    // oso.qeval("x.a matches Integer and x.b matches String");
+}
--- a/languages/rust/oso/tests/test_polar.rs
+++ b/languages/rust/oso/tests/test_polar.rs
@@ -981,13 +981,16 @@ fn test_exclusive_isa() {
         boo(x) if x.a matches String and x.a matches Integer;
         moo(x: Integer) if y matches String and y = x;
         goo(x) if x.a matches String and y.a matches Integer and x = y;
-
+        zoo(x) if x.a.b matches String and y.a.b matches Integer and x = y;
+        roo(x) if x.a.b matches String and x.a.b matches Integer;
     "#;
     oso.load_str(policy);
     oso.qnull("foo(x)");
     oso.qnull("boo(x)");
     oso.qnull("goo(x)");
     oso.qnull("moo(x)");
+    oso.qnull("roo(x)");
+    oso.qnull("zoo(x)");
     // TODO(gw) support expressions
     // oso.qeval("x matches Integer and y matches String");
     // oso.qeval("x.a matches Integer and x.b matches String");
--- a/polar-core/src/bindings.rs
+++ b/polar-core/src/bindings.rs
@@ -425,35 +425,6 @@ impl BindingManager {
     }
 }
 
-fn combine_partials(left: &Symbol, l: Operation, right: &Symbol, r: Operation) -> Operation {
-    use Operator::*;
-    fn unifies_symbols(Operation { operator, args }: &Operation) -> bool {
-        matches!(operator, Operator::Unify if args[0].value().as_symbol().is_ok() && args[1].value().as_symbol().is_ok())
-    }
-
-    let (mut var_unifys, others): (Vec<_>, Vec<_>) = l
-        .constraints()
-        .into_iter()
-        .chain(r.constraints().into_iter())
-        .partition(unifies_symbols);
-    var_unifys.push(op!(Unify, term!(left.clone()), term!(right.clone())));
-
-    let termify = |i: Vec<_>| i.into_iter().map(Term::from).collect();
-    let (var_unifys, others) = (termify(var_unifys), termify(others));
-    Operation {
-        operator: And,
-        args: vec![],
-    }
-    .merge_constraints(Operation {
-        operator: And,
-        args: var_unifys,
-    })
-    .merge_constraints(Operation {
-        operator: And,
-        args: others,
-    })
-}
-
 // Private impls.
 impl BindingManager {
     /// Bind two variables together.
@@ -558,7 +529,7 @@ impl BindingManager {
                     self.add_binding(left, term!(right.clone()));
                     self.add_binding(right, term!(left.clone()));
 
-                    let term = term!(combine_partials(left, lp, right, rp));
+                    let term = term!(op!(And, term!(lp), term!(rp)));
                     Ok(Some(Goal::Query { term }))
                 }
             }
--- a/polar-core/src/bindings.rs
+++ b/polar-core/src/bindings.rs
@@ -3,10 +3,14 @@
 /// Bindings associate variables in the VM with constraints or values.
 use std::collections::{HashMap, HashSet};
 
-use crate::error::RuntimeError;
-use crate::folder::{fold_list, fold_term, Folder};
-use crate::terms::{has_rest_var, Operation, Operator, Symbol, Term, Value};
-use crate::vm::Goal;
+use crate::{
+    error::RuntimeError,
+    folder::{fold_list, fold_term, Folder},
+    formatting::ToPolarString,
+    partial::sub_vars,
+    terms::{has_rest_var, Operation, Operator, Symbol, Term, Value},
+    vm::Goal,
+};
 
 type Result<T> = core::result::Result<T, RuntimeError>;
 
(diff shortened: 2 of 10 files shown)


polar-core/src/bindings.rs
```diff
@@ -425,52 +427,91 @@ impl BindingManager {
     }
 }
 
+fn combine(left: &Symbol, l: Operation, right: &Symbol, r: Operation) -> Operation {
```
gw000, line 430: needs a better name

polar-core/src/bindings.rs
```diff
@@ -425,52 +427,91 @@ impl BindingManager {
...
-                self.add_binding(right, left_value);
-            }
-            (
-                BindingManagerVariableState::Unbound,
-                BindingManagerVariableState::Bound(right_value),
-            ) => {
-                self.add_binding(left, right_value);
+        match (
+            (left, self._variable_state(left)),
+            (right, self._variable_state(right)),
```
gw000, line 493: The purpose of doing the match this way is to avoid having to write duplicate branches for left/right handed versions of the same case.

polar-core/tests/integration_tests.rs
```diff
@@ -49,8 +49,8 @@ fn no_isa(_: Term, _: Symbol) -> bool {
     true
 }
 
-fn no_is_subspecializer(_: u64, _: Symbol, _: Symbol) -> bool {
-    false
+fn no_is_subspecializer(_: u64, a: Symbol, b: Symbol) -> bool {
+    a == b
```
gw000, line 53: Can take this out but it seems to make sense

polar-core/src/vm.rs
```diff
@@ -1172,42 +1172,54 @@ impl PolarVirtualMachine {
...
 
+                use Operator::*;
                 // TODO (dhatch): what if there is more than one var = dot_op constraint?
                 // What if the one there is is in a not, or an or, or something
-                let lhs_of_matches = simplified
+                let mut lhss_of_matches: Vec<_> = simplified
                     .constraints()
                     .into_iter()
-                    .find_map(|c| {
+                    .filter_map(|Operation { operator, args }| match operator {
```
gw000, line 1181: The main change here is to use `filter_map` instead of `find_map`. `choose_conditional` takes a collection as its first arg anyway so it's compatible

polar-core/src/vm.rs
```diff
@@ -1172,42 +1172,54 @@ impl PolarVirtualMachine {
...
-                    .unwrap_or_else(|| left.clone());
+                            matches!(args[0].value().as_expression(), Ok(Operation { operator: Dot, .. }))
+                            => Some(args[0].clone()),
+
+                        _ => None,
+                    }).collect();
+
+                // FIXME(gw) always do this?
+                if lhss_of_matches.is_empty() {
+                    lhss_of_matches.push(left.clone());
```
gw000, line 1202: Corresponds to the `unwrap_or_else` in the old code

polar-core/src/partial/partial.rs
```diff
@@ -1215,7 +1220,7 @@ mod test {
         p.register_constant(sym!("y"), term!(value!(op!(And))))?;
         let mut q = p.new_query_from_term(term!(call!("f", [sym!("x"), sym!("y")])), false);
         let next = next_binding(&mut q)?;
-        assert_partial_expressions!(next, "x" => "_this = y", "y" => "x = _this");
+        assert_eq!(next[&sym!("x")], var!("y"));
```
gw000, line 1223: this seems like a better answer

polar-core/src/partial/partial.rs
```diff
@@ -990,7 +990,12 @@ mod test {
         let p = Polar::new();
         p.load_str("f(x) if x = x + 0;")?;
         let mut q = p.new_query_from_term(term!(call!("f", [sym!("x")])), false);
-        assert_partial_expression!(next_binding(&mut q)?, "x", "_this + 0 = _this");
+        // FIXME(gw) try not to change this test ??
+        assert_partial_expression!(
+            next_binding(&mut q)?,
+            "x",
+            "_this + 0 = __op_1_5 and _this + 0 = _this"
```
gw000, line 997: this isn't ideal

polar-core/src/bindings.rs
samscott89: What's this method doing? A comment would be super helpful.

polar-core/src/vm.rs
```diff
@@ -1222,7 +1234,7 @@ impl PolarVirtualMachine {
 
                 // Run compatibility check.
                 self.choose_conditional(
-                    vec![Goal::Run { runnable }],
+                    runnables,
```
dhatch, line 1237: So this means all LHS must have a compatible isa?

polar-core/tests/integration_tests.rs
dhatch: In the host libraries, `is_subspecializer` returns false for equivalent classes. A class A is only a subspecializer if it is a subclass of B w.r.t. the MRO of an instance i. So I think this should return false even for equivalent classes.

You can see the implementation in python in host.py, I just checked this for reference.

polar-core/src/bindings.rs
dhatch: 👍 that sounds fine. and a comment

polar-core/src/bindings.rs
```diff
@@ -425,52 +427,91 @@ impl BindingManager {
     }
 }
 
+fn combine(left: &Symbol, l: Operation, right: &Symbol, r: Operation) -> Operation {
+    let unis = l
```
dhatch, line 431:
```suggestion
    let unifications = l
```

would make it easier for me to read, I'd know what it is immediately rather than thinking for a few

polar-core/src/bindings.rs
```diff
@@ -425,52 +427,91 @@ impl BindingManager {
...
+                    args[0].value().as_symbol().unwrap().clone(),
+                    args[1].value().as_symbol().unwrap().clone(),
+                ))
+            }
+            _ => None,
+        })
+        .chain(std::iter::once((left.clone(), right.clone())));
+    let sub = crate::data_filtering::partition_equivs(unis)
+        .into_iter()
+        .flat_map(|cls| {
```
dhatch, line 453: What's `cls` mean? I think `class`? but that doesn't seem right to me

polar-core/src/bindings.rs
```diff
@@ -425,52 +427,91 @@ impl BindingManager {
...
+            if let Some(rep) = cls.iter().find(|v| !v.is_temporary_var()) {
```
dhatch, line 454: I'm having a little trouble following what this is doing.

Seems like outputting _x => _x if there is no non-temporary variable in the equivalence set or _t => x if there is a non-temporary?

polar-core/src/bindings.rs
```diff
@@ -425,52 +427,91 @@ impl BindingManager {
...
+            } else {
+                cls.iter()
+                    .map(|pv| (pv.clone(), pv.clone()))
+                    .collect::<Vec<_>>()
+            }
+        })
+        .collect::<HashMap<_, _>>();
+
+    let subbed = |o| {
+        sub_vars(sub.clone(), term!(o))
```
dhatch, line 467: Why can't a ref of `sub` just be passed to `sub_vars`?

polar-core/src/bindings.rs
```diff
@@ -425,52 +427,91 @@ impl BindingManager {
...
+}
+
 // Private impls.
 impl BindingManager {
     /// Bind two variables together.
     fn bind_variables(&mut self, left: &Symbol, right: &Symbol) -> Result<Option<Goal>> {
+        if left == right {
+            return Ok(None);
+        }
+        let pc = |cy: &Vec<Symbol>| cy.iter().map(|t| t.0.clone()).collect::<Vec<_>>();
```
dhatch, line 488: what's this?

dhatch (changes requested):
I'm having a bit of trouble following why this fixes the issue. I can see how that change with binding can help (because all the matches are re-evaluated and we will discover they are incompatible. Maybe we can talk through live.

How can we measure the performance impact of this?

polar-core/src/bindings.rs
gw000: I changed the whole way this works, it's a lot simpler now 👍

gw000: (but it stood for "class" as in "equivalence class")

polar-core/src/bindings.rs
gw000: For debug logging, just forgot to take it out

polar-core/src/bindings.rs
gw000: replaced this entire function with `term!(op!(And, term!(left), term!(right)))`

polar-core/src/partial/partial.rs
gw000: whatever it was it's gone now

languages/rust/oso/tests/test_polar.rs
```diff
@@ -969,3 +969,26 @@ fn test_rule_types() {
...
+        boo(x) if x.a matches String and x.a matches Integer;
+        moo(x: Integer) if y matches String and y = x;
+        goo(x) if x.a matches String and y.a matches Integer and x = y;
+
+    "#;
+    oso.load_str(policy);
+    oso.qnull("foo(x)");
+    oso.qnull("boo(x)");
+    oso.qnull("goo(x)");
+    oso.qnull("moo(x)");
```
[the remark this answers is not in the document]
gw000: Done

polar-core/src/partial/partial.rs
```diff
@@ -2104,12 +2109,7 @@ mod test {
             &[|r: Bindings| {
                 assert_partial_expressions!(r,
                     "x" => "_this > 0 and _this != 1",
-                    // Right now we output "y" => "_this <= 1".
```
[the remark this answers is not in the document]
gw000: It's back to the original now so I'll put the comment back

polar-core/src/vm.rs
gw000: Yes, I think that's correct behavior, isn't it? Before it was just trying to pick one according to a complicated set of rules and saying *that* one has to be compatible. This could possibly affect performance but I don't know how likely that actually is

gw000: Anyway -- the fix actually seems to work without this change, so I'll revert it. But it's obviously not obvious that what we're doing now is correct.

dhatch (approved):

PR #1446, opened 2021-12-16
Data Filtering: forall / not / or

gw000:
PR checklist:

- [x] Added changelog entry.


polar-core/src/normalize.rs
```diff
@@ -0,0 +1,317 @@
...
+            Boolean(b) => self.clone_with_value(Boolean(!b)),
+
+            Expression(Operation { operator, args }) => match operator {
+                // cancel double negatives
+                Not => self.clone_with_value(args[0].value().clone()),
+
+                // swap and/or using De Morgan's to keep the nots nested
+                // as deep as possible
+                And => {
+                    assert_eq!(args.len(), 2);
```
gkaemmer, line 18: Is this guaranteed to be true? At least in the parser side of things we kind of "flatten" arguments to `or` and `and` so that they may have N args each, wondering if that might ever be the case here

saolsen (approved):
This is really cool.
We'll have to add the filter parsing for the new ops to python and go but not as part of this PR.

polar-core/src/normalize.rs
gw000: `btnf` takes care of that 😃

languages/ruby/spec/oso/polar/data_filtering/new_filter_spec.rb
```diff
@@ -35,12 +32,12 @@ class Planet < ActiveRecord::Base
             Person,
             fields: {
               name: String,
-              sign_name: String,
+              sign_id: Integer,
```
gw000, line 35: These were changed so there would be a numeric field for testing inequality operators

polar-core/src/filter.rs
```diff
@@ -541,9 +562,90 @@ where
     std::iter::once(x).collect()
 }
 
+fn normalize(t: Term) -> Vec<Term> {
```
dhatch, line 565: A comment summing up the normalization that this applies would be helpful.

polar-core/src/normalize.rs
```diff
@@ -0,0 +1,317 @@
...
+
+    /// Condition input for dnf/cnf transformation
+    /// - negations fully nested
+    /// - double negatives removed
+    /// - and/or nodes form a binary tree
+    fn pre_norm(&self) -> Self {
+        self.btnf().nnf()
+    }
+
+    pub fn dnf(&self) -> Self {
```
dhatch, line 142: These public functions should not use abbreviations because they may be called anywhere in the code. A caller should be able to tell what this is without having to jump to this file. (And jumping to this file won't help right now because there's not a doc string).

polar-core/src/normalize.rs
```diff
@@ -0,0 +1,317 @@
...
+
+pub fn and_(l: Term, r: Term) -> Term {
+    term!(op!(And, l, r))
+}
+
+pub fn not_(t: Term) -> Term {
+    term!(op!(Not, t))
+}
+
+fn opnp(l: &Term, op: Operator) -> bool {
```
dhatch, line 163: What does this mean? `is_op` seems like an easier to understand and similarly short name.

polar-core/src/normalize.rs
```diff
@@ -0,0 +1,317 @@
...
+                }))
+            }
+        }
+    }
+
+    fn _hs(&self, n: usize) -> Self {
+        self.value().as_expression().unwrap().args[n].clone()
+    }
+
+    fn lhs(&self) -> Self {
```
dhatch, line 86: assert self len is 2? Otherwise LHS & RHS doesn't have meaning.

polar-core/src/normalize.rs
dhatch: Now realizing `p` means predicate. In any case, this is a reason why we should use descriptive names.

polar-core/src/normalize.rs
```diff
@@ -0,0 +1,317 @@
+use crate::terms::*;
+
+impl Term {
+    /// intelligently negate an expression
```
dhatch, line 4: Comment, input must be in `btnf`.

polar-core/tests/integration_tests.rs
```diff
@@ -414,6 +414,32 @@ fn test_jealous() -> TestResult {
     Ok(())
 }
 
+#[test]
+fn test_forall_with_dot_lookup() -> TestResult {
+    let p = polar();
+    p.load_str(
+        r#"
+        foo(x) if forall(y in x.ys, y != 0);
```
dhatch, line 422: What about w/ multiple terms on RHS? `y != 0 and y < 5`

polar-core/src/normalize.rs
```diff
@@ -0,0 +1,317 @@
...
+    fn rhs(&self) -> Self {
+        self._hs(1)
+    }
+
+    /// normalize an expression by fully distributing logical connectives. the
+    /// expression must already be in btnf and nnf. the arguments are two
+    /// predicate / constructor pairs:
+    /// - p1/c1: predicate/constructor for *inner* connective (and for dnf, or for cnf)
+    /// - p2/c2: predicate/constructor for *outer* connective (or for dnf, and for cnf)
+    fn distribute(
```
dhatch, line 98: Is there a reference you could link maybe about this operation? I haven't reviewed it for correctness or thought about edge cases at all -- assuming it's translated from somewhere / well known?

polar-core/src/filter.rs
dhatch: Can this use the code in `normalize` instead?

languages/ruby/spec/oso/polar/data_filtering/new_filter_spec.rb
```diff
@@ -13,19 +13,16 @@
...
+      belongs_to :sign
     end
 
     class Planet < ActiveRecord::Base
-      self.primary_key = :name
-      has_many :signs, foreign_key: :planet_name
+      has_many :signs
     end
 
     context 'astrology' do # rubocop:disable Metrics/BlockLength
```
dhatch, line 28: A test of `not in` independently of `forall` would be great.

polar-core/src/filter.rs
dhatch: Is this fixing something or changing the disjunctive normal form?

gw000: It's just flattening the tree into a vector, so the name isn't really right, I'll change it to `vec_of_ands` or something

polar-core/src/normalize.rs
gw000: Already assumed because of the prior call to `btnf`

polar-core/src/normalize.rs
gw000: No, it's just literally what you do when you distribute the connectives over each other 🤷‍♀️

polar-core/src/normalize.rs
gw000: I'm going to nest the function defs so it's impossible to call them in the wrong way 👍

polar-core/tests/integration_tests.rs
gw000: Sure, it's not really relevant to this test though -- internally that's still just one expression & it works exactly the same

polar-core/src/normalize.rs
gw000: abbreviation expanded

polar-core/src/normalize.rs
gw000: changed

polar-core/tests/integration_tests.rs
dhatch: Oh okay, no worries then

polar-core/src/normalize.rs
dhatch: nice idea

polar-core/src/normalize.rs
dhatch: <3

polar-core/src/filter.rs
dhatch: Ohhhh I totally missed that

dhatch (approved):
🥳

PR #1448, opened 2021-12-20
go data filtering

saolsen:


languages/go/oso.go
```diff
@@ -398,6 +407,111 @@ func (o Oso) AuthorizedFields(actor interface{}, action interface{}, resource in
...
+
+func (o Oso) dataFilter(actor interface{}, action interface{}, resource_type string) (*Query, interface{}, error) {
+	query, err := (*o.p).queryRule("allow", actor, action, types.Variable("resource"))
+	if err != nil {
+		return nil, nil, err
+	}
+
+	query.SetAcceptExpression(true)
+
+	constraint :=
```
gw000, line 422: Wow this vs Ruby 🥲 
```ruby
def type_constraint(var, cls)
  Expression.new(
    'And',
    [Expression.new('Isa', [var, Pattern.new(class_to_name(cls), {})])]
  )
end
```

saolsen: lol yeah, it's wild

languages/go/internal/ffi/ffi.go
```diff
@@ -231,6 +231,38 @@ func (p PolarFfi) RegisterMro(name string, mro []uint64) error {
 	return err
 }
 
+// yeah, not ideal types yet lol
+func (p PolarFfi) BuildDataFilter(user_types map[string]map[string]map[string]map[string]string, partials []map[string]map[string]types.Term, resource_var string, resource_type string) (*types.Filter, error) {
```
gw000, line 235: Really shows the power of Go's type system lol

languages/go/internal/host/host.go
```diff
@@ -427,11 +439,94 @@ func (h Host) ToGo(v types.Term) (interface{}, error) {
...
+				fields_map[k] = map[string]map[string]string{
+					"Relation": {
+						"kind":            t.Kind,
+						"other_class_tag": t.OtherType,
+						"my_field":        t.MyField,
+						"other_field":     t.OtherField,
+					},
+				}
+			default:
+				// error, this aint good
```
gw000, line 479: return an error here?

languages/go/internal/host/host.go
```diff
@@ -427,11 +439,94 @@ func (h Host) ToGo(v types.Term) (interface{}, error) {
...
+
+func (h *Host) BuildQuery(filter *Filter) (interface{}, error) {
+	if h.adapter == nil {
+		return nil, fmt.Errorf("must register an adapter to use data filtering")
+	}
+
+	return (h.adapter).BuildQuery(filter)
+}
+
+func (h *Host) ExecQuery(query interface{}) (interface{}, error) {
```
gw000, line 500: `ExecuteQuery` to be consistent with python/ruby

languages/go/internal/ffi/ffi.go
```diff
@@ -231,6 +231,38 @@ func (p PolarFfi) RegisterMro(name string, mro []uint64) error {
...
+	if err != nil {
+		panic(err)
+	}
+
+	processMessages(p)
+
+	var filter types.Filter
+	err = json.Unmarshal([]byte(*filterJson), &filter)
+	if err != nil {
+		panic(err)
```
gw000, line 260: Could we `return nil, err` or something?

languages/go/types/polar_types.go
```diff
@@ -1039,132 +1039,82 @@ func (result *Operator) UnmarshalJSON(b []byte) error {
...
 
 func (variant Operator) MarshalJSON() ([]byte, error) {
-	switch inner := variant.OperatorVariant.(type) {
+	switch variant.OperatorVariant.(type) {
 
 	case OperatorDebug:
-		return json.Marshal(map[string]OperatorDebug{
-			"Debug": inner,
-		})
+		return json.Marshal("Debug")
```
gw000, line 1045: I don't understand what this was even doing before

saolsen: It was just wrong, Instead of serializing the ops as a string, eg `{"Expression": {"Operator": "Eq", args:  [...]}` It was serializing the operations like this  `{"Expression": {"Operator": {"Debug": {}}, args:  [...]}`

languages/go/internal/host/host.go
```diff
@@ -69,14 +80,15 @@ func (h Host) getClass(name string) (*reflect.Type, error) {
...
-func (h Host) CacheClass(cls reflect.Type, name string, constructor reflect.Value) error {
+func (h Host) CacheClass(cls reflect.Type, name string, constructor reflect.Value, fields map[string]interface{}) error {
 	if v, ok := h.classes[name]; ok {
 		return errors.NewDuplicateClassAliasError(name, cls, v)
 	}
 	h.classes[name] = cls
 	if constructor.IsValid() {
 		h.constructors[name] = constructor
 	}
+	h.fields[name] = fields
```
gw000, line 91: This is different from how we do it now in python/ruby/js, which is to have one map where the values are type info structs. Since we no longer have 3 query callbacks per type it doesn't seem so bad to do it this way for now. I'll make a task for next cycle to bring it in line

languages/go/types/polar_types.go
```diff
@@ -2980,3 +2930,151 @@ func (variant Value) MarshalJSON() ([]byte, error) {
...
+func (Projection) isDatum() {}
+func (Immediate) isDatum()  {}
+
+type FilterRelation struct {
+	FromTypeName  string
+	FromFieldName string
+	ToTypeName    string
+}
+
+// todo
```
gw000, line 3024: looks done?

languages/go/tests/enforcement_test.go
```diff
@@ -196,6 +213,35 @@ func TestAuthorizedActions(t *testing.T) {
...
+	o.SetDataFilteringAdapter(&TestAdapter{})
+
+	o.LoadString("allow(_actor: User, \"get\", resource: Widget) if resource.Parent.Id = 1;")
+
+	actor := User{Name: "Sally"}
+	resource := Widget{Id: 1}
+	_, _ = o.IsAllowed(actor, "get", resource)
+
+	results, err := o.AuthorizedQuery(actor, "get", "Widget")
+	fmt.Printf("%v\n", results)
```
gw000, line 240: Is it possible to add a test based on https://github.com/osohq/oso-go-df-sqlboiler ?

saolsen: Not really, without adding sqlboiler as a dependency (as far as I can tell?)

gw000 (approved):
This looks generally fine but merging with no tests is scary 😨 As long as it's documented as a preview feature

PR #1450, opened 2021-12-21
[WIP] another partial isa/cc fix

gw000:
PR checklist:

- [x] Added changelog entry.


languages/python/oso/tests/data_filtering/test_basic.py
```diff
@@ -555,3 +556,147 @@ class Qux:
...
+        allow(_, "read", _n: Network);
+
+        has_permission(u: User, "read", a: Article{group}) if
+        a matches Article and
+            group matches Group and
+            has_permission(u, "read", group);
+
+        has_permission(u: User, "read", g: Group{network}) if
+        g matches Group and
+            network matches Network and
```
dhatch, line 600: seeing gabe's comment in slack, would be good to remove these `matches` if possible.

dhatch (approved):

languages/python/oso/tests/data_filtering/test_basic.py
gw000: ✂️

PR #1456, opened 2021-12-22
Add valid shorthand rules to KB even when not all shorthand rules are valid

gj:
PR checklist:

- [ ] Added changelog entry.

### DIFF
--- a/polar-core/src/polar.rs
+++ b/polar-core/src/polar.rs
@@ -301,6 +301,8 @@ impl Polar {
 
 #[cfg(test)]
 mod tests {
+    use std::collections::HashSet;
+
     use super::*;
     use crate::error::{
         ErrorKind::{Runtime, Validation},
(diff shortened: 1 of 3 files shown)


gkaemmer (approved):
Looks good. It's a bit scary that this doesn't break any tests -- would it be easy to add a test for the case where _some_ shorthand rules are inserted into the KB despite errors?

gj:
> Looks good. It's a bit scary that this doesn't break any tests -- would it be easy to add a test for the case where _some_ shorthand rules are inserted into the KB despite errors?

Great idea; will do

polar-core/src/polar.rs
```diff
@@ -398,4 +395,65 @@ mod tests {
         );
         assert!(!polar.kb.read().unwrap().has_rules());
     }
+
+    #[test]
+    fn test_valid_shorthand_rules_rewritten_in_presence_of_other_issues() {
+        let polar = Polar::new();
+        let src = r#"
+            allow(actor, action, resource) if
```
gkaemmer, line 403: Not sure I totally understand this test — what’s the error that would have prevented these rules from being rewritten into the KB?

gj: Ah yeah that's a good shout — it's not totally obvious. I'll add a comment explaining it

gj: Simpler test in af960f69

PR #1459, opened 2021-12-23
VS Code extension metrics Set -> Array

gj:


gneray (approved):
Solid engineering, as always.

PR #1460, opened 2021-12-24
VS Code extension: no more syntax highlighting flash

gj:
Being smarter about triggering `didOpen` events.

PR checklist:

- [ ] Added changelog entry.

### DIFF
--- a/vscode/oso/client/src/index.ts
+++ b/vscode/oso/client/src/index.ts
@@ -5,7 +5,6 @@ import { join } from 'path';
 import { debounce } from 'lodash';
 import {
   ExtensionContext,
-  languages,
   RelativePattern,
   TextDocument,
   Uri,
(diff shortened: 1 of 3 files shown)


vscode/oso/client/src/index.ts
```diff
@@ -84,53 +83,36 @@ function polarFilesInWorkspaceFolderPattern(folder: WorkspaceFolder) {
...
 // documents in the current workspace folder.
 //
 // [didOpen]: https://code.visualstudio.com/api/references/vscode-api#workspace.onDidOpenTextDocument
-async function reloadDocument(uri: Uri) {
+async function openDocument(uri: Uri) {
   const uriMatch = (d: TextDocument) => d.uri.toString() === uri.toString();
-  let doc = workspace.textDocuments.find(uriMatch);
-
-  if (doc) {
-    // If `doc` is already open, cycle its `languageId` from `polar` ->
```
samscott89, line 116: So was this logic not actually necessary? Or some other change that made it redundant?

gj: Not actually necessary. If a document is "open" in VS Code (the scare quotes are b/c this doesn't necessarily mean the document is open in an editor window — in practical terms it means that a `DidOpen` event has been emitted and a `DidClose` event has not yet been emitted), then we get `DidChange` events every time it's changed. If an un-opened document is changed, our `createChangeWatcher` will notice the change and, in turn, call `openDocument`, which will re-open the document in VS Code, transmitting the updated state (post-change) to our language server.

samscott89 (approved):

PR #1461, opened 2021-12-26
No metrics when non-Polar files are deleted

gj:


polar-language-server/src/lib.rs
```diff
@@ -141,61 +165,81 @@ impl PolarLanguageServer {
             // Nothing to do when we receive the `Initialized` notification.
             Initialized::METHOD => (),
 
-            _ => log(&format!("on_notification {} {:?}", method, params)),
+            _ => log("unexpected notification"),
```
samscott89, line 168: Super nit: potentially worth logging the method + params still for debugging purposes. Good change to the message though

polar-language-server/src/lib.rs
```diff
@@ -141,61 +165,81 @@ impl PolarLanguageServer {
...
         let uri = doc.uri.clone();
         if self.upsert_document(doc).is_none() {
-            log(&format!("updated untracked doc: {}", uri));
+            log(&format!("\tupdated untracked doc: {}", uri));
         }
         self.reload_kb()
     }
 
-    fn on_did_delete_files(&mut self, uris: Vec<Url>) -> Diagnostics {
+    fn on_did_change_watched_files(&mut self, uris: Vec<Url>) -> Diagnostics {
```
samscott89, line 191: Could you add a comment here saying that we currently only listen for deletions, so this method assumes all changes are deletions?

polar-language-server/src/lib.rs
gj: The method is now logged out here: https://github.com/osohq/oso/blob/9c0400beecd94697e36923e5f29d78fc4ca0ae86/polar-language-server/src/lib.rs#L56

I figured debug-printing the params would be noisy, but I guess it doesn't really matter. We should *never* hit this unless we're adding new LSP functionality. I can add the params back.

polar-language-server/src/lib.rs
gj: Sure, I can link back to this comment: https://github.com/osohq/oso/blob/9c0400beecd94697e36923e5f29d78fc4ca0ae86/polar-language-server/src/lib.rs#L95-L97

polar-language-server/src/lib.rs
samscott89: Cool, that makes sense. (happy with/without params tbh)

polar-language-server/src/lib.rs
samscott89: Yes please. It took me a little while to double-check the logic here and that this wasn't called elsewhere

samscott89 (approved):
lgtm assuming comment gets added.

polar-language-server/src/lib.rs
gj: 3267aca221f90c09afd8c7a114d57cb506f0e3cc

PR #1466, opened 2022-01-03
Elevate all APPLICABLE_RULE & RULE log entries to INFO

patrickod:
Previously we logged the first APPLICABLE_RULES line for the top level query to
INFO and all subsequent entries to TRACE.

This PR reverts the behavior and consolidates all rule-matching debugging logs
to the `INFO` log level where they can be read more easily without having to
parse through TRACE entries

### DIFF
--- a/polar-core/src/vm.rs
+++ b/polar-core/src/vm.rs
@@ -535,7 +535,7 @@ impl PolarVirtualMachine {
             Goal::TraceRule { trace } => {
                 if let Node::Rule(rule) = &trace.node {
                     self.log(
-                        LogLevel::Trace,
+                        LogLevel::Info,
                         || {
                             let source_str = self.rule_source(rule);
                             format!("RULE: {}", source_str)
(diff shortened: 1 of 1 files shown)


polar-core/src/vm.rs
```diff
@@ -535,7 +535,7 @@ impl PolarVirtualMachine {
             Goal::TraceRule { trace } => {
                 if let Node::Rule(rule) = &trace.node {
                     self.log(
-                        LogLevel::Trace,
+                        LogLevel::Info,
```
dhatch, line 538: Is this going to be redundant with the `QUERY RULE` log from L1485?

dhatch (changes requested):
Question about change to TraceRule logging

polar-core/src/vm.rs
patrickod: I don't believe so. The L1485 event is emitted at the top of the query for the outer call to the rule w/ the arguments whereas this L538 entry shows us entering into the body of a specific rule definition w/ accompanying source.

dhatch (approved):

PR #1467, opened 2022-01-03
Fix bug in partial, partial unification

dhatch:
Some changes to @gw000 's fix for part of the isa bug. A user had an issue where the UnhandledPartial exception was being raised for valid policies.

Their policy contained a unification:

```
has_relation(s, r, o) if s = o;
```

Upon inspection of the logs, partials for `o` did not contain all the constraints or the unification with `s`. I realized this is because we are no longer adding the `l = r` constraint on the `Partial`, `Partial` unify case in `BindingManager::bind_variables`.

I changed that case to:
1. merge the two partials
2. express all constraints in the partials in terms of the left variable (arbitrarily chosen)
3. add the `left = right` unification operation to both partials
4. (grounding added by @gw000) query the VM for the residual partial *without* the `left = right` unification to avoid infinite recursion

PR checklist:

- [ ] Added changelog entry.

### DIFF
--- a/languages/python/oso/tests/regression/test_unhandled_partial.py
+++ b/languages/python/oso/tests/regression/test_unhandled_partial.py
@@ -1,0 +1,79 @@
+import pytest
+
+
+def test_unhandled_partial():
+    """Spurious unhandled partial needs fix."""
+    POLICY = """\
+actor User {}
+
+allow(actor, action, resource) if has_permission(actor, action, resource);
+
+resource A {
+    permissions = ["Read"];
+    roles = ["User"];
+
+    "Read" if "User";
+}
+
+has_role(user: User, "User", a: A) if
+    a_role in a.groups and
+    "Read" = a_role.p and
+    a_role.group_id in user.group_ids;
+
+resource Aprime {
+    relations = { a: A };
+    permissions = ["Read"];
+
+    "Read" if "User" on "a";
+}
+
+has_relation(subject: A, "a", object: Aprime) if
+    subject = object;
+    """
+
+    from dataclasses import dataclass
+
+    from oso import Oso, Variable
+    from polar import Pattern, Expression
+
+    @dataclass
+    class Group:
+        permission: str
+        group_id: int
+
+    @dataclass
+    class A:
+        groups: list[Group]
+
+    @dataclass
+    class Aprime(A):
+        pass
+
+    @dataclass
+    class User:
+        group_ids: list[int]
+
+    oso = Oso()
+    oso.register_class(Aprime)
+    oso.register_class(A)
+    oso.register_class(User)
+
+    oso.load_str(POLICY)
+
+    constraint = Expression(
+        "And",
+        [Expression("Isa", [Variable("resource"), Pattern("Aprime", {})])],
+    )
+
+    # Unhandled partial raised here.
+    results = list(
+        oso.query_rule(
+            "allow",
+            User(group_ids=[0]),
+            "Read",
+            Variable("resource"),
+            accept_expression=True,
+            bindings={"resource": constraint},
+        )
+    )
+    print(results)
--- a/languages/python/oso/tests/regression/test_unhandled_partial.py
+++ b/languages/python/oso/tests/regression/test_unhandled_partial.py
@@ -1,5 +1,5 @@
 import pytest
-
+from typing import List
 
 def test_unhandled_partial():
     """Spurious unhandled partial."""
--- a/languages/python/oso/tests/regression/test_unhandled_partial.py
+++ b/languages/python/oso/tests/regression/test_unhandled_partial.py
@@ -2,7 +2,7 @@ import pytest
 
 
 def test_unhandled_partial():
-    """Spurious unhandled partial needs fix."""
+    """Spurious unhandled partial."""
     POLICY = """\
 actor User {}
 
--- a/languages/python/oso/tests/regression/test_unhandled_partial.py
+++ b/languages/python/oso/tests/regression/test_unhandled_partial.py
@@ -1,4 +1,3 @@
-import pytest
 from typing import List
 
 
--- a/languages/python/oso/tests/regression/test_unhandled_partial.py
+++ b/languages/python/oso/tests/regression/test_unhandled_partial.py
@@ -1,11 +1,12 @@
 import pytest
 from typing import List
 
+
 def test_unhandled_partial_gh1467():
     """Test that previously failed due to incorrect partial unification.
-    
+
     Submitted by BAH.
-    
+
     Fixed in https://github.com/osohq/oso/pull/1467."""
     POLICY = """\
 actor User {}
--- a/languages/python/oso/tests/regression/test_unhandled_partial.py
+++ b/languages/python/oso/tests/regression/test_unhandled_partial.py
@@ -65,7 +65,7 @@ has_relation(subject: A, "a", object: Aprime) if
         [Expression("Isa", [Variable("resource"), Pattern("Aprime", {})])],
     )
 
-    # Unhandled partial raised here.
+    # Unhandled partial was raised here before fix.
     results = list(
         oso.query_rule(
             "allow",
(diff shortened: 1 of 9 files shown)


languages/python/oso/tests/regression/test_unhandled_partial.py
```diff
@@ -0,0 +1,80 @@
+import pytest
+
+
+def test_unhandled_partial():
+    """Spurious unhandled partial needs fix."""
```
dhatch, line 5:
```suggestion
    """Spurious unhandled partial."""
```

languages/python/oso/tests/regression/test_unhandled_partial.py
```diff
@@ -0,0 +1,80 @@
...
+    oso.register_class(UserAlchemical)
+
+    oso.load_str(POLICY)
+
+    constraint = Expression(
+        "And",
+        [Expression("Isa", [Variable("resource"), Pattern("AnalyticFilterable", {})])],
+    )
+
+    # Unhandled partial raised here.
```
dhatch, line 69:
```suggestion
    # Unhandled partial was raised here before fix.
```

polar-core/src/bindings.rs
```diff
@@ -521,16 +521,22 @@ impl BindingManager {
...
-                    Ok(Some(Goal::Query { term }))
+                    // Merge the two partials.
+                    let merged = lp.clone().merge_constraints(rp.clone());
+
+                    // Express the partial in terms of lv (bind rv to lv, replacing all rv in partial with lv).
+                    let goal = self.partial_bind(merged, rv, term!(lv.clone()))?;
+
+                    // Unification from lv = rv (remember that vars are equal so that the
+                    // simplifier can later choose the correct one). We do this
+                    // after the partial bind so that we don't recurisvely query the unification.
```
dhatch, line 536:
```suggestion
                    // after the partial bind so that we don't recursively query the unification.
```

languages/python/oso/tests/regression/test_unhandled_partial.py
```diff
@@ -0,0 +1,80 @@
+import pytest
+from typing import List
+
+def test_unhandled_partial():
```
samscott89, line 4: Super nit: update test name/comment/filename

polar-core/src/bindings.rs
```diff
@@ -521,16 +521,22 @@ impl BindingManager {
 
             // partial x partial -- if they already overlap, do nothing.
```
samscott89, line 522: Is this comment still accurate?

samscott89 (changes requested):
Code changes themselves look fine. I think we could also update the test cases for partial unification (e.g. lines https://github.com/osohq/oso/pull/1467/files#diff-4993b79f07ffaef682550d5f5c376dedee62a8990a38455ea45555d2f752efcdR1194-R1198)

I think the reason this bug fell through the cracks is because these tests no longer actually test partial unification due to grounding. Previously, we would have looked for a partial `_this = 1 and _this = 2`, but grounding means these trivially fail. If we have `x > 1 and y > 2 and x = y` we would have caught this earlier.

For example, here's an updated test case that fails correctly:

```rust

    #[test]
    fn test_partial_unification_2() -> TestResult {
        let p = Polar::new();
        p.load_str(
            r#"f(x, y) if x = y;
               f(x, y) if x = y and 1 < x;
               f(x, y) if 2 > y and x = y and x > 1;

               g(x, y) if x > 1 and y < 2;
               g(x, y) if x > 1 and y < 2 and x = y;"#,
        )?;

        // Register `x` as a partial.
        p.register_constant(sym!("x"), term!(value!(op!(And))))?;
        let mut q = p.new_query_from_term(term!(call!("f", [sym!("x"), sym!("y")])), false);
        let next = next_binding(&mut q)?;
        assert_partial_expressions!(next, "x" => "_this = y", "y" => "x = _this");
        let next = next_binding(&mut q)?;
        assert_partial_expressions!(next, "x" => "_this = y and 1 < _this", "y" => "x = _this and 1 < x");
        let next = next_binding(&mut q)?;
        // TODO: This seems wrong. Should be 2 > this and this > 1 for `y` binding
        assert_partial_expressions!(next, "x" => "_this = y and 2 > _this and _this > 1", "y" => "x = _this and 2 > x and x > 1");
        assert_query_done!(q);

        let mut q = p.new_query_from_term(term!(call!("g", [sym!("x"), sym!("y")])), false);
        let next = next_binding(&mut q)?;
        assert_partial_expressions!(next, "x" => "_this > 1", "y" => "_this < 2");
        let next = next_binding(&mut q)?;
        // TODO: This is failing on main since we don't get the x = y unification
        assert_partial_expressions!(next, "x" => "_this = y and _this > 1 and _this < 2", "y" => "x = _this and _this > 1 and _this < 2");
        assert_query_done!(q);

        // Register `y` as a partial.
        p.register_constant(sym!("y"), term!(value!(op!(And))))?;
        let mut q = p.new_query_from_term(term!(call!("f", [sym!("x"), sym!("y")])), false);
        let next = next_binding(&mut q)?;
        assert_eq!(next[&sym!("x")], var!("y"));
        let next = next_binding(&mut q)?;
        assert_eq!(next[&sym!("x")], term!(1));
        assert_eq!(next[&sym!("y")], term!(1));
        assert_query_done!(q);

        let mut q = p.new_query_from_term(term!(call!("g", [sym!("x"), sym!("y")])), false);
        let next = next_binding(&mut q)?;
        assert_eq!(next[&sym!("x")], term!(1));
        assert_eq!(next[&sym!("y")], term!(2));
        assert_query_done!(q);
        Ok(())
    }
```

(aside: I'm struggling to see the difference between `test_partial_unification_1` and `test_partial_unification_2`)

languages/python/oso/tests/regression/test_unhandled_partial.py
dhatch: Updated a bit... this is a regression test so it doesn't really test anything too meaningful... Usually I'd just use issue numbers for this but we don't have that right now in notion. This better?

polar-core/src/bindings.rs
dhatch: Yeah, L525 does this.

dhatch:
Updated test so it's passing, still need to review it for correctness -- there are a few parts that definitely could be simplified better, but I'm not sure that they are wrong.

languages/python/oso/tests/regression/test_unhandled_partial.py
```diff
@@ -0,0 +1,84 @@
...
+        oso.query_rule(
+            "allow",
+            UserAlchemical(group_ids=[0]),
+            "Read",
+            Variable("resource"),
+            accept_expression=True,
+            bindings={"resource": constraint},
+        )
+    )
+    print(results)
```
gw000, line 84: If it's possible to add a partial test to the core that triggers this issue then we should remove this. If not we should at least change the names and remove the extra details like the unused roles etc.

polar-core/src/partial/partial.rs
```diff
@@ -325,17 +325,25 @@ mod test {
...
                     .value()
                     .as_expression()
                     .unwrap()
                     .clone()
                     .into_iter()
-                    .collect::<HashSet<Term>>(),
-                    hashset! { $($args),+ })
+                    .collect::<HashSet<Term>>();
+            let r = hashset! { $($args),+ };
+            let fmt = |hs: &HashSet<Term>| {
```
samscott89, line 338: Nit: this could be:

```rust
let fmt = |hs: &HashSet<Term>| format!("{{ {} }}", hs.iter().map(|t| t.to_polar()).collect::<Vec<String>>().join(", "));

languages/python/oso/tests/regression/test_unhandled_partial.py
samscott89: +1

dhatch: I worked a bit on this and couldn't get a simpler repro in the core. I don't have more time to spend on it right now but I do agree that would be better.

Changed the names and simplified this instead.

samscott89 (approved):

PR #1468, opened 2022-01-03
Constrain Go instance type equality to exact matches

patrickod:
Previously we relied on Go's reflect.ConvertibleTo to determine whether
two instances were of equivalent types, however this results in runtime
type confusion between structs of identical schemas.

PR checklist:

- [x] Added changelog entry.


languages/go/tests/parity_test.go
```diff
@@ -145,12 +145,15 @@ func (MethodVariants) GetEmptyIter() interfaces.Iterator {
 	return IterableClass{}
 }
 
-// TODO: I don't think these make sense. Maybe as interfaces?
 type ParentClass struct{}
 
-type ChildClass struct{}
+// TODO: this kind of inheritance doesn't fit Go conventions
+// There is not way to use this kind of subclassing for composition, and
```
samscott89, line 151:
```suggestion
// There is no way to use this kind of subclassing for composition, and
```

samscott89 (commented):
Needs a breaking change changelog entry.

docs/content/any/project/changelogs/NEXT.md
```diff
@@ -21,9 +21,11 @@ draft: true
   before upgrading.
 {{% /callout %}}
 
-##### Breaking change 1
+##### Updated Go type checking behavior
```
samscott89, line 24: This should replace the top part too, e.g.
```md

## `oso` 0.26.0

### Go

#### Breaking changes

{{% callout "Warning" "orange" %}}
  This release contains breaking changes. Be sure to follow migration steps
  before upgrading.
{{% /callout %}}

```

docs/content/any/project/changelogs/NEXT.md
```diff
@@ -21,9 +21,11 @@ draft: true
...
 
-Summary of breaking change.
+When evaluating whether a given query variable matches a Go type Polar will now use direct instance comparisons instead of Go's `reflect.IsConvertableTo` functionality. This change resolves false-positive type checking results where discrete structs with identical sets of fields were considered to be equivalent.
```
samscott89, line 26:
```suggestion
When evaluating whether a given query variable matches a Go type Polar will now use direct instance comparisons instead of Go's `reflect.ConvertibleTo` functionality. This change resolves false-positive type checking results where discrete structs with identical sets of fields were considered to be equivalent. 
```

docs/content/any/project/changelogs/NEXT.md
```diff
@@ -21,9 +21,11 @@ draft: true
   before upgrading.
 {{% /callout %}}
 
-##### Breaking change 1
+##### Updated Go type checking behavior
 
-Summary of breaking change.
+When evaluating whether a given query variable matches a Go type Polar will now use direct instance comparisons instead of Go's `reflect.IsConvertableTo` functionality. This change resolves false-positive type checking results where discrete structs with identical sets of fields were considered to be equivalent. 
+
+This change has implications for the use of NewTypes in Polar rule definitions. Rules that are defined using NewTypes will now only match instances of the NewType and no longer match the underlying wrapped type.
```
samscott89, line 28: I think an example would be useful here.

docs/content/any/project/changelogs/NEXT.md
```diff
@@ -10,20 +10,41 @@ draft: true
...
+Where previously it was possible to utilize this `Role` type as interchangeable with that of `string`:
+```polar
+has_role(user: User, role String, resource: Resource) if ...
+```
+
+This rule definition must be rewritten as follows:
 
-Summary of breaking change.
+```polar
+has_role(user: User, role: Role, resource: Resource) if ...
```
samscott89, line 46: I'm not sure this rule will be allowed, due to the rule type we have for `has_role`.

docs/content/any/project/changelogs/NEXT.md
```diff
@@ -10,20 +10,41 @@ draft: true
...
+type Role string
+const (
+    Admin Role = "admin"
+    User Role = "user"
+)
+```
+
+Where previously it was possible to utilize this `Role` type as interchangeable with that of `string`:
+```polar
+has_role(user: User, role String, resource: Resource) if ...
```
samscott89, line 40:
```suggestion
has_role(user: User, role: String, resource: Resource) if ...
```

docs/content/any/project/changelogs/NEXT.md
```diff
@@ -17,21 +17,53 @@ draft: true
...
+```go
+type Action string
+const (
+    View Action = "view"
+    Create Action = "create"
+    Update Action = "update"
+)
+```
+
+Where previously it was possible to utilize this `Role` type as interchangeable with that of `string`:
```
samscott89, line 46:
```suggestion
Where previously it was possible to utilize this `Action` type as interchangeable with that of `string`:
```

docs/spelling/allowed_words.txt
```diff
@@ -1,57 +1,59 @@
 ABAC
```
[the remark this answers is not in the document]
samscott89: @patrickod do you have it aliased to `sort -f` or something?

Also, what version do you have? I have:

```
❯ sort --version
sort (GNU coreutils) 8.30
```

Which does case-sensitive sorting by default.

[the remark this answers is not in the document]
patrickod:
```
8:49:43 patrickod@kimkilwhan oso patrickod/go-instance-type-equality-exact-match sort --version                                                                            
sort (GNU coreutils) 8.32                                                        
Copyright (C) 2020 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <https://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by Mike Haertel and Paul Eggert.
```

TIL that the sort order is based on _locale data_. 

My `LC_COLLATE` being set as `en_US.UTF-8` caused the sorting behavior above. Setting `LC_COLLATE=C` while invoking sort returned the expected sorting behavior with uppercase taking precedence over lowercase.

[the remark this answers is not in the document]
patrickod: pushed with a newly sorted `allowed_words.list` with the `NewType(s)` additions.

samscott89 (approved):

test/spec/host_specializers.yml
```diff
@@ -5,17 +5,21 @@ cases:
...
+  #   query: is_parentclass
+  #   input: [{type: ChildClass}]
+  # - description: can specialize on either subclass or class
+  #   query: has_subclass
+  #   input: [{type: ChildClass}]
+  #   result: [{}, {}]
+  # - description: can order rules by specificity
+  #   query: level
+  #   input: [{type: GrandchildClass}, {var: x}]
+  #   result: [{x: 3}, {x: 2}, {x: 1}]
```
gj, line 22: Should these be deleted?

languages/go/tests/parity_test.go
```diff
@@ -145,12 +145,15 @@ func (MethodVariants) GetEmptyIter() interfaces.Iterator {
...
 
-type ChildClass struct{}
+// TODO: this kind of inheritance doesn't fit Go conventions
+// There is no way to use this kind of subclassing for composition, and
+// so it doesn't fit with our model of inheritance. Leaving for posterity
+// but these are not tested
+type ChildClass ParentClass
 
-type GrandchildClass struct{}
+type GrandchildClass ChildClass
```
gj, line 156: Why are we leaving this for posterity if there's no way to do this in Go?

languages/go/internal/host/reflect_test.go
```diff
@@ -0,0 +1,50 @@
...
+		t.Error("other should not be an instance of Unit")
+	}
+
+	stringClass := reflect.TypeOf("")
+	roleClass := reflect.TypeOf(Admin)
+	if !isInstance(Admin, roleClass) || !isInstance(Member, roleClass) {
+		t.Error("roles should be instances of the role class")
+	}
+
+	// TODO: not currently passing since we don't/cannot implement "inheritance" checks
```
gj, line 45: Is this ever going to be passing? The comments elsewhere in this PR make it seem like the answer is 'no'

gj (changes requested):

languages/go/tests/parity_test.go
patrickod: ✂️ 'd out as no longer used in parity test suite.

languages/go/internal/host/reflect_test.go
patrickod: updated comment to reflect the test & behavior change

test/spec/host_specializers.yml
patrickod: ✂️ 'd out

languages/go/internal/host/reflect_test.go
```diff
@@ -42,9 +42,11 @@ func TestIsInstance(t *testing.T) {
 		t.Error("roles should be instances of the role class")
```
gj, line 42: Based on this error message, should L40 be `roleClass := reflect.TypeOf(Role)`?

languages/go/internal/host/reflect.go
```diff
@@ -5,6 +5,16 @@ import (
 	"reflect"
 )
 
+// Checking for type equality is currently done with exact matching
+// We will not check if Go would permit a conversion between types,
+// or if the instance is an instance of a newtype, as doing so allows for type
```
gj, line 10: Is this conjunct correct? I thought we _do_ check "if the instance is an instance of a newtype" — we just don't check if the instance is an instance of the underlying wrapped type?

languages/go/polar.go
```diff
@@ -37,7 +37,7 @@ func newPolar() (*Polar, error) {
 
 	builtinClasses := map[string]reflect.Type{
 		"Boolean":    reflect.TypeOf(true),
-		"Integer":    reflect.TypeOf(int(1)),
+		"Integer":    reflect.TypeOf(int64(1)),
```
gj, line 40: is this change required due to the new instance checking behavior or is it unrelated?

gj (commented):
Couple remaining questions but mostly LGTM

languages/go/polar.go
samscott89: Required

languages/go/internal/host/reflect.go
samscott89: Your understanding is correct, the comment is wrong

languages/go/internal/host/reflect_test.go
patrickod: The error message is misleading but I believe the behavior here is correct when read with the accompanying inverse check on L48. I think `role constants should be instances of the Role class` might be a clearer statement of the test intent. 

We want to test that a constant instance of the newtype will match only the newtype `Role # reflect.TypeOf(Admin)` and not the wrapped type `String`. Prior to the change both of these `isInstance` calls would have returned true, but this had the side-effect of allowing for runtime confusion between unrelated types whose schemas were idnetical.

languages/go/tests/parity_test.go
```diff
@@ -145,13 +145,8 @@ func (MethodVariants) GetEmptyIter() interfaces.Iterator {
 	return IterableClass{}
 }
 
-// TODO: I don't think these make sense. Maybe as interfaces?
 type ParentClass struct{}
```
gj, line 148: Is `ParentClass` still buying us anything in the absence of `ChildClass` & `GrandchildClass`?

languages/go/internal/host/reflect_test.go
gj: So you're saying that `reflect.TypeOf(Admin)` returns the newtype `Role` whereas `reflect.TypeOf(Role)` would return `string`?

samscott89: `reflect.TypeOf(Role)` would make sense, but you can't actually reference types that way in Go, you can only get the type of a concrete value. Same reason in many places we have `myClass := reflect.TypeOf(MyClass{})` <- the `MyClass{}` is creating an instance of MyClass, and then we get the type of that new instance.

gj: > `reflect.TypeOf(Role)` would make sense, but you can't actually reference types that way in Go, you can only get the type of a concrete value. Same reason in many places we have `myClass := reflect.TypeOf(MyClass{})` <- the `MyClass{}` is creating an instance of MyClass, and then we get the type of that new instance.

Okay and at the risk of being obtuse, we can't do `reflect.TypeOf(Role{})` because `Role` is just a new type alias to `string` and `string` is a primitive instead of an interface? So instead we do `reflect.TypeOf(Admin)` because `Admin` points to a concrete (const) instance of `Role`?

gj (approved):
My only blocking concern has been alleviated

languages/go/internal/host/reflect_test.go
patrickod: yep `reflect.TypeOf(Role{})` will barf because as a`NewType` of `string` it has no interface to satisfy this, so instead we use a constant instance which can be reflected.

languages/go/tests/parity_test.go
patrickod: it's used as a stand-in `other` type to be compared against `UnitType` in a comparison parity unit-test.

PR #1470, opened 2022-01-05
Update Go types based on polar-crossgen updates

samscott89:
This PR includes two commits:

- https://github.com/osohq/oso/commit/fab364124d1ad801fe6e48d5acaa12eab1d998d4 updates the enum unit variant serialization logic, and cleans up a few things. This is based on the `sam/polar-crossgen` branch
- https://github.com/osohq/oso/commit/0551e62879c26a1c6dec4d08613864ef251c1765 adds the data filtering types to the codegen, and uses the types throughout. This is based on `sam/polar-crossgen-df`. This change is way bigger, so we may want to back it out.

Before reviewing, we should decide on which commit to take.


languages/go/types/polar_types.go
```diff
@@ -15,6 +15,593 @@ type Call struct {
...
+	case ComparisonGeq:
+		return json.Marshal("Geq")
+	default:
+		return nil, fmt.Errorf("unexpected variant %#v of %v", inner, variant)
+	}
+
+}
+
+// Condition struct
+type Condition struct {
```
samscott89, line 191: This is what tuples look like on the Go side. It's not nice. Would be great if these were structs instead

saolsen: Yeah, we wanted to make that change anyway. I had parsed them into go structs in my version I think.

languages/go/types/polar_types.go
```diff
@@ -1880,6 +2528,68 @@ func (variant QueryEvent) MarshalJSON() ([]byte, error) {
...
+		variant.V0,
+		variant.V1,
+		variant.V2,
+	}
+
+	return json.Marshal(fieldArray)
+}
+
+// ResultSet struct
+type ResultSet struct {
```
saolsen, line 2584: We don't need any of the `FilterPlan` stuff, only the `Filter` stuff.

languages/go/types/polar_types.go
```diff
@@ -15,6 +15,593 @@ type Call struct {
...
+	case ConstraintKindNin:
+		return json.Marshal("Nin")
+	default:
+		return nil, fmt.Errorf("unexpected variant %#v of %v", inner, variant)
+	}
+
+}
+
+// ConstraintValueTerm newtype
+type ConstraintValueTerm Term
```
saolsen, line 368: This is the biggest problem with just automatically converting the core type. The type we want to pass users is almost the same as the core Filter but we want the values to be go values, not a Term object. I think in my manual version I made this an interface{} and had a post parsing pass to call host.toGo() on it. That isn't too great either but we'll have to do something like that.

samscott89: Ahh, okay that makes sense. It might make sense to have _all_ of `polar_types.go` be internal-only, and pass them through `toGo` to get the Go-side type. Most of the `toGo` code will look similar to existing ones: recursively toGo fields/arrays etc. Does allow us to customise things a little if we want.

I'll revert the filter plan changes from this PR though, since it's pretty far from ready based on that.

languages/go/types/polar_types.go
```diff
@@ -405,28 +385,21 @@ func (result *MessageKind) UnmarshalJSON(b []byte) error {
...
 
 func (variant MessageKind) MarshalJSON() ([]byte, error) {
 	switch inner := variant.MessageKindVariant.(type) {
-
 	case MessageKindPrint:
-		return json.Marshal(map[string]MessageKindPrint{
-			"Print": inner,
-		})
-
+		return json.Marshal("Print")
```
samscott89, line 396: This is the main result of updating the codegen: we now serialize unit variants as the string, not with an empty inner thing.

samscott89:
The conflicts with main are not worth trying to resolve.

gj:
For posterity, the conflicts mostly involve adding `Deserialize` bounds onto error variants (`Deserialize` is needed by the codegen b/c it relies on serde but it isn't otherwise needed on the error types b/c we only ever serialize them):

```diff
diff --git a/polar-core/src/error.rs b/polar-core/src/error.rs
index d066875f..5100e0a6 100644
--- a/polar-core/src/error.rs
+++ b/polar-core/src/error.rs
@@ -4 +4 @@ use indoc::formatdoc;
-use serde::Serialize;
+use serde::{Deserialize, Serialize};
@@ -16 +16 @@ pub type PolarResult<T> = Result<T, PolarError>;
-#[derive(Debug, Clone, Serialize)]
+#[derive(Debug, Clone, Serialize, Deserialize)]
@@ -182 +182 @@ impl PolarError {
-#[derive(Clone, Serialize)]
+#[derive(Clone, Serialize, Deserialize)]
@@ -197,5 +197,5 @@ impl From<PolarError> for FormattedPolarError {
-#[derive(Clone, Serialize)]
+#[derive(Clone, Serialize, Deserialize)]
 #[serde(transparent)]
 pub struct ParseError {
     pub kind: ParseErrorKind,
-    #[serde(skip_serializing)]
+    #[serde(skip)]
@@ -226 +226 @@ impl From<ParseError> for PolarError {
-#[derive(AsRefStr, Clone, Debug, Serialize)]
+#[derive(AsRefStr, Clone, Debug, Serialize, Deserialize)]
@@ -318 +318 @@ impl fmt::Display for ParseErrorKind {
-#[derive(AsRefStr, Clone, Debug, Serialize)]
+#[derive(AsRefStr, Clone, Debug, Serialize, Deserialize)]
@@ -472 +472 @@ The expression is: {expr}
-#[derive(AsRefStr, Clone, Debug, Serialize)]
+#[derive(AsRefStr, Clone, Debug, Serialize, Deserialize)]
@@ -480,0 +481 @@ pub enum OperationalError {
+        #[serde(skip)]
@@ -512 +513 @@ impl fmt::Display for OperationalError {
-#[derive(AsRefStr, Clone, Debug, Serialize)]
+#[derive(AsRefStr, Clone, Debug, Serialize, Deserialize)]
diff --git a/polar-core/src/sources.rs b/polar-core/src/sources.rs
index a1684a5e..bf412695 100644
--- a/polar-core/src/sources.rs
+++ b/polar-core/src/sources.rs
@@ -85 +85 @@ impl SourceInfo {
-#[derive(PartialEq, Serialize, Deserialize)]
+#[derive(PartialEq, Serialize, Deserialize, Default)]
```

PR #1472, opened 2022-01-07
{Python,Ruby,JS}: Add `class_id` information to ExternalInstance values; use for in-VM IsA checks

patrickod:
* Introduce a branch within the `VM#isa` implementation to use this `class_id` information where possible to perform IsA checks without having to rely on external events.
* Update the Python, Ruby and Javascript host integration to add `class_id` information to `ExternalInstance` values sent to the core. 

- [ ] Added changelog entry.


languages/js/src/types.ts
```diff
@@ -712,6 +713,12 @@ export interface ClassParams extends DataFilteringQueryParams {
    * `name` property.
    */
   name?: string;
+  /**
+   * Explicit ID to use when registering the class class in Polar. An ID is
```
patrickod, line 717:
```suggestion
   * Explicit ID to use when registering the class in Polar. An ID is
```

languages/js/src/Host.ts
```diff
@@ -507,12 +507,25 @@ export class Host implements Required<DataFilteringQueryParams> {
...
+          classId = instanceId;
+        } else {
+          // pass classId for instances of *registered classes* only
+          if (classRepr !== undefined && this.types.has(classRepr)) {
+            classId = this.getType(classRepr)?.id;
+          }
+        }
+
+        // pass classRepr for *registered* classes only, pass undefined
+        // otherwise
```
gj, line 528: Given the following code:

```js
class A {}
class B extends A {}
oso.registerClass(A);
oso.registerClass(B);
```

should the `classId` & `classRepr` for `B` be those of `A` or `B`?

gj: To expand on that a bit, I'm trying to lay out the cases to make sure we're handling them all correctly:

- `v` is a registered class whose superclass is a registered class.
- `v` is a registered class whose superclass isn't a registered class.
- `v` is an instance of a registered class.
- `v` is anything else.

Does that correspond with your understanding of the exhaustive cases?

gj (changes requested):
Just through JS so far but had some questions before I continue

languages/js/src/Polar.ts
```diff
@@ -242,8 +241,12 @@ export class Polar {
    * @param params An optional object with extra parameters.
    */
   registerClass(cls: Class, params?: ClassParams): void {
+    params = params ? params : {};
+    params.id = this.#ffiPolar.newId();
```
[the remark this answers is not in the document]
patrickod: as mentioned below there was no strong reason to keep these changes so reverted them 😄 to the simpler form.

languages/js/src/Host.ts
patrickod: to answer questions in order:
* as `B` is a registered class, `ExternalInstances` of `B` should have discrete `classID` and `classRepr` values to those `ExternalInstance`s representing `A` 

 * in the following V scenarios
   * `v` is registered class whose superclass _is a registered class_
     *  display `classId` and `classRepr` of the registered `v` class, not its superclass
     * `classId` and `instanceId` are identical
   * `v` is a registered class whose superclass _isn't a registered class_
     * display `classId` and `classRepr` of the registered `v` class, not its superclass
     * `classId` and `instanceId` are identical
   * `v` is an instance of a registered class
     *  display `classId` of registered `v` class with unique `instanceId` value
   * `v` is an instance of a non-registered class
     * `classId` is `None` ; `classRepr` is `None`

languages/js/src/Polar.ts
```diff
@@ -242,8 +241,12 @@ export class Polar {
...
+
     const clsName = this.getHost().cacheClass(cls, params);
     this.registerConstant(cls, clsName);
+    this.getHost().registerMros();
```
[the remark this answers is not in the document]
dhatch: If loadSources isn't called and we run queries directly we need MROs. This is fine since it's idempotent and perf isn't an issue b/c this is just for Oso instance creation.

polar-core/src/kb.rs
```diff
@@ -26,9 +27,9 @@ impl RuleParamMatch {
 pub struct KnowledgeBase {
     /// A map of bindings: variable name → value. The VM uses a stack internally,
     /// but can translate to and from this type.
-    constants: Bindings,
+    constants: Constants,
     /// Map of class name -> MRO list where the MRO list is a list of class instance IDs
-    mro: HashMap<Symbol, Vec<u64>>,
+    pub mro: HashMap<Symbol, Vec<u64>>,
```
dhatch, line 32: for future reference we typically avoid making fields public and instead access through a method. this makes it easier to change the internal representation of data without affecting all callers

polar-core/src/kb.rs
```diff
@@ -574,7 +575,26 @@ impl KnowledgeBase {
...
         }
-        self.constants.insert(name, value);
+
+        if let Value::ExternalInstance(ExternalInstance {
+            class_id,
+            instance_id,
+            ..
+        }) = *value.value()
+        {
+            if class_id.map_or(false, |id| id == instance_id) {
```
dhatch, line 585: This logic seems a bit brittle. Probably better to have a different FFI for registering class ids. Not an issue for now.

dhatch (approved):

polar-core/src/kb.rs
patrickod: 💯 thanks, that absolutely makes sense.

polar-core/src/kb.rs
patrickod: agreed! we are stretching things here and I think it'd make life easier in the other language impls to not have to piggyback atop `ExternalInstance`

polar-core/src/kb.rs
dhatch: we also definitely haven't followed this convention well so far :)

PR #1478, opened 2022-01-10
Prefer Display to ToPolarString

gj:


polar-core/src/debugger.rs
```diff
@@ -325,7 +324,7 @@ impl Debugger {
                     let mut vars = vm
                         .bindings(true)
                         .keys()
-                        .map(|k| k.to_polar())
+                        .map(|k| k.0.clone())
```
gkaemmer, line 327: Tiny nit, but do we need this line? It seems like rust should let us call `join()` somehow without explicitly cloning each element of the list.

gkaemmer (approved):
This looks good to me. Is there a reason to keep the `ToPolarString` trait around at all? Seems like we could replace it's impls with `fmt::Display`.

polar-core/src/formatting.rs
```diff
@@ -568,7 +580,7 @@ pub mod to_polar {
     }
 
     impl Rule {
-        pub fn head_as_string(&self) -> String {
+        pub(crate) fn head_as_string(&self) -> String {
```
patrickod, line 583: is there a major significance to this impl being `pub(crate)` where others in the same module are `pub(super)` ?

gj: `pub(super)` makes `polar_core::formatting::to_polar::ToPolarString` visible in the `polar_core::formatting` scope so that it's visible here: https://github.com/osohq/oso/blob/cf82d03a1a33afe03aab5278145432b93c962bea/polar-core/src/formatting.rs#L166

The others are `pub(crate)` because they're used elsewhere in `polar_core` outside of `polar_core::formatting`.

gj:
> Is there a reason to keep the ToPolarString trait around at all? Seems like we could replace it's impls with fmt::Display.

Not sure we still need it, but I didn't feel like doing the brain work to figure that out as part of this PR 😄

polar-core/src/debugger.rs
gj: It was just me being lazy with the borrow checker since it wasn't happy with my original attempt of `.map(|k| &k.0)`. Updated to use `AsRef` in a153f52b3c549d8373bb26702f16f6f9b5056b83

PR #1481, opened 2022-01-11
new data filtering for node

### LINKED ISSUE #1496: New data filtering in NodeJS library
I've read about the v3 data filtering in a [blog post](https://www.osohq.com/post/authorization-logic-into-sql). I've seen that it's already available for python and I would like to see it implemented in JS as well.

saolsen:
- Adds new data filtering API configured with `setDataFilteringAdapter`
- Removes v1 API
- Includes an example adapter for TypeORM

Closes #1496


languages/js/src/filter.test.ts
```diff
@@ -328,25 +263,150 @@ async function fixtures() {
...
+        return `${proj.typeName}.${proj.fieldName}`
+      }
+    }
+
+    for (let cs of filter.conditions) {
+      let values = {}
+      let ands = ""
+      for (let c of cs) {
+        let op;
+        switch(c.cmp) {
```
gw000, line 342: could replace the switch with a `const ops: {[key: string]: string} = { 'Eq': '=', ... };`?

languages/js/src/filter.test.ts
```diff
@@ -328,25 +263,150 @@ async function fixtures() {
...
+        if (c.lhs.hasOwnProperty("typeName") && (c.lhs as Projection).fieldName == undefined) {
+          c.lhs = {
+            typeName: (c.lhs as Projection).typeName,
+            fieldName: 'id',
+          }
+          c.rhs = {
+            value: (c.rhs as Immediate).value.id
+          }
+        }
+        if (c.rhs.hasOwnProperty("typeName") && (c.rhs as Projection).fieldName == undefined) {
```
gw000, line 377: These two branches could share almost all of their code

languages/js/src/filter.test.ts
```diff
@@ -328,25 +263,150 @@ async function fixtures() {
...
+    'Foo': Foo,
+    'Num': Num,
+    'Log': Log,
+  };
+
+  async function buildQuery(filter: Filter): Promise<SelectQueryBuilder<Resource>> {
+    var repo: Repository<Resource> = repositories[filter.model];
+    let query: SelectQueryBuilder<Resource> = repo.createQueryBuilder(filter.model);
+
+    // Sort relations so they are in a sql valid order
```
gw000, line 299: Should be able to take this out now 🤞

languages/js/src/Host.ts
```diff
@@ -89,32 +61,27 @@ export class UserType<Type extends Class<T>, T = any, Query = any> {
  *
  * @internal
  */
-export class Host implements Required<DataFilteringQueryParams> {
+export class Host<Q, R> {
```
gw000, line 64: The new type variables are for the Query returned by `buildQuery` and the Resource returned by `executeQuery`

languages/js/src/Query.ts
```diff
@@ -145,13 +149,25 @@ export class Query {
     // NOTE(gj): disabling ESLint for following line b/c we're fine if
     // `receiver[rel.myField]` blows up -- we catch the error and relay it to
     // the core in `handleCall`.
-    const value = receiver[rel.myField] as unknown; // eslint-disable-line
+    const value = receiver[rel.myField]; // eslint-disable-line
```
gw000, line 152: rm comments

languages/js/src/filter.test.ts
```diff
@@ -1,6 +1,16 @@
...
+  Filter,
+  FilterCondition,
+  FilterConditionSide,
+  Immediate,
+  Projection,
+  Relation,
+  SerializedRelation,
+  Adapter,
+} from './filter';
+//import type { Filter, FilterKind, FilterField } from './dataFiltering';
```
gw000, line 13: remove

languages/js/src/Query.ts
gj: 18dcc3a0

languages/js/src/Query.ts
```diff
@@ -179,7 +190,7 @@ export class Query {
       )) as NullishOrHasConstructor;
       const rel = this.#host.getType(receiver?.constructor)?.fields?.get(attr);
       if (rel instanceof Relation) {
-        value = await this.handleRelation(receiver, rel);
+        value = await this.handleRelation(receiver as obj, rel);
```
gj, line 193: Won't `receiver` be a `Constructable` since `this.#host.getType(receiver?.constructor)?.fields?.get(attr)` evaluated to a `Relation` instance?

languages/js/src/types.ts
```diff
@@ -749,6 +727,8 @@ export interface UserTypeParams<Type extends Class>
   isaCheck: IsaCheck;
 }
 
+export type Constructable = { constructor: Class };
```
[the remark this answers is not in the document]
gw000: It's a remnant from an earlier way I tried to do the types, will remove

languages/js/src/Query.ts
gw000: We actually require it to be indexable with strings (ie. `obj`) but we test for it by seeing if it has a `constructor: Class`. Js being js means pretty much the latter => the former

languages/js/src/filter.test.ts
```diff
@@ -328,25 +264,85 @@ async function fixtures() {
...
+          )
+        );
+
+        return filter.relations
+          .reduce((query, { fromTypeName, fromFieldName, toTypeName }) => {
+            const { my_field, other_field } = (
+              filter.types[fromTypeName][fromFieldName] as SerializedRelation
+            ).Relation;
+            const join = `${fromTypeName}.${my_field} = ${toTypeName}.${other_field}`;
+            return query.innerJoin(toTypeName, toTypeName, join); // lol typeorm ??
```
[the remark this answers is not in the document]
saolsen: lol yeah that's weird

languages/js/src/Oso.ts
```diff
@@ -2,32 +2,26 @@ import { Polar } from './Polar';
...
 } from './errors';
-import { filterData } from './dataFiltering';
-import type { FilterPlan } from './dataFiltering';
+import { parseFilter, Adapter, FilterJson } from './filter';
 
 /** The Oso authorization API. */
 // TODO(gj): maybe pass DF options to constructor & try to parametrize a
 // `Query` type w/ the return type of the provided buildQuery fn.
 export class Oso<
+  Q = unknown,
```
gj, line 18: `Q` and `R` seem _okay_ in `Polar` & `Host` since they're internal interfaces and there's only two generic type params, but what do you think about calling this one `Query` to match the rest of the spelled-out generic parameters here?

gj: 6c71c2d4

languages/js/src/Oso.ts
```diff
@@ -319,28 +314,20 @@ export class Oso<
...
-  ): Promise<T[]> {
+    resourceCls: Class<Resource>
+  ): Promise<Resource[]> {
     const query = await this.authorizedQuery(actor, action, resourceCls);
     if (!query) return [];
-    const userType = this.getHost().getType(resourceCls);
-    if (userType === undefined)
-      throw new UnregisteredClassError(resourceCls.name);
-    return (await userType.execQuery(query)) as T[];
+    return await this.getHost().adapter.executeQuery(query);
```
gj, line 324: Probably don't need this `await` since `Adapter::executeQuery` returns `Promise<Resource[]>`
```suggestion
    return this.getHost().adapter.executeQuery(query);
```

languages/js/src/Query.ts
```diff
@@ -133,31 +137,38 @@ export class Query {
    *
    * @internal
    */
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  private async handleRelation(receiver: any, rel: Relation): Promise<unknown> {
+  private async handleRelation(receiver: obj, rel: Relation): Promise<unknown> {
```
gj, line 140: Can this return `Promise<Resource | Resource[]>` instead of `unknown`?

languages/js/src/filter.ts
```diff
@@ -0,0 +1,137 @@
...
+import { OsoError } from './errors';
+
+export interface SerializedRelation {
+  Relation: {
+    kind: string;
+    other_class_tag: string;
+    my_field: string;
+    other_field: string;
+  };
+}
```
gj, line 12: nit: This type & the corresponding `Relation::serialize` method don't feel like they're providing much value

languages/js/src/filter.ts
```diff
@@ -0,0 +1,140 @@
...
+  relations: FilterRelation[];
+  conditions: FilterCondition[][];
+  types: { [tag: string]: SerializedFields };
+}
+
+export interface FilterJson {
+  conditions: unknown[][][];
+  relations: string[][];
+  root: string;
+}
```
gj, line 92: Comments (a la https://github.com/osohq/oso/blob/f75566c1a885966646741a43029524b0a283e023/languages/js/src/filter.ts#L16) for at least some of these types would be helpful. E.g., `FilterConditionSide` is pretty obvious but `Projection` & `Immediate` aren't

languages/js/src/typeorm_adapter.ts
```diff
@@ -0,0 +1,94 @@
+import { Connection, SelectQueryBuilder } from 'typeorm';
```
gj, line 1: nit: file should be named `typeOrmAdapter.ts` to match the primary export

gj: Also the `typeorm` dependency listed in `devDependencies` in `languages/js/package.json` won't be installed when someone installs the published `oso` package. I think there'll be an error about `typeorm` missing when anyone imports from `oso` and doesn't already have TypeORM installed in their project, regardless of whether the project actually uses TypeORM. Definitely need to test this prior to shipping.

gj (changes requested):
See `typeorm` dependency concern

languages/js/src/typeorm_adapter.ts
gw000: Ok then I'm just going to keep the adapter in the tests for now and people can copy it if they want to

dhatch: Can we avoid this somehow? In python you can conditionally import and handle failure gracefully. Can we do that here?

gj: Avoided by removing the top-level export. Anyone wishing to can still `import { typeOrmAdapter } from 'oso/dist/src/typeOrmAdapter'`. It's ugly but good enough for the time being.

languages/js/src/filter.ts
gj: added in d3d40b5

languages/js/src/Oso.ts
```diff
@@ -2,32 +2,26 @@ import { Polar } from './Polar';
...
 
 /** The Oso authorization API. */
 // TODO(gj): maybe pass DF options to constructor & try to parametrize a
 // `Query` type w/ the return type of the provided buildQuery fn.
 export class Oso<
+  Query = unknown,
+  Resource = unknown,
   Actor = unknown,
   Action = unknown,
-  Resource = unknown,
```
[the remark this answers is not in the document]
gj: Also this would be a breaking change that we need to call out in the changelog (in addition to the addition of the new `Query` generic parameter)

languages/js/src/Oso.ts
```diff
@@ -319,28 +314,20 @@ export class Oso<
    * @param resourceCls Object type.
    * @returns An array of authorized resources.
    */
-  async authorizedResources<T extends Resource>(
```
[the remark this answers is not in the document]
gw000: I don't think so, so I removed it

languages/js/src/filter.ts
gw000: I replaced `Serialized*` with `obj`.

languages/js/src/Oso.ts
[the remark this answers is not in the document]
gw000: Reverted

gj (approved):
Needs a changelog but that can come later (e.g., as part of the new DF docs work) if you'd prefer. Great work!

PR #1488, opened 2022-01-14
{JS,Python,Ruby} Format ExternalInstance values with user-provided class names if present 

patrickod:
Update the `class_repr` information passed with `ExternalInstance` values to contain the user-provided class name if one exists. Previously these were generated based on object inspection and differed from their user-provided aliases.

PR checklist:

- [ ] Added changelog entry.


dhatch (approved):
lgtm

PR #1489, opened 2022-01-14
DF tests for Go

gw000:
Lots of tests based on the Ruby library. Includes an example adapter for [GORM](https://gorm.io). Bumps minimum supported Go version to 1.14 from 1.12, maybe we can avoid that somehow?


languages/go/internal/host/host.go
```diff
@@ -451,6 +455,14 @@ func (h *Host) SetAcceptExpression(acceptExpression bool) {
 	h.acceptExpression = acceptExpression
 }
 
+func (h *Host) GetRelationFields(rel FilterRelation) (string, string) {
+	switch rec := h.fields[rel.FromTypeName][rel.FromFieldName].(type) {
+	case types.Relation:
+		return rec.MyField, rec.OtherField
+	}
+	return "", "" // FIXME return error
```
gw000, line 463: fix

languages/go/tests/enforcement_test.go
```diff
@@ -220,7 +220,7 @@ func (a TestAdapter) BuildQuery(filter *types.Filter) (interface{}, error) {
 	return nil, nil
 }
 
-func (a TestAdapter) ExecuteQuery(query interface{}) (interface{}, error) {
+func (a TestAdapter) ExecuteQuery(query interface{}) ([]interface{}, error) {
```
gw000, line 223: I think we want this

saolsen: yeah I agree even if its a little weird to create, otherwise can't do the relation on bound thing reliably

languages/go/query.go
```diff
@@ -227,6 +230,33 @@ func (q Query) handleExternalIsa(event types.QueryEventExternalIsa) error {
 	return q.ffiQuery.QuestionResult(event.CallId, isa)
 }
 
+func (q Query) handleExternalIsaWithPath(event types.QueryEventExternalIsaWithPath) error {
```
saolsen, line 233: oh wow oops, well really shows the value of tests

languages/go/tests/data_filtering_test.go
```diff
@@ -0,0 +1,801 @@
+package oso_test
+
+import (
+	"os"
+	"reflect"
+	"strings"
+	"testing"
+
+	"gorm.io/driver/sqlite"
+	"gorm.io/gorm"
```
saolsen, line 10: why gorm?

saolsen (approved):

gj (changes requested):
I haven't reviewed the substantive changes in this PR but as discussed on Slack if we're bumping the version of Go we test against from 1.12 -> 1.14 we should make 1.14 the new minimum Go version we support everywhere (probably mostly docs changes)

languages/go/internal/host/host.go
gj: 855fd04b8c88471eedff38bea7f38212d41463f0

gj (approved):

PR #1490, opened 2022-01-15
Add OperationalError::UnexpectedValue

gj:


polar-core/src/error.rs
```diff
@@ -490,6 +496,10 @@ impl fmt::Display for OperationalError {
         match self {
             Self::InvalidState { msg } => write!(f, "Invalid state: {}", msg),
             Self::Serialization { msg } => write!(f, "Serialization error: {}", msg),
+            Self::UnexpectedValue { expected, received } => write!(
+                f,
+                "Unexpected value.\n  Expected: {expected}\n  Received: {received}"
```
samscott89, line 501: Look at you using the fancy new features

polar-core/src/error.rs
```diff
@@ -475,6 +475,12 @@ pub enum OperationalError {
     InvalidState { msg: String },
     /// Serialization errors in the `polar-c-api` crate.
     Serialization { msg: String },
+    // This should go away once we can constrain the value variant of a particular term in the type
+    // system, e.g., `Term<String>` instead of `Term::value().as_string()`.
+    UnexpectedValue {
+        expected: String,
```
samscott89, line 481: Could make this `&'static str` for no runtime allocations.

gj: 714256235

polar-core/src/error.rs
```diff
@@ -609,6 +619,10 @@ where
     Err(OperationalError::InvalidState { msg }.into())
 }
 
+pub(crate) fn unexpected_value<T>(expected: &'static str, received: Arc<Value>) -> PolarResult<T> {
```
samscott89, line 622: Nit: why doesn't this take Term instead of `Arc<Value>`?

polar-core/src/vm.rs
```diff
@@ -1688,14 +1687,15 @@ impl PolarVirtualMachine {
                     return wrong_arity();
                 }
                 let result = args.pop().unwrap();
-                if !matches!(result.value(), Value::Variable(_)) {
-                    return invalid_state(format!("Not a variable: {}", result));
+                match result.value() {
```
samscott89, line 1690: Nit: could make this:

```rust
let _ = result.as_symbol()?;
```

polar-core/src/vm.rs
```diff
@@ -1962,8 +1962,9 @@ impl PolarVirtualMachine {
         let right = &args[1];
         let result = &args[2];
 
-        if !matches!(result.value(), Value::Variable(_)) {
-            return invalid_state(format!("arithmetic_op_helper: not a variable: {}", result));
+        match result.value() {
```
samscott89, line 1965: Same possible replacement here? Make this `result.as_symbol()?`

polar-core/src/error.rs
gj: Over-optimization to avoid cloning the source info; undone in 012ff8e. Not sure if the term we attach to the error will ever actually have context, but this at least sets us up to handle it appropriately if it does (either now or at some point in the future)

polar-core/src/vm.rs
gj: 7589343e

polar-core/src/vm.rs
gj: 7589343e

nice nit!

polar-core/src/data_filtering.rs
```diff
@@ -201,19 +201,18 @@ impl VarInfo {
...
     fn undot(&mut self, term: &Term) -> Value {
-        let val = term.value();
-        match val.as_expression() {
+        match term.as_expression() {
             Ok(Operation {
                 operator: Operator::Dot,
                 args,
             }) if args.len() == 2 => Value::from(self.dot_var(&args[0], &args[1])),
-            _ => val.clone(),
+            _ => term.value().clone(),
```
gj, line 209: This was the lone instance where moving the `.as_*()` methods from `Value` -> `Term` resulted in more work (`.as_expression()` calls `Term::value` and then we call `Term::value` a second time on this line)

samscott89 (approved):
Nice

PR #1493, opened 2022-01-17
Go version 1.12 -> 1.14

gw000:
For a test dependency 😔

PR checklist:

- [ ] Added changelog entry.


gj (changes requested):
are we planning to update the minimum supported version to 1.14 everywhere as part of this PR?

PR #1494, opened 2022-01-17
[vscode] Support multiple project roots

gj:
PR checklist:

- [x] Added changelog entry.


vscode/oso/client/src/index.ts
```diff
@@ -113,106 +125,159 @@ async function openDocument(uri: Uri) {
...
-  const createChangeWatcher = workspace.createFileSystemWatcher(
-    polarFilesInWorkspaceFolderPattern(folder),
-    false, // ignoreCreateEvents
-    false, // ignoreChangeEvents
-    true // ignoreDeleteEvents
-  );
+async function validateRoots(root: Uri, candidates: string[]) {
+  const projectRoots = [];
+  for (const path of [...new Set(candidates)]) {
+    // Split configured path on POSIX-style separator.
```
samscott89, line 134: Is that so that the configuration always uses `/`? I wonder if windows people would prefer using `\`

vscode/oso/package.json
```diff
@@ -67,6 +67,16 @@
           ],
           "default": "default",
           "markdownDescription": "Share usage data to help us make Oso better. Defaults to the `#telemetry.telemetryLevel#` setting."
+        },
+        "oso.polarLanguageServer.projectRoots": {
+          "type": "array",
+          "items": {
+            "type": "string",
+            "pattern": "^([^/]+\/)*[^/]+\/?$",
```
samscott89, line 75: Nice

samscott89 (approved):
Tested this fairly extensively locally. Seems to work well!

vscode/oso/client/src/index.ts
gj: Discussed offline; we're gonna ship it now with this POSIX bent since the configuration property validation & docs are very clear about it. If we hear from Windows users that this stinks, it probably wouldn't be too much more work to accept Windows and POSIX paths.

PR #1497, opened 2022-01-18
Remove v1 DF API from Python and Ruby

gw000:
1481 removes it from Node.

The number of tests deleted in this PR may appear alarming. The remaining tests were ported from the deleted ones, so the coverage is still similar. We just only have one way of doing things now.

PR checklist:

- [ ] Added changelog entry.


saolsen (approved):
beautiful

PR #1499, opened 2022-01-19
Remove unused feedback.html partial

gj:
Left over from #1498


gj:
Letting it hang around per @dhatch

PR #1505, opened 2022-01-21
Look up relations on bound queries

gw000:
Kind of an important piece! Getting the tests to work required explicitly preloading a ton of relations anyway, which masked the fact that this wasn't working. The new code in `query.go` is exercised by the existing tests, now that they don't preload everything up front.


saolsen (approved):
Oh yikes! def an important piece. Thanks!

PR #1526, opened 2022-02-14
bump Python build/dev/test tooling to current versions

alexhafner:
When working on the Python libraries using MacOS Monterey, there are issues when trying to run the tests, likely due to outdated library versions. This PR updates the tooling versions to more recent ones. In my test, I ran into the following errors

`FileExistsError: [Errno 17] File exists when building .egg files`
--> fixed by updating cffi 1.14 -> 1.15

`ast3/Python/ast.c:4514:9: error: implicit declaration of function '_PyUnicode_DecodeUnicodeEscape' is invalid in C99 [-Werror,-Wimplicit-function-declaration]`
--> fixed by updating black 21.12b -> 22.1.0

`TypeError: required field "lineno" missing from alias`
--> fixed by updating pytest 5.3.5 -> 7.0.1
--> see also https://stackoverflow.com/questions/69528842/an-error-while-trying-to-execute-tests-on-python-3-10-with-pytest
[An error while trying to execute tests on python 3.10 with pytest - Stack Overflow](https://stackoverflow.com/questions/69528842/an-error-while-trying-to-execute-tests-on-python-3-10-with-pytest)

The mypy update to 0.931 requires the installation of type stubs, in this case types-dataclasses==0.6.4. This is because starting with mypy 0.900, most third party package stubs must be installed separately : https://mypy.readthedocs.io/en/stable/getting_started.html

I've also added a step to install the requirements for the oso build in the Makefile in languages/python/oso

After installing this PR and making sure the Rust toolchain is up-to date for the build, the tests on MacOS Monterey using Python 3.10 completed normally.


alexhafner:
The error pattern `FileExistsError: [Errno 17] File exists: '/home/runner/work/oso/oso/languages/python/oso/.eggs/cffi-1.15.0-py3.6-linux-x86_64.egg'` which is now fixed on higher Python versions when running locally now reproduces on lower versions here on the test action

alexhafner:
On a side note, the underlying issue https://github.com/pypa/cibuildwheel/issues/902 mentioned in https://github.com/osohq/oso/blob/59e572af208b3f86373ede28b81d7faed75526b1/.github/workflows/release.yml#L593-L601

appears to be resolved, so it may be worthwhile also re-adding Python 3.10 to that part of the build matrix

gw000:
Hi @alexhafner , sorry for the long turnaround on this, I'm going to take a look today!

samscott89 (approved):

PR #1528, opened 2022-02-16
Update partial.py

jackdreillyvia:
Two bugs:

1. `left_path` can be an empty tuple. And it will throw
```
assert left_path[0] == Variable("_this")
IndexError: tuple index out of range
```
2. In the case that `left` is an ORM object, then you also want to check `isinstance(left, constraint_type)`.

I'm probably missing subtleties, but I can confirm, that on my project, I needed the extra `or` condition, or valid resources returned by `oso.authorized_resources` were dropped.

PR checklist:

- [ ] Added changelog entry.


jackdreillyvia:

Issue #1529, opened 2022-02-16
Support linux aarch64


magicarm22:
I'm trying to run the docker image on a Mac M1 laptop. But when docker container `python:3.10-slim-bullseye` tried to install oso, I got an error:

```
pipenv sync
[pipenv.exceptions.InstallError]: ERROR: Could not find a version that satisfies the requirement oso==0.26.0 (from versions: none)
[pipenv.exceptions.InstallError]: ERROR: No matching distribution found for oso==0.26.0
ERROR: Couldn't install package: oso
 Package installation failed...
 ```

patrickod:
Hey @magicarm22 

we currently don't have official support Linux arm64 builds for Python yet, however one of our community members has created a [set of Python Oso Docker images which support aarch64](https://github.com/multi-py/python-oso) which you can use in these circumstances. In particular I think the documentation on copying over existing aarch64 builds from the [source image into your own](https://github.com/multi-py/python-oso#copy-just-the-packages) might be most relevant to your use-case.

RonquilloAeon:
I was able to resolve this issue by setting `platform: linux/amd64` in my `docker-compose.yml`.

Found it on this page: https://til.simonwillison.net/macos/running-docker-on-remote-m1

patrickod:
@nenaddzambasevic unfortunately I don't have an updated timeline to offer on Linux aarch64 support but I have noted it in our work tracker. Thanks for your patience on this.

fmatzy:
I too have been experimenting with oso's golang library and have noticed that it fails to build with `GOOS=linux GOARCH=arm64`. I would be glad if oso's Linux aarch64 support would be provided. Any update on the roadmap?

amiran-gorgazjan:
I am getting the same issue. Running on a MacBook Pro with an Apple M1 Max chip.

omusil24:
@gj @patrickod would you accept a PR adding Linux arm64 support for Go builds (in the https://github.com/osohq/oso/blob/main/.github/workflows/release.yml)?

omusil24:
@gj made a PR here: https://github.com/osohq/oso/pull/1678

andrewkoltsov:
is it possible to make a new release with this minor change only?

omusil24:
@andrewkoltsov what version of golang are you using? And do you have `CGO_ENABLED=1` set when building?

kevinvalk:
[...]

Issue #808, opened 2021-03-27
Python - Could not find a version that satisfies 


mantissaman:
While trying to run OSO Python sample app or trying to install oso (pip install oso) on Apple M1 architecture I am getting following error:
(venv) ➜  oso-python-quickstart git:(main) pip install -r requirements.txt
ERROR: Could not find a version that satisfies the requirement oso
ERROR: No matching distribution found for oso

saolsen:
Yeah sorry we don't have M1 builds yet. The easiest thing to do is use x64 python via rosetta for development. We hope to have M1 builds available soon.

delicb:
@mantissaman 
I did manage to make it work locally with Python 3.9 on M1 chip, with the following steps.

1. install Rust - `brew install rust` - needed for compile
2. create virtual env - `python -m venv <path-to-your-env>`
3. activate virtual env - `source <path-to-your-env>/bin/activate`
4. install wheel (not sure if required, but some errors have gone after I did this) - `pip install wheel`
4. clone oso repo - `git clone https://github.com/osohq/oso.git`
5. build - `cd oso && make python-build`

The last step builds OSO locally and installs it in the active python environment by creating `/site-packages/oso.egg-link`. 

After that, I was able to import and use OSO. Note that I needed OSO only, not flask or sqlalchemy libs, but you should be able to install those manually once you have OSO core installed. 

```
oso ❯ python
Python 3.9.2 (default, Mar  5 2021, 18:57:15)
[Clang 12.0.0 (clang-1200.0.32.29)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import platform
>>> platform.processor()
'arm'
>>> import oso
>>> o = oso.Oso()
>>> o.load_str('allow("foo", "bar", "baz");')
>>> o.is_allowed("foo", "bar", "baz")
True
>>>
```

I hope this helps.

mantissaman:
@delicb  - Many thanks. It worked like a charm.

rcrowe:
Unable to develop with Go; either the linker fails or if I try forcing the architecture I see...

```
> env GOOS=darwin GOARCH=amd64 go build .
package oso-test
	imports github.com/osohq/go-oso
	imports github.com/osohq/go-oso/internal/ffi: build constraints exclude all Go files in $GOPATH/pkg/mod/github.com/osohq/go-oso@v0.14.0/internal/ffi
```

saolsen:
[...]

PR #1535, opened 2022-02-26
Use a matrix for library build in release workflow

connec:
I'm working on a Go application, using an M1 mac and an Alpine Linux container-based local development environment. I'd like to use Oso, but to do so I need linux arm64 support in the Go package. I've successfully spiked this locally, and started investigating how to bring it upstream.

In the process, I decided to take a stab at simplifying the current build by using a matrix of all the builds, and their shared/static library artifacts. This is intended to make it easier to see what targets are currently being built, and to simplify adding others in future.

It additionally introduces more parallelism, since each target will build in parallel vs. each OS before (and atm each OS has two builds). On the down side, it's not currently possible to depend on specific matrix job instances, so some jobs have to wait for additional builds. In theory, the parallelism should cancel this out, for a net win.

It's possible the naming conventions could be standardised such that the `shared`/`static` fields could just be `boolean`, but this way no changes are needed in the other jobs. Something to consider in future.

I've been [testing it in my fork](https://github.com/connec/oso/actions/workflows/release.yml) but haven't achieved a fully green run yet. I'm not set up to for the m1 mac runner (are there any instructions for this?), and I'm seeing [missing Python 3.6 on Windows](https://github.com/connec/oso/runs/5346106091?check_suite_focus=true#step:4:5), plus a number of seemingly [spurious errors](https://github.com/connec/oso/runs/5346097483?check_suite_focus=true). Not sure if any of these are known issues?

image

I see there's some activity around this already (e.g. https://github.com/osohq/oso/issues/1529 and https://github.com/osohq/oso/issues/808#issuecomment-1051770311), so hopefully this doesn't interfere with any WIP.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -662,7 +662,7 @@ jobs:
     runs-on: windows-latest
     strategy:
       matrix:
-        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
+        python-version: ["3.7", "3.8", "3.9", "3.10"]
     steps:
       - uses: actions/checkout@v2
       - name: Set version env
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -36,9 +36,66 @@ jobs:
           event-type: oso_release
           client-payload: '{"commit": "${{ github.sha }}"}'
 
-  linux_libs:
-    name: Build release libraries on Linux
-    runs-on: ubuntu-latest
+  build_libs:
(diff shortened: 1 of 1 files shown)


connec:

connec:
The release now goes through if I host an m1 runner: https://github.com/connec/oso/actions/runs/1952256611

image

(`dispatch` still fails because it depends on secrets that aren't on my fork.)

connec:
Closing per #1540.

PR #1539, opened 2022-03-08
Fix unused singletons in lib_guide.rs

joshrotenberg:
The `lib_guide` example has a bunch of unused singletons that aren't prefixed. Fixing them here so it can run.

PR checklist:

- [ ] Added changelog entry.


joshrotenberg:
The `types` example now fails with this: `Error: Cannot load additional Polar code -- all Polar code must be loaded at the same time.`

gkaemmer:
Hey @joshrotenberg, thanks for this. I think these examples have gotten a bit out of date because they're not a part of CI. I just pushed up a change that fixes the `Error: Cannot load additional Polar code -- all Polar code must be loaded at the same time.` error. Now it seems to work!

samscott89 (approved):

PR #1556, opened 2022-04-09
refactor: prefer yield from over looping in lexer

kkirsche:
This commit simplifies the docs lexer by using `yield from`rather than a for loop. 

This makes the code slightly shorter and removes the mental overhead and extra variable used by the for loop. Eliminating the for loop also makes the yield from version about 15% faster on average.

PR checklist:

- [ ] Added changelog entry.

### DIFF
--- a/languages/python/docs/lexer.py
+++ b/languages/python/docs/lexer.py
@@ -63,9 +63,7 @@ class GenericShellLexer(Lexer):
                     insertions = []
                 yield match.start(), token.Generic.Output, line
         if curcode:
-            yield from do_insertions(
-                insertions, lexer.get_tokens_unprocessed(curcode)
-            )
+            yield from do_insertions(insertions, lexer.get_tokens_unprocessed(curcode))
 
 
 class JShellLexer(GenericShellLexer):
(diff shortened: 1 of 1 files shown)


kkirsche:
That's not why that happens. It has to do with what parts execute in C and what execute in Python. This is similar to how list comprehension is much faster than an equivalent loop as comprehensions execute in C.

Since they don't JIT it, they don't compile those types of hot loops unless you understand how Python executes various data structures and how they get mapped onto the C APIs. 

I'll run that in a few 👍

kkirsche:
Oh, format requires clang and dev headers and whatnot…this may take longer than I expected

kkirsche:
I cannot run fmt as it throws errors with the C code of the repository with the provided commands.

Please see above for the specific AST code with errors and warnings

gw000:
Yikes! No problem, I'll do it

gw000:
Oh never mind, I can't push to your repo obviously 😅 Could you try shortening it 1 line, ie `yield from do_insertions(...)` ? I bet that's what it's complaining about

kkirsche:
Updated to a single line and updated branch to merge main into the branch.

gw000 (approved):
Thank you!

PR #1557, opened 2022-04-09
refactor: simplify logical expression using De Morgan identities

kkirsche:
When reading logical statements it is important for them to be as simple as possible. This proposal applies De Morgan's laws to simplify code by pushing negations deeper into the statement.

PR checklist:

- [ ] Added changelog entry.

### DIFF
--- a/languages/python/oso/polar/polar.py
+++ b/languages/python/oso/polar/polar.py
@@ -60,7 +60,7 @@ class Polar:
             path = Path(filename)
             extension = path.suffix
             filename = str(path)
-            if not extension == ".polar":
+            if extension != ".polar":
                 raise PolarFileExtensionError(filename)
 
             try:


gw000:
This is fine, but honestly I'd rather remove the error entirely. What do we gain by requiring people to use a specific file extension? @gj @samscott89 any thoughts?

gj:
> This is fine, but honestly I'd rather remove the error entirely. What do we gain by requiring people to use a specific file extension? @gj @samscott89 any thoughts?

Possibly not much, but I'd rather avoid thinking about the answer to that question for now and just ship this as-is. FWIW, it's something that has come up in the past — I believe @plotnick asked the same question & I gave the same answer 😅

gw000:
Alas https://youtu.be/cOeKidp-iWo

gw000 (approved):
Thank you! 🙂

PR #1558, opened 2022-04-13
Make polar-language-server out dir customizable

vrama628:
Changes the `polar-language-server` Makefile to allow for specifying a custom directory to build into.

### DIFF
--- a/polar-language-server/Makefile
+++ b/polar-language-server/Makefile
@@ -1,10 +1,11 @@
 .PHONY: build test
 
 CARGO_FLAGS ?= --dev
+OUT_DIR ?= ../vscode/oso/out
 
 build:
-	wasm-pack --quiet build $(CARGO_FLAGS) --target nodejs --out-dir ../vscode/oso/out
-	rm -f ../vscode/oso/out/.gitignore ../vscode/oso/out/package.json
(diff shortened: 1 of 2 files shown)


vrama628:

vrama628:
recheck

polar-language-server/Makefile
```diff
@@ -3,8 +3,8 @@
 CARGO_FLAGS ?= --dev
 
 build:
-	wasm-pack --quiet build $(CARGO_FLAGS) --target nodejs --out-dir ../vscode/oso/out
-	rm -f ../vscode/oso/out/.gitignore ../vscode/oso/out/package.json
+	wasm-pack --quiet build $(CARGO_FLAGS) --target nodejs --out-dir $(OUT_DIR)
+	rm -f $(OUT_DIR)/.gitignore $(OUT_DIR)/package.json
```
gj, line 7: should we add a default like `CARGO_FLAGS` above?

gj: Or, alternatively, blow up loudly & predictably if `OUT_DIR` isn't set?

vrama628: I had a default at first (see commit before "fix" commit) but removed it because moving forward I think it won't make sense to have one -- I'll make it blow up loudly and predictably

Issue #1561, opened 2022-04-20
Policy fails in Django 3.2.13 (but not in 3.2.12)


MichaelLampe:
[A commit in Django 3.2.13](https://github.com/django/django/commit/93cae5cb2f9a4ef1514cf1a41f714fef08005200) introduced a new filter via `self.check_alias`.  This has caused one of our policies to start failing, with an error of `ValueError: Column aliases cannot contain whitespace characters, quotation marks, semicolons, or SQL comments.`.

The value that was being injected by OSO was `'Variable('_component_290')__exists'`

## Policy
```POLAR
# I think I transcribed this correctly
policy_function(string_value, instance: instance) if
  item_ones in instance.item_ones and
  item_two in item_ones.item_two and
  item_two.string_value == string_value;
```

Where `item_ones` is a reverse reference on  `instance` and `item_two` is a FK attached to that set's item types.

Will try to create a better reproduction, but couldn't find anything (Because the Django release was just made), so figured I'd surface it.

## Larger Stack Trace
```PYTHON
  python3.8/site-packages/django_oso/models.py", line 28, in authorize
    filter = authorize_model(
  python3.8/site-packages/django_oso/auth.py", line 100, in authorize_model
    next_filter = partial_to_query_filter(resource_partial, model)
  python3.8/site-packages/django_oso/partial.py", line 277, in partial_to_query_filter
    return fb.finish()
  python3.8/site-packages/django_oso/partial.py", line 246, in finish
    filtered = subq.model.objects.filter(subq.finish()).values("pk")
  python3.8/site-packages/django_oso/partial.py", line 250, in finish
    objects = objects.annotate(**{name: exists}).filter(**{name: True})
  python3.8/site-packages/django/db/models/query.py", line 1091, in annotate
    return self._annotate(args, kwargs, select=True)
  python3.8/site-packages/django/db/models/query.py", line 1130, in _annotate
    clone.query.add_annotation(
  python3.8/site-packages/django/db/models/sql/query.py", line 1062, in add_annotation
```

## Context
django-oso = "0.25.1"
django = "3.2.13"

samscott89:
Hey @MichaelLampe. Thanks for opening an issue!

In this case, it looks like this might be a bug on our side that already existed, but that Django 3.2.13 is helpfully raising.

[...]

PR #1567, opened 2022-04-27
remove leading and trailing quotes in Variable str

MurphyMarkW:
This attempts to fix the compatibility issue raised in #1561 by removing the leading and trailing single quotes around the wrapped string of the `Variable` class when getting the `str` of `Variable`.

From:
```python
str(Variable("something")) == "Variable('something')"
```

To:
```python
str(Variable("something")) == "Variable(something)"
```

Which conforms to the new checks in the Django ORM regarding no quotation marks on aliases, as long as the underlying string doesn't contain quotes. That said, I suspect this isn't the most correct way to address this issue. Our test suite passes w/ Django updated to 3.2.13 and this change in place pass our tests, but I'm uncertain if there are any subtleties that we might not be aware of in how these strings are used elsewhere. Would appreciate any insights.

### DIFF
--- a/languages/python/oso/polar/variable.py
+++ b/languages/python/oso/polar/variable.py
@@ -5,7 +5,7 @@ class Variable(str):
         return f"Variable({super().__repr__()})"
 
     def __str__(self):
-        return repr(self)
+        return f"Variable({super().__str__()})"
 
     def __eq__(self, other):
         return super().__eq__(other)


samscott89:
Hey @MurphyMarkW ! Thanks for trying that out. I tried an alternative in #1568 which should address the problem more directly. I'll close this in favour of that one.

PR #1576, opened 2022-06-01
Better handling for repeated SQLAlchemy model names.

samscott89:
Catch the errors, and allow users to override it by registering
classes.

### DIFF
--- a/languages/python/sqlalchemy-oso/sqlalchemy_oso/auth.py
+++ b/languages/python/sqlalchemy-oso/sqlalchemy_oso/auth.py
@@ -1,6 +1,6 @@
 from oso import Oso
 from polar import Variable
-from polar.exceptions import PolarRuntimeError
+from polar.exceptions import DuplicateClassAliasError, OsoError, PolarRuntimeError
 from polar.partial import TypeConstraint
 
 from sqlalchemy.orm.query import Query
@@ -14,7 +14,7 @@ from sqlalchemy_oso.compat import iterate_model_classes
 from functools import reduce
 
 
-def polar_model_name(model) -> str:
+def default_polar_model_name(model) -> str:
     """Return polar class name for SQLAlchemy model."""
     return model.__name__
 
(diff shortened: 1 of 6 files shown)


languages/python/sqlalchemy-oso/sqlalchemy_oso/auth.py
```diff
@@ -29,7 +29,20 @@ def register_models(oso: Oso, base_or_registry):
...
+            # skip models that were manually registered
+            continue
+        try:
+            oso.register_class(model, name=default_polar_model_name(model))
+        except DuplicateClassAliasError as e:
+            raise OsoError(
+                (
+                    "Attempted to register two classes with the same name when automatically registering SQLAlchemy models\n"
+                    "To fix this, try manually registering the new class. E.g.\n"
+                    '  oso.register_class(MyModel, "models::MyModel")\n'
```
samscott89, line 42:
```suggestion
                    '  oso.register_class(MyModel, name="models::MyModel")\n'
```

languages/python/sqlalchemy-oso/sqlalchemy_oso/compat.py
```diff
@@ -17,7 +17,15 @@ def iterate_model_classes(base_or_registry):
     try:  # 1.3 declarative base.
         # TODO (dhatch): Not sure this is legit b/c it uses an internal interface?
         models = base_or_registry._decl_class_registry.items()
-        yield from {model for name, model in models if name != "_sa_module_registry"}
+        for name, model in models:
+            if name != "_sa_module_registry":
+                if isinstance(
+                    model, sqlalchemy.ext.declarative.clsregistry._MultipleClassMarker
```
samscott89, line 23: One of the main changes. My understanding is that we're using SQLAlchemy's builtin module registry (it's what allows you to write `ForeignKey("User")` or whatever the syntax is.

If multiple models have the same classname, they're stored inside this multiple class marker, so that SQLAlchemy can tell you to use fully qualified names (`admins.User`)

languages/python/sqlalchemy-oso/sqlalchemy_oso/auth.py
```diff
@@ -29,7 +29,20 @@ def register_models(oso: Oso, base_or_registry):
     """Register all models in registry (SQLAlchemy 1.4) or declarative base
     class (1.3 and 1.4) ``base_or_registry`` with Oso as classes."""
     for model in iterate_model_classes(base_or_registry):
-        oso.register_class(model)
+        if model in oso.host.types:
+            # skip models that were manually registered
```
samscott89, line 33: This makes it possible for users to specify their own class names by registering them manually.

gkaemmer (approved):

Issue #1579, opened 2022-06-09
Python `sqlalchemy_oso.sqlalchemy_utils:97` can trigger recursion error


kkirsche:
There seems to be a loop possibility in 

https://github.com/osohq/oso/blob/main/languages/python/sqlalchemy-oso/sqlalchemy_oso/sqlalchemy_utils.py#L97

I would need to see if I can extract sufficient code to create a reproduction of this, but wanted to at least share it in case the Oso team may know why this would occur.

Traceback:

```python
Traceback (most recent call last):
  File "/Users/kkirsche/Library/Caches/pypoetry/virtualenvs/REDACTED--kEzOSes-py3.10/lib/python3.10/site-packages/flask/app.py", line 1523, in full_dispatch_request
    rv = self.dispatch_request()
  File "/Users/kkirsche/Library/Caches/pypoetry/virtualenvs/REDACTED--kEzOSes-py3.10/lib/python3.10/site-packages/flask/app.py", line 1509, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**req.view_args)
  File "/Users/kkirsche/git/docker/REDACTED/api/src/REDACTED/middleware/iam.py", line 110, in wrapper
    return f(*args, **kwds)
  File "/Users/kkirsche/git/docker/REDACTED/api/src/REDACTED/views/v1/resource_type/request.py", line 163, in get_all_open_requests_for_resource_type
    return controller.get_all_open_requests_paginated()
  File "/Users/kkirsche/git/docker/REDACTED/api/src/REDACTED/middleware/server_timing_api.py", line 173, in wrapper
    return f(*args, **kwds)
  File "/Users/kkirsche/git/docker/REDACTED/api/src/REDACTED/controllers/v1/resource_type/request.py", line 323, in get_all_open_requests_paginated
    paginated_requests = get_paginated_open_requests(
  File "/Users/kkirsche/git/docker/REDACTED/api/src/REDACTED/authorization/authorization_initialization.py", line 65, in wrapper
    return func(*args, **kwargs)
  File "/Users/kkirsche/git/docker/REDACTED/api/src/REDACTED/crud/request.py", line 83, in get_paginated_open_requests
    return PaginationParameters.paginate_query_from_request(
  File "/Users/kkirsche/git/docker/REDACTED/api/src/REDACTED/schemas/v1/pagination.py", line 259, in paginate_query_from_request
    return instance.paginate_query(sort_by_model, query)
[...]

PR #1581, opened 2022-06-09
Fix recursion error in default_load_entities

kkirsche:
Fixes: #1579 

Add a relationship set to the recursion to ensure that if a relationship has already been seen, we don't repeatedly evaluate it.

PR checklist:
- [ ] Added changelog entry.


samscott89 (approved):
Nice catch! And thanks for the fix

PR #1585, opened 2022-06-10
Merge consecutive list appends into single extend

kkirsche:
This change updates the way `libs` is extended so that it uses a single `extend` rather than multiple `append`s and uses f-strings for increased legibility.

PR checklist:

- [ ] Added changelog entry.


samscott89 (approved):

kkirsche:
Test failure is unrelated to this code:

```
HUGO_ENV=production hugo -b https://docs-preview.oso.dev/
Start building sites … 
fatal error: concurrent map read and map write
```

Sounds like hugo is missing a mutex

PR #1587, opened 2022-06-10
Don't assign to builtin variables

kkirsche:
This line overwrites https://docs.python.org/3/library/functions.html#filter which can cause confusion if this is attempted to be used later. Where possible, it's best to avoid the use of builtins as the keys to ensure that there aren't conflicts down the line.

PR checklist:

- [ ] Added changelog entry.


samscott89 (approved):
Thanks!

PR #1601, opened 2022-07-13
Fix Clippy for 0.1.62

gkaemmer:
This is a bit annoying, and I'm not a fan of adding a bunch of `unwrap`s all over the place.

There seem to be some people who agree that clippy's suggestion to use `write!` introduces a new error case that must be handled, and that's a bit unfair for a lint step: https://github.com/rust-lang/rust-clippy/issues/9077

That being said, it feels better than ignoring this particular lint everywhere.


gj (approved):
Got a few more occurrences to clean up, but seems fine to me

PR #1609, opened 2022-08-03
Fix CI for Python builds.

samscott89:
- Update cibuildwheel to support m1 builds
- Add pyproject which for some reason fixes pip builds
- Use fat libs

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,7 +16,7 @@ jobs:
       - name: Set version env
         run: echo "oso_version=$(cat VERSION)" >> $GITHUB_ENV
       - name: Check github ref matches
-        if: github.ref != 'refs/heads/main'
+        if: startsWith(github.ref, 'refs/tags/')
         env:
           github_ref: ${{ github.ref }}
         run: grep "${github_ref/refs\/tags\/v/}" VERSION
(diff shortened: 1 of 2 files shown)


samscott89:
Here's the passing Create Release run: https://github.com/osohq/oso/actions/runs/2797014820

.github/workflows/release.yml
```diff
@@ -331,65 +315,25 @@ jobs:
...
-      - name: Build wheels
-        run: |
-          python -m cibuildwheel --output-dir wheelhouse
-        working-directory: languages/python/oso
-      - uses: actions/upload-artifact@v2
-        with:
-          name: wheel
-          path: languages/python/oso/wheelhouse/*.whl
-
-  build_macos_arm_wheels:
```
samscott89, line 387: No longer needed. cibuildwheel itself can cross-compile wheels, and by using the fat static lib it'll work for both.

.github/workflows/release.yml
```diff
@@ -331,65 +315,25 @@ jobs:
...
           name: wheel
-          path: languages/python/oso/wheelhouse/*.whl
+          path: ./wheelhouse/*.whl
 
   build_macos_wheels:
     name: Build wheels on MacOS
-    # Build on 10.15 so we can still build a 3.6 wheel.
-    # May have to drop once github actions switches to macos-11 by default
-    # because 3.6 isn't supported on 11.
-    runs-on: macos-10.15
```
samscott89, line 347: 10.15 no longer support by github actions

.github/workflows/release.yml
```diff
@@ -596,37 +540,6 @@ jobs:
           python test.py
         working-directory: test
 
-  validate_python_macos_old:
```
samscott89, line 599: 10.15 no longer supported on github actions

.github/workflows/release.yml
```diff
@@ -891,7 +804,7 @@ jobs:
   release:
     name: Create release
     runs-on: ubuntu-latest
-    if: github.ref != 'refs/heads/main'
+    if: startsWith(github.ref, 'refs/tags/')
```
samscott89, line 807: This makes it a little easier to run the release build for a one-off branches, because this will _only_ run for tags.

languages/python/oso/pyproject.toml
```diff
@@ -0,0 +1,3 @@
+[build-system]
```
samscott89, line 1: For some reason this is needed for pip versions >= 21.

It's something to do with them changing builds to happen in the same folder instead of a temporary dir. I don't know why this works, it just did.

saolsen (approved):

PR #1611, opened 2022-08-08
refactor: Add Python type hints to authorization API

kkirsche:
This merge request expands existing type hints. The primary focus of this merge request is on the public authorization API and the calls that this API reaches down to

The use of `object` instead of `Any` may be a bit odd at first, but this is meant to better conform to the recommendations of the Python team managing type information. `object` is for when the code can accept any object for use, but we want errors if it's misused. `Any` disables error checking in a way that's a bit hard to detect a first.

See this for a breakdown from some of their team:
https://github.com/python/typeshed/pull/8469#issuecomment-1203116838 

PR checklist:

- [ ] Added changelog entry.

### DIFF
--- a/languages/python/oso/oso/oso.py
+++ b/languages/python/oso/oso/oso.py
@@ -3,9 +3,10 @@
 __version__ = "0.26.1"
 
 import os
-from typing import List, Any, Set
+from typing import Any, List, Set, Type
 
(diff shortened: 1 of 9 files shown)


kkirsche:
Closing to approach this in a smaller, more focused way to simplify the review

PR #1614, opened 2022-08-10
Formatting/import sort

kkirsche:
This enables isort, https://pycqa.github.io/isort/, and uses plugins to ensure that it's leveraging a black compatible format. The example plugins used are distributed by the isort team.

isort is a Python utility / library to sort imports alphabetically, and automatically separated into sections and by type

By organizing into sections (future, standard library, third party, and internal) it provides a more scannable selection of imports making it easier to locate whether something is imported or not. 

PR checklist:

- [ ] Added changelog entry.


kkirsche:
The check error seems to be unrelated to this work as it's related to Java

```
Run axel-op/googlejavaformat-action@v3
Version of JDK: 11
Downloading Google Java Format
Error: TypeError: Cannot read property 'find' of undefined
```

PR #1615, opened 2022-08-16
Fix JS isaCheck.

samscott89:
PR checklist:

- [ ] Added changelog entry.

### DIFF
--- a/languages/js/src/Oso.ts
+++ b/languages/js/src/Oso.ts
@@ -21,7 +21,7 @@ export class Oso<
   Field = unknown,
   Request = unknown,
   Query = unknown
-> extends Polar<Query, Resource> {
+  > extends Polar<Query, Resource> {
   #notFoundError: CustomError = NotFoundError;
   #forbiddenError: CustomError = ForbiddenError;
   #readAction: unknown = 'read';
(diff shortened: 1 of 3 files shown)


languages/js/src/Oso.ts
```diff
@@ -258,13 +258,18 @@ export class Oso<
   async authorizedQuery(
     actor: Actor,
     action: Action,
-    resourceCls: Class<Resource>
+    resourceCls: Class<Resource> | string
```
samscott89, line 261: This is a nice quality of life improvement. In many cases it's easier to just provide a string for the class name (like with Prisma)

languages/js/src/Polar.ts
```diff
@@ -79,7 +79,7 @@ export class Polar<Query, Resource> {
     this.registerClass(Number, { name: 'Float' });
     this.registerClass(String);
     this.registerClass(Array, { name: 'List' });
-    this.registerClass(Object, { name: 'Dictionary' });
+    this.registerClass(Map, { name: 'Dictionary' });
```
samscott89, line 82: This is the "big" change. By making `Map` the default dictionary choice, we're ensure that all other classes get treated as external instances with an unknown type.

gkaemmer (approved):

docs/content/any/project/changelogs/NEXT.md
```diff
@@ -8,9 +8,9 @@ description: >-
 draft: true
 ---
 
-## `RELEASED_PACKAGE_1` NEW_VERSION
+## `0.26.3` NEW_VERSION
 
-### LANGUAGE (e.g., 'Core' or 'Python' or 'Node.js')
+### Node.js (e.g., 'Core' or 'Python' or 'Node.js')
 
 #### Breaking changes
```
samscott89, line 15: I think this is technically a breaking change, but it's fixing something that was broken.

samscott89: So not sure if it needs a version bump.

gkaemmer (approved):

PR #1622, opened 2022-09-06
Only pull in tracing-subscriber for tests, examples, and CLI binary

ZackPierce:
Makes the `tracing-subscriber` dependency optional and mentions it in the `cli` feature-set.

This means the `oso` library no longer has a hard `tracing-subscriber` dependency, in accordance with best practices for leaving tracing-output selection flexible until the point of executable construction.


ZackPierce:

PR #1623, opened 2022-09-14
do not raise for multiple parents error when results are empty

lp:
### Why

_the relationship handler should not raise a multiple-parent error when results are empty_

### DIFF
--- a/languages/ruby/lib/oso/polar/data/adapter/active_record_adapter.rb
+++ b/languages/ruby/lib/oso/polar/data/adapter/active_record_adapter.rb
@@ -10,7 +10,7 @@ module Oso
             types = filter.types
             query = filter.relations.reduce(filter.model.all) do |q, rel|
               rec = types[rel.left].fields[rel.name]
-              join_type = if rec.kind == "one" then "LEFT" else "INNER" end
+              join_type = rec.kind == 'one' ? 'LEFT' : 'INNER'
               q.joins(
                 "#{join_type} JOIN #{rel.right.table_name} ON " \
                 "#{rel.left.table_name}.#{rec.my_field} = " \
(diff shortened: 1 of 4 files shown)


lp:
env:
    GITHUB_TOKEN: ***
    PERSONAL_ACCESS_TOKEN: ***

lp:

samscott89 (commented):
Thanks for the PR!

This change looks pretty reasonable, though I'm a little wary of merging this without first understanding what behaviour this is fixing. Do you think it might be possible to add a test case for this (to https://github.com/osohq/oso/blob/main/languages/ruby/spec/oso/polar/data_filtering/gitclub_spec.rb), or elsewhere?

Is this as simple as: if you have a relation that can return one _or none_, we're not currently handling the none case properly?

samscott89:
@lp I added a test in https://github.com/osohq/oso/pull/1623/commits/3b4bd0f5b360516a25573ffe9faf4a1ccbf70359 which I think exhibits the behaviour your contribution is fixing. This fails on main, but passes with your PR.

However, there was one gotcha -- because the adapter using inner joins, we ended up filter _out_ correct results. Could you take a look at https://github.com/osohq/oso/pull/1623/commits/bf417e3feced7510587a8f47047ed40a7cb93ff2 and see if that makes sense as a fix?

lp:
> @lp I added a test in [3b4bd0f](https://github.com/osohq/oso/commit/3b4bd0f5b360516a25573ffe9faf4a1ccbf70359) which I think exhibits the behaviour your contribution is fixing. This fails on main, but passes with your PR.
> 
> However, there was one gotcha -- because the adapter using inner joins, we ended up filter _out_ correct results. Could you take a look at [bf417e3](https://github.com/osohq/oso/commit/bf417e3feced7510587a8f47047ed40a7cb93ff2) and see if that makes sense as a fix?

thanks for the test @samscott89
and the joins tweak fully makes sense

samscott89 (approved):
Thanks for finding this and figuring out the problem!

PR #1632, opened 2022-10-07
Add requirement to avoid flask-sqlalchemy 3.0 incompatibility

snstanton:


snstanton:

snstanton:
> @snstanton it looks like [our test suite still installed Flask-SQLAlchemy 3.0.0](https://github.com/osohq/oso/actions/runs/3207987367/jobs/5243477573#step:9:321) since we don't constrain it in [the `requirements-test.txt`](https://github.com/osohq/oso/blob/231f1a8353ab0f3c8abc0d3e9b700b5bd48476e1/languages/python/sqlalchemy-oso/requirements-test.txt#L3) file that we install in [the tox.ini](https://github.com/osohq/oso/blob/231f1a8353ab0f3c8abc0d3e9b700b5bd48476e1/languages/python/sqlalchemy-oso/tox.ini#L9) test setup.
> 
> We can add a constraint to that `requirements-test.txt` file, but is there a way to tell Python / pip to throw some sort of warning/error if `sqlalchemy-oso` is installed alongside `Flask-SQLAlchemy` >= 3.0.0?

We could add a version check to ./sqlalchemy-oso/sqlalchemy_oso/flask.py to check the __version__ of the library and raise a warning if it's 3.0 or later.

Issue #1602, opened 2022-07-18
Broken implementation of the oso::PolarClass derive macro


Siceberg:
I am using both diesel 2.0 and oso in a project, if I define a struct:
```
#[derive(PolarClass)]
#[diesel(table_name = accounts)]
pub struct Account {
    #[polar(attribute)]
    pub id: Uuid,
    pub email: String,
}
```
get error:
```
proc-macro derive panicked
message: called `Result::unwrap()` on an `Err` value: Error("expected literal")
```

this looks like a broken implementation of the oso::PolarClass derive macro. 

Here is [demo](https://github.com/Siceberg/oso-diesel2-conflict-demo)

PR #1642, opened 2022-11-04
avoid conflicts with other macros

Siceberg:
The code changed in this PR is to avoid conflicts with other macros like in [oso-diesel2-conflict-demo](https://github.com/Siceberg/oso-diesel2-conflict-demo).

Since I'm a beginner in Rust, not very familiar with macros, and I'm in a place where using internet resources is not very convenient, I'm not sure if this is a better solution. If you guys have other better ideas please let me know, thanks.

If this is not a correct solution, I hope to promote [Oso](https://github.com/osohq/oso) to solve related problems as soon as possible through this PR


Siceberg:

Siceberg:
#1602

PR #1596, opened 2022-06-27
Update cbindgen, clap, and strum_macros

onalante-msft:
The original goal of this patch was to update `polar-core`'s `heck` transitive dependency to `0.4` so that dependents on `oso` would not have to build `heck 0.3`. Updating `strum_macros` to `0.24` is sufficient to accomplish this. However, the PR as submitted is a larger changeset also including a version bump of `cbindgen` (which removes `heck 0.3` within the `oso` repository) and `clap` (essentially unrelated).

Of the changes, the `clap` update is the most dubious since `criterion` still requires `clap 2` ([for now](https://github.com/bheisler/criterion.rs/pull/543), at least), which may cause developer and CI build times to increase. If only the minimum set of changes (i.e. updating `strum_macros`) is preferred, I can amend the PR to match.

PR checklist:

- [ ] Added changelog entry.
  [PENDING COMMENT ON ACCEPTABLE CHANGES]

PR #1653, opened 2022-12-09
Fix macro namespacing and serialization bugs

onalante-msft:
Though `polar_core` macros are exposed publicly, they are not usable without importing `polar_core::*` since they expect other `polar_core` macros to be in scope.  Using `$crate` references as appropriate fixes this.  Also, fix a `Value::String` string injection bug and `Operator::Dot` bug causing incorrect serialization when the second argument is a `Value::String` that requires quotes.

PR checklist:
- [x] Added changelog entry.


onalante-msft:

onalante-msft:
My apologies about the delayed CLA signing (*Cf.* https://github.com/osohq/oso/pull/1596). I only just got approval to do so.

onalante-msft:
Closing due to lack of bandwidth.

Issue #1655, opened 2022-12-15
Oso does not support Kotlin data classes


unredundant:
Hey 👋 Finally getting around to trialing Oso as an auth solution for a Kotlin application that I'm building. 

However, it seems that Oso does not support Kotlin data classes :( Or, as is always possible... I'm just doing something dumb

I am trying to emulate the Java quickstart example, with a User trying to read from a repository.

I have the following models

```kt
data class Repo(
  val id: UUID,
  val name: String,
  val isPublic: Boolean
)

data class User (
  val id: UUID,
  val email: String,
  val repoRoles: List<RepoRole>
)
```

I have set up OSO with the following 

```kt
private val oso: Oso = Oso()

init {
  // On a tangent... it doesn't seem to even load 
  // unless I explicitly repeat the class name as the second param
  oso.registerClass(Repo::class.java, "Repo")
  oso.registerClass(User::class.java, "User")
  oso.loadStr(
    """
allow(actor, action, resource) if
has_permission(actor, action, resource);

actor User {}

resource Repo {
permissions = ["read", "push", "delete"];
roles = ["contributor", "maintainer", "admin"];

"read" if "contributor";
"push" if "maintainer";
"delete" if "admin";

"maintainer" if "admin";
"contributor" if "maintainer";
}

# This rule tells Oso how to fetch roles for a Repo
has_role(actor: User, role_name: String, Repo: Repo) if
role in actor.repoRoles and
role_name = role.name and
Repo = role.Repo;

has_permission(_actor: User, "read", Repo: Repo) if
Repo.isPublic;

allow(actor, action, resource) if
has_permission(actor, action, resource);
""".trimIndent()
  )
}
```

Just as a test, I have created a repo with `isPublic=true` with name `test`.  However, when I run the following

```kt
fun readByName(name: String): RepoModels.Response {
    val result = Repo(
      id = UUID.randomUUID(),
      name = name,
      isPublic = true
    )
    val user = User(
      id = UUID.randomUUID(),
      email = "admin@bkbn.io",
      repoRoles = listOf(RepoRole(role = "admin", repo = result))
    )
    oso.authorize(user, "read", result)
    return RepoModels.Response.fromRepo(result)
  }
``` 

I get an authorization error from oso

```
com.osohq.oso.Exceptions$NotFoundException: Oso NotFoundException -- The current user does not have permission to read the given resource. You should handle this error by returning a 404 error to the client.
[...]

PR #1662, opened 2023-01-14
PoC kotlin data classes

wcurrie:
Extending https://github.com/osohq/oso/pull/1661 to also handle querying fields of kotlin data classes.

Uses `PropertyUtils.getProperty` to call the `getFoo()` method generated by kotlin when trying to access `bar.foo` from polar. See `OsoKotlinTest.kt` for an example.

### DIFF
--- a/languages/java/Makefile
+++ b/languages/java/Makefile
@@ -16,7 +16,7 @@ test: rust
 	cd oso && mvn $(MVN_FLAGS) clean test
 
 package: rust
-	cd oso && mvn $(MVN_FLAGS) clean package
+	cd oso && mvn $(MVN_FLAGS) clean install
 
 repl: rust
 	cd oso && mvn $(MVN_FLAGS) exec:java -Dexec.mainClass="com.osohq.oso.Oso" -Dexec.args="$(ARGS)"
--- a/languages/java/oso/src/test/java/com/osohq/oso/OsoJavaRecordTest.java
+++ b/languages/java/oso/src/test/java/com/osohq/oso/OsoJavaRecordTest.java
@@ -1,0 +1,155 @@
+package com.osohq.oso;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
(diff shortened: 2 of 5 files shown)


languages/java/Makefile
```diff
@@ -16,7 +16,7 @@ test: rust
 	cd oso && mvn $(MVN_FLAGS) clean test
 
 package: rust
-	cd oso && mvn $(MVN_FLAGS) clean package
+	cd oso && mvn $(MVN_FLAGS) clean install
```
wcurrie, line 19: Just so I could use the locally built `0.26.5-SNAPSHOT` in other projects like a fork of https://github.com/osohq/oso/issues/1655

languages/java/oso/src/test/java/com/osohq/oso/OsoJavaRecordTest.java
```diff
@@ -0,0 +1,155 @@
...
+
+import java.net.URL;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+public class OsoJavaRecordTest {
```
wcurrie, line 14: Copy paste of the existing `OsoTest` with `record User` instead of `class User`

languages/java/oso/src/test/kotlin/com/osohq/oso/OsoKotlinTest.kt
```diff
@@ -0,0 +1,123 @@
+package com.osohq.oso
+
+import com.osohq.oso.Exceptions.OsoException
+import org.junit.jupiter.api.Assertions
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import java.util.Set
+
+class OsoKotlinTest {
```
wcurrie, line 9: (Mostly) output of Intellij "convert to kotlin" of the existing `OsoTest`. The use `data class...` instead of `class`

PR #1673, opened 2023-02-01
Drop support for end-of-life Python 3.6

kkirsche:
- Drops support for Python 3.6 which is end-of-life
  - proof: https://devguide.python.org/versions/
- Upgrades internal code to support Python 3.7+ via PyUpgrade
  - This replaces the use of IOError with OSError as was changed in Python 3.3
  - Source: https://docs.python.org/3/library/exceptions.html#OSError which states:
  - Removes unnecessary calls to set, parenthesis, Python 2 style classes, etc.
> Changed in version 3.3: [EnvironmentError](https://docs.python.org/3/library/exceptions.html#EnvironmentError), [IOError](https://docs.python.org/3/library/exceptions.html#IOError), [WindowsError](https://docs.python.org/3/library/exceptions.html#WindowsError), [socket.error](https://docs.python.org/3/library/socket.html#socket.error), [select.error](https://docs.python.org/3/library/select.html#select.error) and mmap.error have been merged into [OSError](https://docs.python.org/3/library/exceptions.html#OSError), and the constructor may return a subclass.

PR checklist:

- [ ] Added changelog entry.

### DIFF
--- a/languages/python/django-oso/setup.py
+++ b/languages/python/django-oso/setup.py
@@ -10,7 +10,7 @@ here = path.abspath(path.dirname(__file__))
 try:
     with open(path.join(here, "README.md"), encoding="utf-8") as f:
         long_description = f.read()
-except IOError:
+except OSError:
     long_description = ""
 
 install_requires = ""
--- a/languages/python/django-oso/setup.py
+++ b/languages/python/django-oso/setup.py
@@ -46,12 +46,12 @@ setup(
     author="Oso Security, Inc.",
     classifiers=[
         "Development Status :: 4 - Beta",
-        "Programming Language :: Python :: 3.6",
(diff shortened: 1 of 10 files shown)


kkirsche:
I need to update the CI config still to remove Python 3.6

kkirsche:
> I haven't forgotten about this! Sorry for the delay @kkirsche. Going to endeavor to get to this & a couple of your other PRs tomorrow.

No worries, I know you all have a lot going on and are doing the best you can with a small team. Let me know if there is anything that can be done to simplify things for you or your team in the future.

kkirsche:
No worries. The packaging PR may help with things like Black since that version prevents the use of the latest versions over there. Hope you have a wonderful and relaxing weekend

PR #1699, opened 2023-05-02
Improves error reporting

xfbs:
Hey! We ran into an issue where we got an error from Oso. It looked like this:

```
Type error: Expected Instance
```

We've added this little hot patch here so that we get a more useful error message:

```
Type error: Expected Instance of service::auth::resources::Resource got String
```

The offending code was that we were using the wrong type in one place:

```polar
has_role(actor: User, “member”, resource: Resource) if
  actor.is_member_of(plo.id);
```

But this was really hard to track down in a large Polar file with the previous error message.

In the future, we would maybe improve it to add some more context to the errors, such as line numbers. This would greatly improve the debugging experience.

Maybe look over the PR before merging it because I put it together quickly for debugging.

### DIFF
--- a/languages/rust/oso/src/host/class.rs
+++ b/languages/rust/oso/src/host/class.rs
@@ -379,6 +379,10 @@ impl Instance {
             .unwrap_or_else(|_| self.debug_type_name)
     }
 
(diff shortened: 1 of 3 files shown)


alisomay:

Issue #1727, opened 2023-10-06
Not installable on Python 3.12


hynek:
Since oso has non-abi3 wheels (and no sdist) for only up to Python 3.11, trying to install it on Python 3.12 fails with a:

```
ERROR: Could not find a version that satisfies the requirement oso (from versions: none)
ERROR: No matching distribution found for oso
```

I could paste an extremely long `-vv` output, but it's just a long chain of:

```
Skipping link: none of the wheel's tags (pp39-pypy39_pp73-manylinux2014_x86_64, pp39-pypy39_pp73-manylinux_2_17_x86_64) are compatible (run pip debug --verbose to show compatible tags): https://files.pythonhosted.org/packages/7c/55/58be7fe2d37a8c486bd4ba4e1f456548811a9bbf5fa7885fd02fc08fc1dd/oso-0.27.0-pp39-pypy39_pp73-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (from https://pypi.org/simple/oso/) (requires-python:>=3.7)
  Skipping link: none of the wheel's tags (pp39-pypy39_pp73-win_amd64) are compatible (run pip debug --verbose to show compatible tags): https://files.pythonhosted.org/packages/2a/2c/584ecc3cc8cbff9fdb449568ee610b44b707b9308ce90cd6579f04697d61/oso-0.27.0-pp39-pypy39_pp73-win_amd64.whl (from https://pypi.org/simple/oso/) (requires-python:>=3.7)
```

… for all possible versions.

underyx:
of 295 dependencies in our project, oso seems to be the only one not installable on py3.12

kkirsche:
Pushed up a MR for the required release workflow changes based on what I had done for Python 3.11, just need to see if Oso will approve, merge, and push out a release

kkirsche:
Fixed via !1732 in https://pypi.org/project/oso/0.27.2/

PR #1641, opened 2022-11-03
refactor: update release workflow for Python 3.11

kkirsche:
Partial Fix for #1640 

PR checklist:

- [ ] Added changelog entry.

PR #1732, opened 2023-11-16
Add python 3.12 release code

kkirsche:
Based on https://github.com/osohq/oso/pull/1641

Fixes the code required for #1727 — then you just need to push out a release


kkirsche:
I understand that Oso OSS is deprecated, but it would be really useful to expand the supported versions

gneray:
Hey Kevin - looking into it.

kkirsche:
Thank you! 🙏 grateful for your time

PR #1733, opened 2023-12-01
Include an arm64 support in ruby

jdeff:
When running a Ruby on Rails application using Docker for Mac on my Apple Silicon machine, I am unable run `oso` because the shared library for linux included in the gem, `libpolar.so`, only supports x86_64 and my container is arm64. This change compiles both an x86_64 shared lib as well as a new aarch64 shared lib and adds them both to the gem.

PR checklist:

- [ ] Added changelog entry.


jdeff:

jdeff:
Any thoughts on this PR? Maybe @gj?

jdeff:
Another friendly ping on this change. Any questions or concerns @gj? I understand the project is going to be deprecated, but we would still find this change useful.

PR #1738, opened 2024-02-02
add OR support to django-oso FilterBuilder

destos:
I'm fine with trying to add additional tests for the partial changes if desired.

PR checklist:

- [ ] Added changelog entry.


destos:

samscott89:
Thanks @destos !

Yes, a test case would be great. In particular, we normally turn `or`s into disjunctive normal form, so I'm curious to see where those come from.

destos:
@samscott89 I believe this isn't actually needed as I broke the usage of or into different rules, which works fine.

